PR #13, opened 2021-09-29
operators / ground operators

tomsilver:


src/structs.py
```diff
@@ -35,7 +35,7 @@ def __hash__(self) -> int:
 
 
 @dataclass(frozen=True, order=True, repr=False)
-class TypedEntity:
+class _TypedEntity:
     """Struct defining an entity with some type, either an object (e.g.,
     block3) or a variable (e.g., ?block).
```
ronuchit, line 40: Append "Should not be instantiated externally." to this docstring

src/structs.py
```diff
@@ -222,10 +222,16 @@ def variables(self):
         return list(self.entities)
 
     @cached_property
-    def _str(self):
+    def _str(self) -> str:
         return (str(self.predicate) + "(" +
                 ", ".join(map(str, self.variables)) + ")")
 
+    def ground(self, sub) -> GroundAtom:
```
ronuchit, line 229: Would we ever need to go the other way, i.e. GroundAtom to LiftedAtom? We would need this sometimes right, like when we lift effects for clustering?

ronuchit, line 229: Please annotate the type of "sub" parameter

src/structs.py
```diff
@@ -303,3 +309,92 @@ class _Option:
...
+    def __str__(self) -> str:
+        return self._str
+
+    def __repr__(self) -> str:
+        return str(self)
+
+    def __hash__(self) -> int:
+        return self._hash
+
+    def __eq__(self, other) -> bool:
```
ronuchit, line 349: Please annotate the type of "other" (actually, I think this is missing in a few places in this file)

src/structs.py
```diff
@@ -303,3 +309,92 @@ class _Option:
...
+        delete_effects = {atom.ground(sub) for atom in self.delete_effects}
+        sampler = lambda s: self._sampler(s, objects)
+        return _GroundOperator(self, objects, preconditions, add_effects,
+                               delete_effects, self.option, sampler)
+
+
+@dataclass(frozen=True, repr=False, eq=False)
+class _GroundOperator:
+    """A ground operator is an operator + objects."""
+    operator: Operator
```
ronuchit, line 369: Is there a reason GroundOperators need to store a pointer to the lifted Operator? I'm fine with it if we have a reason, but right now it looks like we're just using that lifted Operator's name, so we could just pass the name in instead?

tests/test_structs.py
```diff
@@ -202,3 +203,54 @@ def _terminal(s, p):
     assert np.all(option.policy(state) == np.array(params)*2)
     assert not option.initiable(state)
     assert not option.terminal(state)
+
+
+def test_operators():
+    """Tests for Operators and _GroundOperators.
```
ronuchit, line 209: Operator and _GroundOperator [to avoid confusion]

src/structs.py
```diff
@@ -303,3 +309,92 @@ class _Option:
...
+class Operator:
+    """Struct defining an operator (as in STRIPS). Lifted!
+    """
+    name: str
+    parameters: Sequence[Variable]
+    preconditions: Collection[LiftedAtom]
+    add_effects: Collection[LiftedAtom]
+    delete_effects: Collection[LiftedAtom]
+    option: ParameterizedOption
+    # A sampler maps a state and objects to a option parameters.
```
ronuchit, line 324: typo: "to option parameters"

src/structs.py
tomsilver: so polite

tomsilver: so polite

src/structs.py
tomsilver: this is apparently not recommended:

```
predicators/src/structs.py:356: error: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"
predicators/src/structs.py:356: note: This violates the Liskov substitution principle
predicators/src/structs.py:356: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
predicators/src/structs.py:356: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
predicators/src/structs.py:356: note:     def __eq__(self, other: object) -> bool:
predicators/src/structs.py:356: note:         if not isinstance(other, Operator):
predicators/src/structs.py:356: note:             return NotImplemented
predicators/src/structs.py:356: note:         return <logic to compare two Operator instances>
```

src/structs.py
tomsilver: I can't think of a reason but I would like to keep it

PR #14, opened 2021-09-29
base env and approach

ronuchit:


src/approaches/base_approach.py
```diff
@@ -0,0 +1,74 @@
...
+        """Get the atomic representation of this state (i.e., a set
+        of ground atoms).
+        """
+        atoms = set()
+        for pred in self._predicates:
+            domains = []
+            for var_type in pred.types:
+                domains.append([obj for obj in state if obj.type == var_type])
+            for choice in itertools.product(*domains):
+                if len(choice) != len(set(choice)):
```
tomsilver, line 65: maybe let's remove this...?

src/envs/base_env.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @abc.abstractmethod
+    def get_test_tasks(self) -> List[Task]:
+        """Get an ordered list of tasks for testing / evaluation.
+        """
+        raise NotImplementedError("Override me!")
+
+    @property
+    @abc.abstractmethod
+    def predicates(self) -> Set[Predicate]:
```
tomsilver, line 40: when should we use Sequence vs List and Set vs Collection?

src/structs.py
```diff
@@ -30,7 +30,7 @@ def __call__(self, name) -> TypedEntity:
             return Variable(name, self)
         return Object(name, self)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
```
tomsilver, line 33: oh yeah thanks for doing this...

src/structs.py
```diff
@@ -113,6 +113,22 @@ def vec(self, objects: Sequence[Object]) -> ArrayLike:
             feats.append(self[obj])
         return np.hstack(feats)
 
+    def copy(self) -> State:
```
tomsilver, line 116: so states are not frozen then

src/utils.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+
+from typing import Sequence, Callable, Tuple
+from numpy.typing import ArrayLike
+from predicators.src.structs import _Option, State
+
+
+def option_to_trajectory(
+        init: State, simulator: Callable[[State, ArrayLike], State],
+        option: _Option) -> Tuple[Sequence[State], Sequence[ArrayLike]]:
```
tomsilver, line 11: we should probably add a `max_num_steps` argument

tomsilver (approved):
good stuff

PR #19, opened 2021-09-30
random actions approach

tomsilver:


tests/approaches/test_random_actions_approach.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+from gym.spaces import Box  # type: ignore
+from predicators.src.approaches import RandomActionsApproach
+from predicators.src.structs import State, Type, Predicate, Task
+
+
+def test_random_actions_approach():
+    """Tests for RandomActionsApproach class.
+    """
+    cup_type = Type("cup_type", ["feat1"])
```
ronuchit, line 11: Instead of all this, you can do env = CoverEnv() and then RandomActionsApproach(env.simulate, env.predicates, env.options, env.action_space)

PR #36, opened 2021-10-05
demo+replay data generation

tomsilver:


tests/datasets/test_datasets.py
```diff
@@ -44,4 +44,25 @@ def test_demo_dataset():
...
+    })
+    env = CoverEnv()
+    dataset = create_dataset(env, 123)
+    assert len(dataset) == 5 + 3
+    assert len(dataset[-1]) == 2
+    assert len(dataset[-1][0]) == 2
+    assert len(dataset[-1][1]) == 1
+    for _, actions in dataset:
+        for action in actions:
+            assert action.has_option()
```
ronuchit, line 68: Add this test too:
```
# Test that data does not contain options since approach is random
utils.update_config({
    "env": "cover",
    "approach": "random",
})
env = CoverEnv()
dataset = create_dataset(env, 123)
assert len(dataset) == 5 + 3
assert len(dataset[-1]) == 2
assert len(dataset[-1][0]) == 2
assert len(dataset[-1][1]) == 1
for _, actions in dataset:
    for action in actions:
        assert not action.has_option()
```

ronuchit (approved):

PR #39, opened 2021-10-05
operator learning base code

ronuchit:


src/approaches/operator_learning_approach.py
```diff
@@ -0,0 +1,210 @@
...
+                assert variables == sorted(set(sub.values()))
+            if i == 0:
+                preconditions = lifted_atoms
+            else:
+                preconditions &= lifted_atoms
+
+        return variables, preconditions
+
+    @staticmethod
+    def _unify(
```
tomsilver, line 169: can we lru_cache this? it's going to be called twice I think -- first during effect clustering, and then again during precondition learning

PR #40, opened 2021-10-05
video generation

tomsilver:


src/utils.py
```diff
@@ -29,25 +35,33 @@ def run_policy_on_task(policy: Callable[[State], Action], task: Task,
     atoms = abstract(state, predicates)
     states = [state]
     actions: List[Action] = []
+    video: Video = []
+    if make_video:
+        assert render is not None
+        render = cast(Callable[[State], Image], render)
```
ronuchit, line 41: I think that since you have the assert right above this, you won't need this cast (mypy would be smart enough to realize that the "Optional" should get dropped since it was asserted)

src/args.py
```diff
@@ -13,9 +13,21 @@ def parse_args() -> Dict[str, Any]:
...
-    args = parser.parse_args()
+    parser.add_argument("--make_videos", action='store_true')
+    args, overrides = parser.parse_known_args()
     print_args(args)
-    return vars(args)
+    arg_dict = vars(args)
+    if len(overrides) == 0:
+        return arg_dict
+    # Override global settings
+    assert len(overrides) >= 2
```
ronuchit, line 23: Should also check that it's even

PR #48, opened 2021-10-07
sampler bounds checking

ronuchit:


src/operator_learning.py
```diff
@@ -284,8 +284,12 @@ def _sampler(state: State, rng: np.random.Generator,
             x_lst.extend(state[sub[var]])
         x = np.array(x_lst)
         for _ in range(CFG.max_rejection_sampling_tries):
-            params = regressor.predict_sample(x, rng)
+            params = np.array(regressor.predict_sample(x, rng),
+                              dtype=param_option.params_space.dtype)
             if classifier.classify(np.r_[x, params]):
                 break
+        if not param_option.params_space.contains(params):
```
tomsilver, line 291: what are your thoughts about doing this in the rejection sampling loop instead? so, if params are sampled that pass the classifier, but don't pass the param space check, then we'll sample again if there's still time left?

PR #50, opened 2021-10-10
Teacher dataset for interactive learning

amburger66:


src/settings.py
```diff
@@ -25,6 +25,9 @@ class GlobalSettings:
     video_dir = "videos"
     video_fps = 2
 
+    # teacher dataset parameters
+    frac = 0.3
```
tomsilver, line 29: Can we make this name more specific? Maybe `teacher_dataset_label_ratio`? Or if that's too long, just `teacher_dataset_frac`?

tests/datasets/test_datasets.py
```diff
@@ -91,3 +92,24 @@ def test_demo_replay_dataset():
...
+    utils.update_config({
+        "env": "cover",
+        "approach": "random",
+        "offline_data_method": "demo",
+        "offline_data_planning_timeout": 500,
+        "seed": 0,
+        "num_train_tasks": 7,
+    })
+    env = CoverEnv()
+    dataset = create_dataset(env)
```
tomsilver, line 110: Can you add a test that checks if the new hyperparameter is being used as expected? The easiest thing would be to just assert that the length of the ground atoms in one of the examples is equal to a magic number. Then we'll be alerted in the future if something changes unexpectedly. A more robust test would recompute all groundatoms for a given state and check that the number of groundatoms in the dataset is proportional to the expected frac.

src/datasets/teacher.py
```diff
@@ -0,0 +1,32 @@
...
+
+from typing import List, Set
+import numpy as np
+from predicators.src.envs import BaseEnv
+from predicators.src.structs import Dataset, GroundAtom
+from predicators.src.settings import CFG
+from predicators.src import utils
+
+
+def create_teacher_dataset(env: BaseEnv,
```
tomsilver, line 12: one more very minor thing: maybe we should pass in predicates here instead of `env`, since the only thing from env that we're using is `env.predicates`? but I know that the other dataset functions take in `env`, so maybe it's nice to be parallel. I'll leave this up to you!

amburger66: this function is already different from the other dataset functions, so I think it is ok to pass in just the predicates!

tests/datasets/test_datasets.py
```diff
@@ -91,3 +93,39 @@ def test_demo_replay_dataset():
...
+def test_teacher_dataset():
+    """Test teacher dataset creation with Covers env.
+    """
+    # Test that data does not contain options since approach is random
+    utils.update_config({
+        "env": "cover",
+        "approach": "random",
+        "offline_data_method": "demo",
+        "offline_data_planning_timeout": 500,
+        "seed": 0,
```
ronuchit, line 107: could you use seed 123 here instead of 0? sorry for the nit lol

explanation: i've been trying to do this across all our tests lol, the short explanation is that if we get in the habit of using seed 0 for our tests, then the tests that write out saved models will overwrite the saved models of our regular runs, since we often run `python src/main.py` with `--seed 0` out of habit

@tomsilver

ronuchit:
lgtm :)

PR #57, opened 2021-10-13
basic cluttered table implementation

ronuchit:


src/envs/cluttered_table.py
```diff
@@ -0,0 +1,188 @@
...
+        self._can_type = Type(
+            "can", ["pose_x", "pose_y", "radius", "is_grasped", "is_trashed"])
+        # Predicates
+        self._HandEmpty = Predicate(
+            "HandEmpty", [], self._HandEmpty_holds)
+        self._Holding = Predicate(
+            "Holding", [self._can_type], self._Holding_holds)
+        # Options
+        self._Grasp = ParameterizedOption(
+            "Grasp", [self._can_type], params_space=Box(0, 1, (4,)),
```
tomsilver, line 29: can you add a comment for what these dims represent

src/envs/cluttered_table.py
```diff
@@ -0,0 +1,188 @@
...
+            if grasped_can is not None:
+                next_state.set(grasped_can, "pose_x", -999)
+                next_state.set(grasped_can, "pose_y", -999)
+                next_state.set(grasped_can, "is_grasped", 0.0)
+                next_state.set(grasped_can, "is_trashed", 1.0)
+            return next_state
+        # Handle grasping action.
+        if grasped_can is not None:
+            return next_state  # can't grasp while already grasping
+        start_x, start_y, end_x, end_y = action.arr
```
tomsilver, line 67: maybe i'll figure this out later, but why do we need `start_x` and `start_y`? isn't that determined by the robot's resting state?

src/envs/cluttered_table.py
```diff
@@ -0,0 +1,188 @@
...
+            this_y = state.get(can, "pose_y")
+            this_radius = state.get(can, "radius")
+            if np.linalg.norm([end_x-this_x,
+                               end_y-this_y]) < this_radius:  # type: ignore
+                assert desired_can is None
+                desired_can = can
+        if desired_can is None:
+            return next_state  # end point wasn't at any can
+        vec1 = np.array([end_x-start_x, end_y-start_y])
+        # Handle collision checking.
```
tomsilver, line 80: can you write some more comments for this logic? in particular I'm surprised that we don't need to use the other can's radius

src/envs/cluttered_table.py
tomsilver: oh I see that you did this in action_space, that's fine, nvm

src/envs/cluttered_table.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @property
+    def action_space(self) -> Box:
+        # The action_space is 4-dimensional. The first two dimensions are the
+        # start point of the vector corresponding to the grasp approach. The
+        # last two dimensions are the end point. Dumping is a special action
+        # where all 4 dimensions are 0.
+        return Box(0, 1, (4,))
+
+    def render(self, state: State) -> Image:
```
tomsilver, line 131: wanna create an Issue for this? :) i can do it eventually

PR #58, opened 2021-10-13
RandomOptionsApproach

ronuchit:


src/approaches/random_options_approach.py
```diff
@@ -0,0 +1,43 @@
...
+class RandomOptionsApproach(BaseApproach):
+    """Samples random options (and random parameters for those options).
+    """
+    @property
+    def is_learning_based(self) -> bool:
+        return False
+
+    def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+        options = sorted(self._initial_options, key=lambda o: o.name)
+        plan: List[Action] = []
```
tomsilver, line 21: this is okay because the env is deterministic etc, but it's a little odd -- why do we need to plan at all if we're executing random options?

src/approaches/random_options_approach.py
```diff
@@ -0,0 +1,43 @@
...
+        def _policy(state: State) -> Action:
+            nonlocal plan
+            if plan:
+                return plan.pop(0)
+            for _ in range(CFG.random_options_max_tries):
+                param_opt = options[self._rng.choice(len(options))]
+                types_to_objs = defaultdict(list)
```
tomsilver, line 28: worth moving this logic into a utility function? seems potentially generally useful (sample a random object combination from a state, given a list of types)

src/approaches/random_options_approach.py
```diff
@@ -0,0 +1,42 @@
...
+class RandomOptionsApproach(BaseApproach):
+    """Samples random options (and random parameters for those options).
+    """
+    @property
+    def is_learning_based(self) -> bool:
+        return False
+
+    def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+        options = sorted(self._initial_options, key=lambda o: o.name)
+        cur_option: _Option = DefaultOption
```
tomsilver, line 21: you need the type annotation even though it's getting initialized to DefaultOption?

src/approaches/random_options_approach.py
```diff
@@ -0,0 +1,42 @@
...
+                    params = param_opt.params_space.sample()
+                    opt = param_opt.ground(objs, params)
+                    if opt.initiable(state):
+                        cur_option = opt
+                        cur_option_ind = 0
+                        break
+                else:  # fall back to a random action
+                    return Action(self._action_space.sample())
+            act = cur_option.policy(state)
+            act.set_option((cur_option, cur_option_ind))
```
tomsilver, line 39: it would be ideal if we didn't have to do this everywhere, but that's an issue for another day. (if you agree, maybe make an issue?)

PR #67, opened 2021-10-14
rendering for clutter tables

### LINKED ISSUE #59: add render method to cluttered table environment
@tomsilver

tomsilver:
Closes #59


ronuchit:
TODOs from slack: make colored dots the same color, target should be a different color. to operationalize this, need render() to take in Tasks rather than just states

PR #68, opened 2021-10-14
First attempt at interactive learning approach

amburger66:
- semi-supervision only


src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+from predicators.src.utils import get_object_combinations
+from predicators.src.settings import CFG, get_save_path
+
+
+class InteractiveLearningApproach(TAMPApproach):
+    """An approach that learns predicates from a teacher.
+    """
+    def __init__(self,
+                 simulator: Callable[[State, Action], State],
+                 true_predicates: Set[Predicate],
```
tomsilver, line 26: I'm getting hung up on "true" here -- maybe `all_predicates` instead?

src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+                 types: Set[Type],
+                 action_space: Box,
+                 train_tasks: List[Task]) -> None:
+        # Only the teacher is allowed to know about the initial predicates
+        self._known_predicates = {p for p in true_predicates
+                                  if p.name in CFG.interactive_known_predicates}
+        predicates_to_learn = true_predicates - self._known_predicates
+        self._teacher = _Teacher(true_predicates, predicates_to_learn)
+        # All seen data
+        self.dataset: List[ActionTrajectory] = []
```
tomsilver, line 37: Let's stick with private attributes (names starting with underscores) unless the attribute needs to be used outside the class

src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @property
+    def is_learning_based(self) -> bool:
+        return True
+
+    def _get_current_operators(self) -> Set[Operator]:
+        assert self._operators, "Operators not learned"
+        return self._operators
+
+    def _get_current_predicates(self) -> Set[Predicate]:
```
tomsilver, line 55: Will this / should this include the known predicates? Maybe name `self._predicates` to something more specific to be clear

src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+        return self._predicates
+
+    def learn_from_offline_dataset(self, dataset: Dataset) -> None:
+        # Get data from teacher
+        ground_atom_data = self._teacher.generate_data(dataset)
+        self.dataset.extend(dataset)
+        self.ground_atom_dataset.extend(ground_atom_data)
+        # Learn predicates
+        for pred in self._predicates:
+            print("\nPredicate:", pred)
```
tomsilver, line 65: Remove, or add more words to clarify the printout

src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+            assert pred not in self._known_predicates
+            positive_examples = []
+            negative_examples = []
+            # Positive examples
+            for i, trajectory in enumerate(self.ground_atom_dataset):
```
tomsilver, line 70: It would be possible to reorganize this part of the code so that you only have to go through the dataset once total; right now it'll be 2*|predicates|. But the way that you have it written now is clear, so let's leave it this way for now. If in the future, datasets become so large that this part is slow, we can reorganize then.

src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+            # Negative examples - assume unlabeled is negative for now
+            for (ss, _) in self.dataset:
+                for state in ss:
+                    possible = [state.vec(choice)
+                                for choice in get_object_combinations(
+                                                  list(state),
+                                                  pred.types,
+                                                  allow_duplicates=False)]
+                    negatives = []
+                    for (ex, pos) in itertools.product(possible,
```
tomsilver, line 86: This is also something we can refactor later on if we're seeing slowness during classification dataset generation.

src/approaches/interactive_learning_approach.py
```diff
@@ -0,0 +1,178 @@
...
+            Y = np.array([1 for _ in positive_examples] +
+                         [0 for _ in negative_examples])
+            model = MLPClassifier(X.shape[1])
+            model.fit(X, Y)
+            torch.save(model, f"{save_path}_{pred}.classifier")
+
+            # Construct classifier function, create new Predicate, and save it
+            _classifier = _create_classifier(model)
+            new_pred = Predicate(pred.name, pred.types, _classifier)
+            self._predicates = self._predicates - {pred} | {new_pred}
```
tomsilver, line 109: Can you put parentheses here so the order of operations is clear?

tests/approaches/test_interactive_approach.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def test_interactive_learning_approach():
+    """Tests for InteractiveLearningApproach class.
+    """
+    utils.update_config({"env": "cover", "approach": "interactive_learning",
+                         "timeout": 10, "max_samples_per_step": 10,
+                         "seed": 12345, "classifier_max_itr": 500,
+                         "regressor_max_itr": 500})
+    env = CoverEnv()
```
tomsilver, line 57: How slow is this test? If it's more than a couple of seconds, maybe reduce the hyperparameters, since we want the unit tests to be relatively fast overall

src/approaches/interactive_learning_approach.py
amburger66: it looks like it should include all the predicates?

PR #74, opened 2021-10-20
blocks env implementation

ronuchit:
@tomsilver Please review this blocks implementation. I've verified it works with oracle and operator learning as expected, across multiple seeds. The only thing left to add is unit tests. In particular, can you check the option params spaces & action space, and the policies & samplers?

### DIFF
--- a/src/approaches/oracle_approach.py
+++ b/src/approaches/oracle_approach.py
@@ -36,6 +36,8 @@ def get_gt_ops(predicates: Set[Predicate],
         ops = _get_cover_gt_ops(options_are_typed=True)
     elif CFG.env == "cluttered_table":
         ops = _get_cluttered_table_gt_ops()
+    elif CFG.env == "blocks":
+        ops = _get_blocks_gt_ops()
     else:
         raise NotImplementedError("Ground truth operators not implemented")
     # Filter out excluded predicates/options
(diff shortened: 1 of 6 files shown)


src/approaches/oracle_approach.py
```diff
@@ -208,3 +210,106 @@ def grasp_sampler(state: State, rng: np.random.Generator,
...
+    option = PutOnTable
+    preconditions = {LiftedAtom(Holding, [block])}
+    add_effects = {LiftedAtom(OnTable, [block]),
+                   LiftedAtom(Clear, [block]),
+                   LiftedAtom(GripperOpen, [robot])}
+    delete_effects = {LiftedAtom(Holding, [block])}
+    def putontable_sampler(state: State, rng: np.random.Generator,
+                           objs: Sequence[Object]) -> Array:
+        del state, objs  # unused
+        x = rng.uniform()
```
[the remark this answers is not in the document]
tomsilver: it looks like the dimensions are actually quite different, so why are we limiting put on table to this range?

src/approaches/oracle_approach.py
```diff
@@ -208,3 +210,106 @@ def grasp_sampler(state: State, rng: np.random.Generator,
...
+                     LiftedAtom(Clear, [block]),
+                     LiftedAtom(GripperOpen, [robot])}
+    add_effects = {LiftedAtom(Holding, [block])}
+    delete_effects = {LiftedAtom(OnTable, [block]),
+                      LiftedAtom(Clear, [block]),
+                      LiftedAtom(GripperOpen, [robot])}
+    def pick_sampler(state: State, rng: np.random.Generator,
+                     objs: Sequence[Object]) -> Array:
+        del state, rng, objs  # unused
+        return np.zeros(3, dtype=np.float32)
```
tomsilver, line 244: what's the point of sampling a non-zero-shape vector if it's deterministic? can't we just have the pick option always output the same action, like Dump? same question for the other options (except put on table)

src/envs/blocks.py
```diff
@@ -0,0 +1,439 @@
...
+        # Options
+        self._Pick = ParameterizedOption(
+            # variables: [object to pick]
+            # params: [delta x, delta y, delta z]
+            "Pick", types=[self._block_type], params_space=Box(-1, 1, (3,)),
+            _policy=self._Pick_policy,
+            _initiable=self._Pick_initiable,
+            _terminal=self._Pick_terminal)
+        self._Stack = ParameterizedOption(
+            # variables: [object to stack currently held object onto]
```
tomsilver, line 60: it took me like 30 seconds to parse this, can we say: "object on which to stack currently-held-object"

src/envs/blocks.py
```diff
@@ -0,0 +1,439 @@
...
+            return next_state
+        block = self._get_held_block(state)
+        assert block is not None
+        x, y, z, fingers = action.arr
+        # Check that table surface is clear at this pose
+        poses = [[state.get(b, "pose_x"),
+                  state.get(b, "pose_y"),
+                  state.get(b, "pose_z")] for b in state
+                 if b.type == self._block_type]
+        existing_xys = {(float(p[0]), float(p[1])) for p in poses}
```
tomsilver, line 133: just wondering, why do we need to `float()` these?

src/envs/blocks.py
```diff
@@ -0,0 +1,439 @@
...
+        return {self._block_type, self._robot_type}
+
+    @property
+    def options(self) -> Set[ParameterizedOption]:
+        return {self._Pick, self._Stack, self._PutOnTable}
+
+    @property
+    def action_space(self) -> Box:
+        # dimensions: [x, y, z, fingers]
+        return Box(-100.0, 100.0, (4,))
```
tomsilver, line 202: any reason why these bounds are so big? the experience with -1000 in the covers environment leading to issues with predicate learning for amber made me weary of large magic numbers

src/envs/blocks.py
```diff
@@ -0,0 +1,439 @@
...
+        raise NotImplementedError
+
+    def _get_tasks(self, num_tasks: int, possible_num_blocks: List[int],
+                   rng: np.random.Generator) -> List[Task]:
+        tasks = []
+        for _ in range(num_tasks):
+            num_blocks = rng.choice(possible_num_blocks)
+            piles = self._sample_initial_piles(num_blocks, rng)
+            init_state = self._sample_state_from_piles(piles, rng)
+            atoms = utils.abstract(init_state, self.predicates)
```
tomsilver, line 215: i'm thinking about whether this will be an issue when we turn off predicates for learning. probably not, because we're always going to assume that we have the predicates for defining the goals, and that's all that we need here. right?

src/envs/blocks.py
```diff
@@ -0,0 +1,441 @@
...
+        return {self._Pick, self._Stack, self._PutOnTable}
+
+    @property
+    def action_space(self) -> Box:
+        # dimensions: [x, y, z, fingers]
+        return Box(-100.0, 100.0, (4,))
+
+    def render(self, state: State, task: Task,
+               action: Optional[Action] = None) -> Image:
+        raise NotImplementedError
```
tomsilver, line 206: wanna make me an issue? :)

tests/approaches/test_oracle_approach.py
```diff
@@ -172,3 +172,26 @@ def test_oracle_approach_cluttered_table():
...
+    utils.update_config({"env": "blocks"})
+    env = BlocksEnv()
+    env.seed(123)
+    approach = OracleApproach(
+        env.simulate, env.predicates, env.options, env.types,
+        env.action_space, env.get_train_tasks())
+    assert not approach.is_learning_based
+    approach.seed(123)
+    # Test a couple of train tasks so that we get at least one which
+    # requires resampling placement poses on the table.
```
tomsilver, line 189: nice

src/approaches/oracle_approach.py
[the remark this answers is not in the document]
ronuchit: These are normalized parameters to make learning more straightforward, they get un-normalized in the PutOnTable_policy

src/approaches/oracle_approach.py
ronuchit: these are oracle samplers. we would want our system to learn this (admittedly easy) sampler. i know it's a trivial thing to learn but it's just for sanity checking our sampler learning

src/envs/blocks.py
ronuchit: fixed

PR #78, opened 2021-10-21
blocks rendering

### LINKED ISSUE #75: implement rendering/pybullet for blocks env
either @tomsilver or @ronuchit

tomsilver:
Closes #75

### DIFF
--- a/tests/envs/test_blocks.py
+++ b/tests/envs/test_blocks.py
@@ -40,10 +40,11 @@ def test_blocks():
             assert not (state.get(block, "held") and state.get(block, "clear"))
         if i == 0:
             # Force initial pick to test rendering with holding
+            Pick = [o for o in env.options if o.name == "Pick"][0]
             block = sorted([o for o in state if o.type.name == "block" and \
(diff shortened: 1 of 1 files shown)


tests/envs/test_blocks.py
```diff
@@ -40,12 +38,13 @@ def test_blocks():
...
-        act = Action(env.action_space.sample())
         if i == 0:
-            with pytest.raises(NotImplementedError):
-                env.render(state, task, act)
-        env.simulate(state, act)
-
+            # Force initial pick to test rendering with holding
+            block = sorted([o for o in state if o.type.name == "block" and \
+                            state.get(o, 'clear') > env.clear_tol])[0]
+            pick = env._Pick_policy(state, [block], np.zeros(3))  # pylint: disable=protected-access
```
ronuchit, line 45: instead of this protected access, can we do it the way i did in the function below?
```
Pick = [o for o in env.options if o.name == "Pick"][0]
act = Pick.ground([block1], np.array(
        [0, 0, 0], dtype=np.float32)).policy(state)
```

PR #95, opened 2021-11-01
hierarchical typing

ronuchit:


ronuchit:
There is a question of what to do about the feature names, since technically, the child's feature names should always be a superset of the parent's. I think we can just let users of the Type class deal with that. Thoughts @tomsilver ?

tomsilver:
Re: feature name question, that's interesting... I agree that we should let the users deal with that.

PR #96, opened 2021-11-06
Active learning bugs

amburger66:
Fixes:
- make sure student only asks about predicates to learn
- for states that have no ground atoms samples, just have an empty set instead of failing
- new dataset type to better ensure that # states = # set of ground atoms in each trajectory

### DIFF
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -112,7 +112,7 @@ class InteractiveLearningApproach(OperatorLearningApproach):
                 assert len(states) == len(ground_atom_sets)
                 for (state, ground_atom_set) in zip(states, ground_atom_sets):
                     if len(ground_atom_set) == 0:
-                        continue
+                        continue  # pragma: no cover
                     positives = [state.vec(ground_atom.objects)
                                  for ground_atom in ground_atom_set
                                  if ground_atom.predicate == pred]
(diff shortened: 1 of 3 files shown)


src/approaches/interactive_learning_approach.py
```diff
@@ -116,29 +108,30 @@ def _relearn_predicates_and_operators(self) -> None:
...
             negative_examples = []
             # Positive examples
-            for i, trajectory in enumerate(self._ground_atom_dataset):
-                for j, ground_atom_set in enumerate(trajectory):
-                    state = self._dataset[i][0][j]
+            for (states, _, ground_atom_sets) in self._dataset:
+                assert len(states) == len(ground_atom_sets)
+                for (state, ground_atom_set) in zip(states, ground_atom_sets):
+                    if len(ground_atom_set) == 0:
+                        continue  # pragma: no cover
```
ronuchit, line 115: Is there a reason we can't test this? E.g. if we give data that contains states but we have no predicates, then ground_atom_set should always be empty, and we can test that there are no positives?

ronuchit (commented):
lgtm

tomsilver:
LGTM too!

src/approaches/interactive_learning_approach.py
amburger66: good point! how is this?

ronuchit: if the test is working correctly, we should be able to remove this "pragma: no cover" right?

amburger66: right, i forgot to remove it

Issue #99, opened 2021-11-09
add `option transition model` to API


tomsilver:
In the long term, we imagine learning an option transition model, but not necessarily a low-level simulator. The current code-base does not make a distinction between these. We would like to add to the API an option transition model, which by default will just use the ground truth simulator to step through the option.

PR #101, opened 2021-11-10
option models

ronuchit:


src/settings.py
```diff
@@ -117,6 +117,15 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
...
 
+            # Name of the option model to use.
+            option_model_name=defaultdict(str, {
+                "cover": "default",
+                "cover_typed_options": "default",
+                "cover_hierarchical_types": "default",
+                "cluttered_table": "default",
+                "blocks": "default",
+            })[args["env"]],
+
```
ronuchit, line 128: doing it this way since behavior will need something different

ronuchit:
closes #99

tomsilver:
LGTM! yay!

PR #102, opened 2021-11-12
Boring room playroom env

amburger66:
Design spec here: https://docs.google.com/document/d/1OcLreCaokFE1drLMEg1pUssbCGGYPLpwj8rtYXrrp7E/edit?usp=sharing
The random actions approach runs, and `render` produces a video! The `simulate` code is kind of messy (actually the whole file is messy) so I was wondering if anyone has any feedback for refactoring before I start writing all the unit tests :'D


src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+        cls = PlayroomEnv
+        # Infer which transition function to follow
+        # In boring room next to table
+        if self._NextToTable_holds(state, (self._robot,)):
+            # Interact with blocks
+            if (cls.table_x_lb < x < cls.table_x_ub) \
+                and (cls.table_y_lb < y < cls.table_y_ub) \
+                and self._robot_is_facing_table(action):
+                if fingers < 0.5:
+                    transition_fn = self._transition_pick
```
tomsilver, line 115: rather than this pattern, I suggest just `return self._transition_pick(state, action)`, etc.

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+        x, y, z, rotation, fingers = action.arr
+        cls = PlayroomEnv
+        # Infer which transition function to follow
+        # In boring room next to table
+        if self._NextToTable_holds(state, (self._robot,)):
+            # Interact with blocks
+            if (cls.table_x_lb < x < cls.table_x_ub) \
+                and (cls.table_y_lb < y < cls.table_y_ub) \
+                and self._robot_is_facing_table(action):
+                if fingers < 0.5:
```
tomsilver, line 114: let's try to make all of the magic numbers constants in the class definition at the top

tomsilver:
@amburger66 can you add an example rendering to the design doc, and also label the x and y axes?

src/approaches/interactive_learning_approach.py
```diff
@@ -2,7 +2,7 @@
 """
 
 import itertools
-from typing import Set, Callable, List, Collection, Tuple
+from typing import Set, Callable, List, Collection
```
tomsilver, line 5: can you split off any changes that are not playroom-specific into separate PRs?

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+
+    def _robot_is_facing_door(self, state: State, action: Action) -> bool:
+        door = self._get_door_next_to(state, action)
+        door_x = state.get(door, "pose_x")
+        x, _, _, rotation, _ = action.arr
+        return (x < door_x and (rotation >= 1.5 or rotation <= -1.5)) \
+            or (x >= door_x and 0.5 >= rotation >= -0.5)
+
+    def _robot_can_move(self, state: State, action: Action) -> bool:
+        # Any doors along the path must be open
```
tomsilver, line 837: nice! this turned out simpler than I expected :)

src/approaches/interactive_learning_approach.py
```diff
@@ -2,7 +2,7 @@
 """
 
 import itertools
```
ronuchit, line 4: Are the changes in this file and test_interactive_learning_approach hand-in-hand with the PlayroomEnv? If not, we could separate out the PRs

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+            _initiable=self._Stack_initiable,
+            _terminal=self._Stack_terminal)
+        self._PutOnTable = ParameterizedOption(
+            # variables: [robot]
+            # params: [x, y] (normalized coordinates on the table surface)
+            "PutOnTable", types=[self._robot_type],
+            params_space=Box(0, 1, (2,)),
+            _policy=self._PutOnTable_policy,
+            _initiable=self._PutOnTable_initiable,
+            _terminal=self._PutOnTable_terminal)
```
ronuchit, line 87: Aren't we missing some other predicates?

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+                and (cls.table_y_lb < y < cls.table_y_ub) \
+                and self._robot_is_facing_table(action):
+                if fingers < 0.5:
+                    transition_fn = self._transition_pick
+                elif z < self.table_height + CFG.playroom_block_size:
+                    transition_fn = self._transition_putontable
+                else:
+                    transition_fn = self._transition_stack
+            # Move elsewhere in boring room
+            elif (0.0 < x < 30.0) and (0.0 < y < 30.0):
```
ronuchit, line 121: Let's make these constants instead of magic numbers?

ronuchit: Or pull out this check into a function that's well-named, like "xy_in_boring_room(x, y)"

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+                transition_fn = self._transition_move
+            # Move to hallway or to playroom
+            elif ((30.0 < x < 110.0 and 10.0 < y < 20.0) \
+                    or (110.0 < x < 140.0 and 0.0 < y < 30.0)) \
+                    and self._robot_can_move(state, action):
+                transition_fn = self._transition_move
+            else:  # Invalid
+                return state
+        # In boring room next to door
```
tomsilver, line 130: hmm I'm wondering if we can consolidate some of this so that we just have one chunk of code to handle opening/closing any door. maybe it would be helpful to break it up like: first handle the movement part of the action. then, after moving, check if we're next to any door and open or close it. would that work?

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+            door_y = state.get(self._door6, "pose_y")
+            # Open/close door
+            if (door_x-cls.door_tol < x < door_x + cls.door_tol) and \
+                (door_y-cls.door_tol < y < door_y+cls.door_tol) \
+                and (cls.door_button_z-cls.door_tol < z
+                     < cls.door_button_z+cls.door_tol) \
+                and (rotation >= 1.5 or rotation <= -1.5):
+                transition_fn = self._transition_door
+            # Move elsewhere in playroom
+            elif (110.0 < x < 140.0) and (0.0 < y < 30.0):
```
ronuchit, line 236: Why are we repeating these same kinds of checks so many times? x and y are coming out of the action, so at the start of this function we could just create all the boolean variables we need, e.g. `is_in_playroom = (110.0 < x < 140.0) and (0.0 < y < 30.0)`

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+            door_x = state.get(self._door1, "pose_x")
+            door_y = state.get(self._door1, "pose_y")
+            # Open/close door
+            if (door_x-cls.door_tol < x < door_x + cls.door_tol) \
+                and (door_y-cls.door_tol < y < door_y+cls.door_tol) \
+                and (cls.door_button_z-cls.door_tol < z
+                        < cls.door_button_z+cls.door_tol) \
+                and 0.5 >= rotation >= -0.5:
+                transition_fn = self._transition_door
+            # Move elsewhere in boring room
```
tomsilver, line 142: yeah let's try to remove redundancies here :) I think some redundancies will get removed if we follow the move-first-then-handle-doors/dials/blocks pattern. but if there are any other redundancies remaining let's factor out the common code into helper functions.

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+                               rng=self._train_rng)
+
+    def get_test_tasks(self) -> List[Task]:
+        return self._get_tasks(num_tasks=CFG.num_test_tasks,
+                               rng=self._test_rng)
+
+    @property
+    def predicates(self) -> Set[Predicate]:
+        return {self._On, self._OnTable, self._GripperOpen, self._Holding,
+                self._Clear}
```
ronuchit, line 376: Remember to update all these property methods when you add more predicates/options

ronuchit (commented):
Is there a reason we aren't just subclassing BlocksEnv here and adding on whatever additional stuff? It seems like a lot of the methods and fields are copy-pasted...

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+
+        plt.suptitle(f"Held: {held}, Fingers: {fingers}", fontsize=36)
+        plt.tight_layout()
+        img = utils.fig2data(fig)
+
+        return [img]
+
+    def _get_tasks(self, num_tasks: int, rng: np.random.Generator
+                   ) -> List[Task]:
+        tasks = []
```
tomsilver, line 497: can you add a comment indicating what varies between tasks?

tomsilver (commented):
great work! this is quite the tour de force :)

haha oops I see rohan and I were reviewing at the same time. I agree with his suggestion to subclass what we can, unless you foresee any issues there. my other main suggestion for refactoring the simulate method is: let's first update the position of the robot at the beginning of the function, but don't return yet. then, based on the updated position, do the logic for manipulating the blocks/doors/dials. would that work?

src/approaches/interactive_learning_approach.py
amburger66: hm these changes should already be in master -- i thought i had removed these commits from this branch but i guess not, will fix that

src/envs/playroom.py
amburger66: yeah, will work on those (and operators and options) next!

src/envs/playroom.py
```diff
@@ -0,0 +1,845 @@
...
+    def types(self) -> Set[Type]:
+        return {self._block_type, self._robot_type}
+
+    @property
+    def options(self) -> Set[ParameterizedOption]:
+        return {self._Pick, self._Stack, self._PutOnTable}
+
+    @property
+    def action_space(self) -> Box:
+        # dimensions: [x, y, z, rotation, fingers]
```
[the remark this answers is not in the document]
amburger66: that is true... i tried to explain it in the comment. would it be easier if i just changed it to either radians or degrees?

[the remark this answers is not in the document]
tomsilver: the comment looks good thanks! I think this is fine.

src/structs.py
```diff
@@ -157,6 +157,17 @@ def _copy_state_value(self, val: Any) -> Any:
         assert hasattr(val, "copy")
         return val.copy()
 
+    def allclose(self, other: State) -> bool:
```
amburger66, line 160: I had to grab this from master since my branch didn't have it yet...

src/envs/playroom.py
```diff
@@ -0,0 +1,610 @@
...
+            x = rng.uniform(self.table_x_lb, self.table_x_ub)
+            y = rng.uniform(self.table_y_lb, self.table_y_ub)
+            if self._table_xy_is_clear(x, y, existing_xys):
+                return (x, y)
+
+    def _On_holds(self, state: State, objects: Sequence[Object]) -> bool:
+        block1, block2 = objects
+        if state.get(block1, "held") >= self.held_tol or \
+           state.get(block2, "held") >= self.held_tol:
+            return False
```
amburger66, line 455: this is the only line missing from test coverage, but i'm not sure how to deal with it??

amburger66:
aside from one line missing from test coverage in `playroom.py`, all the other unit test/lint failures should be resolved when i merge in master (this branch started in a bit of a weird state oops)

src/structs.py
ronuchit: You'll have to rebase onto master before we can merge anyway, do you see the "Resolve conflicts" button at the bottom of this PR?

amburger66: I did that, now it's telling me to merge the latest from master into this branch. should I do that?

ronuchit: yeah, you can always do that, this is your branch ;) it's just making sure your changes are applied on top of the most recent master

the latest from master would be bringing in my PR which you reviewed (#107)

ronuchit: oh and so that means you won't need this allclose thing anymore

src/envs/blocks.py
```diff
@@ -376,7 +376,8 @@ def _OnTable_holds(state: State, objects: Sequence[Object]) -> bool:
             (desired_z-cls.pick_tol < z < desired_z+cls.pick_tol)
 
     @staticmethod
-    def _GripperOpen_holds(state: State, objects: Sequence[Object]) -> bool:
+    def _GripperOpen_holds(state: State, objects: Sequence[Object]
+                           ) -> bool:
```
tomsilver, line 380: this is just a new line right? might as well undo, to reduce the number of files involved in the PR

src/envs/playroom.py
```diff
@@ -0,0 +1,631 @@
...
+        self._door2 = Object("door2", self._door_type)
+        self._door3 = Object("door3", self._door_type)
+        self._door4 = Object("door4", self._door_type)
+        self._door5 = Object("door5", self._door_type)
+        self._door6 = Object("door6", self._door_type)
+        self._doors = (self._door1, self._door2, self._door3,
+                       self._door4, self._door5, self._door6)
+        self._dial = Object("dial", self._dial_type)
+
+    def simulate(self, state: State, action: Action) -> State:
```
tomsilver, line 98: wow this got so short, yay!

src/envs/playroom.py
```diff
@@ -0,0 +1,631 @@
...
+        assert self.action_space.contains(action.arr)
+        x, y, z, _, fingers = action.arr
+        # Update robot position
+        if not self._is_valid_loc(action):
```
tomsilver, line 102: recommend changing this so that the input is the location only. I was initially confused because `action` isn't just a "loc".

src/envs/playroom.py
```diff
@@ -0,0 +1,631 @@
...
+        if (dial_x-self.dial_button_tol < x < dial_x+self.dial_button_tol) \
+            and (dial_y-self.dial_button_tol < y
+                    < dial_y+self.dial_button_tol) \
+            and fingers >= self.open_fingers \
+            and self._robot_is_facing_dial(state, action):
+            return self._transition_dial(state)
+
+        return state.copy()
+
+    def _transition_pick(self, state: State, action: Action) -> State:
```
tomsilver, line 143: for this function and the others, I'm wondering if they're identical to the parent class? if so, remove; if not, please add a comment explaining the difference, or, if it involves a small change that would make sense in the parent class too, make the change there.

src/envs/playroom.py
```diff
@@ -0,0 +1,631 @@
...
+    #     robot, dial = objects
+    #     x, y = state.get(robot, "pose_x"), state.get(robot, "pose_y")
+    #     dial_x, dial_y = state.get(dial, "pose_x"), state.get(dial, "pose_y")
+    #     cls = PlayroomEnv
+    #     return (dial_x-cls.dial_r-cls.dial_tol < x
+    #             < dial_x+cls.dial_r+cls.dial_tol) and \
+    #            (dial_y-cls.dial_r-cls.dial_tol < y
+    #             < dial_y+cls.dial_r+cls.dial_tol)
+
+    def _Pick_policy(self, state: State, objects: Sequence[Object],
```
tomsilver, line 552: I'm mostly ignoring this stuff since it'll be for the next PR when you add predicates and options, but I'm also wondering if we can inherit more here

tests/envs/test_playroom.py
```diff
@@ -0,0 +1,297 @@
+"""Test cases for the boring room vs. playroom environment.
+"""
+
+import numpy as np
+from predicators.src.envs import PlayroomEnv
+from predicators.src import utils
+from predicators.src.structs import Action
+
+
```
tomsilver, line 9: absolutely beautiful tests! I'll suggest one more. see here: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/envs/test_cover.py#L170 the idea is to write down a sequence of actions that you think will do all the interesting things in the task. what I like about this is the `make_video` stuff at the end; so you can temporarily set `make_video = True`, run `pytest -s tests/envs/test_playroom.py`, and it should make a video that plays through your actions. The main thing here is to sanity check the video.

tomsilver (commented):
@amburger66 fantastic stuff! I left one medium size comment (about the transition function inheritance) and a few tiny comments. I think we're very close to done (pending also the merge with Rohan's massive PR...)

src/envs/playroom.py
amburger66: the only difference is the action space...I could refactor it by having the functions take in, e.g., `x`, `y`, `z`, and `fingers` instead of `action`? for a couple of the other functions, the only difference is `CFG.blocks_block_size` vs. `CFG.playroom_block_size`, so could I change that to a class variable `block_size` instead?

tomsilver: yes, both of those changes sound great. thank you!!

src/envs/playroom.py
```diff
@@ -0,0 +1,507 @@
...
+                return True
+        return False
+
+    def _get_door_next_to(self, state: State) -> Object:
+        # cannot be next to multiple doors at once
+        for door in self._doors:
+            if self._NextToDoor_holds(state, (self._robot, door)):
+                return door
+        # usage should ensure this is never reached
+        raise RuntimeError("Robot not next to any door")  # pragma: no cover
```
ronuchit, line 488: Is it hard to write a test that reaches this line, just by setting up an impossible state and calling this function directly?

src/envs/playroom.py
```diff
@@ -0,0 +1,507 @@
...
+        ax.add_patch(robot_arrow)
+
+        plt.suptitle(f"Held: {held}, Fingers: {fingers}", fontsize=36)
+        plt.tight_layout()
+        img = utils.fig2data(fig)
+
+        return [img]
+
+    def _get_tasks(self, num_tasks: int, possible_num_blocks: List[int],
+                   rng: np.random.Generator) -> List[Task]:
```
ronuchit, line 297: What is different from the superclass here?

ronuchit (approved):
LGTM but some minor comments

src/envs/playroom.py
amburger66: they're the same right now but this is to remind me to change it later

src/envs/playroom.py
ronuchit: e.g.
env = PlayroomEnv()
state = env.get_train_tasks()[0].init
(change state in some way)
with pytest.raises(RuntimeError):
  env._get_door_next_to(state)  # pylint: disable=protected-access

PR #103, opened 2021-11-15
options and operators for behavior

tomsilver:
* oracle operators for behavior
* navigate option glued up and runs with a single temporary test (see "TEMPORARY" in the behavior env)

TODO next:
1. integrate the other options and perform similar temporary tests. shouldn't be that bad, hopefully.
2. try out the oracle


src/structs.py
```diff
@@ -734,7 +747,7 @@ def _str(self) -> str:
     Delete Effects: {sorted(self.delete_effects, key=str)}
     Side Predicates: {sorted(self.side_predicates, key=str)}
     Option: {self.option}
-    Option Objects: {self.option_objs}"""
+    Option Objects: {self.option_objs}\n"""
```
tomsilver, line 750: !

NishanthJKumar: aha! will fix this and try!

.predicators_pylintrc
```diff
@@ -421,4 +421,4 @@ known-third-party=enchant
 
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception"
-overgeneral-exceptions=Exception
+overgeneral-exceptions=Exception
```
tomsilver, line 424: add back new line? unless this style is preferred by an auto-formatter, let me know if so

mypy.ini
```diff
@@ -31,4 +31,4 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 [mypy-pandas.*]
-ignore_missing_imports = True
+ignore_missing_imports = True
```
tomsilver, line 34: add back new line? unless this style is preferred by an auto-formatter, let me know if so

src/approaches/oracle_approach.py
```diff
@@ -7,17 +7,24 @@
...
+    grasp_obj_param_sampler,
+    place_ontop_obj_pos_sampler,
+)
+from predicators.src.envs import get_env_instance
 
 
 class OracleApproach(TAMPApproach):
-    """A TAMP approach that uses hand-specified NSRTs.
-    """
+    """A TAMP approach that uses hand-specified NSRTs."""
```
tomsilver, line 26: we've (tried to) used the convention that triple quotes are always on the second line -- if you feel strongly that it should be this way, let us know and we can change it everywhere in a separate PR. this comment applies elsewhere in this PR too

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+
+
+def _get_behavior_gt_nsrts() -> Set[NSRT]: # pragma: no cover
+    """Create ground truth nsrts for BehaviorEnv."""
+    env = get_env_instance("behavior")
+
+    type_name_to_type = {t.name: t for t in env.types}
+    pred_name_to_pred = {p.name: p for p in env.predicates}
+
+    def _get_lifted_atom(
```
tomsilver, line 1022: suggest refactor to use `_get_predicate` below inside, to remove redundant lines

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
     nsrts.add(grasp_nsrt)
 
     return nsrts
+
+
+def _get_behavior_gt_nsrts() -> Set[NSRT]: # pragma: no cover
+    """Create ground truth nsrts for BehaviorEnv."""
```
tomsilver, line 1016: can you add a comment somewhere here explaining the type combination stuff, and how this stems from our lack of hierarchical type support?

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+        return LiftedAtom(pred, objects)
+
+    def _get_predicate(
+        base_pred_name: str, types: Sequence[Type]
+    ) -> Predicate:
+        type_names = "-".join(t.name for t in types)
+        pred_name = f"{base_pred_name}-{type_names}"
+        return pred_name_to_pred[pred_name]
+
+    # We start by creating nextTo predicates for all possible type
```
tomsilver, line 1037: minor: end comments with periods, and `nextTo -> NextTo`

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+    # combinations
```
tomsilver, line 1038: explain in the comment here that these predicates are used for side predicates

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+    nextto_predicates = set()
+    for next_to_pred_types in itertools.product(env.types, env.types):
```
tomsilver, line 1040: minor: let's be consistent about `next_to` vs `nextto` in variable names

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+                f"{option.name}-{next(op_name_count)}",
+                parameters,
+                preconditions,
+                add_effects,
+                delete_effects,
+                nextto_predicates,
+                option,
+                option_vars,
+                lambda s, r, o: navigate_to_param_sampler(r,\
+                    [env.object_to_ig_object(o_i) for o_i in o], # type: ignore
```
tomsilver, line 1079: why is the type: ignore needed here?

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+            # Pick from ontop something
+            for surf_obj_type in sorted(env.types):
+                surf_obj = Variable("?surf", surf_obj_type)
+                parameters = [target_obj, agent_obj, surf_obj]
+                option_vars = [target_obj]
+                handempty = _get_lifted_atom("handempty", [])
+                targ_next_to = _get_lifted_atom(
+                    "reachable", [target_obj, agent_obj]
+                )
+                # targ_graspable = _get_lifted_atom(
```
tomsilver, line 1131: remove commented out code

src/approaches/oracle_approach.py
```diff
@@ -1001,3 +1010,183 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+                  f"{option.name}-{next(op_name_count)}",
+                  parameters,
+                  preconditions,
+                  add_effects,
+                  delete_effects,
+                  set(),
+                  option,
+                  option_vars,
+                  lambda s, r, o: place_ontop_obj_pos_sampler(  # type: ignore
+                    env.behavior_env,  # type: ignore
```
tomsilver, line 1182: instead of passing in `env.behavior_env` here, if you need access to an arbitrary behavior env instance within the place sampler, then can you use `env = get_env_instance("behavior")` within the sampler itself?

src/envs/__init__.py
```diff
@@ -27,27 +31,39 @@
...
-    if name == "painting":
-        return PaintingEnv()
-    if name == "playroom":
-        return PlayroomEnv()
-    if name == "behavior":
-        return BehaviorEnv()  # pragma: no cover
-    if name == "repeated_nextto":
-        return RepeatedNextToEnv()
-    raise NotImplementedError(f"Unknown env: {name}")
+        env = CoverEnv() # type: ignore
```
tomsilver, line 39: yikes! why all these type: ignores?

src/envs/__init__.py
```diff
@@ -27,27 +31,39 @@
...
+    elif name == "repeated_nextto":
+        return RepeatedNextToEnv() # type: ignore
+    else:
+        raise NotImplementedError(f"Unknown env: {name}")
+
+    _MOST_RECENT_ENV_INSTANCE[name] = env
+    return env
+
+
+def get_env_instance(name: str) -> BaseEnv: # pragma: no cover
```
tomsilver, line 65: we should cover this in a unit test. in fact, would you mind splitting off the changes to this file (except for the behavior env lines) and creating a small PR with just that?

src/envs/behavior.py
```diff
@@ -7,59 +7,171 @@
 import os
```
tomsilver, line 7: in the spirit of smaller reviews and faster iteration, I'm going to skip this file for now and review it on my next pass through, after my first round comments have been addressed!

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1104 @@
+"""Hardcoded options for BehaviorEnv.
```
tomsilver, line 1: in the spirit of smaller reviews and faster iteration, I'm going to skip this file for now and review it on my next pass through, after my first round comments have been addressed!

src/main.py
```diff
@@ -43,7 +43,7 @@ def main() -> None:
     """Main entry point for running approaches in environments.
     """
     if not os.path.exists("results/"):
-        os.mkdir("results/")
+        os.mkdir("results/") # pragma: no cover
```
tomsilver, line 46: why is this needed?

src/planning.py
```diff
@@ -225,6 +231,7 @@ def _run_low_level_search(
         option = nsrt.sample_option(state, rng_sampler)
         plan[cur_idx] = option
         if option.initiable(state):
+            print(f"Trying option {option}")
```
tomsilver, line 234: to keep the printouts readable in the other environments, let's remove or comment out this new print and the ones below

src/settings.py
```diff
@@ -57,8 +57,10 @@ class GlobalSettings:
     random_options_max_tries = 100
 
     # SeSamE parameters
+    propagate_failures = True
```
tomsilver, line 60: are these two hyperparameters left over from old code?

src/settings.py
```diff
@@ -57,8 +57,10 @@ class GlobalSettings:
...
+    max_samples_per_step = 10
+    max_num_steps_option_rollout = 1000
```
tomsilver, line 62: can you make this environment-specific? (see below in this file for examples)

src/structs.py
```diff
@@ -169,9 +169,16 @@ def allclose(self, other: State) -> bool:
             return False
         for obj in self.data:
             if not np.allclose(self.data[obj], other.data[obj], atol=1e-3):
+                # import ipdb; ipdb.set_trace()
```
tomsilver, line 172: remove

src/structs.py
```diff
@@ -169,9 +169,16 @@ def allclose(self, other: State) -> bool:
...
                 return False
         return True
 
+    def scope(self, objects: Collection[Object]) -> State: # pragma: no cover
```
tomsilver, line 176: remove -- we don't need this anymore with side predicates

tomsilver (commented):
awesome stuff! very satisfying to see those green checks :) and overall looking very good!

in addition to the comments I've left so far, please go through and revert all style changes, or move them to a separate PR if you want to keep them. I personally have no qualms with any of the style changes that you made, but I do think we should keep PRs as small as possible, and we'll be able to remove a few files and focus on the meaty changes by reverting the style changes from this one. in particular, look out for:
* added or removed new lines
* changing the multiline import style from backslashes to parentheses
* similar thing but for function arguments
* making single-line docstring triple quotes on the first line instead of the second

the one style change that I think is actually worth keeping (although @ronuchit may still insist on moving to a different PR) is where you have changed `x+y` to `x + y`, since we do want to use the latter and have failed to in the places you found.

another general comment is, if you're adding any `type: ignore` or `pylint: disable` etc., please add a comment right above explaining why it's needed.

I'll do another pass after these comments are addressed -- in particular, I haven't gone through the biggest files yet. @ronuchit can feel free to wait for me to do mine, or do it in parallel.

.predicators_pylintrc
```diff
@@ -49,7 +49,7 @@ confidence=
...
...tal-literal,import-star-module-level,raw-checker-failed,bad-inline-option,locally-disabled,locally-enabled,file-ignored,suppressed-message,useless-suppression,deprecated-pragma,apply-builtin,basestring-builtin,buffer-builtin,cmp-builtin,coerce-builtin,execfile-builtin,file-builtin,long-builtin,raw_input-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,no-absolute-import,old-division,dict-iter-method,dict-view-method,next-method-called,metaclass-assignment,indexing-exception,raising-string,reload-builtin,oct-method,hex-method,nonzero-method,cmp-method,input-builtin,round-builtin,intern-builtin,unichr-builtin,map-builtin-not-iterating,zip-builtin-not-iterating,range-builtin-not-iterating,filter-builtin-not-iterating,using-cmp-argument,eq-without-hash,div-method,idiv-method,rdiv-method,exception-message-attribute,invalid-str-codec,sys-max-int,bad-python3-import,deprecated-string-function,deprecated-str-translate-call,not-callable,logging-format-interpolation,consider-using-enumerate,invalid-name,eval-used,len-as-condition,raise-missing-from,consider-using-with,cyclic-import,fixme,ungrouped-imports
```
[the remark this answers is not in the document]
NishanthJKumar: The difference is that we added `fixme,ungrouped-imports`. The first (`fixme`) is so that we can leave `TODO` comments in the code and not have the linter fail for that reason. `ungrouped-imports` is because we `try except` imports for the behavior code, and so can't group them in a way that linting will be happy with. Would you like me to add comments mentioning this to the code itself?

.predicators_pylintrc
NishanthJKumar: yup - my bad!

src/approaches/oracle_approach.py
NishanthJKumar: I see - will fix!

.predicators_pylintrc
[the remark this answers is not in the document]
ronuchit: We keep `fixme` out of here intentionally. TODOs generally clutter code, and it's basically a law of physics that when a TODO is merged into master, it will never get addressed. Two resolutions:
1 - fix the TODO in this PR itself (preferred)
2 - open an issue so we can keep track of the things we need to do more easily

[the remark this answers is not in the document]
ronuchit: `ungrouped-imports` is a good check to have in general. Feel free to add `# pylint: disable=ungrouped-imports to the top of the offending file(s)!

src/main.py
NishanthJKumar: without this, we only get 99% coverage

src/settings.py
NishanthJKumar: yup - will remove, good catch!

src/settings.py
NishanthJKumar: should I do this even after deleting the first two params?

src/approaches/oracle_approach.py
NishanthJKumar: I don't need access to an arbitrary behavior env instance, but rather the specific behavior env currently tied to `env`

src/main.py
tomsilver: hmm, but that wasn't true before, so what changed?

src/settings.py
tomsilver: yeah I was only referring to `max_num_steps_option_rollout`, which needs to be larger for behavior than for the other envs

src/approaches/oracle_approach.py
tomsilver: but `env` is just an arbitrary instance in the context of this function. what do you need exactly? if you need it to match up with `s`, then maybe you should use `s.simulator_state`?

src/main.py
NishanthJKumar: good point - I'm not really sure; I just tested this again to verify and pytest complains without it

src/approaches/oracle_approach.py
NishanthJKumar: Sorry, not sure what you mean by this? Could you elaborate/give an example?

tomsilver: sure sorry -- I meant:

```
    def _get_lifted_atom(        
        base_pred_name: str, objects: Sequence[Variable]
    ) -> LiftedAtom:
        pred = _get_predicate(base_pred_name, [o.type for o in objects])
        return LiftedAtom(pred, objects)
```

src/main.py
tomsilver: ah, I think what happens is that `results` doesn't exist on github checks, but it exists for us locally. so I bet if you removed the pragma: no cover here, the checks would pass on github, even though they don't locally. so yeah we can keep this, and @ronuchit said he'll also add it into a PR that he wants to merge soon.

src/envs/__init__.py
NishanthJKumar: yup, will open one!

src/approaches/oracle_approach.py
NishanthJKumar: I understand why we need type combinations here, but unsure how hierarchical types would help? Could you maybe elaborate a bit on this

src/approaches/oracle_approach.py
NishanthJKumar: ah - good point about `env` being arbitrary. I ended up realizing we don't actually need the env, so this is now fixed!

src/approaches/oracle_approach.py
tomsilver: yeah so if we had hierarchical types, we could effectively use the original PDDL operators that you had written, with just one pick operator, one placeontop operator, etc. the way that hierarchical types were used there was that every type inherited from a default object type, so you could have arguments to your PDDL operators that were of type `object`, and then you could use those operators with objects of type `book`, `carrot`, etc. since we don't have support for hierarchical types right now (actually, we might have some support, but let's just finish out this PR and then revisit!), we're hacking around this by one operator for every combination of types, like pick-book, pick-carrot, etc., where each operator has the specific subtypes in its parameters.

src/approaches/oracle_approach.py
tomsilver: ahh I see. okay that's fine, let's just add your answer here as a comment in the code

src/envs/__init__.py
NishanthJKumar: addressed in #248

src/envs/__init__.py
NishanthJKumar: addressed in #248

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+    """Function to detect whether the robot is currently colliding with any
+    object in the scene."""
+    object_in_hand = robot.parts["right_hand"].object_in_hand
+    return (detect_collision(robot.parts["body"].body_id)
+            or detect_collision(robot.parts["left_hand"].body_id)
+            or detect_collision(robot.parts["right_hand"].body_id,
+                                object_in_hand))
+
+
+def get_aabb_volume(lo: np.ndarray, hi: np.ndarray) -> float:
```
ronuchit, line 88: instead of `np.ndarray` for consistency can you use the type `Array`? just `git grep "Array"` to see what I mean, it's defined near the bottom of structs.py

ronuchit: Looks like `np.ndarray` is used a lot in this file, maybe just find-and-replace?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+            [original_orientation[0], original_orientation[1],
+             old_xytheta[2]]))
+    new_pos = np.array([new_xytheta[0], new_xytheta[1], robot_z])
+    new_orn_quat = p.getQuaternionFromEuler(
+        np.array(
+            [original_orientation[0], original_orientation[1],
+             new_xytheta[2]]))
+
+    # Then, simply get the delta position and orientation by multiplying the
+    #  inverse of the old pose by the
```
ronuchit, line 147: extra space?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+    ret_action[0:3] = np.array([delta_pos[0], delta_pos[1], delta_orn[2]])
+
+    return ret_action
+
+
+#################
+
+
+# Navigate To #
+def navigate_to_param_sampler(  # type: ignore
```
ronuchit, line 167: same question, in general i'm not sure what's the motivation for all these `type: ignore`s

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+                        obstacles=get_body_ids(env,
+                                               include_self=True,
+                                               grasping_with_right=True),
+                        rng=rng,
+                    )
+                    p.restoreState(state)
+
+                    # NOTE: This below line is *VERY* important after the
+                    # pybullet state is restored. The hands keep an internal
+                    # track of their state, and if we don't reset their this
```
ronuchit, line 552: typo: "their this"

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+                    # state to mirror the actual pybullet state, the hand will
+                    # think its elsewhere and update incorrectly accordingly
```
ronuchit, line 554: it's*

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+
+    return action
+
+
+def grasp_obj_at_pos(  # type: ignore
+        env,
+        obj,
+        grasp_offset: np.ndarray,
+        rng: Generator = np.random.default_rng(23),
+) -> Union[None, Callable]:
```
ronuchit, line 447: The `Union[None, X]` pattern in mypy is equivalent to the preferable `Optional[X]`

ronuchit: Would it be easy to define the inputs and outputs of Callable here? You can look at how we do that everywhere else `git grep Callable`

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+                    # the hand based on the provided continuous parameters and
+                    # try to create a plan to it.
+                    obj_pos = obj.get_position()
+                    x = obj_pos[0] + grasp_offset[0]
+                    y = obj_pos[1] + grasp_offset[1]
+                    z = obj_pos[2] + grasp_offset[2]
+                    hand_x, hand_y, hand_z = (
+                        env.robots[0].parts["right_hand"].get_position())
+
+                    # # add a little randomness to avoid getting stuck
```
ronuchit, line 479: extra `#`

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+                            closest_point_on_aabb[1] - hand_pos[1],
+                            closest_point_on_aabb[2] - hand_pos[2],
+                        ]
+                        delta_step_to_obj = [
+                            delta_pos / 25.0 for delta_pos in delta_pos_to_obj
+                        ]  # we want to accomplish the motion in 25 timesteps
+
+                        # move the hand along the vector to the object until it
+                        # touches the object
+                        for _ in range(25):
```
ronuchit, line 583: just curious, how important exactly is this 25 parameter to tune correctly? seems like it can't be either too small or too large, else bad things happen?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+                print(f"PRIMITIVE: grasp {obj.name} fail, too far")
+                return None
+            print(f"PRIMITIVE: grasp {obj.name} fail, too big or fixed")
+            return None
+        print(f"PRIMITIVE: grasp {obj.name} fail, no object")
+        return None
+    print(f"PRIMITIVE: grasp {obj.name} fail, agent already has an" +
+          " object in hand!")
+    return None
+
```
ronuchit, line 663: So, in general the code pattern:
```
if not X:
    return None
if not Y:
    return None
if not Z:
    return None
W....
```

is preferable to:
```
if X:
    if Y:
        if Z:
            W
        return None
    return None
return None
```

because the latter is more readable and avoids code from getting over-indented like here. Could you try rewriting this function using this pattern instead?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+                if option_model:
+                    return placeOntopObjectOptionModel
+                return placeOntopObjectOption
+
+            print(f"PRIMITIVE: place {obj_in_hand.name} ontop" +
+                  f"{obj.name} fail, too far")
+            return None
+
+    print("Cannot place; either no object in hand or holding " +
```
ronuchit, line 1007: place option is very complicated lol, i am gonna skip it...

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+        p.removeState(state)
+        print(f"PRIMITIVE: navigate to {obj.name} with params" +
+              f"{pos_offset} failed;" + "birrt failed to sample a plan!")
+        return None
+
+    print("Position commanded is in collision or blocked!")
+    p.restoreState(state)
+    p.removeState(state)
+    print(f"PRIMITIVE: navigate to {obj.name} with params {pos_offset} fail")
+    return None
```
ronuchit, line 369: see my below comment about code pattern, it's also applicable here

ronuchit (changes requested):
whew, lots of good stuff going on here! i unfortunately glossed over most of behavior_options.py, it's very complicated lol. hope the comments help! btw, i'd recommend just splitting off the structs.py / test_structs.py change into its own PR, and then after removing your changes to planning.py, that leaves just 5 files changed (.coveragerc, behavior.py, behavior_options.py, settings.py, oracle_approach.py) which seems good. also if @tomsilver agrees with my comment on settings.py then we should make that a separate PR as well so it's easy to identify any regressions due to it.

src/settings.py
```diff
@@ -179,7 +178,14 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
...
                     "repeated_nextto": 50,
                 })[args["env"]],
-        )
+            max_num_steps_option_rollout=defaultdict(
+                lambda: 100,
+                {
+                    # For the behavior environment, we need
+                    # to execute the options for a larger number of
+                    # timesteps
+                    "behavior": 1000,
```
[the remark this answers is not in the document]
tomsilver: well when we're using the default option models, this parameter will matter right?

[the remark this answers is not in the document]
ronuchit: mmm, only if the option never becomes terminal right? see `utils.option_to_trajectory`. in our other envs options are at most a few steps long. even in behavior too, i think

tomsilver:
well in behavior the option executions can be very long, which is why we
had to make this change. but okay I think you're right, we can change to
1000 default. (I originally suggested this change that Nishanth made here,
sorry Nishanth!)

ronuchit:
Yeah, so if the options are long in behavior but very short in every other env, we can just increase to 1000 across the board.

tests/test_structs.py
```diff
@@ -333,6 +333,9 @@ def _terminal(s, m, o, p):
                                                _policy, _initiable, _terminal)
     assert (repr(parameterized_option) == str(parameterized_option) ==
             "ParameterizedOption(name='Pick', types=[Type(name='type1')])")
+    parameterized_option2 = ParameterizedOption("Pick2", [type1], params_space,
+                                                _policy, _initiable, _terminal)
+    assert parameterized_option2 > parameterized_option
```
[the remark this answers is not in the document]
NishanthJKumar: well I think only `lt` is necessary since if `a > b`, then `b < a` by definition (and I believe Python exploits this); but I'll add the `gt` method for completeness anyways

ronuchit:
Another option could be to just get rid of __gt__ everywhere, it's just in
4 places in structs.py

src/approaches/oracle_approach.py
```diff
@@ -1201,3 +1207,194 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+
+        elif base_option_name == "Grasp":
+            assert len(option_arg_type_names) == 1
+            target_obj_type_name = option_arg_type_names[0]
+            target_obj_type = type_name_to_type[target_obj_type_name]
+            target_obj = Variable("?targ", target_obj_type)
+
+            # Pick from ontop something
+            for surf_obj_type in sorted(env.types):
+                surf_obj = Variable("?surf", surf_obj_type)
```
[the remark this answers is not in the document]
NishanthJKumar: surface

src/approaches/oracle_approach.py
```diff
@@ -1201,3 +1207,194 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+                option_vars = [target_obj]
+                handempty = _get_lifted_atom("handempty", [])
+                targ_nextto = _get_lifted_atom("reachable",
+                                               [target_obj, agent_obj])
+                targ_holding = _get_lifted_atom("holding", [target_obj])
+                preconditions = {handempty, targ_nextto}
+                add_effects = {targ_holding}
+                delete_effects = {handempty}
+                nsrt = NSRT(
+                    f"{option.name}-{next(op_name_count)}",
```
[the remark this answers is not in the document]
NishanthJKumar: yeah; good suggestion!

src/envs/behavior.py
```diff
@@ -1,62 +1,145 @@
...
     from igibson.objects.articulated_object import URDFObject
     from igibson.object_states.on_floor import RoomFloor
     from igibson.robots.behavior_robot import BRBody
-    from igibson.utils.checkpoint_utils import \
-        save_internal_states, load_internal_states
-    from igibson.activity.bddl_backend import SUPPORTED_PREDICATES, \
-        ObjectStateUnaryPredicate, ObjectStateBinaryPredicate
+    from igibson.activity.bddl_backend import SUPPORTED_PREDICATES,\
+        ObjectStateUnaryPredicate,ObjectStateBinaryPredicate
+    from predicators.src.envs.behavior_options import navigate_to_obj_pos,\
```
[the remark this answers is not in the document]
NishanthJKumar: The reason is so that we don't have to make installing igibson and behavior necessary for predicators (if we did, then Ashay and Amber for example would have to install and setup igibson to work on their projects, which have nothing to do with BEHAVIOR atm)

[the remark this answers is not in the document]
ronuchit: I understand why we need the `from igibson` and `from bddl` imports in the `try: except`. This comment was regarding the `from predicators.src.envs.behavior_options import ...` line, which should be able to be moved outside?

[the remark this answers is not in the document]
NishanthJKumar: i see, yes, good point!

src/envs/behavior.py
```diff
@@ -1,62 +1,145 @@
...
 from predicators.src.settings import CFG
 
 
+def get_aabb_volume(lo: np.ndarray, hi: np.ndarray) -> float:
+    """Simple utility function to compute the volume of an aabb."""
+    dimension = hi - lo
+    return dimension[0] * dimension[1] * dimension[2]
+
+
+def make_behavior_option(  # type: ignore
```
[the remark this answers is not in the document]
NishanthJKumar: deprecated reason actually; fixed now!

src/envs/behavior.py
```diff
@@ -1,62 +1,145 @@
...
 
         self._type_name_to_type: Dict[str, Type] = {}
 
         super().__init__()
 
     def simulate(self, state: State, action: Action) -> State:
         assert state.simulator_state is not None
-        load_internal_states(self._env.simulator, state.simulator_state)
+        self.behavior_env.task.reset_scene(state.simulator_state)
         # We can remove this after we're confident in it
```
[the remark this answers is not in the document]
NishanthJKumar: deprecated comment

src/envs/behavior.py
```diff
@@ -226,74 +354,123 @@ def _current_ig_state_to_state(self) -> State:
...
-        bddl_predicate: "bddl.AtomicFormula",
+    def _create_classifier_from_bddl(self,\
+        bddl_predicate: "bddl.AtomicFormula",\
     ) -> Callable[[State, Sequence[Object]], bool]:
 
-        def _classifier(s: State, o: Sequence[Object]) -> bool:
+        def _classifier(_s: State, o: Sequence[Object]) -> bool:
             # Behavior's predicates store the current object states
             # internally and use them to classify groundings of the
             # predicate. Because of this, we will assert that whenever
```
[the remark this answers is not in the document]
NishanthJKumar: yup!

src/envs/behavior.py
```diff
@@ -226,74 +354,123 @@ def _current_ig_state_to_state(self) -> State:
...
             # a predicate classifier is called, the internal simulator
             # state is equal to the state input to the classifier.
-            assert s.allclose(self._current_ig_state_to_state())
+            # assert s.allclose(self._current_ig_state_to_state())
```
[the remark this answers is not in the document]
NishanthJKumar: was removed for debugging a while ago; just put it back in!

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
+"""Hardcoded options for BehaviorEnv."""
+# pylint: disable=import-error,ungrouped-imports
```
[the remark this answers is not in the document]
NishanthJKumar: fixed!

src/envs/behavior_options.py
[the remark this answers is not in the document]
NishanthJKumar: yup, done!

src/envs/behavior_options.py
NishanthJKumar: not particularly; it just dictates the speed at which we move the hand. 25 was my first (and arbitrary) choice.

src/envs/behavior_options.py
```diff
@@ -0,0 +1,1012 @@
...
+        else:
+            if abs(xyz[i] - hi[i]) < abs(xyz[i] - lo[i]):
+                closest_point_on_aabb[i] = hi[i]
+            else:
+                closest_point_on_aabb[i] = lo[i]
+
+    return closest_point_on_aabb
+
+
+def reset_and_release_hand(env) -> None:  # type: ignore
```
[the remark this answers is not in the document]
NishanthJKumar: because the type of env isn't actually a `BehaviorEnv`, but rather an `igibson.envs.behavior_env.BehaviorEnv`, which may not get imported, especially on CI

src/envs/behavior_options.py
NishanthJKumar: See above response on the type of `env`.

src/envs/behavior_options.py
NishanthJKumar: We run into the `BehaviorEnv` import issue again here: the proper type would be `Optional[Callable[[[State, BehaviorEnv], Tuple[Array, bool]]]]`, but `BehaviorEnv` won't get imported on CI because it's an igibson thing...

src/envs/behavior_options.py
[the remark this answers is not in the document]
ronuchit: cool, also keep comments in the code describing this and the other answers you've been giving on github :)

src/utils.py
```diff
@@ -29,6 +29,32 @@
 matplotlib.use("Agg")
 
 
+def get_aabb_volume(lo: Array, hi: Array) -> float:
+    """Simple utility function to compute the volume of an aabb.
+
+    lo refers to the minimum values of the bbox in the x,y and z axes,
```
ronuchit, line 35: add space after comma: `x, y`

ronuchit, line 35: clarify the dimensionality of `lo` and `hi` in the comments

src/utils.py
```diff
@@ -29,6 +29,32 @@
...
+    while hi refers to the highest values
+    """
+    dimension = hi - lo
+    return dimension[0] * dimension[1] * dimension[2]
+
```
ronuchit, line 40: add blank line (hmm, thought yapf handles this...)

src/utils.py
```diff
@@ -29,6 +29,32 @@
...
+def get_closest_point_on_aabb(xyz: List, lo: Array,\
```
ronuchit, line 41: remove backslash so yapf can handle this however it wants

src/utils.py
```diff
@@ -29,6 +29,32 @@
 matplotlib.use("Agg")
 
 
+def get_aabb_volume(lo: Array, hi: Array) -> float:
+    """Simple utility function to compute the volume of an aabb.
+
+    lo refers to the minimum values of the bbox in the x,y and z axes,
+    while hi refers to the highest values
+    """
```
ronuchit, line 37: should we `assert np.all(hi >= lo)`?

src/utils.py
```diff
@@ -29,6 +29,32 @@
...
+    dimension = hi - lo
+    return dimension[0] * dimension[1] * dimension[2]
+
+def get_closest_point_on_aabb(xyz: List, lo: Array,\
+    hi: Array) -> List[float]:
+    """Get the closest point on an aabb from a particular xyz coordinate."""
+    closest_point_on_aabb = [0.0, 0.0, 0.0]
```
ronuchit, line 44: should we `assert np.all(hi >= lo)`?

tests/test_utils.py
```diff
@@ -17,6 +17,21 @@
     _PyperplanHeuristicWrapper
 
 
+def test_aabb_volume():
+    """Tests for get_aabb_volume."""
+    lo = np.array([1.0, 1.5, -1.0])
+    hi = np.array([2.0, 2.5, 0.0])
+    assert utils.get_aabb_volume(lo, hi) == 1.0
```
ronuchit, line 24: add tests for zero volume and negative volume (lo < high, should hit our assert above)

tests/test_utils.py
```diff
@@ -17,6 +17,21 @@
...
+
+
+def test_aabb_closest_point():
+    """Tests for get_closest_point_on_aabb."""
+    xyz = [1.5, 3.0, -2.5]
+    lo = np.array([1.0, 1.5, -1.0])
+    hi = np.array([2.0, 2.5, 0.0])
+    assert utils.get_closest_point_on_aabb(xyz, lo, hi) == [1.5, 2.5, -1.0]
```
ronuchit, line 32: won't this test only hit one code path of your implementation above? add tests to cover all the code paths, and let's split the utils.py changes into a separate PR so that we can verify coverage!

src/structs.py
```diff
@@ -414,10 +414,13 @@ def __hash__(self) -> int:
     def ground(self, objects: Sequence[Object], params: Array) -> _Option:
         """Ground into an Option, given objects and parameter values."""
         assert len(objects) == len(self.types)
+
```
ronuchit, line 417: revert this file

src/approaches/oracle_approach.py
```diff
@@ -6,13 +6,18 @@
 generated at all.
 """
```
ronuchit, line 8: revert changes that aren't relevant to behavior

src/approaches/oracle_approach.py
```diff
@@ -1201,3 +1211,172 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+                nsrt = NSRT(
+                    f"{option.name}-{next(op_name_count_place)}",
+                    parameters,
+                    preconditions,
+                    add_effects,
+                    delete_effects,
+                    set(),
+                    option,
+                    option_vars,
+                    lambda s, r, o:  # type: ignore
```
ronuchit, line 1370: why do we need this type: ignore and the one in the line below?

src/envs/behavior.py
```diff
@@ -1,32 +1,40 @@
 """Behavior (iGibson) environment."""
-# pylint: disable=import-error
+# pylint: disable=import-error,ungrouped-imports
```
ronuchit, line 2: i'm confused, i thought in the previous round of revisions we agreed that we could get rid of disabling the `ungrouped-imports` check?

src/envs/behavior.py
```diff
@@ -36,27 +44,69 @@ def __init__(self) -> None:
...
+        self.behavior_env.task.reset_scene(state.simulator_state)
         a = action.arr
-        self._env.step(a)
+        self.behavior_env.step(a)
+        # a[16] is used to indicate whether to grasp or release the currently-
+        # held object. 1.0 indicates that the object should be grasped, and
+        # -1.0 indicates it should be released
+        if a[16] == 1.0:
+            assisted_grasp_action = np.zeros(28, dtype=float)
+            assisted_grasp_action[26] = 1.0
```
ronuchit, line 95: what is the 26? add more info to comment above

src/envs/behavior.py
```diff
@@ -36,27 +44,69 @@ def __init__(self) -> None:
...
+            _ = (self.behavior_env.robots[0].parts["right_hand"].
+                 handle_assisted_grasping(assisted_grasp_action))
+        elif a[16] == -1.0:
+            released_obj = self.behavior_env.scene.get_objects()[
+                self.behavior_env.robots[0].parts["right_hand"].object_in_hand]
+            # force release object to avoid dealing with stateful AG
```
ronuchit, line 101: write out what AG stands for

src/envs/behavior.py
```diff
@@ -244,56 +308,104 @@ def _classifier(s: State, o: Sequence[Object]) -> bool:
...
+        # _create_classifier_from_bddl
         assert state.allclose(self._current_ig_state_to_state())
         assert len(objs) == 1
         grasped_objs = self._get_grasped_objects(state)
         return objs[0] in grasped_objs
 
-    def _nextto_nothing_classifier(self, state: State,
-                                   objs: Sequence[Object]) -> bool:
-        # Check allclose() here for uniformity with _create_classifier_from_bddl
+    def _nextto_nothing_classifier(self, state: State,\
```
ronuchit, line 397: Why do we still need this, now that we renamed nextto->reachable? What is the difference in the implementation of _reachable_nothing_classifier vs this one?

ronuchit (commented):
behavior_options.py looks very clean now, great stuff! i didn't bother working through it again, though. i left another round of revisions on the other files, but i'd recommend splitting off the utils.py change into a separate PR! and then we can have Tom do the next round of revisions

src/utils.py
NishanthJKumar: yup, great suggestion!

src/utils.py
NishanthJKumar: yeah; I thought so too. I guess not always?

tests/test_utils.py
NishanthJKumar: It actually hits all paths (because the x coord is between lo and hi, the y coord is too big, and the z coord is too small).
I'll add the assertion check though!

src/approaches/oracle_approach.py
NishanthJKumar: turns out we can get rid of it!

src/envs/behavior.py
NishanthJKumar: indeed; idk how it snuck back in here (I must have forgotten to take it out!)

src/envs/behavior.py
NishanthJKumar: good point; this is completely redundant and can be removed!

src/envs/behavior.py
```diff
@@ -244,60 +312,92 @@ def _classifier(s: State, o: Sequence[Object]) -> bool:
...
-            if any(self._env.robots[0].is_grasping(ig_obj)):
+            ig_obj = self.object_to_ig_object(obj)
+
+            # NOTE: The below block is necessary because somehow the body_id
+            # is sometimes a 1-element list...
+            if isinstance(ig_obj.body_id, list):
+                assert len(ig_obj.body_id) == 1
+                ig_obj.body_id = ig_obj.body_id[0]
+
+            if any(self.behavior_env.robots[0].is_grasping(ig_obj.body_id)):
```
tomsilver, line 378: just wondering, why is the `any` necessary? what's the return of `is_grasping`?

src/envs/behavior.py
```diff
@@ -329,7 +429,58 @@ def _bddl_predicate_arity(bddl_predicate: "bddl.AtomicFormula") -> int:
...
     @staticmethod
-    def _create_type_combo_name(original_name: str,
-                                type_combo: Sequence[Type]) -> str:
+    def _create_type_combo_name(original_name: str,\
+        type_combo: Sequence[Type]) -> str:
         type_names = "-".join(t.name for t in type_combo)
         return f"{original_name}-{type_names}"
+
+
+# NOTE: This function must be type ignored because the env argument is untyped.
```
tomsilver, line 438: I think it should be possible to work around this by putting the types in quotes. See `object_to_ig_object(self, obj: Object) -> "ArticulatedObject":`, for example

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+}
+
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def get_body_ids(  # type: ignore
+    env,
+    include_self: bool = False,
+    grasping_with_right: bool = False,
```
tomsilver, line 44: rename to `include_right_hand`?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        ids.append(env.robots[0].parts["body"].get_body_id())
+        ids.append(env.robots[0].parts["eye"].get_body_id())
+        if not grasping_with_right:
+            ids.append(env.robots[0].parts["right_hand"].get_body_id())
+
+    return ids
+
+
+def detect_collision(bodyA: int, object_in_hand: int = None) -> bool:
+    """Detects collisions between objects in the scene (except for the object
```
tomsilver, line 69: can you update this docstring to mention bodyA?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def get_body_ids(  # type: ignore
+    env,
+    include_self: bool = False,
+    grasping_with_right: bool = False,
+) -> List[int]:
+    """Function to return list of body_ids for all objects for collision
```
tomsilver, line 46: why does `detect_collision` not call this function if this function is "for collision checking"?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        ids.append(env.robots[0].parts["left_hand"].get_body_id())
+        ids.append(env.robots[0].parts["body"].get_body_id())
+        ids.append(env.robots[0].parts["eye"].get_body_id())
+        if not grasping_with_right:
+            ids.append(env.robots[0].parts["right_hand"].get_body_id())
+
+    return ids
+
+
+def detect_collision(bodyA: int, object_in_hand: int = None) -> bool:
```
tomsilver, line 68: the type for `object_in_hand` should be `Optional[int]`

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        env.robots[0].parts["right_hand"].set_close_fraction(0)
+        env.robots[0].parts["right_hand"].trigger_fraction = 0
+        p.stepSimulation()
+
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def get_delta_low_level_base_action(  # type: ignore
+    env,
```
tomsilver, line 112: instead of passing env in here, could we just pass `robot_z`?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+    original_orientation: Tuple,
+    old_xytheta: Array,
+    new_xytheta: Array,
+) -> Array:
+    """Get low-level actions from base movement plan."""
```
tomsilver, line 117: this docstring could be improved -- it's unclear what "base movement plan" means

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        env.robots[0].parts["right_hand"].trigger_fraction = 0
+        p.stepSimulation()
+
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def get_delta_low_level_base_action(  # type: ignore
+    env,
+    original_orientation: Tuple,
```
tomsilver, line 113: instead of original orientation, could we pass in the two orientation dimensions that are meant to be unchanged? (they should be unchanged, right?)

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+    # obj is always last in the params list).
+    obj_to_sample_near = objects[-1]
+    closeness_limit = max(
+        [
+            1.5,
+            np.linalg.norm(np.array(\
+                obj_to_sample_near.bounding_box[:2])) + 0.5,  # type: ignore
+        ]
+    )
+    distance = (closeness_limit - 0.01) * rng.random() + 0.03  # type: ignore
```
tomsilver, line 168: can we head off this type: ignore and the one below that uses `distance` by doing `closeness_limit: float = max(...`?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def navigate_to_obj_pos(  # type: ignore
+        env,
+        obj,
+        pos_offset: Array,
+        rng: Generator = np.random.default_rng(23),
```
tomsilver, line 196: generally we want to avoid mutable default arguments. for example, if there are multiple calls to `navigate_to_obj_pos`, they would share a single rng. could we make this kwarg optional and initialize it within the function if it's None?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+) -> Optional[Callable]:
```
tomsilver, line 197: can you make the `Callable` more specific (include inputs and outputs)?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def get_delta_low_level_hand_action(  # type: ignore
+    env,
+    old_pos: Union[Sequence[float], Array],
+    old_orn: Union[Sequence[float], Array],
+    new_pos: Union[Sequence[float], Array],
+    new_orn: Union[Sequence[float], Array],
+) -> Array:
+    """Function to get low level actions from hand-movement plan."""
```
tomsilver, line 375: action*. So the plan is just one step?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
+"""Hardcoded options for BehaviorEnv."""
```
tomsilver, line 1: general comment: I am truly impressed that you figured all of this out, and very grateful that I didn't have to myself 😅

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        print("Position commanded is in collision or blocked!")
+        p.restoreState(state)
+        p.removeState(state)
+        print(
+            f"PRIMITIVE: navigate to {obj.name} with params {pos_offset} fail")
+        return None
+
+    p.restoreState(state)
+    p.removeState(state)
+    state = p.saveState()
```
tomsilver, line 265: what's going on in these 3 lines? why remove the state and then save it again?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+            hand_pos[1] + delta_step_to_obj[1],
+            hand_pos[2] + delta_step_to_obj[2],
+        ]
+        plan.append(new_hand_pos + list(hand_orn))
+        hand_pos = new_hand_pos
+
+    # Setup two booleans to be used as 'memory', as well as
+    # a 'reversed' plan to be used
+    # for our option that's defined below
+    reversed_plan = list(reversed(plan[:]))
```
tomsilver, line 597: I think just `reversed(plan)` will do what you want?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+    def graspObjectOption(_state: State,
+                          env: BehaviorEnv) -> Tuple[Array, bool]:
+        nonlocal plan_executed_forwards
+        nonlocal tried_closing_gripper
+        done_bit = False
+        if (not plan_executed_forwards and not tried_closing_gripper):
+            # Step thru the plan to execute Grasping
+            # phases 1 and 2
+            ret_action = get_delta_low_level_hand_action(
+                env,
```
tomsilver, line 610: is there any hope of passing in some more specific things than env here? it's okay if not

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        rng=rng,
+    )
+    p.restoreState(original_state)
+    p.removeState(original_state)
+
+    # NOTE: This below line is *VERY* important after the
+    # pybullet state is restored. The hands keep an internal
+    # track of their state, and if we don't reset this
+    # state to mirror the actual pybullet state, the hand will
+    # think it's elsewhere and update incorrectly accordingly
```
tomsilver, line 696: what a nightmare! good find!

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+
+    return plan
+
+
+# NOTE: we need a type:ignore because we can't specify the type of obj because
+# it would be dependent on an optional igibson import which won't get imported
+# on CI.
+def place_ontop_obj_pos_sampler(  # type: ignore
+        obj,
+        rng: Generator = np.random.default_rng(23),
```
tomsilver, line 710: same comment here about `rng`, check for any other occurrences

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+    # if return_orn:
+    #     return rnd_params, sampling_results[0][2]
+
+    return rnd_params
+
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def place_ontop_obj_pos(  # type: ignore # pylint: disable=inconsistent-return-statements
```
tomsilver, line 752: whhy is the pylint here necessary?

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+        env,
+        obj,
+        place_rel_pos: Array,
+        place_orn: Optional[Array] = None,
+        option_model: bool = False,
+        rng: Generator = np.random.default_rng(23),
```
tomsilver, line 758: same comment about rng

src/envs/behavior_options.py
```diff
@@ -0,0 +1,976 @@
...
+
+# NOTE: we need a type:ignore because we can't specify the type of env because
+# it would be a igibson.envs.behavior_env.BehaviorEnv, which is an optional
+# import and won't get imported on CI
+def place_ontop_obj_pos(  # type: ignore # pylint: disable=inconsistent-return-statements
+        env,
+        obj,
+        place_rel_pos: Array,
+        place_orn: Optional[Array] = None,
+        option_model: bool = False,
```
tomsilver, line 757: the option model code is not yet ready for prime time, right? can you remove it until it is?

tomsilver (commented):
what a tour de force, great stuff!

src/envs/behavior_options.py
NishanthJKumar: We tried that at first, but that does a shallow copy, which breaks everything because we're popping from `plan` while executing our option forwards

tomsilver:
Interesting — but which part of this does a deep copy?

ronuchit:
Just jumping in here to say, any sort of confusing code like this that Tom
has a question about is a perfect candidate for adding a comment to the
code explaining it clearly :)

Rohan

NishanthJKumar:
@tomsilver the `plan[:]` part does the deep copy

src/envs/behavior.py
NishanthJKumar: it returns an `np.ndarray` by design it seems

src/envs/behavior_options.py
NishanthJKumar: it's used for different collision checkings

src/envs/behavior.py
ronuchit: let's use `np.any` then?

src/envs/behavior_options.py
NishanthJKumar: turns out it isn't!

src/envs/behavior_options.py
NishanthJKumar: that's a relic of old code, thanks for pointing out!

src/envs/behavior_options.py
NishanthJKumar: I think we can, but I'd like to test things before merging that in. I'll split this off into a new PR that we'll open once this is merged; I've opened an issue so we don't lose track of this in the meantime #273

NishanthJKumar (approved):

src/envs/behavior_options.py
tomsilver: just to make it clear that one of the dimensions of orientation is changing and the other two aren't. right now, `original_orientation` has three floats; instead you would pass two

NishanthJKumar: ah gotcha, will do!

ronuchit (approved):
Awesome work!! Looking forward to further results when the option models are in!

Minor: it would be great if you could quickly find-and-replace instances of ",\\" in the PR, instead putting a space: ", \" To find these you can do `git grep ",\\\\"`. I see one instance in `oracle_approach.py` and a bunch in `behavior.py`. Or, just merge and I can open a new PR addressing this :P

PR #106, opened 2021-11-16
change/fix semantics of option initiation and termination

tomsilver:


src/approaches/tamp_approach.py
```diff
@@ -61,3 +52,38 @@ def _get_current_predicates(self) -> Set[Predicate]:
         Defaults to initial predicates.
         """
         return self._initial_predicates
+
+
+def option_plan_to_policy(plan: Sequence[_Option]
+                          ) -> Callable[[State], Action]:
+    """Create a policy that executes the options in order.
+
+    We may want to move this out of here later, but I'm leaving
```
ronuchit, line 61: Let's do this correctly now! We should move this method to utils.py.

We can define a new exception in utils.py, e.g. OptionPlanToPolicyException, and then catch that within TAMPApproach and raise an ApproachFailure with the error message copied over.

ronuchit (changes requested):
Could you also check `option_to_trajectory` in utils.py? And more generally, `git grep "terminal("`

ronuchit:
Is option_to_trajectory in utils.py a special case of your new method? If so, can we unify them?

src/approaches/tamp_approach.py
```diff
@@ -61,3 +52,38 @@ def _get_current_predicates(self) -> Set[Predicate]:
...
+        nonlocal initialized
+        # On the very first state, check initiation condition, and
+        # take the action no matter what.
+        if not initialized:
+            if not queue:
+                raise ApproachFailure("Ran out of options in the plan!")
+            assert queue[0].initiable(state), "Unsound option plan"
+            initialized = True
+        elif queue[0].terminal(state):
+            queue.pop(0)
```
ronuchit, line 84: FYI from my interview prep lol, pop(0) is an O(n) operation, so if we were actually doing this correctly we'd want to reverse the queue, but it doesn't matter since our plans are all miniscule

tomsilver: huh! good to know

src/approaches/tamp_approach.py
tomsilver: FYI there's no variation in the error messages, so I'm not doing the very last thing you said here

ronuchit:
LGTM :) thanks for fixing this!

ronuchit (approved):

PR #108, opened 2021-11-17
Multistep options cover

tomsilver:
This PR should be merged after #106


src/envs/cover.py
```diff
@@ -345,3 +345,372 @@ def __init__(self) -> None:
...
+           grip > self.grasp_thresh:
+            grasp = x - state.get(above_block, "x")
+            next_state.set(above_block, "grasp", grasp)
+        # If we are holding something, place it.
+        # Disallow placing on another block or in free space.
+        # Note: unlike parent env, we also need to check the grip.
+        # Also unlike parent env, we disallow placing except when
+        # below the hand_region_height.
+        if held_block is not None and above_block is None and \
+            grip < self.grasp_thresh and y < self.hand_region_height:
```
ashayathalye, line 490: Don't you also have to make sure the block you are holding is not below the y-axis?

src/envs/cover.py
```diff
@@ -345,3 +345,372 @@ def __init__(self) -> None:
...
+    def _Pick_policy(self, s: State, o: Sequence[Object],  # type: ignore
+                     p: Array) -> Action:
+        # The object is the one we want to pick.
+        assert len(o) == 1
+        obj = o[0]
+        # The parameter is a relative x position to pick at.
+        assert len(p) == 1
+        rel_x = p.item()
+        assert obj.type == self._block_type
+        desired_x = s.get(obj, "x") + rel_x
```
ashayathalye, line 617: Shouldn't the +rel_x be scaled according to the object's width?

src/envs/cover.py
tomsilver: (resolved in conversation)

src/envs/cover.py
```diff
@@ -345,3 +345,372 @@ def __init__(self) -> None:
...
+    controls the gripper "magnet" or "vacuum".
+
+    The state space also needs to change to track x, y, grip.
+
+    The notion of allowed hand regions needs to be adjusted.
+    We will allow the gripper to move anywhere, except when
+    the y dimension is less than a certain threshold, then
+    the x dimension will still need to be within a hand region.
+
+    Currently simulate disallows dropping blocks except when
```
ronuchit, line 365: Is this a current limitation, or an old comment that should be deleted?

src/envs/cover.py
```diff
@@ -345,3 +345,372 @@ def __init__(self) -> None:
...
+            self._targets.append(Object(f"target{i}", self._target_type))
+        self._robot = Object("robby", self._robot_type)
+        # Override the original options to make them multi-step
+        self._Pick = ParameterizedOption(
+            "Pick", types=[self._block_type],
+            params_space=Box(-0.1, 0.1, (1,)),
+            _policy=self._Pick_policy,
+            _initiable=self._Pick_initiable,
+            _terminal=self._Pick_terminal)
+        # Note: there is a change here -- the parameter space is now
```
ronuchit, line 419: Why did you make this change? I'm personally curious, but also you should clarify that in this comment.

src/envs/cover.py
```diff
@@ -345,3 +345,372 @@ def __init__(self) -> None:
...
+        plt.tight_layout()
+        img = utils.fig2data(fig)
+        plt.close()
+        return [img]
+
+    def _create_initial_state(self, rng: np.random.Generator) -> State:
+        # Need to override to account for new low-level state features.
+        # Note: I originally tried to use super(), but ran into issues
+        # because the parent class types and this class types are
+        # actually different (in terms of equality checking).
```
ronuchit, line 577: nice catch.... do we want to change type equality checking to only be based on the type name, not on the features? fine to leave as is too

ronuchit (approved):
Looks reasonable to me, just had some questions!

src/envs/cover.py
tomsilver: this is current. Ashay is working on addressing this limitation, but we'll do that in a separate PR

src/envs/cover.py
tomsilver: I think it's good as is. Without this, I may have not realized that the types are different, and that could've caused problems.

src/envs/cover.py
tomsilver: We made this change because we want to learn options whose parameters correspond to changes in the low-level states of the objects involved in the option's effects. Added to the comment

PR #114, opened 2021-11-18
modifies A* search to make task-planning seeded-deterministic!

NishanthJKumar:
should make A* search at the task-level completely deterministic and reproducible given a particular seed.

### DIFF
--- a/tests/test_structs.py
+++ b/tests/test_structs.py
@@ -385,6 +385,9 @@ def test_nsrts():
     assert isinstance(hash(ground_nsrt), int)
     ground_nsrt2 = nsrt2.ground([cup, plate])
     assert ground_nsrt == ground_nsrt2
+    # Test less than comparison for grounded options
+    assert not ground_nsrt < ground_nsrt2
+    assert not ground_nsrt2 < ground_nsrt
     state = test_state()
     ground_nsrt.sample_option(state, np.random.default_rng(123))
     filtered_nsrt = nsrt.filter_predicates({on})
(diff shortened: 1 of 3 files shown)


ronuchit:
Looks good! Can you add a short test for your __lt__ method? I think it could be as simple as, after this line: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/test_structs.py#L387

adding something like:
```
assert not ground_nsrt < ground_nsrt2
assert not ground_nsrt2 < ground_nsrt
```

(and make sure that your __lt__ method gets used there)

tests/test_structs.py
```diff
@@ -385,6 +385,9 @@ def sampler(s, rng, objs):
     assert isinstance(hash(ground_nsrt), int)
     ground_nsrt2 = nsrt2.ground([cup, plate])
     assert ground_nsrt == ground_nsrt2
+    # Test less than comparison for grounded options
+    assert not ground_nsrt < ground_nsrt2 and not ground_nsrt > ground_nsrt2 # pylint: disable=unneeded-not
```
tomsilver, line 389: strange pylint complaint, but let's see if we can fix it directly... maybe it wants you to do "x >= y" instead of "x < y". could you instead make another ground nsrt that is slightly different from the original and do:
```
assert ground_nsrt < other_ground_nsrt
assert other_ground_nsrt > ground_nsrt
```
thanks!

tomsilver (approved):
LGTM, thank you!

PR #115, opened 2021-11-18
Initial commit of collision checking.

ashayathalye:


src/envs/cover.py
```diff
@@ -435,81 +437,171 @@ def action_space(self) -> Box:
...
+                         state.get(held_block, "height")
+
+        # Ensure neither the gripper nor the possible held block collide with
+        # another block or go below the y-axis.
+        if y + dy < 0 - self.collision_threshold:
+            return state.copy()
+        if held_block is not None:
+            if hy - hh + dy < 0 - self.collision_threshold:
+                return state.copy()
+        p1 = (x, y)
```
tomsilver, line 506: add a comment here previewing what's going on in this chunk

src/envs/cover.py
```diff
@@ -435,81 +437,171 @@ def action_space(self) -> Box:
...
+            ct = self.collision_threshold
+            # These segments defines a slightly smaller rectangle to prevent
+            # floating point arithmetic making us declare a false positive.
+            segments = [
+                ((bx-bw/2+ct, by-ct), (bx+bw/2-ct, by-ct)),  # top
+                ((bx+bw/2-ct, by-ct), (bx+bw/2-ct, by-bh)),  # right
+                ((bx-bw/2+ct, by-ct), (bx-bw/2+ct, by-bh))  # left
+            ]
+            # Only do the check for blocks that are not held.
+            if held_block is None or \
```
tomsilver, line 520: makes sense -- how about we refactor this slightly so that we skip the whole loop over blocks if no object is held? so add before the chunk: `if held_block is not None:`

src/envs/cover.py
```diff
@@ -435,81 +437,171 @@ def action_space(self) -> Box:
...
+        if held_block is not None:
+            if hy - hh + dy < 0 - self.collision_threshold:
+                return state.copy()
+        p1 = (x, y)
+        p2 = (x + dx, y + dy)
+        for block in self._blocks:
+            bx, by = state.get(block, "x"), state.get(block, "y")
+            bw, bh = state.get(block, "width"), state.get(block, "height")
+            ct = self.collision_threshold
+            # These segments defines a slightly smaller rectangle to prevent
```
tomsilver, line 512: nice

src/envs/cover.py
```diff
@@ -435,81 +437,171 @@ def action_space(self) -> Box:
...
+                ((bx-bw/2+ct, by-ct), (bx+bw/2-ct, by-ct)),  # top
+                ((bx+bw/2-ct, by-ct), (bx+bw/2-ct, by-bh)),  # right
+                ((bx-bw/2+ct, by-ct), (bx-bw/2+ct, by-bh))  # left
+            ]
+            # Only do the check for blocks that are not held.
+            if held_block is None or \
+                (held_block is not None and block != held_block):
+                if any(self.intersects(p1, p2, p3, p4) for p3, p4 in segments):
+                    return state.copy()
+            # To check block-block collisions we do: for each of the four
```
tomsilver, line 524: very nice!

src/envs/cover.py
```diff
@@ -435,81 +437,171 @@ def action_space(self) -> Box:
...
+            # vertices of our held block, construct the line segment from its
+            # old location to its new location, and check if this line segment
+            # intersects any segment of the rectangle. Also check if the
+            # segments at the final point intersect any segment of the
+            # rectangle.
+            if held_block is not None and block != held_block:
```
tomsilver, line 530: can also remove the `held_block is None` here after the suggested refactor

src/envs/cover.py
```diff
@@ -366,20 +366,22 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
     over a target region. We should fix this to allow dropping
     blocks anywhere. Also, would be good to check collisions.
```
tomsilver, line 367: please update this comment too, thanks! (also just to check my understanding -- the first two sentences here are still true right? just the third sentence is now obsolete. that's all good, just making sure.)

tomsilver (commented):
@ashayathalye looking very good! left a few small comments. after those and passing the checks, I think we'll be ready to merge, but @ronuchit may want to have a quick look as well.

src/envs/cover.py
```diff
@@ -435,81 +437,171 @@ def action_space(self) -> Box:
         # multi-step options.
         return Box(-0.1, 0.1, (3,))
 
+    @staticmethod
+    def intersects(p1: Tuple[float, float], p2: Tuple[float, float],
```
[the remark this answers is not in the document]
ronuchit: +1, along with some tests added to test_utils.py, these geometric utilities seem very useful!

src/envs/cover.py
ronuchit: the structure `if not X: continue` is more preferable than `if X: <indent everything>`. so the best thing would be, if i'm understanding your logic correctly....
```
if held_block is not None and block == held_block:
    continue
```
and then you don't need either this if statement, or the same one a bit below

src/envs/cover.py
ronuchit: note that under the above refactor, this if statement actually changes to just `if held_block is not None:`

src/envs/cover.py
```diff
@@ -706,11 +805,16 @@ def _get_hand_regions(self, state: State) -> List[Tuple[float, float]]:
 
     @staticmethod
     def _Covers_holds(state: State, objects: Sequence[Object]) -> bool:
-        # Overriding because of the change from "pose" to "x".
+        # Overriding because of the change from "pose" to "x" and because
+        # blocks x position is updated in every step of simulate and not just at
```
ronuchit, line 809: this reads like "blocks cross position" -- do you mean "the block's x-position"?

src/envs/cover.py
ashayathalye: @tomsilver the robot can still collide with a block, so looping through all blocks in the outer loop has to happen?

tomsilver: hmm I see, I didn't realize that you were checking robot-block collisions, I thought only block-block. can you point me to where the robot-block collision checks will happen?

PR #117, opened 2021-11-18
fairly large refactor of the learning pipeline in prep for option learning

tomsilver:
this should finish the open items here:
https://docs.google.com/document/d/1V6scrdAKBESPksjYBTNZC-vajfWRPx5h36CgAh0m40I/edit


src/approaches/iterative_invention_approach.py
```diff
@@ -35,16 +34,25 @@ def _get_current_predicates(self) -> Set[Predicate]:
...
     def learn_from_offline_dataset(self, dataset: Dataset) -> None:
-        # Use the current predicates to generate transitions.
-        transitions_by_option = generate_transitions(
-            dataset, self._get_current_predicates())
+        # Use the current predicates to segment dataset.
+        predicates = self._get_current_predicates()
+        segments = [seg for traj in dataset
+                    for seg in segment_trajectory(traj, predicates)]
+        # If options are known, annotate the segments correctly.
+        if not CFG.do_option_learning:
```
ronuchit, line 42: let's assert that this flag is off, actually...

src/approaches/iterative_invention_approach.py
```diff
@@ -53,104 +61,96 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
             self._learned_predicates.add(neg_new_predicate)
             self._num_inventions += 1
             # Add the new predicate and its negation to all the transitions.
-            for transitions in transitions_by_option.values():
```
ronuchit, line 56: fix unification: for invention it should still include the option args

src/sampler_learning.py
```diff
@@ -85,10 +85,14 @@ def _create_sampler_data(
...
-        for ((state, _, _, option, _, trans_add_effects, trans_delete_effects),
-             obj_to_var) in part_transitions:
-            assert option.parent == param_option
+    for idx, partition in enumerate(partitions):
+        for (segment, obj_to_var) in partition:
+            traj, option, before, after = segment
+            state = traj[0][0]
+            trans_add_effects = after - before
+            trans_delete_effects = before - after
+            if option.parent != param_option:
```
ronuchit, line 94: Why was this assert from before changed to a if not: continue?

tests/test_nsrt_learning.py
```diff
@@ -200,44 +200,3 @@ def test_nsrt_learning_specific_nsrts():
             assert option1.parent.params_space.contains(
                 nsrt.ground([cup0, cup1]).sample_option(
                     state1, np.random.default_rng(123)).params)
-    # The following test checks edge cases of unification with respect to
```
ronuchit, line 203: after the unification fix we can add back this test

ronuchit (commented):
sorry lots of comments lol

src/nsrt_learning.py
```diff
@@ -36,134 +71,185 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
...
+
+def _find_option_for_segment(segment: Segment,
+                             param_option: ParameterizedOption,
+                             opt_vars: Sequence[Variable]) -> _Option:
+    """Figure out which option was executed within the segment.
+
+    At this point, we know which ParameterizedOption was used in the segment,
+    and we know the option_vars, but we don't know what parameters were used.
+    """
+    assert not CFG.do_option_learning, "TODO"
```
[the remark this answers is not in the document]
tomsilver: placeholders for option learning code, coming soon. I added that comment. want to keep them in to make it clear where the option learning code will go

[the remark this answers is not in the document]
ronuchit: i thought there was another TODO which was the placeholder for option learning code. there were two TODOs i saw

src/sampler_learning.py
tomsilver: because before, we had already grouped partitions into options; that's not the case anymore

src/nsrt_learning.py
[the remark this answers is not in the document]
tomsilver: yeah there are two placeholders. the other one is for the main option learning code. this one is: when we're doing option learning, after we've learned the options, we need to go back through the data and associate the learned options to the low-level trajectories. that's necessary for sampler learning.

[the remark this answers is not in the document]
tomsilver: (this could generally be hard, but the way that we're planning to learn options will make this step pretty easy)

src/approaches/iterative_invention_approach.py
```diff
@@ -159,15 +165,15 @@ def _invent_for_nsrt(self, nsrt: NSRT, transitions: List[Transition]
...
         if not data[all_params]["neg"]:
-            print(f"\tNo wrong predictions for NSRT {nsrt.name}")
+            print(f"\tNo wrong predictions for operator {op.name}")
             return None
-        print(f"\tFound a classification problem for NSRT {nsrt.name}")
+        print(f"\tFound a classification problem for operator {op.name}")
         print(f"\t\tData: {len(data[all_params]['pos'])} positives, "
               f"{len(data[all_params]['neg'])} negatives")
-        # For every subset of NSRT parameters, try to fit an MLP classifier.
+        # For every subset of op parameters, try to fit an MLP classifier.
```
ronuchit, line 176: op -> operator

src/nsrt_learning.py
```diff
@@ -19,14 +18,40 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
...
 
+    # Learn strips operators.
+    strips_ops, partitions = learn_strips_operators(segments)
+    assert len(strips_ops) == len(partitions)
+
+    # Learn options, or if known, just look them up.
+    # The order of the options corresponds to the strips_ops.
+    # Each item is a (ParameterizedOption, Sequence[Variable])
+    # with the latter holding the option_vars.
+    option_specs = learn_options(strips_ops, partitions)
```
ronuchit, line 33: In your learning idea, are you clear how the option_vars would fall out? They seem to be tied to the NSRT right?

src/nsrt_learning.py
```diff
@@ -36,134 +61,161 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
...
+            # Make sure the option is consistent within a trajectory.
+            for a in segment.actions:
+                option_a = a.get_option()
+                assert param_option == option_a.parent
+                assert option_vars == [sub[o] for o in option_a.objects]
+        option_specs.append((param_option, option_vars))
+    return option_specs
+
+
+def _find_option_for_segment(segment: Segment,
```
ronuchit, line 196: Let's rename this method since it's modifying. Does it even make sense as its own method? Should we just do these lines where it's being called right now?

src/nsrt_learning.py
```diff
@@ -36,134 +61,161 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
...
+    """Figure out which option was executed within the segment.
+
+    At this point, we know which ParameterizedOption was used in the segment,
+    and we know the option_vars, but we don't know what parameters were used.
+
+    Modifies segment in place.
+    """
+    assert not CFG.do_option_learning, "TODO: implement option learning."
+    assert not segment.has_option()
+    segment.set_option_from_trajectory()
```
ronuchit, line 208: I'm confused, when did the trajectories have options added to them? (In the option learning code path)

ronuchit: In the non-option learning code path, couldn't we just do this in extract_options_from_data?

src/sampler_learning.py
```diff
@@ -5,13 +5,31 @@
...
+    for i, op in enumerate(strips_ops):
+        sampler = learn_sampler(
+            partitions, op.name, op.parameters, op.preconditions,
+            op.add_effects, op.delete_effects, options[i][0], i,
+            do_sampler_learning)
+        samplers.append(sampler)
+    return samplers
+
+
+def learn_sampler(partitions: List[Partition],
```
ronuchit, line 32: Let's make this private?

src/structs.py
```diff
@@ -649,6 +649,141 @@ def unset_option(self) -> None:
...
+        return (option.parent, option_args)
+
+    def add(self, member: Tuple[Segment, ObjToVarSub]) -> None:
+        """Add a new member.
+        """
+        seg, sub = member
+        # Check for consistency.
+        if len(self.members) > 0:
+            # The effects should match.
+            lifted_add_effects = {a.lift(sub) for a in seg.add_effects}
```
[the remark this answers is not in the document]
tomsilver: sorry what do you mean?

src/approaches/iterative_invention_approach.py
tomsilver: op is the name of the variable

src/structs.py
```diff
@@ -649,6 +649,142 @@ def unset_option(self) -> None:
...
+                part_param_option, part_option_args = self.option_spec
+                assert option.parent == part_param_option
+                option_args = [sub[o] for o in option.objects]
+                assert option_args == part_option_args
+        # Add to members.
+        self.members.append(member)
+
+
 # Convenience higher-order types useful throughout the code
 ActionTrajectory = Tuple[List[State], List[Action]]
```
ashayathalye, line 789: I don't know if this is the best name for this. Why not StateActionTrajectory?

ashayathalye (approved):

src/structs.py
tomsilver: @ashayathalye thanks, good suggestion. I'll do this in a separate PR later today.

ronuchit: handled by #121

PR #118, opened 2021-11-19
Poetry refactor

NishanthJKumar:
refactors to switch to poetry for dependency management. Also now uses Poetry for GitHub actions and caches previous installs!


ronuchit (changes requested):
TODO:
make it not change our local workflow
change naming of the checks to match the current ones, if possible

ronuchit:
Thanks, Nishanth! I'll update the required tests so that the names match, and I'll make an announcement to everyone that they should rebase their PRs.

ronuchit (approved):

PR #120, opened 2021-11-20
let options have memory!!!

ronuchit:


NishanthJKumar (commented):
LGTM! I was a bit confused by the new tests at first, but they make a lot of sense now. Great work!  I understand why we need to do this, but one (very simple) thing I'm a bit fuzzy on is what exactly we should put into the memory dict for creating parameterized options for behavior?

tomsilver (approved):
looks perfect, thank you!

PR #124, opened 2021-11-22
small refactor to abstract before segmenting 

tomsilver:
this will make life easier when implementing the approaches described here:
https://docs.google.com/document/d/1O7IqkUz6KCIlG7dZy4GJ8D40kVnuksKG6bzLzLeVQxQ/edit

also, the new utility may be useful for @amburger66 (or not, not sure)


ronuchit (approved):
LGTM! I also noticed that with the Segment/Partition change, the "Transition" type in structs.py is no longer used anywhere. Maybe confirm & remove that with this PR?

PR #125, opened 2021-11-22
add gbfs util

tomsilver:
will be useful for new invention approach

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -453,7 +453,7 @@ def _run_heuristic_search(
 
     Depending on get_priority, can implement A*, GBFS, or UCS.
 
-    If no goal is found, returns the node with the best priority.
+    If no goal is found, returns the state with the best priority.
     """
     queue: List[Tuple[Any, int, _HeuristicSearchNode[_S, _A]]] = []
     state_to_best_path_cost: Dict[_S, float] = \
--- a/src/utils.py
+++ b/src/utils.py
@@ -19,7 +19,7 @@ from predicators.src.args import create_arg_parser
 from predicators.src.structs import _Option, State, Predicate, GroundAtom, \
     Object, Type, NSRT, _GroundNSRT, Action, Task, StateActionTrajectory, \
     OptionTrajectory, LiftedAtom, Image, Video, Variable, PyperplanFacts, \
-    ObjToVarSub, VarToObjSub, Dataset, GroundAtomTrajectory
+    ObjToVarSub, VarToObjSub
 from predicators.src.settings import CFG, GlobalSettings
 matplotlib.use("Agg")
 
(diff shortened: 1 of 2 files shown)


src/utils.py
```diff
@@ -386,6 +386,107 @@ def powerset(seq: Sequence, exclude_empty: bool) -> Iterator[Sequence]:
...
+        # If the goal holds, return.
+        if check_goal(node.state):
+            return _finish_plan(node)
+        # Add node state to explored set.
+        explored.add(node.state)
+        num_expansions += 1
+        # Generate successors.
+        for action, child_state, cost in get_successors(node.state):
+            # If the state is already in explored or frontier, don't bother.
+            if child_state in frontier | explored:
```
ronuchit, line 440: Why do we need both a frontier and an explored? We should be able to just add to explored here when adding a new node, and take out `# Add node state to explored set.` above? Plus the frontier is weird anyway, right, since nothing is being removed from it? (Maybe I'm misunderstanding what the frontier represents.)

src/utils.py
```diff
@@ -386,6 +386,107 @@ def powerset(seq: Sequence, exclude_empty: bool) -> Iterator[Sequence]:
...
+                continue
+            # Add new node
```
ronuchit, line 442: nit: add a period

src/utils.py
```diff
@@ -386,6 +386,107 @@ def powerset(seq: Sequence, exclude_empty: bool) -> Iterator[Sequence]:
...
+    """Helper for _run_heuristic_search.
+    """
+    rev_state_sequence: List[_S] = []
+    rev_action_sequence: List[_A] = []
+
+    while node.parent is not None:
+        action = cast(_A, node.action)
+        rev_action_sequence.append(action)
+        rev_state_sequence.append(node.state)
+        node = node.parent
```
ronuchit, line 471: shouldn't the returned state_sequence be of length 1 more than the returned action_sequence?

tomsilver: oops that was a dumb one

src/utils.py
tomsilver: yeah it's a long and boring story why I had this. I removed and hopefully fixed

tomsilver:
@ronuchit thanks, can you take another quick look?

src/utils.py
```diff
@@ -429,6 +429,109 @@ def powerset(seq: Sequence, exclude_empty: bool) -> Iterator[Sequence]:
...
+    check_goal: Callable[[_S], bool],
+    get_successors: Callable[[_S], Iterator[Tuple[_A, _S, float]]],
+    get_priority: Callable[[_HeuristicSearchNode[_S, _A]], Any],
+    max_expansions: int = 1000
+    ) -> Tuple[List[_S], List[_A]]:
+    """A generic heuristic search implementation.
+
+    Depending on get_priority, can implement A*, GBFS, or UCS.
+
+    If no goal is found, returns the node with the best priority.
```
ronuchit, line 456: Could you update this comment to describe what's actually returned? It's not the node, since the node gets passed through finish_plan first.

PR #127, opened 2021-11-22
ground strips operators

tomsilver:
also will be useful for invention

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -3,7 +3,7 @@
 
 from __future__ import annotations
 from dataclasses import dataclass, field
-from functools import cached_property
+from functools import cached_property, lru_cache
 from typing import Dict, Iterator, List, Sequence, Callable, Set, Collection, \
     Tuple, Any, cast, FrozenSet, DefaultDict, Optional
 import numpy as np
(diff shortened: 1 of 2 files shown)


src/structs.py
```diff
@@ -445,6 +445,22 @@ def make_nsrt(
                     self.add_effects, self.delete_effects, option,
                     option_vars, sampler)
 
+    @lru_cache(maxsize=None)
+    def ground(self, objects: Tuple[Object]) -> _GroundSTRIPSOperator:
+        """Ground into a _GroundSTRIPSOperator, given objects.
+
+        Insist that objects are tuple for hashing in cache.
```
ronuchit, line 452: Let's assert this in the function via `isinstance(..., tuple)`?

src/structs.py
```diff
@@ -471,6 +487,48 @@ def __eq__(self, other: object) -> bool:
...
+    Add Effects: {sorted(self.add_effects, key=str)}
+    Delete Effects: {sorted(self.delete_effects, key=str)}"""
+
+    @cached_property
+    def _hash(self) -> int:
+        return hash(str(self))
+
+    @property
+    def name(self) -> str:
+        """Name of this ground STRIPSOperator.
```
ronuchit, line 514: Should we name these differently, e.g. (GROUND-...)? It's okay to say no

PR #130, opened 2021-11-23
lazy node expansion in search

tomsilver:


src/utils.py
```diff
@@ -496,6 +497,12 @@ def _run_heuristic_search(
             if priority < best_node_priority:
                 best_node_priority = priority
                 best_node = child_node
+                # Optimization: if we've found a better child, immediately
+                # explore the child without expanding the rest of the children.
+                # Accomplish this by putting the parent node back on the queue.
+                if lazy_expansion:
+                    hq.heappush(queue, (priority, next(tiebreak), node))
+                    break
```
ronuchit, line 505: Why is this the right thing to do rather than putting the child onto the queue?

tomsilver: the child is added to the queue already in the lines above. the reason we need to add the parent back onto the queue is that we're not done expanding it. so if it turns out that the good child is actually a dud, we want to return to the parent and explore its other children

ronuchit (approved):

PR #132, opened 2021-11-24
allow search over predicate sets from small to large

tomsilver:
and a few other minor changes, e.g., to printing

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -268,17 +268,16 @@ def _count_positives_for_ops(strips_ops: List[STRIPSOperator],
         if len(atom_sequence) <= 1:
             continue
         objects = set(states[0])
-        ground_ops = [o for op in strips_ops
-                      for o in utils.all_ground_operators(op, objects)]
-        for s, ns in zip(atom_sequence[:-1], atom_sequence[1:]):
-            for ground_op in ground_ops:
-                if not ground_op.preconditions.issubset(s):
-                    continue
-                if ground_op.add_effects == ns - s and \
-                   ground_op.delete_effects == s - ns:
-                    num_true_positives += 1
-                else:
-                    num_false_positives += 1
+        for strips_op in strips_ops:
+            for ground_op in utils.all_ground_operators(strips_op, objects):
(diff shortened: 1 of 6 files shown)


src/utils.py
```diff
@@ -448,7 +448,8 @@ def _run_heuristic_search(
     check_goal: Callable[[_S], bool],
     get_successors: Callable[[_S], Iterator[Tuple[_A, _S, float]]],
     get_priority: Callable[[_HeuristicSearchNode[_S, _A]], Any],
-    max_expansions: int = 1000,
+    max_expansions: int = 10000000,
```
ronuchit, line 451: Let's switch these to float("inf") as the default values?

src/utils.py
```diff
@@ -531,14 +537,15 @@ def run_gbfs(
     check_goal: Callable[[_S], bool],
     get_successors: Callable[[_S], Iterator[Tuple[_A, _S, float]]],
     heuristic: Callable[[_S], float],
-    max_expansions: int = 1000,
+    max_expansions: int = 10000000,
```
ronuchit, line 540: same change to float("inf")

src/approaches/grammar_search_invention_approach.py
```diff
@@ -160,13 +160,32 @@ def _check_goal(s: FrozenSet[Predicate]) -> bool:
...
+            # Start the search with all of the candidates.
+            init = frozenset(candidates)
+        else:
+            assert CFG.grammar_search_direction == "smalltolarge"
+            # Successively consider larger predicate sets.
+            def _get_successors(s: FrozenSet[Predicate]
+                    ) -> Iterator[Tuple[None, FrozenSet[Predicate], float]]:
+                for predicate in sorted(set(candidates) - s):  # determinism
+                    # Actions not needed. Frozensets for hashing.
+                    yield (None, frozenset(s | {predicate}), 1.)
```
ronuchit, line 180: why are the costs always 1 instead of being the cost of removing/adding that predicate?

ronuchit: i mean: the cost of the predicate being added/removed, as produced by generate()

tomsilver: this is just like a "prior" on the predicate. those costs are factored in via true positives and false positives. is that what you mean?

PR #134, opened 2021-11-24
single feature inequality grammar

tomsilver:


tomsilver:
@amburger66 FYI this PR will have a minor impact on your main file (just renaming `all_predicates` to `initial_predicates` for consistency with other things)

tomsilver:
@ronuchit ready for review thanks

src/approaches/grammar_search_invention_approach.py
```diff
@@ -49,44 +50,109 @@ def _generate(self) -> Iterator[Tuple[Predicate, float]]:
...
+class _SingleFeatureInequalitiesPredicateGrammar(_PredicateGrammar):
+    """Generates features of the form "0.feature >= c" or "0.feature <= c".
+    """
+    def _generate(self) -> Iterator[Tuple[Predicate, float]]:
+        # Get ranges of feature values from data.
+        feature_ranges = self._get_feature_ranges()
+        # 0., 1., 0.5, 0.25, 0.75, 0.125, 0.375, ...
+        constant_generator = itertools.chain([0., 1.],
+            _halving_constant_generator(0., 1.))
+        for c in constant_generator:
```
tomsilver, line 113: I think a better way to do this would be a KDTree?

tomsilver: or something.... /shrug. probably fine as is for now.

src/approaches/grammar_search_invention_approach.py
```diff
@@ -49,44 +50,109 @@ def _generate(self) -> Iterator[Tuple[Predicate, float]]:
...
 class _HoldingDummyPredicateGrammar(_PredicateGrammar):
     """A hardcoded cover-specific grammar.
+
     Good for testing with:
         python src/main.py --env cover --approach grammar_search_invention \
             --seed 0 --excluded_predicates Holding
     """
     def _generate(self) -> Iterator[Tuple[Predicate, float]]:
         # A necessary predicate
-        name = "InventedHolding"
```
ronuchit, line 73: What happened to these names? What are the new names going to be? (just curious)

src/settings.py
```diff
@@ -95,7 +95,8 @@ class GlobalSettings:
     grammar_search_false_pos_weight = 1
     grammar_search_size_weight = 1e-2
     grammar_search_pred_complexity_weight = 1
-    grammar_search_grammar_name = "holding_dummy"
+    # grammar_search_grammar_name = "holding_dummy"
```
ronuchit, line 98: remove?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -49,44 +50,109 @@ def _generate(self) -> Iterator[Tuple[Predicate, float]]:
...
-        block_type = [t for t in self.types if t.name == "block"][0]
-        types = [block_type]
-        classifier = _SingleAttributeGEClassifier(0, "is_block", 0.5)
-        yield (Predicate(name, types, classifier), 1.)
+        classifier = _SingleAttributeCompareClassifier(0, "is_block", 0.5,
+                                                       ge, ">=")
+        yield (Predicate(str(classifier), types, classifier), 1.)
+
+
+def _halving_constant_generator(lo: float, hi: float) -> Iterator[float]:
```
ronuchit, line 93: eventually we might want to move this somewhere, but i'm not sure where. will we want to have more generators eventually? should we make a class for them...? (can leave this for the future)

src/approaches/grammar_search_invention_approach.py
ronuchit: yeah doesn't matter, it's not even clear we'd be using it later

src/approaches/grammar_search_invention_approach.py
```diff
@@ -49,44 +50,109 @@ def _generate(self) -> Iterator[Tuple[Predicate, float]]:
...
+        yield r
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _SingleFeatureInequalitiesPredicateGrammar(_PredicateGrammar):
+    """Generates features of the form "0.feature >= c" or "0.feature <= c".
+    """
+    def _generate(self) -> Iterator[Tuple[Predicate, float]]:
+        # Get ranges of feature values from data.
+        feature_ranges = self._get_feature_ranges()
```
ronuchit, line 109: how often is generate() called? bc this feature_ranges can be expensive, and it could be determined right after we initialize the object, since we just need the dataset

if we envision that forever and ever, generate() will only be called once in the whole pipeline, then it's fine as is

src/approaches/grammar_search_invention_approach.py
ronuchit: (maybe include an example of the name of a generated predicate in the test file? how did the coverage check not complain...)

src/approaches/grammar_search_invention_approach.py
```diff
@@ -49,44 +50,109 @@ def _generate(self) -> Iterator[Tuple[Predicate, float]]:
...
+
+
+def _halving_constant_generator(lo: float, hi: float) -> Iterator[float]:
+    mid = (hi + lo) / 2.
+    yield mid
+    left_gen = _halving_constant_generator(lo, mid)
+    right_gen = _halving_constant_generator(mid, hi)
+    for l, r in zip(left_gen, right_gen):
+        yield l
+        yield r
```
ronuchit, line 100: instead of halving_constant_generator, is there any reason you didn't just do np.linspace? is it because you want to prioritize splitting at half points, then quarter points, etc? if so, then why are 0 and 1 the first elements, shouldn't we start with 0.5?

src/approaches/grammar_search_invention_approach.py
tomsilver: the name is now `str(classifier)`. I'll add an example

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah it's just called once

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah that's why. and yeah I guess starting with 0.5 would make sense, but I couldn't think of a way to easily code that. since this is all likely temporary gonna leave it for now

PR #136, opened 2021-11-25
use options to segment when they are available

### LINKED ISSUE #135: extract options crashes when predicate learning does not find a predicate for a certain object argument
to reproduce: `python src/main.py --env blocks --approach grammar_search_invention --seed 0 --excluded_predicates GripperOpen --grammar_search_max_predicates 0`

example traceback:
```
Traceback (most recent call last):
  File "/Users/tom/phd/predicators/src/main.py", line 102, in <module>
    main()
  File "/Users/tom/phd/predicators/src/main.py", line 67, in main
    approach.learn_from_offline_dataset(dataset)
  File "/Users/tom/phd/predicators/src/approaches/grammar_search_invention_approach.py", line 200, in learn_from_offline_dataset
    self._learn_nsrts(dataset)
  File "/Users/tom/phd/predicators/src/approaches/nsrt_learning_approach.py", line 45, in _learn_nsrts
    self._nsrts = learn_nsrts_from_data(
  File "/Users/tom/phd/predicators/src/nsrt_learning.py", line 37, in learn_nsrts_from_data
    option_specs = learn_option_specs(strips_ops, partitions)
  File "/Users/tom/phd/predicators/src/nsrt_learning.py", line 172, in learn_option_specs
    return _extract_options_from_data(partitions)
  File "/Users/tom/phd/predicators/src/nsrt_learning.py", line 186, in _extract_options_from_data
    option_vars = [sub[o] for o in option.objects]
  File "/Users/tom/phd/predicators/src/nsrt_learning.py", line 186, in <listcomp>
    option_vars = [sub[o] for o in option.objects]
KeyError: robby:robot
```

tomsilver:
closes #135

### DIFF
--- a/src/nsrt_learning.py
+++ b/src/nsrt_learning.py
@@ -39,10 +39,16 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
 
     # Now that options are learned, we can update the segments to include
     # which option is being executed within each segment.
-    for partition, option_spec in zip(partitions, option_specs):
-        for (segment, _) in partition:
-            # Modifies segment in place.
-            _update_segment_from_option_specs(segment, option_spec)
+    # If not option learning, the segments will already have the options
+    # after segmentation, so this step can be skipped.
+    assert not CFG.do_option_learning
+    # Uncomment this and _update_segment_from_option_specs for option learning.
+    # Also, change the assertion above to "if CFG.do_option_learning:" and
+    # indent the following lines.
+    # for partition, option_spec in zip(partitions, option_specs):
+    #     for (segment, _) in partition:
+    #         # Modifies segment in place.
+    #         _update_segment_from_option_specs(segment, option_spec)
 
     # Learn samplers.
     # The order of the samplers also corresponds to strips_ops.
@@ -78,7 +84,26 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
         current_segment_traj[0].append(states[t])
         current_segment_traj[1].append(actions[t])
         switch = all_atoms[t] != all_atoms[t+1]
+        # Segment based on option specs if we are assuming that options are
+        # known. If we do not do this, it can lead to a bug where an option
+        # has object arguments that do not appear in the strips operator
+        # parameters. Note also that we are segmenting based on option specs,
+        # rather than option changes. This distinction is subtle but important.
+        # For example, in Cover, there is just one parameterized option, which
+        # is PickPlace() with no object arguments. If we segmented based on
+        # option changes, then segmentation would break up trajectories into
+        # picks and places. Then, when operator learning, it would appear
+        # that no predicates are necessary to distinguish between picking
+        # and placing, since the option changes and segmention have already
+        # made the distinction. But we want operator learning to use predicates
+        # like Holding, Handempty, etc., because when doing symbolic planning,
+        # we only have predicates, and not the continuous parameters that would
+        # be used to distinguish between a PickPlace that is a pick vs a place.
         if actions[t].has_option() and t < len(actions) - 1:
+            # We don't care about the last option in the traj because there's
+            # no next option after it, so we could never possibly want to
+            # segment due to that option. But we still need to check atoms here,
+            # because the length of atoms is one longer.
             option_t = actions[t].get_option()
             option_t1 = actions[t+1].get_option()
             option_t_spec = (option_t.parent, option_t.objects)
@@ -88,12 +113,13 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
         if switch:
             # Include the final state as both the end of this segment
             # and the start of the next segment.
-            # Include the default option here; replaced during option learning.
             current_segment_traj[0].append(states[t+1])
             if actions[t].has_option():
                 segment = Segment(current_segment_traj, all_atoms[t],
                                   all_atoms[t+1], actions[t].get_option())
             else:
+                # If option learning, include the default option here; replaced
+                # during option learning.
                 segment = Segment(current_segment_traj,
                                   all_atoms[t], all_atoms[t+1])
             segments.append(segment)
@@ -155,9 +181,9 @@ def learn_strips_operators(segments: Sequence[Segment], verbose: bool = True,
             new_partition = Partition([(segment, sub)])
             partitions.append(new_partition)
 
-    # We don't need option vars anymore; we'll recover them later.
-    # The only reason to include them is to make sure that params
-    # include the option vars when options are available.
+    # We don't need option vars anymore; we'll recover them later in
+    # `learn_option_specs`. The only reason to include them is to make sure
+    # that params include the option vars when options are available.
     del option_vars
 
     assert len(params) == len(add_effects) == \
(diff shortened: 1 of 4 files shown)


src/nsrt_learning.py
```diff
@@ -75,13 +77,25 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
     for t in range(len(actions)):
         current_segment_traj[0].append(states[t])
         current_segment_traj[1].append(actions[t])
-        if all_atoms[t] != all_atoms[t+1]:
+        switch = all_atoms[t] != all_atoms[t+1]
+        if actions[t].has_option() and t < len(actions) - 1:
```
ronuchit, line 81: This is fixing an otherwise deep bug. Let's include a detailed comment about it.

ronuchit, line 81: I don't fully understand the second condition here, but tell me if this is right:
We don't care about the last option in the traj because there's no next option after it, so we could never possibly want to segment due to that option? But we still need to check atoms here, because the length of atoms is one longer.

src/nsrt_learning.py
```diff
@@ -75,13 +77,25 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+            option_t = actions[t].get_option()
+            option_t1 = actions[t+1].get_option()
+            option_t_spec = (option_t.parent, option_t.objects)
+            option_t1_spec = (option_t1.parent, option_t1.objects)
+            if option_t_spec != option_t1_spec:
+                switch = True
+        if switch:
             # Include the final state as both the end of this segment
             # and the start of the next segment.
             # Include the default option here; replaced during option learning.
```
ronuchit, line 91: This comment can be updated since it's not true in the non-option-learning code path.

src/nsrt_learning.py
```diff
@@ -95,20 +109,28 @@ def learn_strips_operators(segments: Sequence[Segment], verbose: bool = True,
...
+            segment_option_objs = tuple(segment.get_option().objects)
+        else:
+            segment_option_objs = tuple()
         for i in range(len(partitions)):
             # Try to unify this transition with existing effects.
             # Note that both add and delete effects must unify.
+            part_option_vars = option_vars[i]
             part_add_effects = add_effects[i]
             part_delete_effects = delete_effects[i]
             suc, sub = unify_effects_and_options(
```
ronuchit, line 127: Good thing we forgot to rename this method earlier! Now it's the correct name again, lol

src/nsrt_learning.py
```diff
@@ -118,19 +140,26 @@ def learn_strips_operators(segments: Sequence[Segment], verbose: bool = True,
...
             add_effects.append({atom.lift(sub) for atom
                                 in segment.add_effects})
             delete_effects.append({atom.lift(sub) for atom
                                    in segment.delete_effects})
             new_partition = Partition([(segment, sub)])
             partitions.append(new_partition)
 
+    # We don't need option vars anymore; we'll recover them later.
+    # The only reason to include them is to make sure that params
+    # include the option vars when options are available.
```
ronuchit, line 160: To check my understanding, this will happen in learn_option_specs()? If so, we can mention that in the comment.

src/nsrt_learning.py
```diff
@@ -206,8 +235,9 @@ def _update_segment_from_option_specs(segment: Segment,
     Modifies segment in place.
     """
     assert not CFG.do_option_learning, "TODO: implement option learning."
-    assert not segment.has_option()
-    segment.set_option_from_trajectory()
+    # Note: after option learning is implement, use:
```
ronuchit, line 238: implemented*

src/nsrt_learning.py
```diff
@@ -206,8 +235,9 @@ def _update_segment_from_option_specs(segment: Segment,
...
+    #   segment.set_option_from_trajectory()
```
ronuchit, line 239: This comment confuses me --- won't segment.set_option_from_trajectory crash if there are no options? The whole point of this _update_segment_from_option_specs method should be to add in those options...

src/nsrt_learning.py
```diff
@@ -206,8 +235,9 @@ def _update_segment_from_option_specs(segment: Segment,
     Modifies segment in place.
```
ronuchit, line 235: Update this docstring, e.g. "Figure out which option was executed within the segment." is out-of-date. Make it clear that a modification is happening.

src/nsrt_learning.py
ronuchit: Should we get rid of the default argument values for ground_option_args and lifted_option_args, in the `def unify_effects_and_options`

src/nsrt_learning.py
ronuchit: I actually feel that this whole `_update_segment_from_option_specs` method is useless in the non-option-learning code path. Could we just, in the top-level method, not run this method if CFG.do_option_learning is false?

ronuchit: And now i'm even more confused about why segment.set_option_from_trajectory wouldn't just crash, if we're doing option learning

src/nsrt_learning.py
tomsilver: yes that's right, i'll add that as a comment thanks

src/nsrt_learning.py
tomsilver: you're right, that comment was wrong. I'll remove, and yes, we can skip the method entirely

src/nsrt_learning.py
tomsilver: this isn't out of date? (or it will still be true when option learning). lmk if you think it still needs changing. also it says "Modifies segment in place." in the docstring.

tomsilver:

src/nsrt_learning.py
```diff
@@ -39,10 +39,16 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
 
     # Now that options are learned, we can update the segments to include
     # which option is being executed within each segment.
-    for partition, option_spec in zip(partitions, option_specs):
-        for (segment, _) in partition:
-            # Modifies segment in place.
-            _update_segment_from_option_specs(segment, option_spec)
+    # If not option learning, the segments will already have the options
+    # after segmentation, so this step can be skipped.
+    assert not CFG.do_option_learning
```
ronuchit, line 44: Maybe in the next PR, we should add a dummy option learning that just pulls in the ground truth options? And then we can do this. I'm assuming you kept it this way for testing reasons.

ronuchit: In fact, I don't think we're testing the option learning code path at all right now.

src/nsrt_learning.py
```diff
@@ -75,13 +83,45 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+        # known. If we do not do this, it can lead to a bug where an option
+        # has object arguments that do not appear in the strips operator
+        # parameters. Note also that we are segmenting based on option specs,
+        # rather than option changes. This distinction is subtle but important.
+        # For example, in Cover, there is just one parameterized option, which
+        # is PickPlace() with no object arguments. If we segmented based on
+        # option changes, then segmentation would break up trajectories into
+        # picks and places. Then, when operator learning, it would appear
+        # that no predicates are necessary to distinguish between picking
+        # and placing, since the option changes and segmention have already
```
ronuchit, line 97: segmentation*

src/nsrt_learning.py
```diff
@@ -75,13 +83,45 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+        # made the distinction. But we want operator learning to use predicates
+        # like Holding, Handempty, etc., because when doing symbolic planning,
+        # we only have predicates, and not the continuous parameters that would
+        # be used to distinguish between a PickPlace that is a pick vs a place.
```
ronuchit, line 101: Thank you for the detailed comment :)

src/nsrt_learning.py
```diff
@@ -75,13 +83,45 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+        if actions[t].has_option() and t < len(actions) - 1:
+            # We don't care about the last option in the traj because there's
+            # no next option after it, so we could never possibly want to
+            # segment due to that option. But we still need to check atoms here,
+            # because the length of atoms is one longer.
```
ronuchit, line 106: all_atoms*

ronuchit (approved):
LGTM, just some minor typos

PR #138, opened 2021-11-25
new grammar with forall quantifiers

tomsilver:
should be merged after #136 and #137

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -88,7 +88,7 @@ class _ForallClassifier(_NullaryClassifier):
 
     def _classify_state(self, s: State) -> bool:
         for o in utils.get_object_combinations(set(s), self.body.types,
-                                               allow_duplicates=False):
+                                               allow_duplicates=True):
             if not self.body.holds(s, o):
                 return False
         return True
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -23,7 +23,7 @@ from predicators.src.settings import CFG
 ################################################################################
 
 class _ProgrammaticClassifier(abc.ABC):
-    """A classifier implemented an arbitrary program.
+    """A classifier implemented as an arbitrary program.
     """
     @abc.abstractmethod
     def __call__(self, s: State, o: Sequence[Object]) -> bool:
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -22,7 +22,7 @@ from predicators.src.settings import CFG
 #                          Programmatic classifiers                            #
 ################################################################################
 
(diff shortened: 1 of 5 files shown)


ronuchit:
I'll review this after it's rebased on top of merged-in 136 and 137

tomsilver:
@ronuchit thanks, ready now

src/approaches/grammar_search_invention_approach.py
```diff
@@ -18,6 +18,91 @@
...
+        return self.compare(s.get(obj, self.attribute_name), self.constant)
+
+    def __str__(self) -> str:
+        return f"({self.object_index}.{self.attribute_name}" + \
+               f"{self.compare_str}{self.constant:.3})"
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _ForallClassifier(_NullaryClassifier):
+    """Apply a predicate to all objects.
```
ronuchit, line 85: Would a learned Handempty() predicate basically be, "for all objects, their holding bit in the low-level state is false"?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -18,6 +18,91 @@
 from predicators.src.settings import CFG
 
 
+################################################################################
+#                          Programmatic classifiers                            #
+################################################################################
+
+class _ProgrammaticClassifier(abc.ABC):
```
ronuchit, line 25: Really loving this abstraction / overall code setup! :)

src/approaches/grammar_search_invention_approach.py
```diff
@@ -18,6 +18,91 @@
...
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _ForallClassifier(_NullaryClassifier):
+    """Apply a predicate to all objects.
+    """
+    body: Predicate
+
+    def _classify_state(self, s: State) -> bool:
+        for o in utils.get_object_combinations(set(s), self.body.types,
+                                               allow_duplicates=False):
```
ronuchit, line 91: We've been burned by mis-setting allow_duplicates before. Are we sure we don't want to consider them here?

ronuchit: Actually I'm confused here. Why wouldn't the types list always be empty for a ForAllClassifier, since it's subclassing a NullaryClassifier?

ronuchit (approved):
LGTM, just got some questions for my own understanding

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah

src/approaches/grammar_search_invention_approach.py
tomsilver: it's like `Forall ?x ?y... body(?x, ?y)`, which calls the body predicate's classifier. good question about allowed duplicates... yeah for consistency elsewhere, this should be true. will fix

PR #142, opened 2021-11-27
fix potential issue with segment lengths, add a check

ronuchit:


ronuchit:
I believe that segments should always obey the invariant: len(states) == len(actions) + 1. I've added a check to the post_init of the Segment class to verify this, and fixed what I believe to be a bug in the segment generation. In particular, the previous code was adding on the initial state of each segment twice, resulting in things like:
states: [s0, s0, s1]
actions: [a0]

tomsilver (approved):
LGTM

tomsilver:
@ashayathalye FYI you'll want to pull in this change as soon as it's merged, since it's fixing a bug in the data collection for option learning

Issue #139, opened 2021-11-26
set up ground truth option learning approach


ronuchit:
This will require a minor change to the API in `nsrt_learning.py`. We will create a new file, `option_learning.py`, that has classes implementing `learn_option_specs` and `_update_segment_from_option_specs`, because the implementations of these methods are linked.

PR #143, opened 2021-11-27
refactor option learning strategies into a new file

ronuchit:


ronuchit:
closes #139

src/structs.py
```diff
@@ -841,6 +842,7 @@ def add(self, member: Tuple[Segment, ObjToVarSub]) -> None:
 # Convenience higher-order types useful throughout the code
 StateActionTrajectory = Tuple[List[State], List[Action]]
 OptionTrajectory = Tuple[List[State], List[_Option]]
+OptionSpec = Tuple[ParameterizedOption, List[Variable]]
```
tomsilver, line 845: no action required, but I just want to remark that this is basically an "action template" per definition 1 of LOFT paper https://arxiv.org/pdf/2103.00589.pdf

src/option_learning.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class _OracleOptionLearner(_OptionLearnerBase):
+    """The option learner that just cheats by looking up ground truth options
+    from the environment. Useful for testing.
+    """
+    def learn_option_specs(
+            self, strips_ops: List[STRIPSOperator],
+            partitions: List[Partition]) -> List[OptionSpec]:
+        ### THIS HAS ONLY BEEN TESTED FOR THE COVER AND BLOCKS ENVIRONMENTS!!!
```
tomsilver, line 102: we should add and test with CoversMultistepOptions soon, since it's the only env we have right now with multistep options. do you want to do that in this PR, or wait until the next? I'm okay with waiting, but it may reveal bugs

tomsilver (commented):
looking good, pending changes discussed in slack

src/option_learning.py
ronuchit: as discussed on slack, this is hard. requires reverse engineering the option policies which are complicated

PR #147, opened 2021-11-30
 purge trivial learning approach 

### LINKED ISSUE #146: purge trivial_learning
Now that we have actual learning-based approaches, we can and should remove trivial_learning, which was only put in for ease of testing.

tomsilver:
Closes #146


ronuchit (approved):
Much appreciated!

PR #150, opened 2021-12-01
Playroom options and predicates

amburger66:


src/envs/playroom.py
```diff
@@ -460,22 +560,140 @@ def _PutOnTable_policy(self, state: State, memory: Dict,
...
+    def _OpenDoor_policy(self, state: State, memory: Dict,
+                         objects: Sequence[Object], params: Array) -> Action:
+        del memory  # unused
+        door, = objects
+        door_pose = np.array([state.get(door, "pose_x"),
+                              state.get(door, "pose_y"),
+                              self.door_button_z])
+        arr = np.r_[door_pose+params[:-1], params[-1], 1.0].astype(np.float32)
+        arr = np.clip(arr, self.action_space.low, self.action_space.high)
+        return Action(arr)
```
tomsilver, line 593: if the opendoor and closedoor options are indeed the same, I recommend just writing the code once (call it `ToggleDoor` or something) and use the same init/policy/term in constructing the two different ParameterizedOptions.

src/envs/playroom.py
```diff
@@ -460,22 +560,140 @@ def _PutOnTable_policy(self, state: State, memory: Dict,
...
+    def _TurnOnDial_policy(self, state: State, memory: Dict,
+                           objects: Sequence[Object], params: Array) -> Action:
+        del memory  # unused
+        dial, = objects
+        dial_pose = np.array([state.get(dial, "pose_x"),
+                              state.get(dial, "pose_y"),
+                              self.dial_button_z])
+        arr = np.r_[dial_pose+params[:-1], params[-1], 1.0].astype(np.float32)
+        arr = np.clip(arr, self.action_space.low, self.action_space.high)
+        return Action(arr)
```
tomsilver, line 639: same note as above about toggling

src/envs/playroom.py
```diff
@@ -186,17 +249,18 @@ def types(self) -> Set[Type]:
...
+                self._OpenDoor, self._CloseDoor, self._TurnOnDial,
+                self._TurnOffDial}
 
     @property
     def action_space(self) -> Box:
         # dimensions: [x, y, z, rotation, fingers]
         # x, y, z location for the robot's disembodied hand
-        # robot's heading is the angle (rotation * pi/2) in standard position
-        lowers = np.array([self.x_lb, self.y_lb, 0.0, -2.0, 0.0],
+        # robot's heading is the angle (rotation * pi) in standard position
```
tomsilver, line 260: any particular reason why this is changing? fine by me, just curious

src/envs/playroom.py
```diff
@@ -154,7 +217,7 @@ def _transition_door(self, state: State, door: Object) -> State:
         # opens/closes a door that the robot is next to and facing
         assert door.type == self._door_type
         next_state = state.copy()
-        if state.get(door, "open") < 0.5:
+        if state.get(door, "open") < self.door_open:
```
tomsilver, line 220: good replacement; maybe `self.door_open_thresh` or `self.door_open_tol` instead though

tomsilver (commented):
looks great so far! I would recommend starting with a unit test that executes a handcoded sequence of options, and you can make a video and see if it matches what you want. example: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/envs/test_cover.py#L244

src/envs/playroom.py
amburger66: i kept confusing myself because i thought it was just a factor of pi but then i'd have to remember that it was actually pi/2 :P i think this is more intuitive

tests/envs/test_playroom.py
```diff
@@ -260,23 +262,76 @@ def test_playroom_options():
...
+    Stack = [o for o in env.options if o.name == "Stack"][0]
+    PutOnTable = [o for o in env.options if o.name == "PutOnTable"][0]
+    Move = [o for o in env.options if o.name == "Move"][0]
+    OpenDoor = [o for o in env.options if o.name == "OpenDoor"][0]
+    CloseDoor = [o for o in env.options if o.name == "CloseDoor"][0]
+    TurnOnDial = [o for o in env.options if o.name == "TurnOnDial"][0]
+    TurnOffDial = [o for o in env.options if o.name == "TurnOffDial"][0]
+    option_sequence = [
+        Move.ground([robot], [2.0, 30.0, 0.0]),
+        Move.ground([robot], [2.0, 30.0, 0.0]),
```
amburger66, line 297: this option is here twice because the `option_policy` defined below seems to always "bypass" the first option (since terminal is always true). not sure if I should do something about this?

tomsilver: good point -- the option_policy code that I referred you to is outdated. can you change both this one and the other unit test so that they instead use `utils.option_plan_to_policy`? then repeating the first option should not be necessary. thanks & sorry!

tomsilver (commented):
@amburger66 looking great! one small change requested. i'll leave it up to you whether you want to add more tests for failures now. it would also be fine with me to merge this and then move on to writing the oracle operators in a next PR. let me know what you prefer.

PR #151, opened 2021-12-02
painting environment

### LINKED ISSUE #140: add painting environment
And test grammar predicate learning in it! Depending on data distribution (whether we saw holding in both the shelf and the box, or just one), we should see either HoldingSide and HoldingTop get learned, or just Holding.

ronuchit:
closes #140

### DIFF
--- a/src/envs/painting.py
+++ b/src/envs/painting.py
@@ -1,8 +1,7 @@
 """Painting domain, which allows for two different grasps on an object
 (side or top). Side grasping allows for placing into the shelf, and top
-grasping allows for placing into the box. The box has a lid which
-may need to be opened; this lid is NOT modeled by any of the given
-predicates, but could be modeled by a learned predicate.
+grasping allows for placing into the box. The box has a lid which may
+need to be opened; this lid is NOT modeled by any of the given predicates.
 """
 
 from typing import List, Set, Sequence, Dict, Tuple, Optional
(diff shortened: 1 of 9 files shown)


src/envs/painting.py
```diff
@@ -0,0 +1,561 @@
+"""Painting domain, which allows for two different grasps on an object
+(side or top). Side grasping allows for placing into the shelf, and top
+grasping allows for placing into the box. The box has a lid which
+may need to be opened; this lid is NOT modeled by any of the given
+predicates, but could be modeled by a learned predicate.
```
tomsilver, line 5: but we know not to expect this anytime soon because of the action at a distance issue, so maybe just remove the latter clause

src/option_learning.py
```diff
@@ -101,10 +101,10 @@ def learn_option_specs(
         env = create_env(CFG.env)
         option_specs: List[OptionSpec] = []
         if CFG.env == "cover":
-            assert len(strips_ops) == 2
+            assert len(strips_ops) == 3
```
tomsilver, line 104: what happened here?

tomsilver (commented):
looks good, but let me know if I should look more closely at anything. I'm assuming that most of the logic is just copied from the original env.

src/option_learning.py
ronuchit: i increased the number of replay data from 10 to 500 for all methods, which made covers learn a new operator: placing without covering

PR #154, opened 2021-12-02
pretty print states

tomsilver:


ronuchit (approved):
This is awesome! Thanks a bunch for doing this. Should we make pretty_str a cached_property?

tomsilver:
@ronuchit I think we better not cache it because States are mutable

ronuchit:
> @ronuchit I think we better not cache it because States are mutable

Sounds good, merge at will!

PR #156, opened 2021-12-02
Get rid of poetry

NishanthJKumar:
To make everyone's life easier, I realized we might as well just remove poetry. However, we can still cache things and make our CI fast! This PR incorporates these changes.

Sorry for all the trouble with that poetry business, @ronuchit was right to have stopped with AP Lit ;)


ronuchit (approved):
LGTM, thanks a bunch Nishanth!

PR #157, opened 2021-12-03
fix subtle bug in grammar approach

tomsilver:
I found this bug in investigating for Painting why `IsWet,IsDry` weren't getting selected when they are manually proposed (after ablation). After this fix, they are selected as expected.

When counting true positives and false positives, we should be looking at segments, not individual transitions within segments. The way that this issue was manifesting was that transitions within segments that had no change at the symbolic level were being counted as "true positives" for an operator that has empty effects. That operator was only arising for the empty candidate predicate set, so that empty set was getting a better score than the set containing `IsWet,IsDry`.

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -14,7 +14,8 @@ from predicators.src.approaches import NSRTLearningApproach
 from predicators.src.nsrt_learning import segment_trajectory, \
     learn_strips_operators
 from predicators.src.structs import State, Predicate, ParameterizedOption, \
-    Type, Task, Action, Dataset, Object, GroundAtomTrajectory, STRIPSOperator
+    Type, Task, Action, Dataset, Object, GroundAtomTrajectory, STRIPSOperator, \
+    Segment
 from predicators.src.settings import CFG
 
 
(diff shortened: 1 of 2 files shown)


src/approaches/grammar_search_invention_approach.py
```diff
@@ -363,27 +364,27 @@ def _get_successors(s: FrozenSet[Predicate]
...
+        covered_by_some_op = False
+        for ground_op in ground_ops:
+            if not ground_op.preconditions.issubset(segment.init_atoms):
+                continue
+            if ground_op.add_effects == segment.add_effects and \
+               ground_op.delete_effects == segment.delete_effects:
+                covered_by_some_op = True
+            else:
+                num_false_positives += 1
+        if covered_by_some_op:
```
ronuchit, line 386: Why are we only incrementing once if this segment is covered by any op, rather than incrementing once per op that covers it (as we were doing before)?

ronuchit (approved):
lgtm, small question

src/approaches/grammar_search_invention_approach.py
tomsilver: From a prediction error perspective, I thought this would make more sense. We wouldn't want to give a higher score if we hypothetically had duplicate operators, for example, but the old way would've done that.

PR #158, opened 2021-12-03
fix to nsrt learning unify

tomsilver:
in the case where options are known, we should use the parameterizedoption (and not just the option vars) when unifying in nsrt learning.

### DIFF
--- a/src/nsrt_learning.py
+++ b/src/nsrt_learning.py
@@ -5,7 +5,8 @@ import functools
 from typing import Set, Tuple, List, Sequence, FrozenSet
 from predicators.src.structs import Dataset, STRIPSOperator, NSRT, \
     GroundAtom, LiftedAtom, Variable, Predicate, ObjToVarSub, \
-    StateActionTrajectory, Segment, Partition, Object, GroundAtomTrajectory
+    StateActionTrajectory, Segment, Partition, Object, GroundAtomTrajectory, \
+    DefaultOption, ParameterizedOption
 from predicators.src import utils
 from predicators.src.settings import CFG
 from predicators.src.sampler_learning import learn_samplers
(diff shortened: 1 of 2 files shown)


src/nsrt_learning.py
```diff
@@ -234,13 +243,18 @@ def unify_effects_and_options(
...
         ground_option_args: Tuple[Object, ...],
-        lifted_option_args: Tuple[Variable, ...],
+        lifted_option_args: Tuple[Variable, ...]
 ) -> Tuple[bool, ObjToVarSub]:
     """Wrapper around utils.unify() that handles option arguments, add effects,
     and delete effects. Changes predicate names so that all are treated
     differently by utils.unify().
     """
+    # Can't unify if the parameterized options are different.
+    if ground_param_option != lifted_param_option:
```
ronuchit, line 256: Note, of course, we could directly check this in the loop above. But fine to keep all the unification logic in one place, even if it's trivial as in this case.

tests/test_nsrt_learning.py
```diff
@@ -319,7 +322,26 @@ def test_unify_effects_and_options():
...
         lifted_option_args)
     assert not suc
     assert not sub
+    # The following test is for an edge case where everything is identical
+    # except for the name of the parameterized option. We do not want to
+    # unify in this case.
+    param_option1 = ParameterizedOption(
+        "dummy1", [cup_type], Box(0.1, 1, (1,)), lambda s, m, o, p: Action(p),
+        lambda s, m, o, p: False, lambda s, m, o, p: False)
+    suc, sub = unify_effects_and_options(
```
ronuchit, line 337: To hammer home this test, let's also include a copy-paste of it where you pass in param_option0 twice, and all the other arguments are the same -- in this case, `suc` should be True

ronuchit:
Thank you for catching this!

PR #162, opened 2021-12-04
Remove clear bit from blocks and update predicate grammar accordingly

### LINKED ISSUE #159: remove clear bit from blocks environment
and test grammar invention (will require extending to binary predicates)

tomsilver:
Closes #159

and should be merged after #163

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -123,7 +123,7 @@ class _UnaryFreeForallClassifier(_UnaryClassifier):
     Examples:
         - ForAll ?x. On(?x, ?y)
         - Forall ?y. On(?x, ?y)
-        - ForAll ?z. Between(?x, ?z, ?y)
+        - ForAll ?x, ?y. Between(?x, ?z, ?y)
     """
     body: Predicate  # Must be arity 2 or greater.
     free_variable_idx: int
@@ -138,6 +138,7 @@ class _UnaryFreeForallClassifier(_UnaryClassifier):
                 if i != self.free_variable_idx]
 
     def _classify_object(self, s: State, obj: Object) -> bool:
+        assert obj.type == self.body.types[self.free_variable_idx]
         for o in utils.get_object_combinations(set(s), self._quantified_types):
             o_lst = list(o)
             o_lst.insert(self.free_variable_idx, obj)
@@ -367,11 +368,25 @@ def _create_grammar(grammar_name: str, dataset: Dataset,
         sfi_grammar = _SingleFeatureInequalitiesPredicateGrammar(dataset)
         return _NegationPredicateGrammarWrapper(sfi_grammar)
     if grammar_name == "forall_single_feat_ineqs":
+        # We start with the given predicates because we want to allow
+        # negated and quantified versions of the given predicates, in
+        # addition to negated and quantified versions of new predicates.
         given_grammar = _GivenPredicateGrammar(given_predicates)
         sfi_grammar = _SingleFeatureInequalitiesPredicateGrammar(dataset)
+        # This chained grammar has the effect of enumerating first the
+        # given predicates, then the single feature inequality ones.
         chained_grammar = _ChainPredicateGrammar([given_grammar, sfi_grammar])
+        # For each predicate enumerated by the chained grammar, we also
+        # enumerate the negation of that predicate.
         negated_grammar = _NegationPredicateGrammarWrapper(chained_grammar)
+        # For each predicate enumerated, we also enumerate foralls for
+        # that predicate.
         forall_grammar = _ForallPredicateGrammarWrapper(negated_grammar)
+        # Finally, we don't actually need to enumerate the given predicates
+        # because we already have them in the initial predicate set,
+        # so we just filter them out from actually being enumerated.
+        # But remember that we do want to enumerate their negations
+        # and foralls, which is why they're included originally.
         return _SkipGrammar(forall_grammar, given_predicates)
     raise NotImplementedError(f"Unknown grammar name: {grammar_name}.")
 
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -444,6 +444,9 @@ class GrammarSearchInventionApproach(NSRTLearningApproach):
             op_size = _get_operators_size(strips_ops)
             # Also add a penalty based on predicate complexity.
             pred_complexity = sum(candidates[p] for p in s)
+            # Useful for debugging:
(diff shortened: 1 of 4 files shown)


src/approaches/grammar_search_invention_approach.py
```diff
@@ -103,6 +116,43 @@ def __str__(self) -> str:
...
+class _UnaryFreeForallClassifier(_UnaryClassifier):
+    """Universally quantify all but one variable in a multi-arity predicate.
+
+    Examples:
+        - ForAll ?x. On(?x, ?y)
+        - Forall ?y. On(?x, ?y)
+        - ForAll ?z. Between(?x, ?z, ?y)
+    """
+    body: Predicate  # Must be arity 2 or greater.
+    free_variable_idx: int
```
ronuchit, line 129: If we're universally quantifying "all but one variable" then shouldn't this field actually be nonfree_variable_idx? That would make more sense since this classifier is unary.

src/approaches/grammar_search_invention_approach.py
```diff
@@ -103,6 +116,43 @@ def __str__(self) -> str:
...
+
+    def __post_init__(self) -> None:
+        assert self.body.arity >= 2
+        assert self.free_variable_idx < self.body.arity
+
+    @cached_property
+    def _quantified_types(self) -> List[Type]:
+        return [t for i, t in enumerate(self.body.types)
```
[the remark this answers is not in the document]
ronuchit: This is all related to my general confusion above.

src/approaches/grammar_search_invention_approach.py
```diff
@@ -103,6 +116,43 @@ def __str__(self) -> str:
...
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class _UnaryFreeForallClassifier(_UnaryClassifier):
+    """Universally quantify all but one variable in a multi-arity predicate.
+
+    Examples:
+        - ForAll ?x. On(?x, ?y)
+        - Forall ?y. On(?x, ?y)
+        - ForAll ?z. Between(?x, ?z, ?y)
```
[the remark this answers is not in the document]
ronuchit: Your docstring says "all but one variable" but this example is only quantifying over 1 of the 3 variables.

src/approaches/grammar_search_invention_approach.py
```diff
@@ -236,16 +350,29 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
     if grammar_name == "holding_dummy":
         return _HoldingDummyPredicateGrammar(dataset)
     if grammar_name == "single_feat_ineqs":
-        return _SingleFeatureInequalitiesPredicateGrammar(dataset)
+        sfi_grammar = _SingleFeatureInequalitiesPredicateGrammar(dataset)
+        return _NegationPredicateGrammarWrapper(sfi_grammar)
     if grammar_name == "forall_single_feat_ineqs":
-        base = _SingleFeatureInequalitiesPredicateGrammar(dataset)
-        return _ForallPredicateGrammarWrapper(dataset, base)
+        given_grammar = _GivenPredicateGrammar(given_predicates)
```
ronuchit, line 370: What's the idea here behind proposing given predicates? Aren't they already going to be given to / used by the operator learning always?

ronuchit: Okay now I'm just confused due to the SkipGrammar below going and removing the given_predicates again. This portion needs some comments, I think.

src/approaches/grammar_search_invention_approach.py
```diff
@@ -236,16 +350,29 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
             yield (predicate, cost)
             classifier = _ForallClassifier(predicate)
             yield (Predicate(str(classifier), [], classifier), cost)
+            if predicate.arity >= 2:
```
ronuchit, line 353: When will this ever happen, since we're only considering UnaryPredicates right now?

ronuchit (changes requested):

src/approaches/grammar_search_invention_approach.py
```diff
@@ -103,6 +116,43 @@ def __str__(self) -> str:
...
+    Examples:
+        - ForAll ?x. On(?x, ?y)
+        - Forall ?y. On(?x, ?y)
+        - ForAll ?z. Between(?x, ?z, ?y)
+    """
+    body: Predicate  # Must be arity 2 or greater.
+    free_variable_idx: int
+
+    def __post_init__(self) -> None:
+        assert self.body.arity >= 2
```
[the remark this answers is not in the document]
tomsilver: no -- for example, we could have `ForAll ?x, ?y. Between(?x, ?z, ?y)` (fixed the example in the docstring)

src/approaches/grammar_search_invention_approach.py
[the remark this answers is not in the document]
tomsilver: yes, I'll add an assert

src/approaches/grammar_search_invention_approach.py
tomsilver: "free variable" refers to the variable that is exposed in the new predicate. free is the opposite of quantified

src/approaches/grammar_search_invention_approach.py
tomsilver: The given predicates can have arity 2 (e.g. `On`)..

tomsilver:
@ronuchit thanks, can you take another look?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -273,9 +416,12 @@ def _get_current_predicates(self) -> Set[Predicate]:
     def learn_from_offline_dataset(self, dataset: Dataset) -> None:
         # Generate a candidate set of predicates.
         print("Generating candidate predicates...")
-        grammar = _create_grammar(CFG.grammar_search_grammar_name, dataset)
+        grammar = _create_grammar(CFG.grammar_search_grammar_name, dataset,
+                                  self._initial_predicates)
         candidates = grammar.generate(max_num=CFG.grammar_search_max_predicates)
-        print(f"Done: created {len(candidates)} candidates.")
+        print(f"Done: created {len(candidates)} candidates:")
+        for predicate in candidates:
```
ronuchit, line 423: Won't this be a massive printout?

ronuchit (approved):
Thank you! I got confused between free and quantified. Great to see this working with the clear bit removed! Big step in the right direction.

src/approaches/grammar_search_invention_approach.py
tomsilver: right now it's very manageable because we're not proposing that many. it's been helpful for me but we can remove later if it's too much.

PR #163, opened 2021-12-04
always allow duplicates

tomsilver:
it's confusing to allow duplicates in operator/nsrt parameters but not within predicates. I want to purge `allow_duplicates=False`.

what led me down this path was a weird binary predicate being learned by the grammar approach in blocks. after digging for a long time, I realized that the predicate was basically acting as `neq` to enforce that operator parameters were different. I want to avoid similar misadventures in the future.

`allow_duplicates=True` is also consistent with standard first-order logic semantics. and we can always capture the same behavior by introducing an explicit `neq` predicate.

I verified that oracle and nsrt_learning still work as expected for cover, blocks, and painting.

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -90,8 +90,7 @@ class _ForallClassifier(_NullaryClassifier):
     body: Predicate
 
     def _classify_state(self, s: State) -> bool:
-        for o in utils.get_object_combinations(set(s), self.body.types,
-                                               allow_duplicates=True):
+        for o in utils.get_object_combinations(set(s), self.body.types):
             if not self.body.holds(s, o):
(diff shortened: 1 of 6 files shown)


ronuchit (approved):
LGTM. You may want to `git grep duplicate` to verify the purge. Question: In your PR description, you say that `allow_duplicates=False` is consistent with typical FOL semantics. Does that mean we're now going against these semantics, since your change here is always allowing duplicates?

tomsilver:
> LGTM. You may want to `git grep duplicate` to verify the purge. Question: In your PR description, you say that `allow_duplicates=False` is consistent with typical FOL semantics. Does that mean we're now going against these semantics, since your change here is always allowing duplicates?
Oops sorry, that was a typo -- fixed. And yes I grepped

PR #168, opened 2021-12-07
fix to segmentation leading to change in planning re: empty effects

tomsilver:


src/nsrt_learning.py
```diff
@@ -94,16 +94,25 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+                option_t_spec = (option_t.parent, option_t.objects)
+                option_t1_spec = (option_t1.parent, option_t1.objects)
+                if option_t_spec != option_t1_spec:
+                    switch = True
+            # Special case: assume that trajectory data never interrupts the
+            # final option (we will assert this). Under this assumption, we
+            # can safely segment without using any continuous info. Note that
+            # excluding the final option from the data is highly problematic
+            # when using demo+replay with the default 1 option per replay
+            # because all of the replay data would get excluded (unless there
```
ronuchit, line 111: update this comment to make it more clear -- "because the replay data which causes no change in the symbolic state would get excluded."

ronuchit (approved):
lgtm

PR #169, opened 2021-12-07
minor cleanup: unify trivial initiable/terminal into BaseEnv class

ronuchit:


tomsilver (commented):
good to refactor, but can we put these in utils.py instead? my thinking is that it's nice that our base env right now is a very clean illustration of the environment API we're assuming. if we start adding helper methods to it, that API will get muddied a bit

tomsilver (approved):
lgtm thanks

PR #170, opened 2021-12-07
use option specs in score function for grammar approach

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -360,6 +363,87 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+# # Make sure to uncomment return _DebugGrammar(skip_grammar) in
+# # _create_grammar and the call to _run_debug_analysis.
+# _DEBUG_PREDICATE_STRS = [
+#     "((0:obj).wetness<=0.5)",
+#     "NOT-((0:obj).wetness<=0.5)",
+# ]
+
+# @dataclass(frozen=True, eq=False, repr=False)
+# class _DebugGrammar(_PredicateGrammar):
+#     """A grammar that omits given predicates from being enumerated.
```
ronuchit, line 378: Is this really omitting, or is this docstring erroneously copied from elsewhere?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -514,28 +609,69 @@ def _get_successors(s: FrozenSet[Predicate]
...
+                             ) -> Tuple[int, int,
+                                        List[Set[int]], List[Set[int]]]:
+    """Returns num true positives, num false positives, and for each strips op,
+    lists of segment indices that contribute true or false positives.
     """
+    assert len(strips_ops) == len(option_specs)
     num_true_positives = 0
     num_false_positives = 0
-    for segment in segments:
+    # The following two lists are just useful for debugging with
```
ronuchit, line 622: Clarify this in the docstring as well?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -514,28 +609,69 @@ def _get_successors(s: FrozenSet[Predicate]
...
+            assert len(option_vars) == len(option_objects)
+            option_var_to_obj = dict(zip(option_vars, option_objects))
+            # We want to get all ground operators whose corresponding
+            # substitution is consistent with the option vars for this
+            # segment. So, determine all of the operator variables
+            # that are not in the option vars, and consider all
+            # groundings of them.
+            types = [p.type for p in op.parameters
+                     if p not in option_vars]
+            for choice in utils.get_object_combinations(objects, types):
```
ronuchit, line 647: Would this code be simpler if we did the same thing as before, but then filter out when the option arguments don't match?
```
for op in strips_ops:
    for ground_op in utils.all_ground_operators(op, objects):
        if the option_vars of the ground_op don't match the option_objects:
            continue
```

Just a thought... if we don't do this, then we should move the functionality you've got here into utils.py, maybe into a method called `all_ground_operators_consistent_with_option`

src/approaches/grammar_search_invention_approach.py
```diff
@@ -360,6 +363,87 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
                     yield (uff_predicate, cost)
 
 
+### Useful for debugging ###
+
+# # Replace these strings with anything you want to exclusively enumerate.
```
ronuchit, line 368: `# #` -> `#` in these 3 lines

src/approaches/grammar_search_invention_approach.py
tomsilver: I don't think we want to do the first thing because that would increase the computational complexity. I'll make the utility

src/approaches/grammar_search_invention_approach.py
tomsilver: But then if you block uncomment, you have to re-comment these lines

src/utils.py
```diff
@@ -585,6 +585,32 @@ def all_ground_operators(operator: STRIPSOperator,
     return ground_operators
 
 
+def all_ground_operators_given_partial(operator: STRIPSOperator,
```
ronuchit, line 588: Definitely should add a test or two for this function!

tomsilver:
@ronuchit I decided to split off the large debugging code into a new top level `analysis/` directory, which we can put other analysis scripts into. let me know what you think.

analysis/grammar_search_analysis.py
```diff
@@ -0,0 +1,146 @@
+"""Debugging script for grammar search invention approach.
+"""
+from collections import defaultdict
```
ronuchit, line 3: blank line for consistency w other files

analysis/grammar_search_analysis.py
```diff
@@ -0,0 +1,146 @@
+"""Debugging script for grammar search invention approach.
```
ronuchit, line 1: maybe let's move analysis/ into src/ so that mypy looks at it? we can exclude it from coverage requirements

ronuchit (approved):
LGTM, just some minor things!

analysis/grammar_search_analysis.py
ronuchit: (by adding an exclusion rule for `src/analysis/` somehow)

PR #171, opened 2021-12-08
Clip cover_multistep_options's option policies with limits of action space. 

ashayathalye:
This allows for flexibility in defining variable step policies in cover_multistep_options because the Pick() and Place() ground truth option policies now clip the low-level-action by the bounds of the action space. If we want policies that take more steps to go from point to point, we can reduce the limits of the action space.

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -18,6 +18,9 @@ class GlobalSettings:
     cover_block_widths = [0.1, 0.07]
     cover_target_widths = [0.05, 0.03]
 
+    # cover_multistep_options parameters
+    action_space_limits = [-np.inf, np.inf]
+
     # cluttered table env parameters
     cluttered_table_num_cans_train = 5
     cluttered_table_num_cans_test = 10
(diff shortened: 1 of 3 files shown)


src/settings.py
```diff
@@ -18,6 +18,9 @@ class GlobalSettings:
     cover_block_widths = [0.1, 0.07]
     cover_target_widths = [0.05, 0.03]
 
+    # cover_multistep_options parameters
+    action_space_limits = [-np.inf, np.inf]
```
tomsilver, line 22: can you change this to `cover_multistep_options_action_space_limits`? or if that's too long, `cover_multistep_action_limits`

tomsilver (commented):
one small comment, otherwise lgtm thanks!

ronuchit:
LGTM after tom's suggestion

src/settings.py
ashayathalye: Will do the latter.

PR #173, opened 2021-12-08
Playroom env oracle approach

amburger66:
* added oracle NSRTs for playroom env
* `test_oracle_approach_playroom` doesn't work right now because planning is hard; right now the approach works if all the door start out open


tomsilver:
lgtm, great work! just waiting to see if checks pass after I merged in master.

ronuchit (approved):

PR #174, opened 2021-12-09
repeated NextTo env

### LINKED ISSUE #141: set up simple environment with repeated predicates
We expect an issue with operator learning in the behavior environment is that there are repeated NextTo facts in the state, and so the operator set will blow up. We may want to consider an algorithmic way to resolve this issue, but for starters we will make a simple environment that isolates this issue.

ronuchit:
Toy environment for isolating the problem of a robot being next to potentially differing numbers of things as it moves around an environment. Currently, both planning and learning work only because we have set the threshold for being next to an object to be very low, such that whp the robot is only next to one object at any given time. Increasing this `nextto_thresh = 0.02` threshold would quickly break planning (and learning).

closes #141


tomsilver (approved):
good stuff. tested rendering, oracle, and nsrt_learning locally and all look good.

PR #176, opened 2021-12-09
refactor heuristic function in grammar approach, prep for new ones

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,124 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+    return (segments, strips_ops, option_specs)
+
+
+def _prediction_error_heuristic(initial_predicates: Set[Predicate],
+                                atom_dataset: List[GroundAtomTrajectory],
+                                candidates: Dict[Predicate, float],
+                                s: FrozenSet[Predicate]) -> float:
+    """Score a predicate set by learning operators and counting false positives.
+    """
+    print("Scoring predicates:", s)
```
ronuchit, line 479: This seems a little strange to me. Aren't we redoing a lot of computation on every separate s, even if we don't need to? I think we should have these functions be called e.g. `_make_prediction_error_heuristic` which returns a Callable[_S, float].

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,124 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+        del s  # unused
+        return False
+
+    if CFG.grammar_search_direction == "largetosmall":
+        # Successively consider smaller predicate sets.
+        def _get_successors(s: FrozenSet[Predicate]
+                ) -> Iterator[Tuple[None, FrozenSet[Predicate], float]]:
+            for predicate in sorted(s):  # sorting for determinism
+                # Actions not needed. Frozensets for hashing.
+                # The cost of 1. is irrelevant because we're doing GBFS
```
ronuchit, line 521: change to 1.0

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,124 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+                # and not A* (because we don't care about the path).
+                yield (None, frozenset(s - {predicate}), 1.)
```
ronuchit, line 523: change to 1.0

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,124 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+        # Start the search with all of the candidates.
+        init = frozenset(candidates)
+    else:
+        assert CFG.grammar_search_direction == "smalltolarge"
+        # Successively consider larger predicate sets.
+        def _get_successors(s: FrozenSet[Predicate]
+                ) -> Iterator[Tuple[None, FrozenSet[Predicate], float]]:
+            for predicate in sorted(set(candidates) - s):  # determinism
+                # Actions not needed. Frozensets for hashing.
+                # The cost of 1. is irrelevant because we're doing GBFS
```
ronuchit, line 534: change to 1.0

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,124 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+                # and not A* (because we don't care about the path).
+                yield (None, frozenset(s | {predicate}), 1.)
```
ronuchit, line 536: change to 1.0

src/approaches/grammar_search_invention_approach.py
tomsilver: What do you mean? Of course we need to relearn operators for each new predicate set

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,124 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+        ) -> Set[Predicate]:
+    """Perform a greedy search over predicate sets.
+    """
+
+    # There are no goal states for this search; run until exhausted.
+    def _check_goal(s: FrozenSet[Predicate]) -> bool:
+        del s  # unused
+        return False
+
+    if CFG.grammar_search_direction == "largetosmall":
```
ronuchit, line 515: Could we actually just combine these two similar get_successors into one? We can just assign a different lambda function that produces the target we're gonna iterate over (`for predicate in ...`). So like:
```
if CFG.grammar_search_direction == "largetosmall":
    predicates_fn = lambda s: sorted(s)
    init = frozenset(candidates)
else:
    assert CFG.grammar_search_direction == "smalltolarge"
    predicates_fn = lambda s: sorted(set(candidates) - s)
    init = frozenset()
```

and then you only have to define one `_get_successors`

tomsilver: Sure can do. Or we can just remove the large to small one because we probably won't use it again anytime soon. Lmk what you prefer

ronuchit: Actually for consistency, it would be good to change the `assert` in this code snippet above to a separate `else` condition that raises a NotImplementedError (like we do for the `grammar_search_heuristic` setting), and this can easily be covered via a test that calls `select_predicates_to_keep` with an invalid `grammar_search_direction`.

src/approaches/grammar_search_invention_approach.py
ronuchit: But we don't have to relearn on every new state! A heuristic-making function should return a function from state to float.

ronuchit: Oh wait nvm I get it

tomsilver: What? States are predicate sets here. Maybe I'm misunderstanding

ronuchit: ya i got confused :D

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,110 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+    pred_complexity = sum(candidates[p] for p in s)
+    total_score = \
+        CFG.grammar_search_false_pos_weight * num_false_positives + \
+        CFG.grammar_search_true_pos_weight * (-num_true_positives) + \
+        CFG.grammar_search_size_weight * op_size + \
+        CFG.grammar_search_pred_complexity_weight * pred_complexity
+    # Useful for debugging:
+    # print("TP/FP/S/C/Total:", num_true_positives, num_false_positives,
+    #       op_size, pred_complexity, total_score)
+    # Lower is better.
```
ronuchit, line 499: move this comment before the calculation of the score

src/approaches/grammar_search_invention_approach.py
```diff
@@ -432,93 +432,110 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
-            print(pred)
-
-        return set(kept_predicates)
+
+def _create_heuristic_function(initial_predicates: Set[Predicate],
+        atom_dataset: List[GroundAtomTrajectory],
+        candidates: Dict[Predicate, float]
+        ) -> Callable[[FrozenSet[Predicate]], float]:
+    if CFG.grammar_search_heuristic == "prediction_error":
+        return partial(_prediction_error_heuristic,
```
ronuchit, line 452: clarify with a comment that this returns a function which takes in the last argument of _prediction_error_heuristic. It's not intuitive at a glance.

ronuchit (approved):
LGTM! minor comments

PR #177, opened 2021-12-09
refactor StateActionTrajectory into its own class called LowLevelTrajectory

ronuchit:
This new class contains goals.

### DIFF
--- a/src/datasets/demo_replay.py
+++ b/src/datasets/demo_replay.py
@@ -20,15 +20,15 @@ def create_demo_replay_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
     # all states into one list, is that we want to compute
     # all ground NSRTs once per trajectory only, rather
     # than once per state.
-    weights = np.array([len(traj) for traj in demo_dataset])
+    weights = np.array([len(traj.states) for traj in demo_dataset])
     weights = weights / sum(weights)
     # Ground all NSRTs once per trajectory
     all_nsrts = get_gt_nsrts(env.predicates, env.options)
     ground_nsrts = []
-    for (ss, _) in demo_dataset:
-        objects = sorted(ss[0])
+    for traj in demo_dataset:
+        objects = sorted(traj.states[0])
         # Assumes objects should be the same within a traj
-        assert all(set(objects) == set(s) for s in ss)
+        assert all(set(objects) == set(s) for s in traj.states)
         ground_nsrts_traj: List[_GroundNSRT] = []
         for nsrt in all_nsrts:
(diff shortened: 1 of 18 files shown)


src/structs.py
```diff
@@ -740,17 +740,66 @@ def unset_option(self) -> None:
         assert not self.has_option()
 
 
+@dataclass(frozen=True, repr=False, eq=False)
```
tomsilver, line 743: I like that these are frozen, that means we can probably do some caching that wasn't possible before...

src/structs.py
```diff
@@ -740,17 +740,66 @@ def unset_option(self) -> None:
...
+class LowLevelTrajectory:
+    """A structure representing a low-level trajectory, containing a
+    state sequence, action sequence, and optional goal.
+    Invariant 1: If a goal is included, the trajectory achieves that goal.
```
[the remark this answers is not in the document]
tomsilver: i see nvm

src/nsrt_learning.py
```diff
@@ -76,12 +76,12 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
     segments = []
-    states, actions, all_atoms = trajectory
-    assert len(states) == len(actions) + 1 == len(all_atoms)
-    current_segment_traj : StateActionTrajectory = ([], [])
-    for t in range(len(actions)):
-        current_segment_traj[0].append(states[t])
-        current_segment_traj[1].append(actions[t])
+    traj, all_atoms = trajectory
+    assert len(traj.states) == len(all_atoms)
+    current_segment_traj : Tuple[List[State], List[Action]] = ([], [])
```
tomsilver, line 81: can you change this so that it's now two lists, `current_segment_states` and `current_segment_actions`?

src/datasets/demo_replay.py
```diff
@@ -20,15 +20,15 @@ def create_demo_replay_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
     # all states into one list, is that we want to compute
     # all ground NSRTs once per trajectory only, rather
     # than once per state.
-    weights = np.array([len(traj) for traj in demo_dataset])
+    weights = np.array([len(traj.states) for traj in demo_dataset])
```
tomsilver, line 23: 😓

tomsilver (approved):
this is great! thanks a lot for doing it

tomsilver:
@ashayathalye @amburger66 note the new `LowLevelTrajectory` class that is being added here. may be useful to you!

src/nsrt_learning.py
ronuchit: yes, good idea!

PR #183, opened 2021-12-11
hadd optimizations

tomsilver:
I tested this by making sure that the heuristic values are the same as the original implementation on cover, blocks, and painting with oracle operators. It seems to get a 1.5-2x speedup.

The main thing was realizing that when this line was true in the original implementation: `if tmp_dist < neighbor.distance` then `neighbor.distance` was always inf. I think that would not necessarily be true for domains where there are non-unitary costs.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -861,12 +861,6 @@ class HAddHeuristic:
         """
         return sum([self.facts[fact].distance for fact in self.goals])
 
-    def finished(self, achieved_goals: Set[Tuple[str, ...]],
-                 queue: List[Tuple[float, float, RelaxedFact]]) -> bool:
-        """This function gives a stopping criterion for the Dijkstra search.
-        """
-        return achieved_goals == self.goals or not queue
-
     def run_forward_pass(self, initial_state: PyperplanFacts) -> None:
         """Calculate the distance to each goal fact from the initial state
         under the delete relaxation.
--- a/src/utils.py
+++ b/src/utils.py
@@ -766,8 +766,6 @@ class RelaxedOperator:
     name: str
     preconditions: PyperplanFacts
     add_effects: PyperplanFacts
-    # Cost of applying this operator.
-    cost: int = field(default=1)
     # Alternative method to check whether all preconditions are True.
     counter: int = field(init=False, default=0)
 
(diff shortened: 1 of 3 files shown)


src/utils.py
```diff
@@ -876,50 +861,37 @@ def calc_goal_h(self) -> float:
...
+        new_facts = reachable_facts.copy()
+        while new_facts:
+            # Check if the goal is reached.
+            if self.goals.issubset(reachable_facts):
+                return
+            next_reachable_facts = set()
+            for fact in new_facts:
                 # Iterate over all operators this fact is a precondition of.
-                for operator in fact.precondition_of:
+                for operator in self.facts[fact].precondition_of:
```
ronuchit, line 877: This change looks strange, what's going on here?

src/utils.py
```diff
@@ -876,50 +861,37 @@ def calc_goal_h(self) -> float:
...
-                 queue: List[Tuple[float, float, RelaxedFact]]) -> bool:
-        """This function gives a stopping criterion for the Dijkstra search.
-        """
-        return achieved_goals == self.goals or not queue
-
-    def dijkstra(self, queue: List[Tuple[float, float, RelaxedFact]]) -> None:
-        """This function is an implementation of a Dijkstra search.
-        For efficiency reasons, it is used instead of an explicit graph
-        representation of the problem.
+    def run_forward_pass(self, initial_state: PyperplanFacts) -> None:
```
ronuchit, line 864: Is the basic idea that we are doing a fixed point iteration instead of a dijkstra now, like we do for `is_dr_reachable`?Can clarify this in docstring.

src/utils.py
```diff
@@ -876,50 +861,37 @@ def calc_goal_h(self) -> float:
...
+        """Calculate the distance to each goal fact from the initial state
+        under the delete relaxation.
```
ronuchit, line 866: Since this method returns None, clarify in the docstring what it's mutating.

ronuchit:
Actually, can we just wipe the whole cost thing in the hAdd implementation? We know that we can get it back from pyperplan codebase if we ever need it, and it just serves to confuse right now... esp. in light of your statement that it may invalidate your optimization.

tomsilver:

ronuchit (approved):
lgtm

src/utils.py
```diff
@@ -868,58 +854,49 @@ def get_cost(self, operator: RelaxedOperator) -> float:
...
+        while new_facts:
+            # Check if the goal is reached.
+            if self.goals.issubset(reachable_facts):
+                return
+            next_reachable_facts = set()
+            for fact in new_facts:
+                relaxed_fact = self.facts[fact]
                 # Iterate over all operators this fact is a precondition of.
-                for operator in fact.precondition_of:
+                for operator in relaxed_fact.precondition_of:
```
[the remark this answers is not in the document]
tomsilver: `fact` is just a tuple, it doesn't have `.precondition_of`

PR #184, opened 2021-12-13
revert hadd optimization and convert hadd lookahead to logspace 

tomsilver:


ronuchit:
Just so we remember for the future, the bug in this optimization was that it isn't properly considering operators with no preconditions. See the comment "# Handle operators that have no preconditions." in pyperplan: https://github.com/aibasel/pyperplan/blob/master/pyperplan/heuristics/relaxation.py#L125

PR #185, opened 2021-12-14
allow option for excluding all predicates

ronuchit:


tomsilver (approved):
thanks, long overdue in retrospect

PR #186, opened 2021-12-14
branching factor grammar search heuristic

tomsilver:
this is a bad heuristic, but we may want to use it as a baseline in the future.


tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -422,3 +426,53 @@ def test_hadd_lookahead_heuristic():
...
+    }
+    dataset = create_dataset(env, next(env.train_tasks_generator()))
+    atom_dataset = utils.create_ground_atom_dataset(dataset,
+        env.goal_predicates | set(candidates))
+    heuristic = _BranchingFactorHeuristic(env.goal_predicates, atom_dataset,
+                                          candidates)
+    holding_h = heuristic.evaluate({Holding})
+    forall_not_covers_h = heuristic.evaluate({forall_not_covers0,
+                                              forall_not_covers1})
+    # This is just to illustrate that the cost for these two bad predicates are
```
ronuchit, line 475: costs* ?

ronuchit (approved):
lgtm thanks

tests/approaches/test_grammar_search_invention_approach.py
tomsilver: no I meant the heuristic value for the set containing both heuristics

tomsilver: oh i see the grammar issue, will fix

PR #189, opened 2021-12-16
have run_testing return results dict so we can more easily save it

tomsilver:
this is one of a few PRs that I am going to make so that we can automatically generate the proxy analysis tables


ronuchit (approved):
lgtm

PR #190, opened 2021-12-16
 make option specs more readable and immediate 

tomsilver:


ronuchit (approved):
lgtm but should we not print out the STRIPS operators then? i guess if we're doing option learning, it could do a bunch of stuff between STRIPS operator printing and this new printout... but in the main method that will be instantaneous

tomsilver:
@ronuchit good point, I changed it to print in both cases only if option learning

PR #191, opened 2021-12-16
clean up grammar analysis code so it auto-generates proxy analysis table

tomsilver:


analysis/grammar_search_analysis.py
```diff
@@ -2,263 +2,188 @@
...
+    print(df)
+    csv_filepath = os.path.join(outdir, "proxy_analysis.csv")
+    df.to_csv(csv_filepath)
+    print(f"Wrote out to {csv_filepath}.")
 
 
 if __name__ == "__main__":
-    # _run_analysis()
-    _run_proxy_analysis()
+    ENV_NAMES = [
```
ronuchit, line 172: It's a bit strange to have all these constants (all-caps names) defined down here. Two options here:
1) Move all this into a new method called `def _main():` so you can name these constants as regular variables instead, and then just do `if __name__ == "__main__": _main()`
2) Define all these constants at top of file

tomsilver:
@ronuchit thanks, anything else?

ronuchit (approved):
lgtm

Issue #175, opened 2021-12-09
get RepeatedNextToEnv working with nsrt_learning


ronuchit:
Currently, both planning and learning work only because we have set the threshold for being next to an object to be very low, such that whp the robot is only next to one object at any given time. Increasing this `nextto_thresh = 0.02` threshold would quickly break planning (and learning).

Requires some new ideas for planning or operator learning (or both).

ronuchit:
planning was solved in #192

PR #192, opened 2021-12-16
possible way to address repeatednextto

tomsilver:
possible path toward addressing planning (half of #175)

@ronuchit @NishanthJKumar @wmcclinton

### DIFF
--- a/src/planning.py
+++ b/src/planning.py
@@ -208,10 +208,10 @@ def _run_low_level_search(
                 traj[cur_idx+1] = next_state
                 cur_idx += 1
                 # Check atoms against expected atoms_sequence constraint.
-                assert len(traj) == len(atoms_sequence)
                 # The NSRT defines a scope: objects that are relevant for this
                 # transition. All objects outside the scope will be ignored when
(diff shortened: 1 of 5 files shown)


src/planning.py
```diff
@@ -209,9 +209,17 @@ def _run_low_level_search(
                 cur_idx += 1
                 # Check atoms against expected atoms_sequence constraint.
                 assert len(traj) == len(atoms_sequence)
-                atoms = utils.abstract(traj[cur_idx], predicates)
-                if atoms == {atom for atom in atoms_sequence[cur_idx]
-                             if atom.predicate.name != _NOT_CAUSES_FAILURE}:
+                # The NSRT defines a scope: objects that are relevant for this
```
ronuchit, line 212: Move this comment above the `assert len(traj) == len(atoms_sequence)` line, so that it's attached with the `# Check atoms against expected atoms_sequence constraint.` comment, otherwise that comment seems like it's only for the subsequent line.

ronuchit (approved):
lgtm, one comment

PR #194, opened 2021-12-21
prep for task planning predicate invention: pass train_tasks through to heuristics, and expose task planning only

ronuchit:


tests/test_planning.py
```diff
@@ -21,8 +22,16 @@ def test_sesame_plan():
     nsrts = get_gt_nsrts(env.predicates, env.options)
     task = next(env.train_tasks_generator())[0]
     option_model = create_option_model(CFG.option_model_name, env.simulate)
-    plan = sesame_plan(task, option_model, nsrts, env.predicates, 1, 123)
+    plan, _ = sesame_plan(task, option_model, nsrts, env.predicates,
```
tomsilver, line 25: I'm confused, was this a bug in the old unit test?

tomsilver (approved):
thanks for doing this

tests/test_planning.py
ronuchit: yes lol the metrics return wasn't being considered

src/planning.py
```diff
@@ -126,6 +135,8 @@ def _run_search(task: Task,
         if task.goal.issubset(node.atoms):
             # If this skeleton satisfies the goal, run low-level search.
             metrics["num_skeletons_optimized"] += 1
+            if not do_low_level_search:
```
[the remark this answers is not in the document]
ronuchit: that would be very tough! let's leave it...

PR #195, opened 2021-12-22
 task planning heuristic for grammar search approach 

tomsilver:
we know that this is bad (see unit test), but we want to include it for completeness

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -17,7 +17,7 @@ from predicators.src.approaches import NSRTLearningApproach, ApproachFailure, \
     ApproachTimeout
 from predicators.src.nsrt_learning import segment_trajectory, \
     learn_strips_operators
-from predicators.src.option_model import create_option_model
+from predicators.src.option_model import DummyOptionModel
 from predicators.src.planning import sesame_plan
 from predicators.src.structs import State, Predicate, ParameterizedOption, \
     Type, Task, Action, Dataset, Object, GroundAtomTrajectory, STRIPSOperator, \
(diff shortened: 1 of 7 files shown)


src/approaches/grammar_search_invention_approach.py
```diff
@@ -558,6 +564,47 @@ def _evaluate_with_operators(self, predicates: FrozenSet[Predicate],
...
+
+    def _evaluate_with_operators(self, predicates: FrozenSet[Predicate],
+                                 pruned_atom_data: List[GroundAtomTrajectory],
+                                 segments: List[Segment],
+                                 strips_ops: List[STRIPSOperator],
+                                 option_specs: List[OptionSpec]) -> float:
+        del pruned_atom_data, segments  # unused
+        score = 0.0
+        node_expansion_upper_bound = 1e7
+        option_model = create_option_model("default", lambda s, a: s)  # dummy
```
ronuchit, line 583: Let's just make a `DummyOptionModel = ...` in option_model.py and pass that directly into sesame_plan below?

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -487,3 +490,48 @@ def test_branching_factor_heuristic():
...
+    }
+    train_tasks = next(env.train_tasks_generator())
+    dataset = create_dataset(env, train_tasks)
+    atom_dataset = utils.create_ground_atom_dataset(dataset,
+        env.goal_predicates | set(candidates))
+    heuristic = _TaskPlanningHeuristic(env.goal_predicates, atom_dataset,
+                                       train_tasks, candidates)
+    all_included_h = heuristic.evaluate({Holding, HandEmpty})
+    none_included_h = heuristic.evaluate(set())
+    # This is terrible!
```
ronuchit, line 527: Why exactly is this happening? Is it because the operators with no added predicates just think you can always achieve Covers, which works at the task planning level?

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -487,3 +490,48 @@ def test_branching_factor_heuristic():
...
+    assert none_included_h < all_included_h
+    # Test cases where operators cannot plan to goal.
+    utils.update_config({
+        "min_data_for_nsrt": 10000,
+    })
+    assert heuristic.evaluate(set()) == len(train_tasks) * 1e7
```
ronuchit, line 533: Can we also check that this happens with `heuristic.evaluate({Holding, HandEmpty})`?

tests/approaches/test_grammar_search_invention_approach.py
tomsilver: yeah, exactly

ronuchit (approved):
lgtm

PR #197, opened 2021-12-22
rename DefaultOption -> DummyOption

ronuchit:
It's a bit strange that DefaultOption and DefaultOptionModel were separate conceptual ideas. So, doing a simple renaming to make DummyOption sound more similar to DummyOptionModel, which are the same conceptual idea.


tomsilver (approved):
lgtm thanks

PR #203, opened 2021-12-23
refactor hadd/exact match/lookahead score functions

ronuchit:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -702,6 +705,82 @@ def _evaluate_atom_trajectory(self, atoms_sequence: List[Set[GroundAtom]],
...
+            utils.atoms_to_tuples(goal),
+            relaxed_operators)
+        def _hadd_fn_h(init_atoms: Set[GroundAtom]) -> float:
+            return hadd_fn(utils.atoms_to_tuples(init_atoms))
+        return _hadd_fn_h
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _ExactHeuristicBasedScoreFunction(_HeuristicBasedScoreFunction):  # pylint:disable=abstract-method
+    """Implement _generate_heuristic() with task planning.
```
tomsilver, line 734: just curious, what's going on with the pylint thing here? i don't see any abstract methods

src/approaches/grammar_search_invention_approach.py
```diff
@@ -702,6 +705,82 @@ def _evaluate_atom_trajectory(self, atoms_sequence: List[Set[GroundAtom]],
...
+                            ground_ops: Set[_GroundSTRIPSOperator]
+                            ) -> Callable[[Set[GroundAtom]], float]:
+        relaxed_operators = frozenset({utils.RelaxedOperator(
+            op.name, utils.atoms_to_tuples(op.preconditions),
+            utils.atoms_to_tuples(op.add_effects)) for op in ground_ops})
+        hadd_fn = utils.HAddHeuristic(
+            utils.atoms_to_tuples(init_atoms),
+            utils.atoms_to_tuples(goal),
+            relaxed_operators)
+        def _hadd_fn_h(init_atoms: Set[GroundAtom]) -> float:
```
tomsilver, line 727: can you change `init_atoms` to something else here to avoid confusion with the thing in the outer scope?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -702,6 +705,82 @@ def _evaluate_atom_trajectory(self, atoms_sequence: List[Set[GroundAtom]],
...
+    """Implement _generate_heuristic() with task planning.
+    """
+    def _generate_heuristic(self, init_atoms: Set[GroundAtom],
+                            objects: Set[Object],
+                            goal: Set[GroundAtom],
+                            strips_ops: Sequence[STRIPSOperator],
+                            option_specs: Sequence[OptionSpec],
+                            ground_ops: Set[_GroundSTRIPSOperator]
+                            ) -> Callable[[Set[GroundAtom]], float]:
+        def _task_planning_h(init_atoms: Set[GroundAtom]) -> float:
```
tomsilver, line 743: same comment about `init_atoms`

src/approaches/grammar_search_invention_approach.py
```diff
@@ -702,6 +705,82 @@ def _evaluate_atom_trajectory(self, atoms_sequence: List[Set[GroundAtom]],
...
+                    init_atoms, objects, goal, strips_ops, option_specs,
+                    CFG.seed, CFG.grammar_search_task_planning_timeout)
+            except (ApproachFailure, ApproachTimeout):
+                return float("inf")
+            return float(len(plan))
+        return _task_planning_h
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _HAddHeuristicMatchBasedScoreFunction(
```
tomsilver, line 758: nice use of multiple inheritance

tomsilver (approved):
great stuff! also if you don't mind, can you add the new score function to the analysis file?

src/approaches/grammar_search_invention_approach.py
ronuchit: pylint complains when you make a class that only partially implements an abstract parent class. another solution is to copy-paste the abstract method from parent, and keep it abstract here

src/approaches/grammar_search_invention_approach.py
ronuchit: thanks missed that

tomsilver (approved):

PR #205, opened 2021-12-23
hmax heuristic

tomsilver:
hmax is significantly worse in cover, blocks, and painting than hadd, so we don't want to make it the default. but in the playroom domain, hmax can succeed where hadd completely fails. so it's good to give ourselves the option to toggle.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -935,7 +935,7 @@ class _RelaxationHeuristic:
     def get_cost(self, operator: RelaxedOperator) -> float:
         """This function calculates the cost of applying an operator.
         """
-        # Sum over the heuristic values of all preconditions.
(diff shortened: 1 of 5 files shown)


amburger66 (approved):
lgtm!

src/utils.py
```diff
@@ -962,6 +993,22 @@ def dijkstra(self, queue: List[Tuple[float, float, RelaxedFact]]) -> None:
                 fact.expanded = True
 
 
+class HAddHeuristic(_RelaxationHeuristic):
```
ronuchit, line 996: Let's make HAddHeuristic and HMaxHeuristic private so create_heuristic is the only interface

ronuchit (approved):
lgtm after one minor fix

PR #208, opened 2021-12-23
filter static and empty effect operators in the operator-based score

tomsilver:
…functions for grammar search

also  switch from inf to epsilon greedy in lookahead score function

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -666,26 +666,27 @@ def extract_preds_and_types(nsrts: Collection[NSRT]) -> Tuple[
     return preds, types
 
 
-def filter_static_nsrts(ground_nsrts: Collection[_GroundNSRT],
-                        atoms: Collection[GroundAtom]) -> List[
-                            _GroundNSRT]:
-    """Filter out ground NSRTs that don't satisfy static facts.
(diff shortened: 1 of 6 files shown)


src/utils.py
```diff
@@ -666,26 +666,27 @@ def extract_preds_and_types(nsrts: Collection[NSRT]) -> Tuple[
     return preds, types
 
 
-def filter_static_nsrts(ground_nsrts: Collection[_GroundNSRT],
-                        atoms: Collection[GroundAtom]) -> List[
-                            _GroundNSRT]:
-    """Filter out ground NSRTs that don't satisfy static facts.
+_O = TypeVar('_O', bound=Union[_GroundNSRT, _GroundSTRIPSOperator])
```
tomsilver, line 669: @ronuchit I think we can probably refactor a number of things using this pattern, yeah?

tomsilver:
this currently appears to be a step in the wrong direction, so going to close for now.

src/utils.py
ronuchit: this pattern doesn't work as intended bc `bound` isn't what we actually want, but see #220

PR #209, opened 2021-12-23
reduce amount of demos in general and for heuristic-based grammar search

ronuchit:


tomsilver (approved):
lgtm after changing num training tasks to 15 across the board

PR #212, opened 2021-12-27
learning operators with side predicates

### LINKED ISSUE #175: get RepeatedNextToEnv working with nsrt_learning

tomsilver:
closes #175


src/nsrt_learning.py
```diff
@@ -309,3 +436,136 @@ def unify_effects_and_options(
...
+                if ground_op.add_effects != segment.add_effects or \
+                   ground_op.delete_effects != segment.delete_effects:
+                    continue
+                # Operator covers the segment.
+                skeleton.append((ground_op.operator.name,
+                                 tuple(ground_op.objects)))
+                break
+            else:
+                continue
+            break
```
ronuchit, line 483: wtf

tomsilver: 😬

src/nsrt_learning.py
```diff
@@ -146,9 +142,157 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+                print("Skeletons still covered; keeping conversion.")
+        # Consider converting each delete effect.
+        for effect in op.delete_effects:
+            if verbose:
+                print(f"Considering delete effect: {effect} from {op.name}")
+            current_op = name_to_strips_op[op.name]
+            assert current_op.name == op.name
+            # Tentatively replace the current operator.
+            remaining_params = {p for atom in current_op.preconditions | \
+                (current_op.add_effects - {effect}) | \
```
ronuchit, line 228: this was a bug right? this should be removed from delete_effects

tomsilver: yeah good catch

ronuchit:
closing bc outdated, we can always revisit later in the closed PRs section

PR #213, opened 2021-12-28
allow hadd lookahead to actually perform a >0 depth lookahead

tomsilver:
this allows us to explore intermediate methods between the original hadd lookahead and exact planning.

proxy analysis CSV file generated from this PR: 
[proxy_analysis.csv](https://github.com/Learning-and-Intelligent-Systems/predicators/files/7785563/proxy_analysis.csv)

* hadd lookahead depth 1 looks very good
* hadd lookahead depth 2 also looks good *except* the number for painting with all predicates given is way high. not sure why, will investigate (edit: see notes below).

## investigating painting all predicates

consider the first state in a trajectory where the first action taken is open-lid. here are some on and off-demo heuristic values for the respective methods:
  
**hadd lookahead (depth 0)**
off-demo h: 17
off-demo h: 19
off-demo h: 19
off-demo h: 21
on-demo h: 17
on-demo h: 17
off-demo h: 21
on-demo h: 17
on-demo h: 17
on-demo h: 17
on-demo h: 17
on-demo h: 17
off-demo h: 19

**hadd lookahead depth 1**
off-demo h: 18.0
off-demo h: 19.0
off-demo h: 19.0
off-demo h: 21.0
on-demo h: 18.0
on-demo h: 18.0
off-demo h: 21.0
on-demo h: 18.0
on-demo h: 18.0
on-demo h: 18.0
on-demo h: 18.0
on-demo h: 18.0
off-demo h: 20.0
   
**hadd lookahead depth 2**
off-demo h: 18.0
**off-demo h: 14.0**
off-demo h: 20.0
off-demo h: 22.0
**on-demo h: 19.0**
on-demo h: 19.0
off-demo h: 14.0
on-demo h: 19.0
on-demo h: 19.0
on-demo h: 19.0
on-demo h: 19.0
on-demo h: 19.0
off-demo h: 20.0
   
**exact planning**
off-demo h: 10.0
off-demo h: 10.0
off-demo h: inf
off-demo h: inf
on-demo h: 11.0
on-demo h: 11.0
off-demo h: 10.0
on-demo h: 11.0
on-demo h: 11.0
on-demo h: 11.0
on-demo h: 11.0
on-demo h: 11.0
off-demo h: inf

so I think the issue stems from the bolded values under hadd lookahead depth 2. basically in this domain, looking ahead 2 steps is enough to realize that doing something is better than noop, but it's not enough to realize that opening the box lid (which appears to be a noop) is only 1 worse than picking some object at the task planning level.


ronuchit:
I understand what you're trying to do in this PR, but I don't know why we're doing it. Is the idea just to experiment?

- I see that the depth=1 values look good, but they don't seem better to me than the depth=0 values. Is that expected?
- Why are the times for depth=1 so similar to those for depth=0? Is it just because we generally have very few successors in all our domains?
- Why exactly is this naive, exhaustive lookahead strategy allowing for a middle ground toward exact lookahead? Is it because exact lookahead (which runs A* task planning) can be interpreted as doing this same naive search? Does our A* task planning always give the same results as this naive search would, with the `min`ing over the successors?

tomsilver:
@ronuchit good questions...

> I see that the depth=1 values look good, but they don't seem better to me than the depth=0 values. Is that expected?

that is my understanding yeah. but in the limit of depth=inf, this would be exact planning, which we know to be sometimes better than depth=0.

>Why are the times for depth=1 so similar to those for depth=0? Is it just because we generally have very few successors in all our domains?

yeah and we're caching the heuristic evaluations, so that helps.

>Why exactly is this naive, exhaustive lookahead strategy allowing for a middle ground toward exact lookahead? Is it because exact lookahead (which runs A* task planning) can be interpreted as doing this same naive search? Does our A* task planning always give the same results as this naive search would, with the mining over the successors?

right, the second sentence is the answer to the first sentence, basically. I don't understand the third sentence -- what do you mean by mining over the successors?

ronuchit:
By "mining over the successors" I mean `min`-ing -- does that help? You're doing a `result = 1.0 + min(successor_hs)`. I'm wondering if that's more or less equivalent to what our A* planner is doing.

ronuchit:
Okay, this change is good, so rebase/prepare this PR and then I can review it properly?

tomsilver:
> By "mining over the successors" I mean `min`-ing -- does that help? You're doing a `result = 1.0 + min(successor_hs)`. I'm wondering if that's more or less equivalent to what our A* planner is doing.

oh lol I did not parse out "min" from that. I think it is equivalent to A* with an admissible heuristic, in that both are guaranteed to find an optimal plan.

tomsilver:
@ronuchit thanks, it should be ready now

src/utils.py
```diff
@@ -714,6 +714,25 @@ def apply_operator(op: GroundNSRTOrSTRIPSOperator,
     return new_atoms
 
 
+def get_successors_from_ground_ops(atoms: Set[GroundAtom],
+        ground_ops: Collection[_GroundSTRIPSOperator],
```
ronuchit, line 718: The type of this should be `GroundNSRTOrSTRIPSOperator`, matching `get_applicable_operators` and `apply_operator`. Also, this will let us change the two uses of get_applicable_operators/apply_operator: one in planning.py and the other in grammar_search_invention_approach.py. They can both use this function instead (remember to use `unique=False` though!).

ronuchit (approved):
one minor comment

src/utils.py
tomsilver: the one in planning.py needs the `nsrt`, not just the successor atoms, so I'm leaving that one untouched.

PR #214, opened 2021-12-28
pruning redundant predicates in grammar search approach

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -373,6 +381,42 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+                continue
+            # Found a new predicate.
+            seen[pred_id] = predicate
+            yield (predicate, cost)
+
+    def _get_predicate_identifier(self, predicate: Predicate) -> int:
+        """Returns a hash of frozensets of groundatoms for each data point.
+        """
+        # Get atoms for this predicate alone on the dataset.
+        atom_dataset = utils.create_ground_atom_dataset(self.dataset,
```
ronuchit, line 410: lol, there's no way this hacky hashing works... does it really!?!

src/approaches/grammar_search_invention_approach.py
```diff
@@ -54,12 +54,18 @@ def _create_grammar(grammar_name: str, dataset: Dataset,
         # For each predicate enumerated, we also enumerate foralls for
         # that predicate, along with appropriate negations.
         forall_grammar = _ForallPredicateGrammarWrapper(negated_grammar)
-        # Finally, we don't actually need to enumerate the given predicates
+        # Prune proposed predicates by checking if they are equivalent to
```
ronuchit, line 57: it's a bit weird to do this for `forall_single_feat_ineqs` but not `single_feat_ineqs`. since these names are extremely out of date, can we get rid of the latter and just have the former (with a better name)? In fact, I don't think we need holding_dummy anymore, it's served its purpose. Should we just get rid of `CFG.grammar_search_grammar_name` altogether?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -293,7 +301,7 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
         feature_ranges = self._get_feature_ranges()
         # 0.5, 0.25, 0.75, 0.125, 0.375, ...
         constant_generator = _halving_constant_generator(0.0, 1.0)
-        for c in constant_generator:
+        for n, c in enumerate(constant_generator):
```
ronuchit, line 304: What's going on here? Why should 0.25 have a lower cost than 0.75? Maybe let's make halving_constant_generator return reasonable costs instead? E.g. 0.5 is lowest cost, then 0.25 and 0.75 are the next-lowest, etc?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -373,6 +381,42 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+@dataclass(frozen=True, eq=False, repr=False)
+class _PrunedGrammar(_DataBasedPredicateGrammar):
+    """A grammar that prunes redundant predicates.
+    """
+    base_grammar: _PredicateGrammar
+
+    def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
+        # Predicates are identified based on their evaluation across
+        # all states in the dataset.
+        seen = {}  # maps identifier to previous predicate, for debugging
```
ronuchit, line 393: Why "for debugging"? Isn't this necessary for checking redundancy, which is the whole point of this class?

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -61,7 +61,16 @@ def test_predicate_grammar():
...
+    # There are only so many unique predicates possible under the grammar.
+    # Non-unique predicates are pruned. Note that with a larger dataset,
+    # more predicates would appear unique.
+    assert len(forall_grammar.generate(max_num=10)) == 7
+    # Test CFG.grammar_search_predicate_cost_upper_bound.
+    default = CFG.grammar_search_predicate_cost_upper_bound
+    utils.update_config({"grammar_search_predicate_cost_upper_bound": 0})
+    assert len(single_ineq_grammar.generate(max_num=10)) == 0
+    # Reset to default just in case.
+    utils.update_config({"grammar_search_predicate_cost_upper_bound": default})
```
ronuchit, line 73: Could you add a test that with an empty dataset, there should only ever be one predicate generated, since all predicates would have the same hash?

src/approaches/grammar_search_invention_approach.py
tomsilver: wait what's hacky about it? the reason i'm storing `hash(thing)` rather than `thing` is because the `thing`s could be large, but I can probably just store the things anyway if this is confusing or seems bad

tomsilver: eh nvm I'll just do the other way

src/approaches/grammar_search_invention_approach.py
```diff
@@ -254,39 +247,14 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
-@dataclass(frozen=True, eq=False, repr=False)
-class _HoldingDummyPredicateGrammar(_DataBasedPredicateGrammar):
-    """A hardcoded cover-specific grammar.
-
-    Good for testing with:
-        python src/main.py --env cover --approach grammar_search_invention \
-            --seed 0 --excluded_predicates Holding
+def _halving_constant_generator(lo: float, hi: float, cost: float = 1.0
+                                ) -> Iterator[Tuple[float, float]]:
+    """The second element of the tuple is a cost.
```
ronuchit, line 252: Add to this docstring an explanation of what the output constants and costs will be. (Basically copied from the test_halving_constant_generator)

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -126,9 +121,11 @@ def test_halving_constant_generator():
     """Tests for _halving_constant_generator().
     """
     expected_sequence = [0.5, 0.25, 0.75, 0.125, 0.625, 0.375, 0.875]
+    expected_costs = [1.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0]
     generator = _halving_constant_generator(0., 1.)
-    for i, x in zip(range(len(expected_sequence)), generator):
-        assert abs(expected_sequence[i] - x) < 1e-6
+    for i, (constant, cost) in zip(range(len(expected_sequence)), generator):
```
ronuchit, line 126: You can zip 3 lists in python:
`for (expected_constant, expected_cost, (constant, cost)) in zip(expected_sequence, expected_costs, generator):`

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -126,9 +121,11 @@ def test_halving_constant_generator():
     """Tests for _halving_constant_generator().
     """
     expected_sequence = [0.5, 0.25, 0.75, 0.125, 0.625, 0.375, 0.875]
```
ronuchit, line 123: rename to `expected_constants` to match the function?

ronuchit (approved):
lgtm after minor comments

src/approaches/grammar_search_invention_approach.py
```diff
@@ -303,7 +279,7 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
                     if abs(lb - ub) < 1e-6:
                         continue
                     # Scale the constant by the feature range.
-                    k = (c + lb) / (ub - lb)
+                    k = (constant + lb) / (ub - lb)
                     # Only need one of (ge, le) because we can use negations
```
ronuchit, line 283: Does this comment need updating since we removed `ge`?

ronuchit: i guess not, nvm

PR #215, opened 2021-12-28
truncated score function for grammar search

tomsilver:
this is still too darn slow to be usable in painting, but I want to remember that I tried it...


tomsilver:
@ronuchit we should talk about whether it's worth merging both this and #213 , because they are basically two different ways to do very similar things.

ronuchit:
I'm not a fan of this, I like #213 a lot more, especially because #213 doesn't rely on running the task planner at all. Let's close this PR and focus on merging that one instead?

tomsilver:
closing in favor of #213

PR #216, opened 2021-12-29
added case showing fallability of exact lookahead in painting

tomsilver:
this shows a case where adding an apparently spurious predicate -- `NOT-Forall[0:obj,1:shelf].[NOT-InShelf(0,1)]()` -- to the full set in painting leads to a better `exact_lookahead` score, but worse planning (41-45/50 successes, depending on timeout).

despite the original title of this PR, after digging, the failure appears to be related to sampler learning rather than predicate learning.

in particular, the sampler for this NSRT is looking bad, in the sense that it's sampling places that are not in the shelf:

```
NSRT-Op5:
    Parameters: [?x0:obj, ?x1:shelf, ?x2:robot]
    Preconditions: [Holding(?x0:obj), HoldingSide(?x2:robot), OnTable(?x0:obj)]
    Add Effects: [GripperOpen(?x2:robot), InShelf(?x0:obj, ?x1:shelf), NOT-Forall[0:obj,1:shelf].[NOT-InShelf(0,1)]()]
    Delete Effects: [Holding(?x0:obj), HoldingSide(?x2:robot), OnTable(?x0:obj)]
    Option Spec: Place(?x2:robot)
```

now consider this other NSRT:
```
NSRT-Op9:
    Parameters: [?x0:obj, ?x1:shelf, ?x2:robot]
    Preconditions: [Holding(?x0:obj), HoldingSide(?x2:robot), NOT-Forall[0:obj,1:shelf].[NOT-InShelf(0,1)](), OnTable(?x0:obj)]
    Add Effects: [GripperOpen(?x2:robot), InShelf(?x0:obj, ?x1:shelf)]
    Delete Effects: [Holding(?x0:obj), HoldingSide(?x2:robot), OnTable(?x0:obj)]
    Option Spec: Place(?x2:robot)
```

I believe that all of the data that fits the second NSRT will be viewed as negative examples for the first NSRT's sampler's classifier...

so for a (state, action) where the robot is holding an object with a side grasp and the action correctly places that object into the shelf, if this is the second, third, etc. object that's getting placed into the shelf (so `NOT-Forall[0:obj,1:shelf].[NOT-InShelf(0,1)]()` is true), then `NOT-Forall[0:obj,1:shelf].[NOT-InShelf(0,1)]()` will not appear in the add effects, so that transition would look like a failure from the perspective of the first NSRT.

this is normally fine, because what we would expect is that the sampler classifier should be able to distinguish between the two cases. however, here, the feature that distinguishes between the two cases involves universal quantification, which would mean that the sampler has to use the full `state` and not just the objects that are named in the NSRT.

one possible workaround would be to give the sampler classifiers the symbolic preconditions, and have them learn a "delta" on those.

----

there's a second issue here, which is that the sampler regressor is sometimes outputting garbage, so it's not just the classifier's fault. this seems to be a data issue -- the bounds for being on table are [-10.0, -0.2]; the range of y values seen in the data for on table are about [-9.5, -1.5]; and the sampler is getting queried on -0.22, which is significantly outside of that data range.

re-running with the original 50 train tasks instead of 15, we get 50/50 performance. so that suggests that we should increase the number of train tasks back up, but we should probably still fix the issue with the classifier discussed above, even if it's not preventing a good final score.

----

finally, there's the question of why this spurious predicate is getting selected in the first place, even if it doesn't degrade final performance. unlike in the past, this does not appear to be a case of a surprisingly good predicate -- the total number of nodes expanded increases from about 3900 to about 5000 with the addition of this predicate.

verified that this is not related to the 15 vs. 50 train tasks.

I thought it might be that the operators with the spurious predicate are getting "extra credit" for predicting the same on-demo successor multiple times. I experimented with this change:

```
diff --git a/src/approaches/grammar_search_invention_approach.py b/src/approaches/grammar_search_invention_approach.py
index dc4f4f4..90e03de 100644
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -688,9 +688,14 @@ class _HeuristicLookaheadBasedScoreFunction(_HeuristicBasedScoreFunction):  # py
             atoms, next_atoms = atoms_sequence[i], atoms_sequence[i+1]
             ground_op_demo_lpm = -np.inf  # total log prob mass for demo actions
             ground_op_total_lpm = -np.inf  # total log prob mass for all actions
+            seen_successors = set()
             for ground_op in utils.get_applicable_operators(ground_ops, atoms):
                 # Compute the next state under the operator.
                 predicted_next_atoms = utils.apply_operator(ground_op, atoms)
+                frozen_predicted_next_atoms = frozenset(predicted_next_atoms)
+                if frozen_predicted_next_atoms in seen_successors:
+                    continue
+                seen_successors.add(frozen_predicted_next_atoms)
```

but that did not resolve the issue; the scores change a bit, but the score for the spurious predicate is still better (431.5 for spurious vs. 460.5 for ground truth).

okay: it's because in the demonstration data, there are no plans where the box item is placed before a shelf item. why does this happen? for some reason, the hAdd value given to pickside is better than that given to picktop in the first step of the plan... this is surprising because I thought everything would be symmetric between pickside and picktop. example:

```
initial atoms: {IsClean(obj2:obj), IsDry(obj1:obj), IsClean(obj1:obj), IsDry(obj0:obj), IsDirty(obj0:obj), IsDry(obj2:obj), OnTable(obj1:obj), OnTable(obj0:obj), OnTable(obj2:obj), GripperOpen(robby:robot)}

root node
('InBox', 'obj2:obj', 'receptacle_box:box') 3
('IsBoxColor', 'obj2:obj', 'receptacle_box:box') 2
('IsShelfColor', 'obj0:obj', 'receptacle_shelf:shelf') 4
('InShelf', 'obj1:obj', 'receptacle_shelf:shelf') 3
('IsShelfColor', 'obj1:obj', 'receptacle_shelf:shelf') 2
('InShelf', 'obj0:obj', 'receptacle_shelf:shelf') 3

after PickFromTop [obj2:obj, robby:robot]:
('InBox', 'obj2:obj', 'receptacle_box:box') 1
('IsBoxColor', 'obj2:obj', 'receptacle_box:box') 1
('IsShelfColor', 'obj0:obj', 'receptacle_shelf:shelf') 6
('InShelf', 'obj1:obj', 'receptacle_shelf:shelf') 5
('IsShelfColor', 'obj1:obj', 'receptacle_shelf:shelf') 3
('InShelf', 'obj0:obj', 'receptacle_shelf:shelf') 5
total f: 22

after PickFromSide [obj0:obj, robby:robot]:
('InBox', 'obj2:obj', 'receptacle_box:box') 5
('IsBoxColor', 'obj2:obj', 'receptacle_box:box') 3
('IsShelfColor', 'obj0:obj', 'receptacle_shelf:shelf') 2
('InShelf', 'obj1:obj', 'receptacle_shelf:shelf') 3
('IsShelfColor', 'obj1:obj', 'receptacle_shelf:shelf') 3
('InShelf', 'obj0:obj', 'receptacle_shelf:shelf') 1
total f: 18
```

alright I think this is just an hAdd fluke... `HoldingSide(robot)` is a precondition for more relevant operators than `HoldingTop(robot)`, because there are more objects that need to be put into the shelf than into the box, so this means that accomplishing `HoldingSide(robot)` first always looks better than `HoldingTop(robot)`. this explains behavior we saw before, where the pick order always has top in the second to last or last position.

so after understanding all that, I think no action is required from this third chunk of stuff; it's just a fluke of our heuristic that is ultimately not very important.

but to restate this more generally: the supposition of `exact_lookahead` is that the data was generated by an agent sampling plans from the energy-based distribution. under this distribution, any optimal plan gets an equal probability mass. but in fact, the data was not generated in this way; it was generated by the planner, and because of the heuristic, it happens that the data distribution does _not_ give every optimal plan equal probability mass. the probability of "sampling" a plan under this data distribution is higher when the plan starts out with pickside than when it starts out with picktop.

arguably, the conclusion of the above paragraph is that `exact_lookahead` is not a perfect proxy for our real objective, efficiency issues aside...


ronuchit:
i didn't get why "there are no plans where the box item is placed before a shelf item" implies that this predicate would get selected. why does this predicate make the demonstrations seem more rational?

tomsilver:
> i didn't get why "there are no plans where the box item is placed before a shelf item" implies that this predicate would get selected. why does this predicate make the demonstrations seem more rational?

yeah good question. I think it's at least in part because of this operator:

```
STRIPS-Op8:
    Parameters: [?x0:obj, ?x1:box, ?x2:robot]
    Preconditions: [Holding(?x0:obj), HoldingTop(?x2:robot), NOT-Forall[0:obj,1:shelf].[NOT-InShelf(0,1)](), OnTable(?x0:obj)]
    Add Effects: [GripperOpen(?x2:robot), InBox(?x0:obj, ?x1:box)]
    Delete Effects: [Holding(?x0:obj), HoldingTop(?x2:robot), OnTable(?x0:obj)]
    Option Spec: Place(?x2:robot)
```

which is the only operator for placing into the box, and which has "shelf is not empty" as a precondition. since all of the demonstrations do have placing into the box occur when the shelf is not empty, this operator looks more rational than one where the predicate is not included in the preconditions. I think.

ronuchit:
> this is normally fine, because what we would expect is that the sampler classifier should be able to distinguish between the two cases.
Maybe we can test that if we add a hacked low-level feature to the environment about whether there is an object in the shelf, then things should work well.

ronuchit:
Should we close this PR? I don't think there's any action items here?

tomsilver:
this has served its PRpose

ronuchit:
Or did you want to think harder about how we can fix the samplers?

PR #217, opened 2021-12-30
implement hff heuristic

tomsilver:
on oracle, compared to hadd:
* very slightly worse for cover
* slightly better for blocks
* significantly worse for painting
* significantly better for playroom (@amburger66 please verify)


amburger66:
verified! tried a handful of combinations of open/closed doors, and all were solvable :0 very cool

src/utils.py
```diff
@@ -1014,6 +1022,52 @@ def _accumulate(distances: Collection[float]) -> float:
...
 
 
+class _HFFHeuristic(_RelaxationHeuristic):
+    """Implements the HFF delete relaxation heuristic.
+    """
+    @staticmethod
+    def _accumulate(distances: Collection[float]) -> float:
+        return sum(distances)
+
+    def calc_goal_h(self) -> float:
```
ronuchit, line 1032: is this copied from pyperplan?

src/settings.py
```diff
@@ -57,7 +57,7 @@ class GlobalSettings:
     random_options_max_tries = 100
 
     # SeSamE parameters
-    task_planning_heuristic = "hadd"  # hadd or hmax
+    task_planning_heuristic = "hadd"  # hadd or hmax or hff
```
ronuchit, line 60: based on your comment, let's just make this env-specific for the sake of progress? i'd be fine with that

ronuchit: predicate invention envs are all using hadd, while amber project is using hff, so it's not an issue paper-wise

tomsilver:
@ronuchit stop working! you're on vacation!

src/utils.py
tomsilver: mostly yeah, a few very minor changes for style

src/settings.py
tomsilver: sounds good

PR #218, opened 2021-12-30
 hmax and hff score functions for grammar search 

tomsilver:
allow grammar search score functions to use hmax, hff, etc., instead of just hadd.

## notes from `hff_lookahead` experiments

I spent a long time trying to understand why an apparently spurious feature in Cover was getting selected by `hff_lookahead`. the feature is: `NOT-Forall[0:target].[NOT-IsTarget(0)]` which is always true.

after digging, my best understanding is that hFF is a fundamentally nondeterministic algorithm, and when we resolve the nondeterminism arbitrarily, as a function of the predicates/operators, (e.g. using `sorted` like here: https://github.com/Learning-and-Intelligent-Systems/predicators/pull/217/commits/b99c0813b7cd4ac17d181f7d731e216fd6352603) the hFF values can vary drastically, and you can end up preferring predicates/operators that just happened to have a "lucky" tiebreak.

for example...
* before I fixed the nondeterminism in the linked commit, I was seeing the score for `hff_lookahead` given all ground truth predicates jump between ~15 and ~105. after, it happened to settle at ~15.
* the `hff_lookahead` score for all ground truth predicates + the spurious predicate above is about ~5.
* however, when I change the `sorted(...)` to `sorted(..., reverse=True)`, the score for ground truth + spurious jumps up to ~16.

so possible conclusion from all this is: "hff considered harmful" and we should not use it in the grammar search.


src/approaches/grammar_search_invention_approach.py
```diff
@@ -750,28 +758,28 @@ def _generate_heuristic(self, init_atoms: Set[GroundAtom],
...
                             ground_ops: Set[_GroundSTRIPSOperator]
                             ) -> Callable[[Set[GroundAtom]], float]:
-        hadd_fn = utils.create_heuristic("hadd", init_atoms, goal, ground_ops)
+        h_fn = utils.create_heuristic(self.heuristic_name, init_atoms, goal,
+                                      ground_ops)
         del init_atoms  # unused after this
         cache: Dict[Tuple[FrozenSet[GroundAtom], int], float] = {}
-        def _hadd_fn_h(atoms: Set[GroundAtom],
-                       depth: int=0) -> float:
+        def _fn_h(atoms: Set[GroundAtom], depth: int=0) -> float:
```
ronuchit, line 765: rename to `_relaxation_h`

src/approaches/grammar_search_invention_approach.py
```diff
@@ -807,19 +815,19 @@ def _task_planning_h(atoms: Set[GroundAtom]) -> float:
...
 
 
 @dataclass(frozen=True, eq=False, repr=False)
-class _HAddHeuristicLookaheadBasedScoreFunction(
-        _HAddHeuristicBasedScoreFunction,
+class _RelaxationHeuristicLookaheadBasedScoreFunction(
+        _RelaxationHeuristicBasedScoreFunction,
         _HeuristicLookaheadBasedScoreFunction):
-    """Implement _generate_heuristic() with HAdd and
+    """Implement _generate_heuristic() with delete relaxation heuristic and
```
ronuchit, line 830: with a*

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -479,7 +491,7 @@ def _evaluate_atom_trajectory(
             return heuristic_fn(atoms_sequence[0])
 
     score_function = _MockHAddLookahead(initial_predicates, atom_dataset,
-        train_tasks, candidates, lookahead_depth=1)
+        train_tasks, candidates, "hadd", lookahead_depth=1)
     assert score_function.evaluate(set(candidates)) == float("inf")
```
ronuchit, line 495: Let's add tests for the two new approaches? We can just copy the first few tests from `def test_hadd_lookahead_score_function():` (before the depth stuff)

PR #220, opened 2022-01-01
get rid of redundant code via smarter mypy typing

ronuchit:
desperately needed...

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from functools import cached_property, lru_cache
 from typing import Dict, Iterator, List, Sequence, Callable, Set, Collection, \
-    Tuple, Any, cast, FrozenSet, DefaultDict, Optional
+    Tuple, Any, cast, FrozenSet, DefaultDict, Optional, TypeVar
 import numpy as np
 from gym.spaces import Box
 from numpy.typing import NDArray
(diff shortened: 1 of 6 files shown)


src/structs.py
```diff
@@ -1001,3 +1002,9 @@ def add(self, member: Tuple[Segment, ObjToVarSub]) -> None:
 ObjToVarSub = Dict[Object, Variable]
 VarToObjSub = Dict[Variable, Object]
 Metrics = DefaultDict[str, float]
+LiftedOrGroundAtom = TypeVar(
```
ronuchit, line 1005: Note: it's important to do it this way instead of e.g. TypeVar("LiftedOrGroundAtom", bound=Union[LiftedAtom, GroundAtom]). Don't ask me why...

ronuchit:
Also updated all the methods in utils.py that don't actually need to take in NSRTs to be able to take in either NSRTs or operators.

ronuchit:
With this PR, we don't have/need "Union"s anymore in the mypy typing, hooray!

ronuchit:
@tomsilver any other code we want to apply this change to?

src/structs.py
tomsilver: i was curious and found this: https://stackoverflow.com/questions/59933946/difference-between-typevart-a-b-and-typevart-bound-uniona-b

tomsilver (approved):
love it!

tomsilver:
> @tomsilver any other code we want to apply this change to?

not that I know of

PR #222, opened 2022-01-02
Multistep cover learned equivalent ground truth

ashayathalye:
This PR adds two more options to cover_multistep_options: LearnedEquivalentPick and LearnedEquivalentPlace. These options use the same parameterization that the options learned through _SimpleOptionLearner use. Corresponding "ground truth" samplers for these two options are added to oracle_approach. A few minor changes are made to update the control flow for how the cover NSRTs are defined in oracle_approach, as well as a few runtime bug fixes such as the cover_multistep_option non-learned-equivalent samplers not properly doing relative positioning (it was over-constraining the support of the uniform distribution).


tomsilver (commented):
@ashayathalye looking very good, thank you! let's do what you mentioned on slack as part of this PR -- refactor the oracle covers stuff to improve readability.

ronuchit:
I'll review after @tomsilver 's suggested refactor!

src/envs/cover.py
```diff
@@ -631,7 +655,7 @@ def _create_initial_state(self, rng: np.random.Generator) -> State:
             # [is_block, is_target, width, x]
             data[target] = np.array([0.0, 1.0, width, x])
         # [x, y, grip, holding]
-        data[self._robot] = np.array([0.0, self.initial_robot_y, 0.0, -1.0])
+        data[self._robot] = np.array([0.0, self.initial_robot_y, -0.1, -1.0])
```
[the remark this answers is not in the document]
ashayathalye: If the grip value is greater than 0, the gripper is in the ON position. If it is less than or equal to 0, the gripper is in the OFF position. I thought it's less confusing if the value we set is not the threshold value, since someone may not remember if the logic is is > or >= to set it to ON. 

I now changed it to +1.0 and -1.0 instead of +/- 0.1, as 0.1 is more arbitrary than +/- 1.0

src/approaches/oracle_approach.py
```diff
@@ -169,20 +209,45 @@ def pick_sampler(state: State, rng: np.random.Generator,
...
+                else len(objs) == 2
+            block, robot, target = objs
+            assert block.is_instance(block_type)
+            assert robot.is_instance(robot_type)
+            assert target.is_instance(target_type)
+            tx, tw = state.get(target, "x"), state.get(target, "width")
+            desired_x = rng.uniform(tx-tw/2, tx+tw/2)
+            desired_x = float(tx)
+            bw, bh = state.get(block, "width"), state.get(block, "height")
+            desired_y = bh + 1e2
```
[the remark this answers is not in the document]
ashayathalye: Good point, I added checks as you suggested.

src/approaches/oracle_approach.py
```diff
@@ -93,96 +88,167 @@ def _get_options_by_names(env_name: str,
     return _get_from_env_by_names(env_name, names, "options")
 
 
-def _get_cover_gt_nsrts(options_are_typed: bool,
-                        include_robot_in_holding: bool = False,
-                        place_sampler_relative: bool = False) -> Set[NSRT]:
-    """Create ground truth NSRTs for CoverEnv.
+def _get_cover_gt_nsrts() -> Set[NSRT]:
```
ronuchit, line 91: Thank you for refactoring this! It had been annoying to me how complicated it was getting. Looks really clean now!!

src/approaches/oracle_approach.py
```diff
@@ -93,96 +88,167 @@ def _get_options_by_names(env_name: str,
...
+            # x, y, grip, holding
+            # grip changes from -1.0 to 1.0
+            # holding changes from -1 to 1
+            robot_param = [desired_x, by, 1.0, 1.0]
+            param = block_param + robot_param
+            return np.array(param, dtype=np.float32)
+    else:
+        def pick_sampler(state: State, rng: np.random.Generator,
+                         objs: Sequence[Object]) -> Array:
+            assert len(objs) == 2 if CFG.env == "cover_multistep_options" \
```
ronuchit, line 166: A bit confusing, would prefer it written out normally:
```
if CFG.env == "cover_multistep_options":
    assert len(objs) == 2
else:
    assert len(objs) == 1
```

src/approaches/oracle_approach.py
```diff
@@ -93,96 +88,167 @@ def _get_options_by_names(env_name: str,
...
-            ub = float(state.get(t, "width")/2)  # relative positioning only
-        else:
-            lb = float(state.get(t, "pose") - state.get(t, "width")/10)
-            lb = max(lb, 0.0)
-            ub = float(state.get(t, "pose") + state.get(t, "width")/10)
-            ub = min(ub, 1.0)
-        return np.array(rng.uniform(lb, ub, size=(1,)), dtype=np.float32)
+        def place_sampler(state: State, rng: np.random.Generator,
+                          objs: Sequence[Object]) -> Array:
+            assert len(objs) == 3 if CFG.env == "cover_multistep_options" \
```
ronuchit, line 238: A bit confusing, would prefer it written out normally:
```
if CFG.env == "cover_multistep_options":
    assert len(objs) == 3
else:
    assert len(objs) == 2
```

tests/envs/test_cover.py
```diff
@@ -172,8 +172,9 @@ def test_cover_multistep_options():
             assert len(obj.type.feature_names) == len(task.init[obj])
     # Predicates should be {IsBlock, IsTarget, Covers, HandEmpty, Holding}.
     assert len(env.predicates) == 5
-    # Options should be {Pick, Place}.
-    assert len(env.options) == 2
+    # Options should be {Pick, Place, LearnedEquivalentPick,
+    # LearnedEquivalentPlace}.
```
ronuchit, line 176: indent this line to align with `Pick` in the previous line

ronuchit (approved):
LGTM! just a few minor comments. but since i generally don't know what's going on, please wait for @tomsilver's approval before merging :)

src/approaches/oracle_approach.py
```diff
@@ -93,96 +88,167 @@ def _get_options_by_names(env_name: str,
...
-            ub = min(ub, 1.0)
-        return np.array(rng.uniform(lb, ub, size=(1,)), dtype=np.float32)
+
+    if CFG.env in ("cover", "cover_hierarchical_types"):
+        option = PickPlace
+        option_vars = []
+    elif CFG.env in ("cover_typed_options", "cover_multistep_options"):
+        option = Pick
+        option_vars = [block]
+    if CFG.env == "cover_multistep_options" and \
```
tomsilver, line 137: minor, can you reorder these so that we don't end up creating and then immediately overriding `option` and `option_vars`? one possibility is to move this case up before the previous one, so:

```
if CFG.env in ("cover", "cover_hierarchical_types"):
...
elif CFG.env == "cover_multistep_options" and CFG.cover_multistep_use_learned_equivalents:
...
else:
    assert CFG.env in ("cover_typed_options", "cover_multistep_options")
```

this comment also applies to a similar pattern with Place

tomsilver (approved):
excellent work! we can merge after addressing the few minor points

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -52,14 +54,18 @@ def _test_approach(env_name, approach_name, excluded_predicates="",
...
 
 def test_nsrt_learning_approach():
     """Tests for NSRTLearningApproach class.
     """
     _test_approach(env_name="cover", approach_name="nsrt_learning")
+    # Sampler learning requires more data than we are allowing for fast
+    # unit tests, so don't try solving.
     _test_approach(env_name="cover_multistep_options",
-                   approach_name="nsrt_learning")
+                   approach_name="nsrt_learning", try_solving=False)
```
ronuchit, line 68: After you rebase to most recent master, I'd recommend also adding `sampler_learner="random"` in addition to `try_solving=False` for extra speed.

ronuchit (approved):
LGTM

PR #230, opened 2022-01-04
do not modify frozen groundnsrts with notcausesfailure

tomsilver:
currently we modify ground NSRTs in-place in the failure discovery part of planning, even though ground NSRTs are frozen. @czeng6 has started encountering issues with this, so this PR is meant to fix the hack.

### DIFF
--- a/src/planning.py
+++ b/src/planning.py
@@ -304,7 +304,7 @@ def _update_nsrts_with_failure(
             if ground_nsrt == discovered_failure.failing_nsrt:
                 new_ground_nsrt = ground_nsrt.copy_with(
                     preconditions=ground_nsrt.preconditions | {atom})
-            # Update the effects of all nsrts that use this object.
+            # Update the effects of all NSRTs that use this object.
             # Note that this is an elif rather than an if, because it would
             # never be possible to use the failing NSRT's effects to set
             # the _NOT_CAUSES_FAILURE precondition.
--- a/src/planning.py
+++ b/src/planning.py
@@ -92,7 +92,8 @@ def sesame_plan(task: Task,
                     return plan, metrics
         except _DiscoveredFailureException as e:
             metrics["num_failures_discovered"] += 1
-            _update_nsrts_with_failure(e.discovered_failure, ground_nsrts)
+            ground_nsrts = _update_nsrts_with_failure(e.discovered_failure,
+                                                      ground_nsrts)
 
 
 def task_plan(init_atoms: Set[GroundAtom],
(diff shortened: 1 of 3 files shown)


src/planning.py
```diff
@@ -287,19 +288,33 @@ def _run_low_level_search(
...
-        # Update the effects of all nsrts that use this object.
-        for nsrt in ground_nsrts:
-            if obj in nsrt.objects:
-                nsrt.add_effects.add(atom)
+        for ground_nsrt in ground_nsrts:
+            # Update the preconditions of the failing NSRT.
+            if ground_nsrt == discovered_failure.failing_nsrt:
+                new_ground_nsrt = ground_nsrt.copy_with(
+                    preconditions=ground_nsrt.preconditions | {atom})
+            # Update the effects of all nsrts that use this object.
```
ronuchit, line 307: NSRTs*

src/planning.py
```diff
@@ -287,19 +288,33 @@ def _run_low_level_search(
...
-    DiscoveredFailure.
+    DiscoveredFailure. Returns a new list of ground NSRTs to replace the input
+    one, where all ground NSRTs that need modification are replaced with new
+    ones (because _GroundNSRTs are frozen).
     """
+    new_ground_nsrts = []
     for obj in discovered_failure.env_failure.offending_objects:
         atom = GroundAtom(Predicate(_NOT_CAUSES_FAILURE, [obj.type],
                                     _classifier=lambda s, o: False), [obj])
-        # Update the preconditions of the failing NSRT.
```
ronuchit, line 297: How was any of this working before if NSRTs are frozen?

src/planning.py
```diff
@@ -287,19 +288,33 @@ def _run_low_level_search(
...
-        for nsrt in ground_nsrts:
-            if obj in nsrt.objects:
-                nsrt.add_effects.add(atom)
+        for ground_nsrt in ground_nsrts:
+            # Update the preconditions of the failing NSRT.
+            if ground_nsrt == discovered_failure.failing_nsrt:
+                new_ground_nsrt = ground_nsrt.copy_with(
+                    preconditions=ground_nsrt.preconditions | {atom})
+            # Update the effects of all nsrts that use this object.
+            # Note that this is an elif rather than an if, because it would
```
ronuchit, line 308: This is a slight logic change, but it makes sense to me. Verify that cluttered_table continues to work (`python src/main.py --env cluttered_table --approach oracle --seed 0` should give >45/50).

ronuchit (approved):

src/planning.py
tomsilver: because even though the dataclass itself is marked as frozen, you can change the sets that are attributes of the dataclass instance and python won't realize

src/planning.py
tomsilver: yup I did

PR #231, opened 2022-01-04
switch from GBFS to hill climbing

ronuchit:


tomsilver (approved):
good stuffs. i actually did review closely this time but find no faults!

PR #232, opened 2022-01-04
switch grammar_search_score_function to hff_lookahead_depth1

ronuchit:
now with fancy regex

### DIFF
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -179,12 +179,12 @@ def _main() -> None:
     ]
     score_function_names = [
         "prediction_error",
-        "hadd_lookahead",
+        "hadd_lookahead_depth0",
         "exact_lookahead",
         "hadd_lookahead_depth1",
         "hadd_lookahead_depth2",
     ]
-    run_planning = True
+    run_planning = False
 
     outdir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                           "results")
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -184,7 +184,7 @@ def _main() -> None:
         "hadd_lookahead_depth1",
         "hadd_lookahead_depth2",
     ]
-    run_planning = False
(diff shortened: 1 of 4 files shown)


src/approaches/grammar_search_invention_approach.py
```diff
@@ -470,23 +471,13 @@ def _create_score_function(
     if score_function_name == "hadd_match":
         return _RelaxationHeuristicMatchBasedScoreFunction(
             initial_predicates, atom_dataset, train_tasks, candidates, "hadd")
-    if score_function_name == "hadd_lookahead":
+    match = re.match(r"(\w+)_lookahead_depth(\d+)", score_function_name)
```
tomsilver, line 474: add a comment describing what values are supported

src/settings.py
```diff
@@ -120,7 +120,7 @@ class GlobalSettings:
     grammar_search_pred_complexity_weight = 1
     grammar_search_max_predicates = 50
     grammar_search_predicate_cost_upper_bound = 6
-    grammar_search_score_function = "hadd_lookahead_depth1"
+    grammar_search_score_function = "hff_lookahead_depth1"
```
tomsilver, line 122: i'm still nervous about using hff with lookahead, but I guess we're going to immediately investigate this so it's fine

analysis/grammar_search_analysis.py
```diff
@@ -179,12 +179,12 @@ def _main() -> None:
...
     score_function_names = [
         "prediction_error",
-        "hadd_lookahead",
+        "hadd_lookahead_depth0",
         "exact_lookahead",
         "hadd_lookahead_depth1",
         "hadd_lookahead_depth2",
     ]
-    run_planning = True
+    run_planning = False
```
tomsilver, line 187: is this intentional?

tomsilver (approved):

src/settings.py
ronuchit: yup, exactly. it's the only one that's doing the right thing in painting

PR #233, opened 2022-01-05
add flags for grammar generation; consolidate get_config_path_str() and get_save_path_str(); prep for supercloud

ronuchit:


analysis/run_supercloud_experiments.sh
```diff
@@ -2,22 +2,32 @@
...
+    # blocks
+    python $FILE --env blocks --approach oracle --seed $SEED
+    python $FILE --env blocks --approach nsrt_learning --seed $SEED
+    python $FILE --env blocks --approach grammar_search_invention --seed $SEED
+    python $FILE --env blocks --approach grammar_search_invention --seed $SEED --excluded_predicates all
+
+    # painting
+    python $FILE --env painting --approach oracle --seed $SEED
+    python $FILE --env painting --approach nsrt_learning --seed $SEED
+    python $FILE --env painting --approach grammar_search_invention --seed $SEED --grammar_search_grammar_includes_givens 0 --grammar_search_grammar_includes_foralls 0
```
tomsilver, line 23: Just checking, will it be problematic that the save config str thing does not take into account commandline overrides?

tomsilver (approved):
Looks good. Feel free to ignore, but I think it would be nice to dump two things into the logs (i.e. print) for our future reference:
1. The full CFG (are we doing this already?)
2. The git commit (assuming that we are always going to run with a clean git diff on supercloud)

analysis/run_supercloud_experiments.sh
ronuchit: as we usually do, the responsibility is on the user to not override their own results files. if you want to run different experiments and make sure everything is saved out, you'll need to add fields to the save/config str.

ronuchit:
> Looks good. Feel free to ignore, but I think it would be nice to dump two things into the logs (i.e. print) for our future reference:
> 
> 1. The full CFG (are we doing this already?)
> 2. The git commit (assuming that we are always going to run with a clean git diff on supercloud)

Done, thanks for the suggestion.

PR #240, opened 2022-01-05
rewrite nsrt_learning.py from scratch

ronuchit:


src/nsrt_learning.py
```diff
@@ -146,118 +149,135 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
+        for nid in nids:
             # Try to unify this transition with existing effects.
             # Note that both add and delete effects must unify,
             # and also the objects that are arguments to the options.
-            part_param_option = parameterized_options[i]
-            part_option_vars = option_vars[i]
-            part_add_effects = add_effects[i]
-            part_delete_effects = delete_effects[i]
+            (nid_param_option, nid_option_vars) = nid.option_spec
+            nid_add_effects = nid.op.add_effects
```
tomsilver, line 169: can we just plug these in directly, like we do for segment.add_effects etc, rather than making new variables?

src/nsrt_learning.py
```diff
@@ -309,3 +329,40 @@ def unify_effects_and_options(
             f_new_ground_delete_effects,
         f_lifted_option_args | f_new_lifted_add_effects | \
             f_new_lifted_delete_effects)
+
+
+@dataclass(repr=False)
+class _NSRTIntermediateData:
```
tomsilver, line 335: hmm I really don't know how to think about this class. I think the fact that we have to call it NID is indicative of it not being a semantically coherent thing :P it's somewhere between a Partition and a partially-formed NSRT, and it's confusing for me to keep the relationship between all of these things straight.

I see the motivation though. can we talk about alternatives? some ideas:

1. make NSRTs not always frozen, but add a `freeze` method on them, which we call at the end of learning, and after which we can treat them as frozen. then build up NSRTs piece by piece, starting with just the effects, then adding the preconditions, then the option specs, etc. We could potentially get rid of `STRIPSOperator` if we do this....
2. extend Partition so that it basically becomes what NID is now.
3. others...

tomsilver (commented):
(will look at the rest after we resolve the big point)

src/structs.py
```diff
@@ -1025,10 +1016,11 @@ def add(self, member: Tuple[Segment, ObjToVarSub]) -> None:
 Image = NDArray[np.uint8]
 Video = List[Image]
 Array = NDArray[np.float32]
-NSRTSampler = Callable[[State, np.random.Generator, Sequence[Object]], Array]
 PyperplanFacts = FrozenSet[Tuple[str, ...]]
 ObjToVarSub = Dict[Object, Variable]
 VarToObjSub = Dict[Variable, Object]
+NSRTDatastore = List[Tuple[Segment, ObjToVarSub]]
```
tomsilver, line 1022: this name is more precise than `Datastore`, but I'm getting confused between `NSRTDatastore` and PNAD. can we change this to just `Datastore` (or something else that is just more distinct from PNAD)?

src/structs.py
```diff
@@ -952,70 +952,61 @@ def set_option(self, option: _Option) -> None:
...
                 option = seg.get_option()
                 part_param_option, part_option_args = self.option_spec
                 assert option.parent == part_param_option
                 option_args = [sub[o] for o in option.objects]
                 assert option_args == part_option_args
         # Add to members.
-        self.members.append(member)
+        self.datastore.append(member)
+
+    def get_nsrt(self) -> NSRT:
```
tomsilver, line 996: rename to `make_nsrt`

src/nsrt_learning.py
```diff
@@ -34,52 +33,41 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
...
     #         are predicates whose truth value becomes unknown (for *any*
     #         grounding not explicitly in effects) upon operator application.
     if CFG.learn_side_predicates:
-        _learn_nid_side_predicates(nids)
+        _learn_pnad_side_predicates(pnads)
 
-    # STEP 5: Learn options (option_learning.py) and update NIDs.
-    _learn_nid_options(nids)
+    # STEP 5: Learn options (option_learning.py) and update PNADs.
+    _learn_pnad_options(pnads)
```
tomsilver, line 52: so PNADs are created before option learning, but PNADs have (not optional) option specs in them. before option learning, I guess these will just be default options, but is that weird? should be just make it optional? not sure

src/nsrt_learning.py
```diff
@@ -34,52 +33,41 @@ def learn_nsrts_from_data(dataset: Dataset, predicates: Set[Predicate],
...
     return set(nsrts)
 
 
-def learn_strips_operators(segments: Sequence[Segment], verbose: bool = True,
-                           ) -> Tuple[List[STRIPSOperator], List[Partition]]:
-    """A light wrapper around _get_initial_nids() that returns the operators
-    and partitions themselves, rather than exposing the internal data structure.
+def learn_strips_operators(segments: Sequence[Segment], verbose: bool = True
+                           ) -> List[PartialNSRTAndDatastore]:
+    """An alias for _get_initial_pnads.
```
tomsilver, line 68: I'm confused, are you just renaming the function then? what's the purpose?

src/nsrt_learning.py
```diff
@@ -149,10 +137,10 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
     return segments
 
 
-def _get_initial_nids(segments: Sequence[Segment], verbose: bool
-                      ) -> List[_NSRTIntermediateData]:
-    # Partition the segments according to common effects.
-    nids = []
+def _get_initial_pnads(segments: Sequence[Segment], verbose: bool
```
tomsilver, line 140: `create` instead of `get`? to me, `get` has the connotation of a fast lookup

src/nsrt_learning.py
```diff
@@ -220,64 +207,65 @@ def _get_initial_nids(segments: Sequence[Segment], verbose: bool
                 preconditions = lifted_atoms
             else:
                 preconditions &= lifted_atoms
-        nid.op = STRIPSOperator(
-            nid.op.name, nid.op.parameters, preconditions, nid.op.add_effects,
-            nid.op.delete_effects, nid.op.side_predicates)
+        pnad.op = STRIPSOperator(
```
tomsilver, line 210: it's pretty weird to create an initial STRIPSOperator before this and then destroy it now. can we avoid this somehow, and create the STRIPSOperator just once?

tomsilver: for example, keep the effects in the PNAD, and then add preconditions, and then call `make_strips_operator` on the PNAD or something?

tomsilver (commented):
looks better thanks. I would like to make the unusual request that we sleep on this and look at it with fresh eyes again tomorrow before merging, since it's a pretty big change, it's pretty late for me, and I don't yet have the full feeling of catharsis that we sometimes get with refactors like this (so maybe there are still opportunities for improved clarity out there)

ronuchit:
Sounds good. I'll incorporate your feedback and keep the PR ready to merge but will leave it alone after that. Then, you can feel free to merge it in tomorrow morning, or make any additional changes you want directly.

src/structs.py
ronuchit: changed to Datastore

src/nsrt_learning.py
ronuchit: Yup. I tried to make that clear in "STEP 3" near the top of nsrt_learning.py. I tried a bit to make it an optional field, but that made things weird with the actual clustering loop, since anyway we need a dummy option + specs for the purposes of running `unify_effects_and_options`. Don't think it's worth it, imo.

src/nsrt_learning.py
ronuchit: fixed

PR #241, opened 2022-01-05
Playroom simulator bug fix

amburger66:
- robot should not be able to do some other action (e.g. picking a block, opening a door, toggling the dial) after moving to a new location

### DIFF
--- a/tests/envs/test_playroom.py
+++ b/tests/envs/test_playroom.py
@@ -329,7 +329,6 @@ def test_playroom_action_sequence_video():
     env.seed(123)
     # Run through a specific plan of low-level actions.
     task = next(env.train_tasks_generator())[0]
-    print("task init:", task.init)
     action_arrs = [
         # Pick up a block
         np.array([11.8, 18, 0.45, -0.15, 0]).astype(np.float32),
--- a/tests/envs/test_playroom.py
+++ b/tests/envs/test_playroom.py
@@ -200,8 +200,8 @@ def test_playroom_simulate_doors_and_dial():
             dial = item
     assert robot is not None
     assert dial is not None
-    # Move somewhere without interacting with anything
-    act = Action(np.array([25, 25, 1, 0, 1]).astype(np.float32))
+    # Move to boring room door
+    act = Action(np.array([29.6, 12, 1, 0, 1]).astype(np.float32))
     next_state = env.simulate(state, act)
     for o in state:
         if o.type != robot_type:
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
thanks! lgtm, but can you add a test or two that verifies that nothing happens if the robot tries to pick and move in the same timestep (as you described being disallowed in the PR comment)?

tests/envs/test_playroom.py
```diff
@@ -213,6 +213,13 @@ def test_playroom_simulate_doors_and_dial():
     assert np.any(state[robot] != next_state[robot])
     assert np.any(state[door1] != next_state[door1])
     state = next_state
+    # Move to playroom door
+    act = Action(np.array([110.2, 15, 3, 1, 1]).astype(np.float32))
+    next_state = env.simulate(state, act)
+    for o in state:
+        if o.type != robot_type:
+            assert np.all(state[o] == next_state[o])
```
ronuchit, line 221: Might want to use allclose here for safety against precision issues? `np.allclose(state[o], next_state[o])`

ronuchit: Do you want to also add an `else: assert not np.allclose(state[o], next_state[o])`?

tests/envs/test_playroom.py
```diff
@@ -222,6 +229,13 @@ def test_playroom_simulate_doors_and_dial():
     act = Action(np.array([105, 15, 3, 1, 1]).astype(np.float32))
     with pytest.raises(EnvironmentFailure):
         next_state = env.simulate(state, act)
+    # Move to dial
+    act = Action(np.array([125, 15.1, 1, -0.5, 1]).astype(np.float32))
+    next_state = env.simulate(state, act)
+    for o in state:
+        if o.type != robot_type:
+            assert np.all(state[o] == next_state[o])
```
ronuchit, line 237: same as above

tests/envs/test_playroom.py
amburger66: good idea, thanks

amburger66:
> thanks! lgtm, but can you add a test or two that verifies that nothing happens if the robot tries to pick and move in the same timestep (as you described being disallowed in the PR comment)?

the tests are already there! I will label them to be more clear about it

tomsilver:
> > thanks! lgtm, but can you add a test or two that verifies that nothing happens if the robot tries to pick and move in the same timestep (as you described being disallowed in the PR comment)?
> 
> the tests are already there! I will label them to be more clear about it

oh just me being dumb, no need. thanks :)

PR #245, opened 2022-01-06
Check oracle operator parameters

amburger66:
- new test `test_check_nsrt_parameters()` makes sure 1) option vars are a subset of nsrt parameters, and 2) every parameter appears in the nsrt's option vars or operator add/delete effects
- corresponding fixes to playroom oracle nsrts + tests


tests/approaches/test_oracle_approach.py
```diff
@@ -62,6 +63,37 @@ def test_get_gt_nsrts():
...
+    envs = {"cover": CoverEnv(), "cover_typed_options": CoverEnvTypedOptions(),
+            "cover_hierarchical_types": CoverEnvHierarchicalTypes(),
+            "cluttered_table": ClutteredTableEnv(), "blocks": BlocksEnv(),
+            "painting": PaintingEnv(), "playroom": PlayroomEnv(),
+            "cover_multistep_options": CoverMultistepOptions(),
+            "repeated_nextto": RepeatedNextToEnv()}
+    for name, env in envs.items():
+        utils.update_config({"env": name})
+        nsrts = get_gt_nsrts(env.predicates, env.options)
+        check_nsrt_parameters(nsrts)  # pylint: disable=protected-access
```
tomsilver, line 94: hmm why is this pylint disable needed?

tests/approaches/test_oracle_approach.py
```diff
@@ -62,6 +63,37 @@ def test_get_gt_nsrts():
         get_gt_nsrts(set(), set())
 
 
+def test_check_nsrt_parameters():
+    """Checks all of the oracle operators for all envs.
+    """
+    def check_nsrt_parameters(nsrts: Set[NSRT]) -> None:
```
tomsilver, line 69: let's just move this function out (e.g. put it on line 64) and make it private

tests/approaches/test_oracle_approach.py
```diff
@@ -62,6 +63,37 @@ def test_get_gt_nsrts():
...
+            precond_vars: Set[Variable] = set()
+            for lifted_atom in nsrt.add_effects:
+                effects_vars |= set(lifted_atom.variables)
+            for lifted_atom in nsrt.delete_effects:
+                effects_vars |= set(lifted_atom.variables)
+            for lifted_atom in nsrt.preconditions:
+                precond_vars |= set(lifted_atom.variables)
+            assert set(nsrt.option_vars).issubset(nsrt.parameters), \
+                f"Option variables is not a subset of parameters in {nsrt}"
+            for var in nsrt.parameters:
```
tomsilver, line 81: let's add one more test that the parameters are equal to the union of all four things

src/approaches/oracle_approach.py
```diff
@@ -851,9 +853,11 @@ def movedoortodial_sampler(state: State, rng: np.random.Generator,
     door = Variable("?door", door_type)
     region = Variable("?region", region_type)
     parameters = [robot, dial, door, region]
-    option_vars = [robot]
+    option_vars = [robot, door, region]
```
tomsilver, line 856: if you still have trouble with learning, one reason may be that if we add these option vars, they need to be used somehow by the option itself. otherwise, in the data, you could see strange things like `Move(robot1, door1, region1)` resulting in a change to `door5` and `region6`. probably the easiest way to prevent this would be to change the `initiable` for the options so that they use the option vars somehow. for example, you could check that `door` and `region` are in the playroom, in this case.

tomsilver (commented):
looking good!

tests/approaches/test_oracle_approach.py
amburger66: oops forgot to take it out

ronuchit (approved):
lgtm, this is a great check to add, thank you so much!! feel free to convert from draft to normal PR!

PR #248, opened 2022-01-06
includes modifications to __init__.py to support env instance caching

NishanthJKumar:
small PR for change necessary in #103


NishanthJKumar (commented):
LGTM!

tomsilver (approved):

src/envs/__init__.py
```diff
@@ -51,3 +54,23 @@ def create_env(name: str) -> BaseEnv:
...
+
+def create_env(name: str) -> BaseEnv:
+    """Create an environment given its name.
+    """
+    env = _create_new_env_instance(name)
+    _MOST_RECENT_ENV_INSTANCE[name] = env
+    return env
+
+
+def get_env_instance(name: str) -> BaseEnv:
```
ronuchit, line 67: I'm a bit confused why we need this. Why can't the user of `create_env` just keep track of the returned environment instance themselves, rather than relying on this cache? Would it make more sense to have `create_env` just return what's in the cache if something already exists, rather than overwriting the cache?

tests/envs/test_base_env.py
```diff
@@ -39,13 +40,15 @@ def test_base_env():
...
-    """Tests for create_env.
+    """Tests for create_env() and get_env_instance().
     """
     for name in ["cover", "cover_typed_options", "cover_hierarchical_types",
                  "cluttered_table", "blocks", "playroom", "painting",
                  "repeated_nextto"]:
         env = create_env(name)
         assert isinstance(env, BaseEnv)
+        other_env = get_env_instance(name)
+        assert env == other_env
```
ronuchit, line 51: change to `assert env is other_env` since they're the literal same object

src/envs/__init__.py
ronuchit: It's confusing to have `get_env_instance` be able to create a new instance, because that's redundant with the functionality of `create_env`. I'd like to propose renaming `get_env_instance` to `get_cached_env_instance` with the following implementation:
```
assert name in _MOST_RECENT_ENV_INSTANCE
return _MOST_RECENT_ENV_INSTANCE[name]
```

The user would be responsible for making sure they call `create_env` before ever calling `get_cached_env_instance`, and then it's clear that these two methods have non-overlapping functionalities.

tomsilver: thanks, the new version looks good. to your first question, that would've been another way to go, but we would need to change `get_gt_nsrts` so that it takes an environment instance as an argument. what we have now seems at least as good to me.

src/envs/__init__.py
```diff
@@ -51,3 +54,23 @@ def create_env(name: str) -> BaseEnv:
...
+
+def create_env(name: str) -> BaseEnv:
+    """Create an environment given its name.
+    """
+    env = _create_new_env_instance(name)
+    _MOST_RECENT_ENV_INSTANCE[name] = env
+    return env
+
+
+def get_cached_env_instance (name: str) -> BaseEnv:
```
tomsilver, line 67: remove space before parens

src/envs/__init__.py
```diff
@@ -51,3 +54,23 @@ def create_env(name: str) -> BaseEnv:
...
+    """Get the most recent env instance, or make a new one.
+    """
+    assert name in _MOST_RECENT_ENV_INSTANCE
+    # It is not easy to cover this because we don't control the order that
```
tomsilver, line 71: it should now be possible to remove this comment and the pragma below, hopefully

tomsilver (approved):
lgtm after two more minor things

src/envs/__init__.py
ronuchit: that was my secret goal the whole time :)

src/envs/__init__.py
```diff
@@ -51,3 +54,18 @@ def create_env(name: str) -> BaseEnv:
     if name == "repeated_nextto":
         return RepeatedNextToEnv()
     raise NotImplementedError(f"Unknown env: {name}")
+
+
+def create_env(name: str) -> BaseEnv:
+    """Create an environment given its name.
```
ronuchit, line 60: update docstring

src/envs/__init__.py
```diff
@@ -51,3 +54,18 @@ def create_env(name: str) -> BaseEnv:
...
+    """
+    env = _create_new_env_instance(name)
+    _MOST_RECENT_ENV_INSTANCE[name] = env
+    return env
+
+
+def get_cached_env_instance(name: str) -> BaseEnv:
```
ronuchit, line 67: update docstring

ronuchit (changes requested):

ronuchit (approved):
LGTM! hit merge after checks pass~

PR #249, opened 2022-01-06
integrate with pyperplan so we can use all of their heuristics directly

tomsilver:


tomsilver:
@ronuchit I'm still working on this, but there is a mypy issue that I can't figure out -- could you take a look?

tomsilver:
@ronuchit that was hard, finally ready for your review...

the real motivation for this is so that we can use other pyperplan heuristics easily. I haven't tried those, but will after we finish this.

```
Oracle timing before & after

Cover: Before (0.4) After (0.4)
Blocks: Before (7.2) After (7.1)
Painting: Before (5.4) After (5.4)
Playroom: Before (13.2) After (13.7)
Cluttered Table: Before (8.6) After (8.4)
Repeated NextTo: Before (0.7) After (0.7)

Grammar search timing before & after

Cover / all: Before (90.0) After (85.5)
Painting / all (with extra flags): Before (374.5) After (333.1)
```

src/approaches/grammar_search_invention_approach.py
```diff
@@ -653,7 +653,8 @@ def _evaluate_with_operators(self, predicates: FrozenSet[Predicate],
             ground_ops = {op for strips_op in strips_ops for op
```
ronuchit, line 653: Btw, why does grammar search get noticeably faster on painting with this PR? Is it bc of the reachable_nsrts change?

requirements.txt
```diff
@@ -8,3 +8,4 @@ pandas
 torch
 scipy
 tabulate
+pyperplan
```
ronuchit, line 11: Is this installable via `pip`? I remember we had some issues before where we actually needed to install from source? I guess the checks pass so it's okay...

src/utils.py
```diff
@@ -797,303 +832,157 @@ def ops_and_specs_to_dummy_nsrts(strips_ops: Sequence[STRIPSOperator],
 
 
 def create_heuristic(heuristic_name: str,
```
ronuchit, line 834: Let's rename create_heuristic to create_task_planning_heuristic?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -751,10 +753,11 @@ def _generate_heuristic(self, init_atoms: Set[GroundAtom],
                             goal: Set[GroundAtom],
                             strips_ops: Sequence[STRIPSOperator],
                             option_specs: Sequence[OptionSpec],
-                            ground_ops: Set[_GroundSTRIPSOperator]
+                            ground_ops: Set[_GroundSTRIPSOperator],
+                            predicates: Collection[Predicate],
```
ronuchit, line 757: remove comma at end to keep signature same as above

src/planning.py
```diff
@@ -71,13 +66,17 @@ def sesame_plan(task: Task,
         # that initially has empty effects may later have a _NOT_CAUSES_FAILURE.
         nonempty_ground_nsrts = [nsrt for nsrt in ground_nsrts
                                  if nsrt.add_effects | nsrt.delete_effects]
-        if check_dr_reachable and \
-           not utils.is_dr_reachable(nonempty_ground_nsrts, atoms, task.goal):
+        all_reachable_atoms = utils.get_reachable_atoms(nonempty_ground_nsrts,
+                                                        atoms)
+        if check_dr_reachable and not task.goal.issubset(all_reachable_atoms):
             raise ApproachFailure(f"Goal {task.goal} not dr-reachable")
+        reachable_nsrts = [nsrt for nsrt in nonempty_ground_nsrts if
```
ronuchit, line 73: This is an interesting change. I wonder if it has any substantial effect on oracle planning? Is the answer no because we typically don't have any unreachable NSRTs in oracle_approach.py?

src/planning.py
```diff
@@ -122,19 +122,28 @@ def task_plan(init_atoms: Set[GroundAtom],
     ground_nsrts = utils.filter_static_operators(ground_nsrts, init_atoms)
     nonempty_ground_nsrts = [nsrt for nsrt in ground_nsrts
                              if nsrt.add_effects | nsrt.delete_effects]
-    if not utils.is_dr_reachable(nonempty_ground_nsrts, init_atoms, goal):
+    all_reachable_atoms = utils.get_reachable_atoms(nonempty_ground_nsrts,
+                                                    init_atoms)
```
ronuchit, line 126: Let's rename `atoms`-> `init_atoms` in sesame_plan() above, partially for consistency with task_plan() here, and also because it's a more clear name.

src/planning.py
```diff
@@ -30,11 +30,6 @@ class _Node:
     skeleton: List[_GroundNSRT]
     atoms_sequence: List[Collection[GroundAtom]]  # expected state sequence
     parent: Optional[_Node]
-    pyperplan_facts: PyperplanFacts = field(
-        init=False, default_factory=frozenset)
-
-    def __post_init__(self) -> None:
-        self.pyperplan_facts = utils.atoms_to_tuples(self.atoms)
```
ronuchit, line 37: Can we purge atoms_to_tuples from the repo entirely?

src/planning.py
```diff
@@ -149,8 +158,9 @@ def _skeleton_generator(task: Task,
                       atoms_sequence=[init_atoms], parent=None)
     rng_prio = np.random.default_rng(seed)
     heuristic = utils.create_heuristic(CFG.task_planning_heuristic,
-                                       init_atoms, task.goal, ground_nsrts)
-    hq.heappush(queue, (heuristic(root_node.pyperplan_facts),
+                                       init_atoms, task.goal, ground_nsrts,
```
ronuchit, line 161: Lots of effort in this file to get the predicates to this point. But why does the pyperplan heuristic need to be told the predicates? Can't it extract them from the NSRTs via `extract_preds_and_types`?

src/utils.py
```diff
@@ -305,6 +307,32 @@ def get_object_combinations(
         yield list(choice)
 
 
+@functools.lru_cache(maxsize=None)
+def get_all_ground_atoms_for_predicate(predicate: Predicate,
```
ronuchit, line 311: Would this be useful somehow in the side predicate planning stuff? (fine if not)

src/utils.py
```diff
@@ -706,20 +734,27 @@ def filter_static_operators(ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
                for op in ground_ops):
             continue
         static_preds.add(pred)
-    static_facts = {atom for atom in atoms if atom.predicate in static_preds}
+    static_atoms = {atom for atom in atoms if atom.predicate in static_preds}
+    return static_preds, static_atoms
+
+
+def filter_static_operators(ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
```
ronuchit, line 741: These are different from the reachable_nsrts stuff you added, right? Are you clear what the relationship between these things is?

src/utils.py
```diff
@@ -797,303 +832,157 @@ def ops_and_specs_to_dummy_nsrts(strips_ops: Sequence[STRIPSOperator],
...
+                                            predicates, objects,
+                                            static_atoms)
+    pyperplan_heuristic = pyperplan_heuristic_cls(pyperplan_task)
+    pyperplan_goal = _atoms_to_tuples(goal - static_atoms)
+    return _PyperplanHeuristicWrapper(heuristic_name, init_atoms, goal,
+                                      ground_ops, static_atoms,
+                                      pyperplan_heuristic, pyperplan_goal)
+
 
+_PyperplanFacts = FrozenSet[Tuple[str, ...]]
```
ronuchit, line 888: Ah I see. This didn't go away, it just got moved to be internal.

src/utils.py
```diff
@@ -797,303 +832,157 @@ def ops_and_specs_to_dummy_nsrts(strips_ops: Sequence[STRIPSOperator],
...
-    def _accumulate(distances: Collection[float]) -> float:
-        return sum(distances)
+    all_atoms = get_all_ground_atoms(frozenset(predicates), frozenset(objects))
+    # Note: removing static atoms.
+    pyperplan_facts = _atoms_to_tuples(all_atoms - static_atoms)
+    pyperplan_state = _atoms_to_tuples(init_atoms - static_atoms)
+    pyperplan_goal = _atoms_to_tuples(goal - static_atoms)
+    pyperplan_operators = set()
+    for op in ground_ops:
+        # Note: the pyperplan operator must include the objects, because hFF
```
ronuchit, line 958: I don't understand why the "because..." implies this note.

src/utils.py
```diff
@@ -797,303 +832,157 @@ def ops_and_specs_to_dummy_nsrts(strips_ops: Sequence[STRIPSOperator],
...
-    @staticmethod
-    def _accumulate(distances: Collection[float]) -> float:
-        return sum(distances)
+    all_atoms = get_all_ground_atoms(frozenset(predicates), frozenset(objects))
+    # Note: removing static atoms.
+    pyperplan_facts = _atoms_to_tuples(all_atoms - static_atoms)
+    pyperplan_state = _atoms_to_tuples(init_atoms - static_atoms)
+    pyperplan_goal = _atoms_to_tuples(goal - static_atoms)
+    pyperplan_operators = set()
+    for op in ground_ops:
```
ronuchit, line 957: Have the unreachable NSRTs already been removed here?

src/planning.py
ronuchit: Edit: Ah, I see they're used internally in utils.py now.

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -376,20 +376,27 @@ def test_relaxation_lookahead_score_function():
...
-    name_to_pred = {}
-    for p in env.predicates:
-        if p.name in ablated:
-            name_to_pred[p.name] = p
-        else:
-            initial_predicates.add(p)
+    class _MockLookahead(_RelaxationHeuristicLookaheadBasedScoreFunction):
+        """Mock class.
+        """
+
```
ronuchit, line 382: remove blank line

tests/test_utils.py
```diff
@@ -832,26 +833,13 @@ def test_is_dr_reachable():
...
-    assert not utils.is_dr_reachable(
-        ground_nsrts, atoms, {pred2([cup2, plate2])})
-    assert not utils.is_dr_reachable(
-        ground_nsrts, atoms, {pred3([cup2, plate1])})
-    assert not utils.is_dr_reachable(
-        ground_nsrts, atoms, {pred3([cup2, plate2])})
+    reachable_atoms = utils.get_reachable_atoms(ground_nsrts, atoms)
+    assert {pred1([cup1, plate1]), pred1([cup1, plate2]), pred2([cup1, plate1]),
+            pred2([cup1, plate2])}.issubset(reachable_atoms)
+    assert not reachable_atoms & {pred3([cup1, plate1]), pred3([cup1, plate2]),
```
ronuchit, line 839: I got confused by order of operations here. Can you make it more clear how the `not` and the set intersection relate?

tests/test_utils.py
```diff
@@ -1034,183 +1022,25 @@ def test_operator_application():
...
-    heuristic = _HFFHeuristic(initial_state, goals, operators)
-    assert heuristic(initial_state) == 2
-    assert heuristic(goals) == 0
-    # Test unreachable goal.
-    goals = frozenset({("Covers", "block0:block", "target0:target")})
-    heuristic = _HFFHeuristic(initial_state, goals, [])
-    assert heuristic(initial_state) == float("inf")
+        utils.create_heuristic("not a real heuristic", set(), set(), set(),
+                               set(), set())
+    # Cover _Heuristic base class.
```
ronuchit, line 1040: Why did we delete all these tests? Are we not planning to test things like hadd/hmax/hff-lookahead anymore just because we don't have to cover them, since they're no longer in our repo?

Oh, maybe these tests you deleted are on the heuristic themselves which we don't care about, but the lookahead tests are unchanged?

ronuchit (commented):
sorry lots of questions lol

src/utils.py
ronuchit: Are the operators that would get filtered out by static atoms check always a subset of the operators that would get filtered out by reachable set?

src/approaches/grammar_search_invention_approach.py
tomsilver: I don't think it's that, because I'm not using reachable_nsrts in the grammar search (because I was nervous, I think in theory it should be fine...). It might be because of all the filtering out of static atoms that I'm doing now, which we weren't doing before?

requirements.txt
tomsilver: yeah it is, right

src/approaches/grammar_search_invention_approach.py
ronuchit: hmm, maybe we should use it there too... right?

src/planning.py
tomsilver: I didn't see any substantial effect. that's a good possible explanation. what do you think, worth keeping?

src/planning.py
tomsilver: resolved on slack

src/utils.py
tomsilver: resolved on slack (probably not)

src/utils.py
tomsilver: yes, so I will remove `filter_static_operators`

src/utils.py
tomsilver: I'll expand the comment:
```
        # Note: the pyperplan operator must include the objects, because hFF
        # uses the operator name in constructing the relaxed plan, and the
        # relaxed plan is a set. If we instead just used op.name, there would
        # be a very nasty bug where two ground operators in the relaxed plan
        # that have different objects are counted as just one.
```

src/utils.py
tomsilver: yeah

tests/test_utils.py
tomsilver: right, the lookahead tests are unchanged, this is just testing the heuristics themselves. so really we would just be testing pyperplan's code. I can add back the equivalent tests if you want, but it would be annoying because the inputs are no longer relaxed facts, they're ground atoms.

tomsilver:

src/utils.py
```diff
@@ -745,10 +769,14 @@ def extract_preds_and_types(
...
 
-def filter_static_operators(
+def get_static_preds_atoms(
         ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
-        atoms: Collection[GroundAtom]) -> List[GroundNSRTOrSTRIPSOperator]:
-    """Filter out ground operators that don't satisfy static facts."""
+        atoms: Collection[GroundAtom]
+) -> Tuple[Set[Predicate], Set[GroundAtom]]:
+    """Get predicates and atoms that are static w.r.t.
+
```
ronuchit, line 777: autoformat fail! just write out "with respect to" so it doesn't think it's a new sentence?

ronuchit: (i'm curious then if it gets rid of this blank line)

src/utils.py
ronuchit: no idea how you figured that out, great stuff!

tests/test_utils.py
```diff
@@ -928,99 +929,24 @@ def test_static_filtering():
...
+    atoms = {pred1([cup1, plate1]), pred1([cup1, plate2])}
+    reachable_atoms = utils.get_reachable_atoms(ground_nsrts, atoms)
+    assert {
         pred1([cup1, plate1]),
         pred1([cup1, plate2]),
         pred2([cup1, plate1]),
-        pred2([cup1, plate2]),
+        pred2([cup1, plate2])
+    }.issubset(reachable_atoms)
+    assert reachable_atoms & {
```
ronuchit, line 940: What's up with this test, why can't you just check whatever you think reachable_atoms should be with a `==`?

ronuchit (approved):
lgtm, minor

src/utils.py
tomsilver: ha

tomsilver: it changed it to:

```
    """Get predicates and atoms that are static with respect to the
    operators."""
```

(two lines.) is this correct style? note that it's two lines, but the triple quotes are on the same line. I guess if both lint and autoformatting pass, that's the rule of law now? @ronuchit

ronuchit: yeah, it's the rule of law now :) i don't understand the rules for docstring quotations but whatever

Issue #246, opened 2022-01-06
add `oracle` option for `sampler_learner`


tomsilver:
when `do_sampler_learning` is False, we use uniformly random samplers (and so would expect this to do badly in many cases) when `do_option_learning` is False, we use _oracle_ options (and so would expect this to be as good as possible). this is confusing; let's rename one of them. @ronuchit what's your preference?

ashayathalye:
relatedly, it might be nice to have something in the pipeline to make it easy to use oracle / ground truth samplers. Right now I am just hardcoding them into `nsrt_learning.py` for my specific use case.

ronuchit:
@tomsilver Here's my preference for this. Like we do for some other settings, we should consolidate `do_option_learning` and `option_learner` into one flag named the latter, with choices `{"no_learning", "oracle"}`. And then we can do the same with `sampler_learner` having choices `{"neural", "random", "oracle"}`. The last one would be something new as per @ashayathalye's request. However, note that it's not as simple as it might seen initially: if you are learning the symbolic operators but you want to use oracle samplers, then that's only going to work if all your symbolic operators exactly match up with some oarcle operator, so that you can use that oracle operator's sampler. We'll need to throw an error otherwise.

tomsilver:
sounds good to me. thanks!

ronuchit:
first half of this is addressed by #250

PR #250, opened 2022-01-06
remove do_*_learning flags

ronuchit:
addresses the first half of #246


tomsilver (approved):
lgtm after minor, thank you!

PR #251, opened 2022-01-06
clean up mypy typing for unify(), move unify_effects_and_options() to utils

ronuchit:


src/nsrt_learning.py
```diff
@@ -150,7 +150,8 @@ def learn_strips_operators(segments: Sequence[Segment], verbose: bool = True,
             # Note that both add and delete effects must unify,
             # and also the objects that are arguments to the options.
             (pnad_param_option, pnad_option_vars) = pnad.option_spec
-            suc, sub = unify_effects_and_options(
+            suc, ent_to_ent_sub = utils.unify_preconds_effects_options(
```
ronuchit, line 153: Note that this name has to be different from `sub` below which is why I called it `ent_to_ent_sub`. Mypy doesn't like the pattern `x = cast(<different type>, x)`

tomsilver (approved):
You're right, this does make me happy! Just one question -- why did you add preconditions to the unify function? Are you planning to use them in side predicate learning?

ronuchit:
I added preconditions bc we'll need them for oracle sampler learning, which is the motivation for this whole PR. To do oracle sampler learning we'll need to be able to unify (all the components of) learned STRIPS operators with oracle STRIPS operators.

tomsilver:
Makes sense. Thanks!

PR #255, opened 2022-01-07
oracle samplers

### LINKED ISSUE #246: add `oracle` option for `sampler_learner`

ronuchit:
Implement oracle samplers for nsrt_learning pipeline by matching to
ground truth NSRTs in oracle_approach.py. We assert that every ground
truth NSRT must match a learned operator (but not the other way around,
since the pipeline could easily learn extra operators that are unimportant).

Works for cover and blocks, but not for painting because learned operators
don't exactly match ground truth ones.

Switched from pickle to dill because samplers defined in
oracle_approach.py are local functions, which cannot be pickled.
Verified that saving/loading work properly in all cases.

closes #246

### DIFF
--- a/src/sampler_learning.py
+++ b/src/sampler_learning.py
@@ -36,9 +36,10 @@ def learn_samplers(strips_ops: List[STRIPSOperator],
     return samplers
 
 
-def _extract_oracle_samplers(strips_ops: List[STRIPSOperator],
-                             option_specs: List[OptionSpec],
-                             ) -> List[NSRTSampler]:
+def _extract_oracle_samplers(
+    strips_ops: List[STRIPSOperator],
+    option_specs: List[OptionSpec],
+) -> List[NSRTSampler]:
     """Extract the oracle samplers matching the given STRIPSOperator objects
     from the ground truth operators defined in approaches/oracle_approach.py.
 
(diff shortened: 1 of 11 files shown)


src/sampler_learning.py
```diff
@@ -5,22 +5,36 @@
...
     """Learn all samplers for each operator's option parameters."""
     samplers = []
+    if sampler_learner == "oracle":
+        env = create_env(CFG.env)
+        # We don't need to match ground truth NSRTs with no
+        # continuous parameters, so filter them out.
+        gt_nsrts = {
+            nsrt
+            for nsrt in get_gt_nsrts(env.predicates, env.options)
+            if nsrt.option.params_space.shape != (0, )
```
tomsilver, line 28: remove space?

src/sampler_learning.py
```diff
@@ -29,9 +43,55 @@ def learn_samplers(strips_ops: List[STRIPSOperator],
...
+            param_option, nsrt.option, tuple(option_vars),
+            tuple(nsrt.option_vars))
+        if suc:
+            matching_nsrt = nsrt
+            matching_sub = sub
+            break
+    if matching_nsrt is None:
+        # Fall back to random sampler.
+        return _RandomSampler(param_option).sampler
+    # We found a matching operator. Use the matching_sub dictionary to
```
tomsilver, line 78: dang, well done...

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -85,6 +92,28 @@ def test_nsrt_learning_approach():
                        learn_side_predicates=True)
 
 
+def test_oracle_samplers():
+    """Test NSRTLearningApproach with oracle samplers."""
+    # Oracle sampler learning should work (and be fast) in cover and blocks.
```
tomsilver, line 97: woohoo!

src/sampler_learning.py
```diff
@@ -29,9 +43,55 @@ def learn_samplers(strips_ops: List[STRIPSOperator],
...
+            frozenset(op.preconditions), frozenset(nsrt.preconditions),
+            frozenset(op.add_effects), frozenset(nsrt.add_effects),
+            frozenset(op.delete_effects), frozenset(nsrt.delete_effects),
+            param_option, nsrt.option, tuple(option_vars),
+            tuple(nsrt.option_vars))
+        if suc:
+            matching_nsrt = nsrt
+            matching_sub = sub
+            break
+    if matching_nsrt is None:
```
tomsilver, line 75: i'm a little worried that we won't know when this is happening. do you know if this happens in cover / blocks right now? can you make an option so that an error is raised here instead of using a random sampler, and set that option to true for at least cover?

tomsilver (commented):
thanks a lot for doing this!!

src/sampler_learning.py
ronuchit: hah, you wish. we're at the mercy of YAPF now!

src/structs.py
```diff
@@ -678,6 +678,10 @@ def __gt__(self, other: object) -> bool:
         assert isinstance(other, NSRT)
         return str(self) > str(other)
 
+    def get_sampler(self) -> NSRTSampler:
```
tomsilver, line 681: is there a reason to do this instead of making the sampler public, like the other NSRT attributes? that would also make `nsrt.get_sampler()(state, rng, reordered_objs)` look less weird

tomsilver (approved):
lgtm after minor

src/structs.py
ronuchit: changed to @property like everywhere else

PR #256, opened 2022-01-07
still buggy side predicates

ronuchit:


src/nsrt_learning.py
```diff
@@ -228,14 +227,168 @@ def learn_strips_operators(
...
+                    atoms = utils.apply_operator(ground_op, segment.init_atoms)
+                    if not atoms.issubset(segment.final_atoms):
+                        continue
+                    # This segment belongs in this datastore, so add it.
+                    # TODO: this is an edge case we need to fix by changing
+                    # ObjToVarSubs in the partition to VarToObjSubs.
+                    if len(set(ground_op.objects)) != len(ground_op.objects):
+                        continue
+                    sub = dict(zip(ground_op.objects, pnad.op.parameters))
+                    pnad.add_to_datastore((segment, sub))
```
tomsilver, line 389: after fixing the bug below, this line will throw an error, because the assertions that we have in this method no longer make sense. you can just remove them or update them.

ronuchit:
closing bc outdated, we can always revisit later in the closed PRs section

PR #257, opened 2022-01-07
add support for other pyperplan heuristics

tomsilver:
realized that pyperplan facts are actually strings, not tuples. this wasn't an issue for hmax/hadd/hff but led to an error for lmcut.


ronuchit (approved):
lgtm thanks! wondering if we should put in the other heuristics too (other than landmark), or should we just keep the ones which we have tried?

tomsilver:
here are all of them: {blind,hadd,hff,hmax,hsa,landmark,lmcut}

so the only other one we could add is hsa. I will do that I guess, can't hurt.

PR #259, opened 2022-01-07
combine multiple heuristics within lookahead score functions for grammar search

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -799,8 +808,11 @@ def _relaxation_h(atoms: Set[GroundAtom], depth: int = 0) -> float:
 class _ExactHeuristicBasedScoreFunction(_HeuristicBasedScoreFunction):  # pylint:disable=abstract-method
     """Implement _generate_heuristic() with task planning."""
 
+    heuristic_names: Sequence[str] = ("exact", )
```
ronuchit, line 811: This would be a class variable instead of an instance variable right? Should it be `heuristic_names: Sequence[str] = field(default=("exact", ))`?

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -209,7 +209,13 @@ def test_create_score_function():
                                         {})
     assert isinstance(score_func,
                       _RelaxationHeuristicLookaheadBasedScoreFunction)
-    assert score_func.heuristic_name == "hff"
+    assert score_func.heuristic_names == ["hff"]
+    score_func = _create_score_function("lmcut,hff_lookahead_depth0", set(),
```
ronuchit, line 213: it's a bit sad that my brain parses `_` as taking precedence over `,` due to normal coding patterns, but oh well

src/approaches/grammar_search_invention_approach.py
```diff
@@ -466,19 +466,23 @@ def _create_score_function(
...
+            initial_predicates, atom_dataset, train_tasks, candidates,
+            ["hadd"])
+    match = re.match(r"([a-z\,]+)_lookahead_depth(\d+)", score_function_name)
     if match is not None:
-        # heuristic_name can be any of {"hadd", "hmax", "hff", "hsa", "lmcut"}
-        # depth can be any non-negative integer
-        heuristic_name, depth = match.groups()
+        # heuristic_name can be any of {"hadd", "hmax", "hff", "hsa", "lmcut"},
+        # or it can be multiple heuristic names that are comma-separated, such
+        # as hadd,hmax or hmadd,hmax,lmcut.
```
ronuchit, line 475: typo `hmadd`

ronuchit (approved):
lgtm, minor

src/approaches/grammar_search_invention_approach.py
ronuchit: and maybe also `init=False`?

PR #260, opened 2022-01-07
enforced hill climbing

tomsilver:
generalizes hill climbing to keep searching even if no immediate improvement can be found

it's very slow with depth>0, as you can imagine. i'm not sure if will use this, or if it's better than GBFS, but maybe? it can't hurt to have in our arsenal...

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -130,7 +130,7 @@ class GlobalSettings:
     grammar_search_heuristic_based_max_demos = 5
     grammar_search_lookahead_based_temperature = 10.
     grammar_search_task_planning_timeout = 1.0
-    grammar_search_hill_climbing_depth = 1
+    grammar_search_hill_climbing_depth = 0
 
     @staticmethod
     def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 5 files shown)


src/settings.py
```diff
@@ -130,6 +130,7 @@ class GlobalSettings:
     grammar_search_heuristic_based_max_demos = 5
     grammar_search_lookahead_based_temperature = 10.
     grammar_search_task_planning_timeout = 1.0
+    grammar_search_hill_climbing_depth = 0
```
ronuchit, line 133: is there a reason we are keeping this at 0? wouldn't it be strictly better to have it higher since otherwise we'd just terminate and be sad? (unless having it higher is causing us to add _more_ predicates once we already have the ideal set......)

ronuchit (commented):
Implementation LGTM, can you add a quick test for this? Should be easy with the gridworld tests we already have set up for this function

src/settings.py
tomsilver: setting it to 1 slows everything down very significantly

tomsilver:
@ronuchit good thing you made me test, I found some issues... (not affecting anything we were doing yet though)

PR #263, opened 2022-01-08
new evaluation metrics

ronuchit:
redo & consolidate evaluation metrics, now we have:
* num test tasks solved
* num test tasks total
* average test time for _successes_
* average number of node expansions for problems where a plan is found
* average plan length for problems where a plan is found
* total learning time


tomsilver (approved):
Thanks for doing this!

PR #265, opened 2022-01-09
basic MLP regressor torch model

tomsilver:
this is @ashayathalye 's code, just creating a separate PR for it (+ tests and a few minor changes)


src/settings.py
```diff
@@ -84,6 +84,10 @@ class GlobalSettings:
 
     # option learning parameters
     option_learner = "no_learning"  # "no_learning" or "oracle"
+    mlp_regressor_max_itr = 10000
```
ronuchit, line 87: let's start a new "section" in settings.py with a comment `# torch model parameters` and move these there, along with the `classifier_*` and `regressor_*` that are currently under `# sampler learning parameters`. Also can you rename some of those to make it clear which ones are using what? Like `learning_rate` can stay as is since everyone uses that, but the `classifier_hid_sizes = [32, 32]` under `# sampler learning parameters` should be renamed to `mlp_classifier_hid_sizes`, similarly with `regressor_hid_sizes` -> `neural_gaus_regressor_hid_sizes`, etc.

src/torch_models.py
```diff
@@ -17,6 +17,117 @@
...
+    def forward(self, inputs: Array) -> Tensor:
+        """Pytorch forward method."""
+        x = torch.from_numpy(np.array(inputs, dtype=np.float32))
+        for _, linear in enumerate(self._linears[:-1]):
+            x = F.relu(linear(x))
+        x = self._linears[-1](x)
+        return x
+
+    def predict(self, inputs: Array) -> Array:
+        """Normalize, predict, and convert to array."""
```
ronuchit, line 50: `"""Normalize, predict, un-normalize, and convert to array."""`

tests/test_models.py
```diff
@@ -2,11 +2,33 @@
...
+    num_samples = 5
+    model = MLPRegressor()
+    X = np.ones((num_samples, input_size))
+    Y = np.zeros((num_samples, output_size))
+    model.fit(X, Y)
+    x = np.ones(input_size)
+    mean = model.predict(x)
+    expected_y = np.zeros(output_size)
+    assert mean.shape == expected_y.shape
+    assert np.allclose(mean, expected_y, atol=1e-2)
```
ronuchit, line 29: Add another test where Y is some non-zero constants like `Y = 75 * np.ones((num_samples, output_size))`, since this is a regressor

ronuchit (changes requested):

tomsilver:

ronuchit (approved):
lgtm

PR #266, opened 2022-01-09
terminate option execution immediately if it's stuck

tomsilver:
because both env.simulate and option policies are deterministic


ronuchit (commented):
shouldn't we keep a visited set instead? for loops like A->B->C->A?

tomsilver:
> shouldn't we keep a visited set instead? for loops like A->B->C->A?
good question -- since states aren't hashable, I think this is more trouble than it's worth. in the meantime, I fixed the comments to clarify.

ronuchit (approved):
sigh that's sad lol. lgtm

PR #267, opened 2022-01-09
first option learning PR

tomsilver:
This is ready for an initial review. I'll add unit tests tomorrow.

Note: this is mostly @ashayathalye 's code; moving to a PR and made some changes (hopefully didn't add any bugs).

### Fixed tasks / degenerate oracle sampler
```python src/main.py --env cover_multistep_options_fixed_tasks --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle --cover_multistep_degenerate_oracle_samplers 1```
Tasks solved: 50 / 50
Average time for successes: 0.01555 seconds
Main script terminated in 14.00903 seconds

### Fixed tasks / regular oracle sampler
```python src/main.py --env cover_multistep_options_fixed_tasks --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle ```
Tasks solved: 50 / 50
Average time for successes: 0.10347 seconds
Main script terminated in 20.19477 seconds

### Fixed tasks / learned sampler
```python src/main.py --env cover_multistep_options_fixed_tasks --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple```
Tasks solved: 23 / 50
Average time for successes: 0.47498 seconds
Main script terminated in 311.16379 seconds

-----
_Note on sampler learning_: I realized that the sampler distribution for the pick option needs to be extremely bimodal, because there are two different kinds of picking:
1. move from the gripper’s initial position to pick the first block. this will be a delta y of something like 0.3.
2. after placing the first block, move to pick the second block. this will be a delta y of approximately 0.0!

our sampler learning code right now is not very well equipped to deal with learning bimodal distributions like this. so that could be one source of issue…

-----

### Changing tasks / degenerate oracle sampler
```python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle --cover_multistep_degenerate_oracle_samplers 1```
Tasks solved: 29 / 50
Average time for successes: 0.17606 seconds
Main script terminated in 229.00142 seconds

### Changing tasks / regular oracle sampler
```python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle```
Tasks solved: 39 / 50
Average time for successes: 0.44019 seconds
Main script terminated in 124.30103 seconds

### Changing tasks / learned sampler
```python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple```
Tasks solved: 30 / 50
Average time for successes: 0.03025 seconds
Main script terminated in 265.64459 seconds

### Bonus: Changing tasks / regular oracle sampler / 100 training tasks (instead of 15)
```python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle --num_train_tasks 100```
Tasks solved: 48 / 50
Average time for successes: 0.29034 seconds
Main script terminated in 60.43579 second

### CODE REVIEW

File: src/envs/cover.py [hunk 1]
```diff
@@ -779,7 +783,7 @@ def _Pick_learned_equivalent_policy(
         y = s.get(self._robot, "y")
         by = s.get(obj, "y")
         desired_x = p[7]
-        desired_y = by
+        desired_y = by + 1e-3
```
tomsilver, line 786: @ashayathalye note: I added this because I saw learned policies colliding with the blocks they were trying to pick very frequently; this seemed to help

src/option_learning.py
```diff
@@ -212,3 +219,205 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+                               objects: Sequence[Object],
+                               params: Array) -> bool:
+        del memory, params  # unused
+        # The hope is that we terminate in the effects.
+        grounded_op = self._operator.ground(tuple(objects))
+        abs_state = utils.abstract(state, self._predicates)
+        if grounded_op.add_effects.issubset(abs_state) and \
+            not grounded_op.delete_effects & abs_state:
+            return True
+        # If we fall outside of the space where the preconditions hold, also
```
tomsilver, line 303: @ashayathalye note: this is a new idea. seems to help with fast failing, at least, which is good for planning.

src/option_learning.py
```diff
@@ -212,3 +219,205 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+            del changing_parameter_set  # not used after this
+            assert changing_parameters == [
+                v for v in op.parameters if v in changing_parameters
+            ]
+
+            for segment, sub in datastore:
+                inv_sub = {v: o for o, v in sub.items()}
+                all_objects_in_operator = [inv_sub[v] for v in op.parameters]
+
+                # First, determine the absolute goal vector for this segment.
```
tomsilver, line 362: @ashayathalye this "absolute goal vector" and "relative goal vector" idea is new. I'm hoping it can be the best of both relative and absolute. see also the corresponding change in the option initiable and policy. let me know if it makes sense.

src/envs/cover.py
tomsilver: I realized belatedly that I did not update the sampler accordingly 🤦 and your check in the terminal didn't catch this because the tolerance is 1e-2 and this was 1e-3. actually, I want to revisit why the tolerance is so high, that seems potentially problematic...

tomsilver: I fixed the other one -- see the change in the oracle place sampler, and the reduction in tolerances in the terminal functions.

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
                                   dtype=np.float32)
                 option = param_opt.ground([robby, block], params)
                 segment.set_option(option)
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _LearnedSimpleParameterizedOption(ParameterizedOption):
+    """A convenience class for holding a learned parameterized option.
```
ronuchit, line 227: I would wipe this entire docstring and replace it with a description of the algorithm behind this "simple" approach! It's def not a convenience class lol

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+
+    Prefer to use this because it is pickleable.
+    """
+    _operator: STRIPSOperator
```
ronuchit, line 231: The docstring above should explain what these three components are (and why we need them)

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+    _regressor: MLPRegressor
+    _changing_parameter_idxs: Sequence[int]
+
+    def __init__(self, name: str, operator: STRIPSOperator,
```
ronuchit, line 235: Why do you have both a dataclass and an `__init__`? I would say just have the `__init__` so you can avoid the hacks below.

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+        object.__setattr__(self, "_changing_parameter_idxs",
+                           changing_parameter_idxs)
+        super().__init__(name,
+                         types,
+                         params_space,
+                         _policy=self._regressor_based_policy,
+                         _initiable=self._precondition_based_initiable,
+                         _terminal=self._effect_based_terminal)
+
+    @cached_property
```
ronuchit, line 261: Note: you have to be careful with cached_property if you're expecting that the self._operator might change at any point

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+                self._operator.add_effects | self._operator.delete_effects}
+
+    def _precondition_based_initiable(self, state: State, memory: Dict,
+                                      objects: Sequence[Object],
+                                      params: Array) -> bool:
+        # The memory here is used to store the absolute params, based on
+        # the relative params and the object states.
+        memory["params"] = params  # store for sanity checking in policy
+        changing_objects = [objects[i] for i in self._changing_parameter_idxs]
+        memory["absolute_params"] = state.vec(changing_objects) + params
```
ronuchit, line 274: What's going on here with absolute_params and relative_params? Is this a general algorithm or specific to some properties of your cover env?

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+
+
+class _SimpleOptionLearner(_OptionLearnerBase):
+    """The option learner that learns options from fixed-length input vectors.
+
+    That is, the the input data only includes objects whose state
+    changes from the first state in the segment to the last state in the
+    segment. The input data does not include data from other objects in
+    the scene, where the number of other objects would differ across
+    different instiantiations of the env.
```
ronuchit, line 321: instantiations*

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+        for op, datastore in zip(strips_ops, datastores):
+            print(f"\nLearning option for NSRT {op.name}")
+
+            X_regressor: List[Array] = []
+            Y_regressor = []
+
+            # The superset of all objects whose states we may include in the
+            # option params is op.parameters. But we only want to include
+            # those objects that have state changes (in at least some data).
+            changing_parameter_set = self._get_changing_parameters(datastore)
```
ronuchit, line 350: Why? Is this just an efficiency thing or an absolute necessity?

src/option_learning.py
```diff
@@ -212,3 +220,206 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+            regressor = MLPRegressor()
+            print(f"Fitting regressor with X shape: {X_arr_regressor.shape}, "
+                  f"Y shape: {Y_arr_regressor.shape}.")
+            regressor.fit(X_arr_regressor, Y_arr_regressor)
+
+            # Construct the ParameterizedOption for this operator.
+            name = f"{op.name}LearnedOption"
+            parameterized_option = _LearnedSimpleParameterizedOption(
+                name, op, regressor, changing_parameters)
+            option_specs.append((parameterized_option, list(op.parameters)))
```
ronuchit, line 404: Why not just the changing_parameters here? I guess I'm still confused why exactly we want those.

ronuchit (commented):
it's a bit hard to review this PR right now bc i'm unclear what is going on in _LearnedSimpleParameterizedOption, hopefully after more comments are added i can take a second pass!

ashayathalye:
Latest results as of [79b66bd](https://github.com/Learning-and-Intelligent-Systems/predicators/pull/267/commits/79b66bdb1c81c61d200e7911c939cc48e4cfee95)

# Fixed tasks / degenerate oracle sampler
`python src/main.py --env cover_multistep_options_fixed_tasks --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle --cover_multistep_degenerate_oracle_samplers 1`
Tasks solved: 50 / 50
Average time for successes: 0.02815 seconds
Main script terminated in 24.21318 seconds

# Fixed tasks / regular oracle sampler
`python src/main.py --env cover_multistep_options_fixed_tasks --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle`
Tasks solved: 49 / 50
Average time for successes: 0.26415 seconds
Main script terminated in 35.25844 seconds

# Fixed tasks / learned sampler
`python src/main.py --env cover_multistep_options_fixed_tasks --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple`
Tasks solved: 24 / 50
Average time for successes: 0.54679 seconds
Main script terminated in 342.23633 seconds

# Changing tasks / degenerate oracle sampler
`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle --cover_multistep_degenerate_oracle_samplers 1`
Tasks solved: 17 / 50
Average time for successes: 0.01759 seconds
Main script terminated in 315.61848 seconds

# Changing tasks / regular oracle sampler
`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle`
Tasks solved: 38 / 50
Average time for successes: 0.56492 seconds
Main script terminated in 109.92936 seconds

# Changing tasks / learned sampler
`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple`
Tasks solved: 26 / 50
Average time for successes: 0.14899 seconds
Main script terminated in 279.18247 seconds

# Bonus: Changing tasks / regular oracle sampler / 100 training tasks (instead of 15)
`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --remake_data --offline_data_method demo --option_learner simple --sampler_learner oracle --num_train_tasks 100`
Tasks solved: 48 / 50
Average time for successes: 0.38921 seconds
Main script terminated in 61.65736 seconds

src/option_learning.py
tomsilver: the operator shouldn't change

src/option_learning.py
tomsilver: it's a general algorithm; hopefully the forthcoming changes to the docstring will make this clear

src/option_learning.py
tomsilver: debatable

src/option_learning.py
tomsilver: it's a learning efficiency thing, I'll indicate this in the comment

src/option_learning.py
tomsilver: let me know if this is cleared up after I push, if not I can elaborate more

tomsilver:
@ronuchit thanks, ready for another review (no rush)

tomsilver:
because of #329, we should update this PR to include a check that the option isn't stuck at the same state

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+        # The hope is that we terminate in the effects.
+        grounded_op = self._operator.ground(tuple(objects))
+        abs_state = utils.abstract(state, self._predicates)
+        if grounded_op.add_effects.issubset(abs_state) and \
+            not grounded_op.delete_effects & abs_state:
+            return True
+        # If we fall outside of the space where the preconditions hold, also
+        # terminate. The assumption is that the option should take us from one
+        # abstract state to another without hitting any others in between.
+        if not grounded_op.preconditions.issubset(abs_state):
```
ronuchit, line 334: this check is strange to me --- why could you not go through other abstract states? like if you have an option whose pre is "InKitchen" and effect is "InBedroom" you certainly will have to go through being in the hallway, and there's nothing wrong with that

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+            return True
+        # Not yet done.
+        return False
+
+
+class _NeuralOptionLearner(_OptionLearnerBase):
+    """Learn _LearnedNeuralParameterizedOptions by behavior cloning.
```
ronuchit, line 341: _LearnedNeuralParameterizedOption objects*

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+
+    See the docstring for _LearnedNeuralParameterizedOptions for a description
```
ronuchit, line 343: _LearnedNeuralParameterizedOption*

src/option_learning.py
ronuchit: i'm still unclear on this but let me try to explain and you tell me if i'm right:
even though the option's policy is only tasked with changing a subset of the objects' features, checking preconditions & effects might require checking other objects' features, so the option still needs to include all the operator parameters.

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+
+    Note that the option terminal, which corresponds to the operator effects,
+    already describes how we want the objects to change. But these effects only
+    characterize a *set* of desired state changes. For example, a Pick operator
+    with Holding(?x) in the add effects says that we want to end up in *some*
+    state where we are holding ?x, but there are in general an infinite number
+    of such states. The role of the continuous parameters is to identify one
+    specific state in this set.
+
+    The hope is that by sampling different continuous parameters, the option
```
ronuchit, line 247: so the idea is that the sampler is responsible for figuring out which objects' features it wants to change, and how much it wants to change them by? will it be easy to learn samplers like that downstream? hopefully...

ronuchit: have you done any tests on this? just curious

src/option_learning.py
ronuchit: i think the answer to this refers back to our operator learning algorithm, and that makes me sad but whatever

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+
+    * The option objects correspond to the operator parameters.
+    * The option initiable corresponds to the operator preconditions.
+    * The option terminal corresponds to the operator effects.
+    * The option policy is implemented as a neural regressor that takes in the
+      states of the option objects and the input continuous parameters (see
+      below), all concatenated into a 1D vector, and outputs an action.
+
+    The continuous parameters of the option are the main thing to note: they
+    correspond to a desired change in state for a subset of the option objects.
```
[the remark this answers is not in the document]
tomsilver: yeah definitely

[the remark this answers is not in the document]
tomsilver: right

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+    Also note that the parameters correspond to a state *change*, rather than
+    an absolute state. This distinction is useful for learning; relative changes
+    lead to better data efficiency / generalization than absolute ones. However,
+    it's also important to note that the change here is a delta between the
+    initial state that the option is executed from and the final state where the
+    option is terminated. In the middle of executing the option, the desired
+    delta between the current state and the final state is different from what
+    it was in the initial state. To handle this, we save the *absolute* desired
+    state in the memory of the option during initialization, and compute the
+    updated delta on each call to the policy.
```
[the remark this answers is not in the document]
tomsilver: we figured it out the hard way 😆

src/option_learning.py
```diff
@@ -212,3 +219,235 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+        """Helper for initiable and terminal."""
+        return {a.predicate for a in self._operator.preconditions | \
+                self._operator.add_effects | self._operator.delete_effects}
+
+    def _precondition_based_initiable(self, state: State, memory: Dict,
+                                      objects: Sequence[Object],
+                                      params: Array) -> bool:
+        # The memory here is used to store the absolute params, based on
+        # the relative params and the object states.
+        memory["params"] = params  # store for sanity checking in policy
```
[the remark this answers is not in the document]
tomsilver: yeah they should be. we assert this in the policy; i'll add an assertion in terminal too.

src/option_learning.py
tomsilver: you're right, and Leslie mentioned this too. I'm going to actually remove this now because I don't think we need it.

tomsilver:
>i'm still unclear on this but let me try to explain and you tell me if i'm right:
even though the option's policy is only tasked with changing a subset of the objects' features, checking preconditions & effects might require checking other objects' features, so the option still needs to include all the operator parameters.

I can't reply to that comment directly for some reason so doing so here. it's less about checking preconditions and effects (which correspond to initiable and terminal) and more about the input to the policy. let's say that there is some object like "table" that appears in the symbolic effects of the operator like Place (on table), but that doesn't actually itself have any changing low-level state features. we could still want to use features of the table for the policy, for example, the table's vertical position / height.

src/option_learning.py
tomsilver: yeah that's the idea. we have run this all with sampler learning (you can see ashay's results above) and it works sometimes, but not as well as with an oracle sampler, as you'd expect

tomsilver:
@ronuchit thanks, incorporated the feedback

src/option_learning.py
ronuchit: looks like this wasn't fixed

ronuchit: (there is no class called `_LearnedNeuralParameterizedOptions` so it could be confusing)

src/option_learning.py
tomsilver: oh and maybe you were asking whether we've tested whether the diversity thing works out as we want it to. no we haven't yet -- we should do this soon. making the covers environment hand regions nontrivial would help.

ronuchit (approved):
lgtm, minor! very happy with the direction here, great job @ashayathalye and @tomsilver

src/option_learning.py
```diff
@@ -212,3 +216,226 @@ def update_segment_from_option_spec(self, segment: Segment,
...
+        for segment, sub in datastore:
+            start = segment.states[0]
+            end = segment.states[-1]
+            for o, v in sub.items():
+                if not np.array_equal(start[o], end[o]):
+                    all_changing_variables.add(v)
+        return all_changing_variables
+
+    def update_segment_from_option_spec(self, segment: Segment,
+                                        option_spec: OptionSpec) -> None:
```
NishanthJKumar, line 436: small thing, but it'd be nice to have a quick docstring here to explain what exactly this is doing

NishanthJKumar (approved):
LGTM with very minor comment! Great work on this; exciting to have this ready-to-go :100:

src/option_learning.py
ronuchit: This is overriding a parent class method. See docstring in https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/option_learning.py.

NishanthJKumar: ah didn't notice that; thanks for the pointer!

src/option_learning.py
```diff
@@ -16,6 +18,8 @@ def create_option_learner() -> _OptionLearnerBase:
         return _KnownOptionsOptionLearner()
     if CFG.option_learner == "oracle":
         return _OracleOptionLearner()
+    if CFG.option_learner == "neural":
+        return _NeuralOptionLearner()
```
ashayathalye, line 22: Writing this here cuz I can't write this below. @line 26: we should have _OptionLearnerBase inherit from abc.ABC. I had this in a previous commit but it got lost in the back and forth. The reason is: https://stackoverflow.com/questions/56008847/when-should-one-inherit-from-abc

ronuchit: This would need to happen in a lot of places. It's really just a matter of convention so I don't mind if we want to not do it. But in either case, we should do it in a separate PR if we want to do it at all.

tomsilver: thanks both! sounds good, yeah let's save it for another PR if we do want to do it.

ronuchit: it's a good idea for sure, helps pylint be less confused too. see #340

PR #268, opened 2022-01-09
let oracle sampler learner work with option learning

tomsilver:


ronuchit (approved):
i thought i had tested this before with oracle option learner, what the heck! was it breaking?

ronuchit:
hmm, i tested on master and it works fine? i'm misunderstanding something

tomsilver:
@ronuchit I think it only works because the oracle option learner actually has the correct names of the options. I hit the error with learned options (not yet merged)

ronuchit:
ok, gotcha!

PR #270, opened 2022-01-10
Playroom nsrt learning

amburger66:
changes implemented to make nsrt learning work:
- make sure for each option, all option vars are used somewhere (either in policy or initiable fnc)
- make sure that each "move transition" by the robot can only be explained by one option -- so now there are 3 different move options, each with a different and more specific policy for movement
- simulator no longer allows the robot to move in ways that the oracle operators do not allow for (e.g. transitioning from region 1 directly to region 3)
- got rid of “robot must be facing x to interact with x” requirement in simulator b/c this made sampling for pick, stack, putontable too hard


tomsilver (approved):
~@amburger66 excellent work! can you summarize the changes that were necessary to get learning to work? also, any reason that this is still a draft?~ (this is done now)

tomsilver:
@ronuchit these changes only touch playroom, so I'm going to merge

PR #271, opened 2022-01-10
low-hanging optimizations for exact lookahead

tomsilver:
ground the NSRTs only once per task, rather than on every call to `task_plan`. this seems to be a 2-3x speedup (based on running the analysis script before vs after).

### DIFF
--- a/src/approaches/grammar_search_invention_approach.py
+++ b/src/approaches/grammar_search_invention_approach.py
@@ -619,14 +619,15 @@ class _TaskPlanningScoreFunction(_OperatorLearningBasedScoreFunction):
             init_atoms = utils.abstract(task.init,
                                         predicates | self._initial_predicates)
             objects = set(task.init)
-            ground_nsrts = task_plan_grounding(init_atoms, objects, strips_ops,
-                                               option_specs)
+            ground_nsrts, reachable_atoms = task_plan_grounding(
+                init_atoms, objects, strips_ops, option_specs)
             heuristic = utils.create_task_planning_heuristic(
                 CFG.task_planning_heuristic, init_atoms, task.goal,
                 ground_nsrts, predicates | self._initial_predicates, objects)
             try:
                 _, _, metrics = task_plan(
-                    init_atoms, task.goal, ground_nsrts, heuristic, CFG.seed,
+                    init_atoms, task.goal, ground_nsrts, reachable_atoms,
+                    heuristic, CFG.seed,
                     CFG.grammar_search_task_planning_timeout)
                 node_expansions = metrics["num_nodes_expanded"]
                 assert node_expansions < node_expansion_upper_bound
@@ -832,8 +833,8 @@ class _ExactHeuristicBasedScoreFunction(_HeuristicBasedScoreFunction):  # pylint
 
         # It's important for efficiency that we only ground once, and create
         # the heuristic once, for every task.
-        ground_nsrts = task_plan_grounding(init_atoms, objects, strips_ops,
-                                           option_specs)
+        ground_nsrts, reachable_atoms = task_plan_grounding(
+            init_atoms, objects, strips_ops, option_specs)
         heuristic = utils.create_task_planning_heuristic(
             CFG.task_planning_heuristic, init_atoms, goal, ground_nsrts,
             set(predicates) | self._initial_predicates, objects)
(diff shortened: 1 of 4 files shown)


src/planning.py
```diff
@@ -101,12 +105,35 @@ def sesame_plan(
...
+    filtering out ones that are unreachable or have empty effects."""
+    nsrts = utils.ops_and_specs_to_dummy_nsrts(strips_ops, option_specs)
+    ground_nsrts = []
+    for nsrt in nsrts:
+        for ground_nsrt in utils.all_ground_nsrts(nsrt, objects):
+            ground_nsrts.append(ground_nsrt)
+    nonempty_ground_nsrts = [
+        nsrt for nsrt in ground_nsrts if nsrt.add_effects | nsrt.delete_effects
+    ]
+    all_reachable_atoms = utils.get_reachable_atoms(ground_nsrts, init_atoms)
```
ronuchit, line 124: Here, we're using ground_nsrts as the first argument, whereas before it was nonempty_ground_nsrts. Intentional?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -619,10 +619,15 @@ def _evaluate_with_operators(self, predicates: FrozenSet[Predicate],
             init_atoms = utils.abstract(task.init,
                                         predicates | self._initial_predicates)
             objects = set(task.init)
+            ground_nsrts = task_plan_grounding(init_atoms, objects, strips_ops,
```
ronuchit, line 622: I guess we can't really squeeze efficiency out of this call like we can for the `generate_heuristic` one below, right? Bc the predicates are always changing? But we only call this once per task instead of once per step in the task, so maybe okay?

src/planning.py
```diff
@@ -101,12 +105,35 @@ def sesame_plan(
...
     objects: Set[Object],
-    goal: Set[GroundAtom],
     strips_ops: Sequence[STRIPSOperator],
     option_specs: Sequence[OptionSpec],
+) -> List[_GroundNSRT]:
+    """Ground all operators for task planning into dummy _GroundNSRTs,
+    filtering out ones that are unreachable or have empty effects."""
+    nsrts = utils.ops_and_specs_to_dummy_nsrts(strips_ops, option_specs)
+    ground_nsrts = []
+    for nsrt in nsrts:
```
ronuchit, line 118: when you merge in #276 remember to change this to `sorted(nsrts)`

ronuchit (approved):
LGTM, will be some trickiness in merging with #276. One thing is definitely make sure to `git grep "for nsrt in nsrts"` to make sure it's not in planner.py (it should be `for nsrt in sorted(nsrts)`). The other thing is that I made some calls to task_plan in my new test, which you'll need to update.

src/planning.py
```diff
@@ -118,37 +145,21 @@ def task_plan(
...
-    ground_nsrts = []
-    for nsrt in nsrts:
-        for ground_nsrt in utils.all_ground_nsrts(nsrt, objects):
-            ground_nsrts.append(ground_nsrt)
-    nonempty_ground_nsrts = [
-        nsrt for nsrt in ground_nsrts if nsrt.add_effects | nsrt.delete_effects
-    ]
-    all_reachable_atoms = utils.get_reachable_atoms(nonempty_ground_nsrts,
-                                                    init_atoms)
+    all_reachable_atoms = utils.get_reachable_atoms(ground_nsrts, init_atoms)
```
ronuchit, line 148: It's a bit sad that this has to happen twice, right?

src/planning.py
```diff
@@ -101,12 +105,35 @@ def sesame_plan(
             predicates |= new_predicates
 
 
-def task_plan(
+def task_plan_grounding(
```
ronuchit, line 108: In `sesame_plan`, the grounding is part of the overall timeout. But that's no longer the case here. Should we fix this?

src/planning.py
tomsilver: it wouldn't make a difference, but not intentional, will fix

src/approaches/grammar_search_invention_approach.py
tomsilver: right, ideally we would find some optimizations for this too, but it's not obvious how because the predicates are changing

src/planning.py
tomsilver: hmm. I'll leave it to the user, which in this case would be grammar search. and for now I'm not going to change it because I think adding more timeouts could only hurt us at this point (where things are not yet working). but we can revisit in the future once things are working and we want optimizations

ronuchit:
can we add some comments to task_plan/task_plan_grounding regarding usage?

PR #275, opened 2022-01-10
don't allow __call__ on zero arity preds

ronuchit:


tomsilver (approved):
Lol do I even want to know what led to this

ronuchit:
we only use this in testing, so it wasn't too bad, i just tried to create a goal with a zero-arity pred and it got treated as a LiftedAtom, so it crashed

PR #276, opened 2022-01-10
planning determinism + test

ronuchit:
Two line edit, 100 line test, lol. @tomsilver note that we don't have to sort `objects` because all the grounding utils methods eventually use `get_object_combinations`, which sorts the objects.

### DIFF
--- a/tests/test_planning.py
+++ b/tests/test_planning.py
@@ -9,8 +9,7 @@ from predicators.src.envs import CoverEnv
 from predicators.src.planning import sesame_plan, task_plan
 from predicators.src import utils
 from predicators.src.structs import Task, NSRT, ParameterizedOption, _Option, \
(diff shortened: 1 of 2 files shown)


tests/test_planning.py
```diff
@@ -152,3 +153,115 @@ def test_sesame_plan_uninitiable_option():
                     timeout=500,
                     seed=123)
     assert "Planning reached max_skeletons_optimized!" in str(e.value)
+
+
+def test_planning_determinism():
+    """Tests that planning is deterministic when there are multiple ways of
+    achieving a goal."""
+    utils.update_config({"env": "cover"})
+    robot_type = Type("robot_type", ["asleep", "cried"])
```
tomsilver, line 162: lol

tomsilver (approved):
are we at that point in the project where the robot starts crying?

tests/test_planning.py
ronuchit: my two favorite things

PR #279, opened 2022-01-11
behavior options pseudocode

ronuchit:


ronuchit:
as discussed offline, we will instead change `_initiable()` in `make_behavior_option` to include a `model_controller` key in the memory dictionary, and then for option_model.py we can add:
```
class _BehaviorOptionModel(_OptionModel):
    """An oracle option model for the BEHAVIOR env."""

    def get_next_state(self, state: State, option: _Option) -> State:
        env = get_cached_env("behavior")
        assert option.memory["model_controller"] is not None
        option.memory["model_controller"](state, env)
```

PR #280, opened 2022-01-11
minor typo & let yapf handle formatting instead of backslashes

ronuchit:


NishanthJKumar (approved):
LGTM! Thanks for doing this!

PR #283, opened 2022-01-11
Sets up interface and incorporates option models for BEHAVIOR!

NishanthJKumar:
Sets up interface (i.e, `_BehaviorOptionModel` in `option_models.py`) as well as option models (included in `behavior_options.py`) for BEHAVIOR!

Planning time is significantly faster (though still cannot solve test task due to sampler/other issues that will be addressed in a separate PR). Bottleneck is no longer `env.step()` calls, but rather time between option model executions (roughly 1 min between each action).

### DIFF
--- a/src/envs/behavior.py
+++ b/src/envs/behavior.py
@@ -443,9 +443,8 @@ def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
     def _policy(state: State, memory: Dict, _objects: Sequence[Object],
                 _params: Array) -> Action:
         assert "has_terminated" in memory
-        assert memory.get(
-            "policy_controller"
-        ) is not None  # must call initiable() first, and it must return True
+        # must call initiable() first, and it must return True
+        assert memory.get("policy_controller") is not None
         assert not memory["has_terminated"]
         action_arr, memory["has_terminated"] = memory["policy_controller"](
             state, env)
--- a/src/envs/behavior.py
+++ b/src/envs/behavior.py
@@ -112,7 +112,7 @@ class BehaviorEnv(BaseEnv):
                 linearVelocity=[0, 0, 0],
                 angularVelocity=[0, 0, 0],
             )
-        next_state = self._current_ig_state_to_state()
+        next_state = self.current_ig_state_to_state()
         return next_state
 
     def train_tasks_generator(self) -> Iterator[List[Task]]:
@@ -128,7 +128,7 @@ class BehaviorEnv(BaseEnv):
             # hacky workaround for that.
             np.random.seed(rng.integers(0, (2**32) - 1))
             self.igibson_behavior_env.reset()
-            init_state = self._current_ig_state_to_state()
+            init_state = self.current_ig_state_to_state()
             goal = self._get_task_goal()
             task = Task(init_state, goal)
             tasks.append(task)
@@ -288,7 +288,9 @@ class BehaviorEnv(BaseEnv):
                 return pred
         raise ValueError(f"No predicate found for name {name}.")
 
-    def _current_ig_state_to_state(self) -> State:
+    def current_ig_state_to_state(self) -> State:
+        """Function to create a predicators State from the current underlying
+        iGibson simulator state."""
         state_data = {}
         for ig_obj in self._get_task_relevant_objects():
             obj = self._ig_object_to_object(ig_obj)
@@ -313,7 +315,7 @@ class BehaviorEnv(BaseEnv):
             # predicate. Because of this, we will assert that whenever
             # a predicate classifier is called, the internal simulator
             # state is equal to the state input to the classifier.
-            assert s.allclose(self._current_ig_state_to_state())
+            assert s.allclose(self.current_ig_state_to_state())
             arity = self._bddl_predicate_arity(bddl_predicate)
             if arity == 1:
                 assert len(o) == 1
@@ -339,7 +341,7 @@ class BehaviorEnv(BaseEnv):
                               objs: Sequence[Object]) -> bool:
         # Check allclose() here for uniformity with
         # _create_classifier_from_bddl
-        assert state.allclose(self._current_ig_state_to_state())
+        assert state.allclose(self.current_ig_state_to_state())
         assert len(objs) == 2
         ig_obj = self.object_to_ig_object(objs[0])
         ig_other_obj = self.object_to_ig_object(objs[1])
@@ -350,7 +352,7 @@ class BehaviorEnv(BaseEnv):
     def _reachable_nothing_classifier(self, state: State,
                                       objs: Sequence[Object]) -> bool:
         # Check allclose() here for uniformity with _create_classifier_from_bddl
-        assert state.allclose(self._current_ig_state_to_state())
+        assert state.allclose(self.current_ig_state_to_state())
         assert len(objs) == 1
         for obj in state:
             if self._reachable_classifier(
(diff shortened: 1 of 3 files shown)


src/envs/behavior.py
```diff
@@ -435,24 +435,26 @@ def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
...
-        if memory.get("controller") is None:
+        if memory.get("policy_controller") is None:
             # We want to reset the state of the environment to
             # the state in the init state so that our options can
             # run RRT/plan from here as intended!
             if state.simulator_state is not None:
                 env.task.reset_scene(state.simulator_state)
             controller = controller_fn(env, igo[0], params, rng=rng)
-            memory["controller"] = controller
+            memory["policy_controller"] = controller
```
ronuchit, line 457: you'll need to add memory_controller here too with your bit being True

NishanthJKumar: right, but that's not implemented yet, so i wasn't sure whether to include this in this PR

NishanthJKumar: should i add the option model code to this PR as well? Or should i open a separate PR that includes that stuff in `behavior_options.py` first and merge that first?

src/option_model.py
```diff
@@ -46,3 +49,14 @@ def get_next_state(self, state: State, option: _Option) -> State:
...
         return traj.states[-1]
+
+
+class _BehaviorOptionModel(_OptionModel):
+    """An oracle option model for the BEHAVIOR env."""
+
+    def get_next_state(self, state: State,
+                       option: _Option) -> State:  # pragma: no cover
+        env = get_cached_env_instance("behavior")
+        assert option.memory["model_controller"] is not None
```
ronuchit, line 60: change to `option.memory.get["model_controller"]` so that if the key doesn't exist, we'll hit this assert

src/option_model.py
```diff
@@ -46,3 +49,14 @@ def get_next_state(self, state: State, option: _Option) -> State:
...
+        final_state = option.memory["model_controller"](state, env)
```
ronuchit, line 61: `final_state` -> `next_state`

src/envs/behavior.py
```diff
@@ -435,24 +435,26 @@ def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
     def _policy(state: State, memory: Dict, _objects: Sequence[Object],
                 _params: Array) -> Action:
         assert "has_terminated" in memory
-        assert ("controller" in memory and memory["controller"] is not None
+        assert ("policy_controller" in memory
+                and memory["policy_controller"] is not None
```
ronuchit, line 439: this assert can be written as `assert memory.get("policy_controller") is not None`

src/envs/behavior.py
ronuchit: i'd say add it to this PR. but you can do the changing of the argument type from iGibsonEnv to BehaviorEnv in a separate PR

src/envs/behavior.py
```diff
@@ -441,26 +443,38 @@ def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
     def _policy(state: State, memory: Dict, _objects: Sequence[Object],
                 _params: Array) -> Action:
         assert "has_terminated" in memory
-        assert ("controller" in memory and memory["controller"] is not None
-                )  # must call initiable() first, and it must return True
+        assert memory.get(
+            "policy_controller"
+        ) is not None  # must call initiable() first, and it must return True
```
ronuchit, line 448: formatting is a bit weird, thanks yapf -_- can you move this comment to its own line directly preceding the line of code, so that  `assert memory.get("policy_controller") is not None` fits in one line?

src/envs/behavior.py
```diff
@@ -441,26 +443,38 @@ def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
...
+                                              params,
+                                              ret_option_model=False,
+                                              rng=rng)
+            memory["policy_controller"] = policy_controller
+            model_controller = controller_fn(env,
+                                             igo[0],
+                                             params,
+                                             ret_option_model=True,
+                                             rng=rng)
+            memory["model_controller"] = model_controller
```
ronuchit, line 475: great to see that this worked!

src/envs/behavior_options.py
```diff
@@ -931,12 +1011,37 @@ def placeOntopObjectOptionPolicy(_state: State,
...
+
 def place_ontop_obj_pos(
     env: "BehaviorEnv",
     obj: Union["URDFObject", "RoomFloor"],
     place_rel_pos: Array,
+    ret_option_model: bool = False,
     rng: Optional[Generator] = None,
-) -> Optional[Callable[[State, "BehaviorEnv"], Tuple[Array, bool]]]:
+) -> Optional[Union[Callable[[State, "BehaviorEnv"], Tuple[Array, bool]],
+                    Callable[[State, "BehaviorEnv"], None]]]:
```
ronuchit, line 1044: can you go through and make sure all the docstrings are correct now? for instance, the one for this method says "If the placement is infeasible, returns an indication to this effect (None). Otherwise, returns a function that can be stepped like an option to output actions at each timestep." but this is out-of-date, because now we have the possibility that it can return a model as well. can you describe in the comment exactly what's going on with this crazy-looking return type? (for all the controller fns, not just this one)

ronuchit (approved):
glad this ended up being quite a simple change at the end of the day! my only major thing is to go through the comments and (a) make sure they're correct, (b) describe exactly what the crazy-looking return types are for the controller_fns! let @tomsilver review before merging.

src/envs/behavior.py
```diff
@@ -441,26 +443,38 @@ def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
...
                 env.task.reset_scene(state.simulator_state)
-            controller = controller_fn(env, igo[0], params, rng=rng)
-            memory["controller"] = controller
+            policy_controller = controller_fn(env,
+                                              igo[0],
+                                              params,
+                                              ret_option_model=False,
+                                              rng=rng)
+            memory["policy_controller"] = policy_controller
+            model_controller = controller_fn(env,
```
tomsilver, line 470: if I understand correctly, this is going to run planning twice -- once in creating the policy controller, and once in creating the model controller. is that right? can we avoid that if so?

this is making me want more the refactor where we avoid the controller functions altogether, and just create options directly in `behavior_options.py`... but if there's an easy way to address this concern in the current paradigm that's fine too, and maybe I or someone else can try the refactor in a later PR.

src/envs/behavior_options.py
tomsilver: +1

tomsilver (commented):
looking good thanks! one other small thing is I don't love the name "model_controller", but I can't think of a better alternative

src/envs/behavior.py
NishanthJKumar: Yes, that is exactly right (and something I discovered only after running it lol).

I don't really know an easy way to address this in the current paradigm unfortunately... Would love to refactor to create options directly, though not sure if we should merge this and do that in a separate PR?

tomsilver: I'll leave it up to you (or @ronuchit if he feels strongly). If there's any chance that this refactor could also help clear up the bugs you're seeing with the current code, that would be reason to prioritize the refactor. But it sounds like you've localized the issue to the grasp sampler, so it might not.

ronuchit: can we just make each controller fn cache the results of the planning/RRT part? that way it can be re-used?

PR #284, opened 2022-01-11
Interactive learning approach

amburger66:
- changed the ask strategy so robot cannot ask teacher about past states: for each state in a newly-explored trajectory, the robot considers each possible ground atom and queries the teacher about the atoms that score higher than the best atom seen previously (note that this is currently flawed b/c the robot will only ask in the first state and never again)
- updated settings

### DIFF
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -92,7 +92,8 @@ class InteractiveLearningApproach(NSRTLearningApproach):
                 max_steps=CFG.interactive_max_steps)
             # Decide whether to ask at each state during exploration
             for s in traj.states:
-                abstract_state = utils.abstract(s, self._get_current_predicates())
+                abstract_state = utils.abstract(s,
+                                                self._get_current_predicates())
                 score = score_goal(self._dataset_with_atoms, abstract_state)
                 # Ask about this state if it is the best seen so far
                 if score > best_score:
@@ -100,7 +101,8 @@ class InteractiveLearningApproach(NSRTLearningApproach):
                     ground_atoms = utils.all_possible_ground_atoms(
                         s, self._predicates_to_learn)
                     atom_set = get_best_goal_by_score(self._dataset_with_atoms,
-                                                      [{ga} for ga in ground_atoms])
+                                                      [{ga}
+                                                       for ga in ground_atoms])
                     assert len(atom_set) == 1
                     atom = atom_set.pop()
                     if self._ask_teacher(s, atom):
@@ -262,8 +264,7 @@ def score_goal(dataset_with_atoms: List[GroundAtomTrajectory],
 
 def get_best_goal_by_score(dataset_with_atoms: List[GroundAtomTrajectory],
                            goals: List[Set[GroundAtom]]) -> Set[GroundAtom]:
-    """Return the best goal out of `goals` using the score function.
-    """
+    """Return the best goal out of `goals` using the score function."""
     scores = [score_goal(dataset_with_atoms, g) for g in goals]
     goals_and_scores = list(zip(goals, scores))
     goals_and_scores.sort(key=lambda tup: tup[1], reverse=True)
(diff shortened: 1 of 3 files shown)


src/approaches/interactive_learning_approach.py
```diff
@@ -88,28 +90,27 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
-            if self._ask_teacher(best_state, list(best_atom_set)[0]):
-                # Add this atom if it's a positive example
-                self._dataset_with_atoms.append(  # pragma: no cover
-                    (LowLevelTrajectory([best_state], []), [best_atom_set]))
-                # Still need to implement a way to use negative examples
+            # Decide whether to ask at each state during exploration
+            for s in traj.states:
+                abstract_state = utils.abstract(s,
+                                                self._get_current_predicates())
+                score = score_goal(self._dataset_with_atoms, abstract_state)
```
tomsilver, line 97: so `score_goal` was originally intended for use with goals sampled from glib, right? but now we're passing in the whole `abstract_state` instead of just a goal.

furthermore, based on our slack conversation just now, i think we should refactor score functions so that they take in single atoms, rather than multiple. in doing so, we can change this function to `score_atom`. then we don't need this two-step process where we first decide whether to ask about a whole state, and then decide about each atom -- instead, we just make one judgement about each atom.

src/approaches/interactive_learning_approach.py
```diff
@@ -88,28 +90,27 @@ def learn_from_offline_dataset(self, dataset: Dataset,
...
+                    ground_atoms = utils.all_possible_ground_atoms(
+                        s, self._predicates_to_learn)
+                    atom_set = get_best_goal_by_score(self._dataset_with_atoms,
+                                                      [{ga}
+                                                       for ga in ground_atoms])
+                    assert len(atom_set) == 1
+                    atom = atom_set.pop()
+                    if self._ask_teacher(s, atom):
+                        # Add this atom if it's a positive example
+                        self._dataset_with_atoms.append(  # pragma: no cover
```
tomsilver, line 110: I probably asked about this before but I forget -- why this `pragma: no cover`? also please add a comment to appease my goldfish memory

tomsilver (commented):
thanks for the fast turnarounds :)

src/approaches/interactive_learning_approach.py
```diff
@@ -92,27 +92,20 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
-                                                       for ga in ground_atoms])
-                    assert len(atom_set) == 1
-                    atom = atom_set.pop()
-                    if self._ask_teacher(s, atom):
-                        # Add this atom if it's a positive example
-                        self._dataset_with_atoms.append(  # pragma: no cover
-                            (LowLevelTrajectory([s], []), [{atom}]))
-                        # Still need to implement a way to use negative examples
-                    best_score = score
+                ground_atoms = utils.all_possible_ground_atoms(
```
tomsilver, line 97: double check that the order of `ground_atoms` is deterministic

src/approaches/interactive_learning_approach.py
```diff
@@ -92,27 +92,20 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
+                    s, self._predicates_to_learn)
+                for atom in ground_atoms:
+                    score = score_atom(self._dataset_with_atoms, atom)
```
tomsilver, line 100: add a comment so we don't forget: although this score function currently does not use `s`, we imagine that future score functions importantly will.

tomsilver (approved):
lgtm after minor. thanks!

src/approaches/interactive_learning_approach.py
amburger66: sorry what do you mean by this? utils.all_possible_ground_atoms() returns the list in sorted order

tomsilver:
Yup that’s what I meant, so we’re good. Thanks for checking!

PR #285, opened 2022-01-11
refactors behavior_options.py

NishanthJKumar:
* splits off functions that return policy function so integrating option model will be significantly easier
* includes some helpful status prints within option execution
* renames behavior_env attr of BehaviorEnv to igibson_behavior_env to be clearer

### DIFF
--- a/src/envs/behavior_options.py
+++ b/src/envs/behavior_options.py
@@ -176,6 +176,78 @@ def navigate_to_param_sampler(rng: Generator,
     return np.array([x, y])
 
 
+def create_navigate_policy(
+    plan: List[Tuple[float, float,
+                     float]], original_orientation: Tuple[float, float, float]
+) -> Callable[[State, "BehaviorEnv"], Tuple[Array, bool]]:
+    """Instantiates and returns a navigation option policy given an RRT
+    plan."""
+
+    def navigateToOptionPolicy(_state: State,
+                               env: BehaviorEnv) -> Tuple[Array, bool]:
+        atol_xy = 1e-2
+        atol_theta = 1e-3
+        atol_vel = 1e-4
+
+        # 1. Get current position and orientation
+        current_pos = list(env.robots[0].get_position()[0:2])
+        current_orn = p.getEulerFromQuaternion(
+            env.robots[0].get_orientation())[2]
+
+        expected_pos = np.array(plan[0][0:2])
+        expected_orn = np.array(plan[0][2])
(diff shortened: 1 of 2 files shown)


src/envs/behavior_options.py
```diff
@@ -588,61 +662,12 @@ def grasp_obj_at_pos(
...
-
-        return ret_action, done_bit
-
     p.restoreState(state)
     p.removeState(state)
 
-    return graspObjectOption
+    print(f"PRIMITIVE: grasp {obj.name} success! Plan found with " +
+          f"continuous params {grasp_offset}. Executing now.")
+    return create_grasp_policy(plan)
```
ronuchit, line 670: i see, so the idea will be if we're in model mode, we shouldn't use this policy, but rather just extract the final state from the plan? and teleport to it?

is that final state from the plan the same as this variable `state`? or is `state` the state before we screw things up with all this planning?

src/envs/behavior_options.py
```diff
@@ -651,7 +676,7 @@ def place_obj_plan(
     original_state: int,
     place_rel_pos: Array,
     rng: Optional[Generator] = None,
-) -> List[List[float]]:
+) -> Optional[List[List[float]]]:
```
ronuchit, line 679: why this change?

ronuchit (approved):

src/envs/behavior_options.py
NishanthJKumar: turns out it actually was always optional and this was just a bug (caught thanks to the refactor)!

NishanthJKumar: (it's optional because the plan might be None)

src/envs/behavior_options.py
NishanthJKumar: so the idea in model mode will be actually that we'll call a function `create_grasp_model` similar to `create_grasp_policy` that will take in the `plan` and then teleport the world to the right end configuration (not exactly, since we'll be doing some extra legwork to make it more realistic so that we can be sure that if the option model works, the actual option policy execution will also work)

`state` is a saved state from before we screw things up with planning that we can use to reset the world before returning our function (that's why there's the `p.restoreState(state)` line before the `return`). The plan is actually a series of waypoints that might be for the hand, or for the body depending on what planning function is called, and so we can't simply return the last element, we have to do some work to teleport the robot (and objects it might be holding, etc.)

PR #286, opened 2022-01-12
don't give approaches redundant train_tasks argument

ronuchit:
A simple code quality change here. Passing train_tasks into
learn_from_offline_dataset is a relic from the past, when we didn't have
our demonstration trajectories annotated with goals. All the information
in train_tasks is already contained in the Dataset object itself.

Instead of:
```
for task in train_tasks:
    print(task.init, task.goal)
```
one can do:
```
for traj in dataset:
    if not traj.is_demo:
        continue
    print(traj.states[0], traj.goal)
```

### DIFF
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -14,7 +14,7 @@ from predicators.src.approaches.grammar_search_invention_approach import \
 from predicators.src.approaches.oracle_approach import _get_predicates_by_names
 from predicators.src.main import _run_testing
 from predicators.src import utils
-from predicators.src.structs import Predicate, Dataset, Task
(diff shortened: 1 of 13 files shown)


ronuchit:
@amburger66 this will affect your current PR

tomsilver (approved):
fine by me, but please wait for @amburger66 to confirm that this won't cause headaches in her stuff

ronuchit:
Yup, will wait. It shouldn't cause a headache hopefully, it's the exact same information, just need to access it a different way as illustrated above, or in my changes to interactive_learning_approach.py

amburger66 (approved):
cool, thanks for letting me know!

PR #287, opened 2022-01-13
Catch approach failures while testing

amburger66:
Catch approach failures, specifically OptionPlanExhausted, during policy execution so it continues to the next task and doesn't crash the whole thing.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -131,6 +131,10 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
             print(f"Task {i+1} / {len(test_tasks)}: Environment failed "
                   f"with error: {e}")
             continue
+        except ApproachFailure as e:
+            print(f"Task {i+1} / {len(test_tasks)}: Approach failed "
+                  f"with error: {e}")
+            continue
         if solved:
             print(f"Task {i+1} / {len(test_tasks)}: SOLVED")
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -131,6 +131,10 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
             print(f"Task {i+1} / {len(test_tasks)}: Environment failed "
                   f"with error: {e}")
             continue
+        except ApproachFailure as e:
+            print(f"Task {i+1} / {len(test_tasks)}: Approach failed "
+                  f"with error: {e}")
```
ronuchit, line 136: can you say "Approach failed at policy execution time with error:" to distinguish from above?

also, except (ApproachTimeout, ApproachFailure)

tomsilver: @ronuchit re the second thing, what would `ApproachTimeout` mean at execution time?

ronuchit: approach failed to compute the next action within the time limit, imagine it's like an online planner for example

ronuchit: think of it not as execution time, but as policy inference time (same thing but slightly diff perspective)

tomsilver: got it, makes sense

tomsilver (approved):
Looks great, thank you!

PR #289, opened 2022-01-14
clean up random seeding

tomsilver:
@ronuchit the change to `update_config` is not super ideal, but the alternatives -- updating all current and future unit tests to explicitly set `seed`, or keep things the way they are now -- seem less good. lmk what you think though.

cc @NishanthJKumar , who originally pointed out the seed(0) grossness.

### DIFF
--- a/src/approaches/base_approach.py
+++ b/src/approaches/base_approach.py
@@ -7,6 +7,7 @@ import numpy as np
 from gym.spaces import Box
 from predicators.src.structs import State, Task, Predicate, Type, \
     ParameterizedOption, Action, Dataset, Metrics
+from predicators.src.settings import CFG
 
 
 class BaseApproach:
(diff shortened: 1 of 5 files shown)


tomsilver:
oops this seems fundamentally broken, disregard until further notice...

ronuchit (approved):
lgtm

tomsilver:
okay unfortunately we can't seed the action space / options in the environment for the reason I specified in the new comment. things I tried that didn't work:
1. use `__post_init__` -- doesn't work because environments aren't dataclasses, so this method is never called
2. change all environments to call `super().__init__()` at the end of their `__init__()` rather than the beginning, but this quickly got messy because some envs use rng inside of `__init__()`.

ronuchit:
okay np

PR #292, opened 2022-01-17
Add "experiment id" flag

amburger66:
The interactive learning approach can't use the excluded predicates flag, so I need some other way to distinguish between different experiment runs. This changes the config and save path strings.

The changes to `analysis/analyze_supercloud_experiments.py` mean the new script won't work on previously saved results.

### DIFF
--- a/analysis/analyze_supercloud_experiments.py
+++ b/analysis/analyze_supercloud_experiments.py
@@ -11,18 +11,18 @@ def _main() -> None:
     # Gather data.
     all_data = []
     column_names = [
-        "ENV", "APPROACH", "EXCLUDED_PREDICATES", "SEED", "NUM_SOLVED",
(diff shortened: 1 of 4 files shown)


tomsilver:
thank you for doing this! genuine question, i'm not sure: what do you think about making this backwards compatible by doing the following... if `experiment_id` is empty, then don't add the "__" at the end. a downside is that when parsing the filenames, you'd have to do in cases, because sometimes that would be N "__"s and other times there would be N+1. @ronuchit @amburger66

ronuchit (approved):
i think it's fine to break backward compatibility

tomsilver:
cool, @amburger66 let me know if you want to make any changes, otherwise I will merge!

amburger66:
ok, no more changes from me!

PR #294, opened 2022-01-17
refactor utils policy-running methods; fix option semantics in one-step case

ronuchit:
no changes to any API or user usage. this PR quickly got out of
control..... :(

### DIFF
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -3,7 +3,7 @@
 from typing import List
 from predicators.src.approaches import create_approach
 from predicators.src.envs import BaseEnv
-from predicators.src.structs import Dataset, Task
+from predicators.src.structs import Dataset, Task, LowLevelTrajectory
 from predicators.src.settings import CFG
 from predicators.src import utils
 
@@ -21,11 +21,11 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
             task,
             env.simulate,
             env.predicates,
-            CFG.max_num_steps_check_policy,
-            annotate_traj_with_goal=True)
+            CFG.max_num_steps_check_policy)
+        assert solved, "Oracle failed on training task."
+        traj = LowLevelTrajectory(traj.states, traj.actions, task.goal)
         if CFG.option_learner != "no_learning":
             for act in traj.actions:
                 act.unset_option()
-        assert solved, "Oracle failed on training task."
         dataset.append(traj)
     return dataset
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -17,10 +17,7 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
         policy = oracle_approach.solve(
             task, timeout=CFG.offline_data_planning_timeout)
         traj, _, solved = utils.run_policy_on_task(
-            policy,
-            task,
-            env.simulate,
-            env.predicates,
+            policy, task, env.simulate, env.predicates,
             CFG.max_num_steps_check_policy)
         assert solved, "Oracle failed on training task."
         traj = LowLevelTrajectory(traj.states, traj.actions, task.goal)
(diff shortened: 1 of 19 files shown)


src/datasets/demo_only.py
```diff
@@ -21,11 +21,11 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
             task,
             env.simulate,
             env.predicates,
-            CFG.max_num_steps_check_policy,
-            annotate_traj_with_goal=True)
+            CFG.max_num_steps_check_policy)
+        assert solved, "Oracle failed on training task."
+        traj = LowLevelTrajectory(traj.states, traj.actions, task.goal)
```
ronuchit, line 26: add goal to traj, this is temporary since we're gonna change all this soon anyway

src/utils.py
```diff
@@ -232,60 +232,82 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+        max_num_steps: int,
 ) -> Tuple[LowLevelTrajectory, Video, bool]:
-    """Execute a policy on a task until goal or max steps.
-
-    Return the low-level trajectory (optionally annotated with the
-    goal), and a bool for whether the goal was satisfied at the end.
+    """Execute a policy from an initial state, using a simulator.
+    Terminates when any of these conditions hold:
+    (1) the termination_function returns True,
+    (2) the policy throws OptionPlanExhausted,
```
ronuchit, line 245: it's sad that (2) needs to be included in this function, but alas, it's all because we have the option weirdness where terminal can be True on the first step and it still works :(

src/utils.py
```diff
@@ -232,92 +238,98 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+    """A light wrapper around run_policy_on_task that returns whether the given
+    policy solves the given task."""
     _, _, solved = run_policy_on_task(policy, task, simulator, predicates,
                                       CFG.max_num_steps_check_policy)
     return solved
 
 
-def option_to_trajectory(init: State, simulator: Callable[[State, Action],
-                                                          State],
+def option_to_trajectory(init_state: State,
```
tomsilver, line 323: maybe we can actually get rid of this function altogether after you change the thing i commented above? users would need to make sure that the option is initiable, but i'm guessing that's already happening...

tomsilver (commented):
lookin good

tomsilver (approved):
lgtm after change to initiable discussed on slack

src/nsrt_learning.py
```diff
@@ -91,7 +91,7 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
         # like Holding, Handempty, etc., because when doing symbolic planning,
         # we only have predicates, and not the continuous parameters that would
         # be used to distinguish between a PickPlace that is a pick vs a place.
-        if traj.actions[t].has_option():
+        if not switch and traj.actions[t].has_option():
```
ronuchit, line 94: this `if` can only make `switch` become True, so i think this minor optimization is ok

src/nsrt_learning.py
```diff
@@ -100,13 +100,13 @@ def segment_trajectory(trajectory: GroundAtomTrajectory) -> List[Segment]:
...
-            # Special case: if the final option terminates in the state, we
-            # can safely segment without using any continuous info. Note that
+            # Special case: if the final option terminates in the final state,
+            # we can safely segment without using any continuous info. Note that
             # excluding the final option from the data is highly problematic
             # when using demo+replay with the default 1 option per replay
             # because the replay data which causes no change in the symbolic
             # state would get excluded.
-            elif traj.actions[t].get_option().terminal(traj.states[t]):
+            elif traj.actions[t].get_option().terminal(traj.states[t + 1]):
```
ronuchit, line 109: i think this was a bug earlier

tomsilver: hmm agreed

tests/test_main.py
```diff
@@ -175,3 +183,21 @@ def test_tamp_approach_failure():
     task = next(env.train_tasks_generator())[0]
     approach.solve(task, timeout=500)
     _run_testing(env, approach)
+
+
+def test_env_failure():
```
tomsilver, line 188: which change led to this test?

tomsilver (approved):
lgtm just a minor question

tests/test_main.py
ronuchit: i have no idea

PR #296, opened 2022-01-18
Cluttered place

czeng6:
Adding "place" version of cluttered_table, which places colliding cans instead of dumping. Made more challenging by limiting the starting range of the action.


src/envs/cluttered_table.py
```diff
@@ -267,3 +242,114 @@ def _any_intersection(pose: Array, radius: float,
             if distance <= (radius + other_radius):
                 return True
         return False
+
+    # Handle collision checking. We'll just threshold the angle between
```
tomsilver, line 246: thanks for refactoring this into a new function. could you move this comment inside the function itself?

src/envs/cluttered_table.py
```diff
@@ -267,3 +242,114 @@ def _any_intersection(pose: Array, radius: float,
...
+                dist = np.linalg.norm(vec2)  # type: ignore
+                if dist > colliding_can_max_dist:
+                    colliding_can_max_dist = dist
+                    colliding_can = can
+        if colliding_can is not None:
+            raise EnvironmentFailure("collision", {colliding_can})
+
+
+class ClutteredTablePlaceEnv(ClutteredTableEnv):
+    """Toy cluttered table domain."""
```
tomsilver, line 283: can you update this docstring to explain the difference from the parent class?

tomsilver (commented):
very nice job! just a few tiny things

ronuchit:
Congrats on the first PR!! I'll review after @tomsilver is happy & checks are passing :)

tomsilver:
@czeng6 looking good to me! there are some new merge conflicts to sort out -- once you resolve those, tag Rohan in a comment and he can start his review. thank you!

czeng6:
@ronuchit Hi! I think checks are passing and Tom is happy. Please take a look when you get the chance!

src/envs/cluttered_table.py
```diff
@@ -267,3 +242,122 @@ def _any_intersection(pose: Array, radius: float,
...
+        if colliding_can is not None:
+            raise EnvironmentFailure("collision", {colliding_can})
+
+
+class ClutteredTablePlaceEnv(ClutteredTableEnv):
+    """Toy cluttered table domain (place version).
+
+    This version places grasped cans instead of dumping them. As an
+    additional challenge, the action space is restricted so that actions
+    can only begin from within a 0.2 x 0.2 corner
```
ronuchit, line 290: won't this make the env impossible?

src/envs/cluttered_table.py
```diff
@@ -267,3 +242,122 @@ def _any_intersection(pose: Array, radius: float,
...
+                                          _initiable=utils.always_initiable,
+                                          _terminal=utils.onestep_terminal)
+
+    @property
+    def options(self) -> Set[ParameterizedOption]:
+        return {self._Grasp, self._Place}
+
+    @property
+    def action_space(self) -> Box:
+        # The action_space is 4-dimensional. The first two dimensions are the
```
ronuchit, line 316: instead of copy-pasting the parent class's comment, describe what's different here in your modified env. the responsibility is on the user to go read the parent class's comment if they want to understand the functionality of the main env

src/envs/cluttered_table.py
ronuchit: maybe would be good to clarify what are the implications of this change planning-wise (it's not super clear to me)

ronuchit (commented):
just minor stuff -- great work overall!!

src/envs/cluttered_table.py
czeng6: Done! Ty for the pointer

src/envs/cluttered_table.py
czeng6: I probably wasn't clear enough here. The goal is to learn to pick up colliding cans between the robot (in the 0.2x0.2 corner) and the desired can, and then place them somewhere where they are not in the way. I updated the comment in the file to say so! (The actual learning about the failure is something Tom and I have yet to work on and is the motivation for this environment).

ronuchit (approved):
Thanks for the clarifications!

PR #300, opened 2022-01-18
option to parallelize hill climbing

### LINKED ISSUE #264: implement parallelized version of hill climbing
we can evaluate all successors in parallel, which will help when the heuristic evaluation is the bottleneck (which it is for predicate learning)

ronuchit:
it's significantly slower on my machine (2-5x slower) on both cover and
painting with grammar_search_invention approach, so leaving this
disabled for now, but the code is here if we ever want to enable it in
the future

closes #264

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -132,12 +132,13 @@ class GlobalSettings:
     grammar_search_pred_complexity_weight = 1
     grammar_search_max_predicates = 50
     grammar_search_predicate_cost_upper_bound = 6
-    grammar_search_score_function = "task_planning"
+    grammar_search_score_function = "hff_lookahead_depth0"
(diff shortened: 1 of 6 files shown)


src/settings.py
```diff
@@ -132,12 +132,13 @@ class GlobalSettings:
     grammar_search_pred_complexity_weight = 1
     grammar_search_max_predicates = 50
     grammar_search_predicate_cost_upper_bound = 6
-    grammar_search_score_function = "task_planning"
+    grammar_search_score_function = "hff_lookahead_depth0"
```
ronuchit, line 135: for some reason we changed this a week ago. not sure why. changing it back to the best setting now.

tomsilver (approved):
perf thanks!

PR #301, opened 2022-01-18
clean up instances of unnecessary calls to issubset() for goal checking

ronuchit:
it's more efficient to just check that every goal atom holds

### DIFF
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -190,8 +190,7 @@ def _main() -> None:
 
     outdir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                           "results")
-    if not os.path.exists(outdir):
-        os.makedirs(outdir)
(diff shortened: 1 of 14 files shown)


analysis/grammar_search_analysis.py
```diff
@@ -190,8 +190,7 @@ def _main() -> None:
 
     outdir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                           "results")
-    if not os.path.exists(outdir):
-        os.makedirs(outdir)
+    os.makedirs(outdir, exist_ok=True)
```
ronuchit, line 193: this is a better pattern that we should prefer going forward to avoid coverage issues @tomsilver

tomsilver: cool

tomsilver (approved):
nice thanks, glad to see all of the `predicates` args getting removed too

PR #303, opened 2022-01-18
allow heuristic based lookahead score functions to use nondemos

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -656,12 +657,20 @@ def _evaluate_with_operators(self, predicates: FrozenSet[Predicate],
...
         scores = {name: 0.0 for name in self.heuristic_names}
         seen_demos = 0
+        seen_nondemos = 0
+        max_demos = CFG.grammar_search_heuristic_based_max_demos
+        max_nondemos = CFG.grammar_search_heuristic_based_max_nondemos
         for traj, atoms_sequence in pruned_atom_data:
-            if not traj.is_demo:  # we only care about demonstrations
+            # Skip this trajectory if it's not a demo and we only want demos,
+            # or if we've exceeded the budget for demos or nondemos.
+            if (self.demos_only and not traj.is_demo) or (
```
ronuchit, line 666: confusing; let's write as 3 separate `if ...: continue`

src/approaches/grammar_search_invention_approach.py
```diff
@@ -647,6 +647,7 @@ class _HeuristicBasedScoreFunction(_OperatorLearningBasedScoreFunction):
     against the demonstrations.
     """
     heuristic_names: Sequence[str]
+    demos_only: bool = field(default=True)
```
ronuchit, line 650: are we planning to make a CFG flag for this later?

src/settings.py
```diff
@@ -135,6 +135,7 @@ class GlobalSettings:
     grammar_search_score_function = "hff_lookahead_depth0"
     grammar_search_heuristic_based_weight = 10.
     grammar_search_heuristic_based_max_demos = 5
+    grammar_search_heuristic_based_max_nondemos = 50
```
ronuchit, line 138: is there no hope of it ever running with the full 500, even in cover?

src/approaches/grammar_search_invention_approach.py
tomsilver: I think 2 is clear, so I'm going to do that

src/approaches/grammar_search_invention_approach.py
tomsilver: I'm planning to make a different score function that uses `demos_only=False` later

ronuchit: that would be a subclass of this one?

src/settings.py
tomsilver: we can make it env-dependent later if you want, I imagine 500 would be fine for cover, but 50 is also seemingly fine for cover

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah well this would be a mixin, like we do for the existing lookahead score functions

ronuchit (approved):
lgtm

PR #305, opened 2022-01-18
counting based score functions

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -467,26 +467,42 @@ def _create_score_function(
     if score_function_name == "hadd_match":
         return _RelaxationHeuristicMatchBasedScoreFunction(
             initial_predicates, atom_dataset, candidates, ["hadd"])
-    match = re.match(r"([a-z\,]+)_lookahead_depth(\d+)", score_function_name)
+    match = re.match(r"([a-z\,]+)_(\w+)_depth(\d+)", score_function_name)
     if match is not None:
         # heuristic_name can be any of {"hadd", "hmax", "hff", "hsa", "lmcut"},
         # or it can be multiple heuristic names that are comma-separated, such
         # as hadd,hmax or hadd,hmax,lmcut.
+        # score_name can be {"lookahead", "count"}.
```
ronuchit, line 475: can be any of*

src/approaches/grammar_search_invention_approach.py
```diff
@@ -772,6 +792,45 @@ def _evaluate_atom_trajectory(
...
+            atoms, next_atoms = atoms_sequence[i], atoms_sequence[i + 1]
+            best_h = float("inf")
+            on_sequence_h = float("inf")
+            for predicted_next_atoms in utils.get_successors_from_ground_ops(
+                    atoms, ground_ops, unique=False):
+                # Compute the heuristic for the successor atoms.
+                h = heuristic_fn(predicted_next_atoms)
+                best_h = min(h, best_h)
+                if predicted_next_atoms == next_atoms:
+                    assert on_sequence_h in [h, float("inf")]
```
ronuchit, line 823: is it really even possible to have the same predicted_next_atoms appear more than once in this loop over successors...? won't that be impossible due to the way we cluster effects? (fine to keep this general though)

ronuchit: but we could check correctness of the above by changing this assert to `assert on_sequence_h == float("inf")` just for testing

src/approaches/grammar_search_invention_approach.py
```diff
@@ -772,6 +792,45 @@ def _evaluate_atom_trajectory(
...
+                    on_sequence_h = h
+            # Bad case 1: transition is optimal and sequence is not a demo.
+            if on_sequence_h == best_h and not is_demo:
+                score += CFG.grammar_search_off_demo_count_penalty
+            # Bad case 2: transition is not optimal and sequence is a demo.
+            elif on_sequence_h > best_h and is_demo:
+                score += CFG.grammar_search_on_demo_count_penalty
```
ronuchit, line 830: checking myself before i go further: i think on_penalty should be higher than off_penalty

ronuchit: correct hooray

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -606,6 +589,50 @@ def test_exact_lookahead_score_function():
...
+    train_tasks = next(env.train_tasks_generator())
+    dataset = create_dataset(env, train_tasks)
+    atom_dataset = utils.create_ground_atom_dataset(dataset, env.predicates)
+    for name in ["exact_lookahead", "lmcut_count_depth0"]:
+        score_function = _create_score_function(name, initial_predicates,
+                                                atom_dataset, candidates)
+        all_included_s = score_function.evaluate(set(candidates))
+        none_included_s = score_function.evaluate(set())
+        assert all_included_s < none_included_s  # good!
+        # Cover bad case 1: transition is optimal and sequence is not a demo.
```
ronuchit, line 631: why not bad case 2 too?

ronuchit (approved):
lgtm, minor

PR #306, opened 2022-01-19
suspicious states idea for count based score functions

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -676,6 +676,10 @@ def _evaluate_with_operators(self, predicates: FrozenSet[Predicate],
         seen_nondemos = 0
         max_demos = CFG.grammar_search_heuristic_based_max_demos
         max_nondemos = CFG.grammar_search_heuristic_based_max_nondemos
+        demo_atom_sets = {
```
ronuchit, line 679: i don't get what this is... what is `a`? is this just a set of all abstract states seen in the demos?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -802,22 +810,38 @@ class _HeuristicCountBasedScoreFunction(_HeuristicBasedScoreFunction):  # pylint
     the transition is optimal and the sequence is not a demo, that's
     assumed to be bad; if the transition is not optimal and the sequence
     is a demo, that's also assumed to be bad.
+
+    Also: for each successor that is one step off the atoms_sequence, if the
+    state is optimal, then check if the state is "suspicious", meaning that it
+    does not "match" any state in the demo data. The definition of match is
+    currently based on utils.unify(). It may be that this definition is too
+    strong, so we could consider others. The idea is to try to distinguish
+    states that are actually impossible (suspicious) from ones that are simply
```
ronuchit, line 819: this whole thing is quite suspicious lol. there isn't really either direction of an iff between "state is suspicious per our definition" and "state is bad for optimally achieving the goal"

src/approaches/grammar_search_invention_approach.py
```diff
@@ -802,22 +810,38 @@ class _HeuristicCountBasedScoreFunction(_HeuristicBasedScoreFunction):  # pylint
...
         for i in range(len(atoms_sequence) - 1):
             atoms, next_atoms = atoms_sequence[i], atoms_sequence[i + 1]
             best_h = float("inf")
             on_sequence_h = float("inf")
+            optimal_successors = set()
             for predicted_next_atoms in utils.get_successors_from_ground_ops(
                     atoms, ground_ops, unique=False):
                 # Compute the heuristic for the successor atoms.
                 h = heuristic_fn(predicted_next_atoms)
+                if h < best_h:
```
ronuchit, line 841: I'm confused how the logic is working here. Won't best_h only be guaranteed to be set correctly once the `get_successors_from_ground_ops` loop has gone all the way through? Why is it okay to compare to a best_h that you're still working on figuring out the value of?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -828,8 +852,24 @@ def _evaluate_atom_trajectory(self, atoms_sequence: List[Set[GroundAtom]],
             # Bad case 2: transition is not optimal and sequence is a demo.
             elif on_sequence_h > best_h and is_demo:
                 score += CFG.grammar_search_on_demo_count_penalty
+            # Bad case 3: there is a "suspicious" optimal state.
+            for successor in optimal_successors:
+                # We already know that next_atoms is not suspicious.
+                if successor == frozenset(next_atoms):
```
ronuchit, line 858: i guess this is like a minor efficiency to avoid calling _is_suspicious in the case where successor would trivially unify with next_atoms in the _is_suspicious loop anyway. Right?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -828,8 +852,24 @@ def _evaluate_atom_trajectory(self, atoms_sequence: List[Set[GroundAtom]],
...
+                    continue
+                if self._is_suspicious(successor, demo_atom_sets):
+                    score += CFG.grammar_search_suspicious_penalty
         return score
 
+    @staticmethod
+    def _is_suspicious(successor: FrozenSet[GroundAtom],
```
ronuchit, line 865: i guess one way of thinking about this is that we're asserting that the set of lifted states seen in the demonstrations is the set of "good" states, and any candidate predicate set which believes that a state outside this set is also good would be a bad predicate set.

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah right

src/approaches/grammar_search_invention_approach.py
tomsilver: right exactly

ronuchit (approved):
lgtm

PR #309, opened 2022-01-19
BEHAVIOR: Gets oracle approach to run on sorting books task!

NishanthJKumar:
Several changes to *finally* get our oracle SeSaMe approach to run on BEHAVIOR pick-and-place tasks (currently, just works on sorting books task, but should be straightforward to add new tasks).  

Summary of major changes:
* fixes the grasp sampler
* adds an error-correcting loop to movement of the hand within the grasp option model, leading to a much greater grasp success rate
* adjusts navigation sampler to better complement grasp sampler and lead to more grasp successes (by navigating closer to objects)
* changes state loading/saving to actually require file I/O. This (1) reduces memory usage and prevents our process from getting killed by the OS for using up too much RAM (which had been happening) and (2) allows perfect resets of the environment (including the state of all objects, whereas previously, some `URDFObjects` were being not reset, leading to planning failures


src/envs/behavior_options.py
```diff
@@ -514,48 +510,102 @@ def graspObjectOptionPolicy(_state: State,
...
+            current_pos = list(
+                env.robots[0].parts["right_hand"].get_position())
+            current_orn = list(
+                p.getEulerFromQuaternion(
+                    env.robots[0].parts["right_hand"].get_orientation()))
+            expected_pos = np.array(plan[hand_i][0:3])
+            # If we're not where we expect in the plan, take some corrective
+            # action
+            if not np.allclose(current_pos, expected_pos, atol=atol_xyz):
+                low_level_action = (get_delta_low_level_hand_action(
```
tomsilver, line 555: can the outer parentheses be removed?

src/envs/behavior_options.py
```diff
@@ -514,48 +510,102 @@ def graspObjectOptionPolicy(_state: State,
...
+                hand_i += 1
+            ec_loop_counter += 1
 
-        # 2 Simulate Grasp
+        # 3. Close hand and simulate grasp
         a = np.zeros(17, dtype=float)
         a[16] = 1.0
         assisted_grasp_action = np.zeros(28, dtype=float)
         assisted_grasp_action[26] = 1.0
+        if isinstance(obj_to_grasp.body_id, List):
```
tomsilver, line 588: i'm more used to seeing `isinstance(x, list)` than `isinstance(x, List)`, but I can't think of any reason to prefer the former... hopefully they're equivalent? i googled around for a minute but didn't learn anything \shrug

src/envs/behavior_options.py
```diff
@@ -1023,10 +1074,22 @@ def placeOntopObjectOptionModel(_init_state: State,
         env.robots[0].parts["right_hand"].set_position_orientation(
             target_pos, p.getQuaternionFromEuler(target_orn))
         env.robots[0].parts["right_hand"].force_release_obj()
+        obj_to_place.force_wakeup()
         # this is running a zero action to step simulator
         env.step(np.zeros(17))
+        # reset the released object to zero velocity so it doesn't
+        # fly away because of residual warp speeds from teleportation!
```
tomsilver, line 1081: warp speeds lol

tomsilver (commented):
looking good!

.gitignore
```diff
@@ -14,4 +14,5 @@ results
 videos
 saved_approaches
 saved_datasets
-analysis/results
\ No newline at end of file
+analysis/results
+tmp_behavior_states
```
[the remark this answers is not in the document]
ronuchit: note: see uses of "tempfile" in torch_models.py for a usage example

ronuchit:
I'll review after Tom's feedback is incorporated. Btw, are there any updates/pulls we have to make on any other repo? I remember we're using some fork of igibson or something...

.gitignore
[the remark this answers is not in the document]
tomsilver: Actually instead of creating a temporary directory, maybe we can just use tempfile NamedTemporaryFile and that's it; there's no need for the files to be in any particular directory.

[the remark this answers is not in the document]
NishanthJKumar: sounds good! thanks for this suggestion!

NishanthJKumar:
yeah - to run this code, you'd need to pull our fork of iGibson and BDDL.
I'll update the README with instructions for behavior in a separate PR

.gitignore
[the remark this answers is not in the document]
NishanthJKumar: i think we need a temp directory, since each `save_checkpoint` call saves two different files

src/envs/behavior.py
```diff
@@ -329,8 +333,32 @@ def current_ig_state_to_state(self) -> State:
...
+
+        def save_unique_checkpoint(sim: "Simulator",
+                                   save_dir: str) -> Tuple[str, str]:
+            unique_id = int(np.random.randint(size=1, low=0, high=10**6))
+            path = save_dir + str(unique_id) + "/"
+            isExist = os.path.exists(path)
+            if not isExist:
+                os.makedirs(path)
+            simulator_state = save_checkpoint(sim, path)
+            return str(simulator_state), str(unique_id)
```
tomsilver, line 345: I just took another look here and noticed the `str(simulator_state)`. what's going on there? isn't `simulate_state` like a really large and gross dictionary?

.gitignore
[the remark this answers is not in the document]
NishanthJKumar: i'm not sure how/when to automatically delete these directories though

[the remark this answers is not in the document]
tomsilver: cool. so then `tempfile.TemporaryDirectory` is probably what we want.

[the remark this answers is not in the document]
tomsilver: or maybe `dirpath = tempfile.mkdtemp()`. and btw, we might want to automatically delete these to avoid clogging up our storage. see: https://stackoverflow.com/questions/3223604/how-to-create-a-temporary-directory-and-get-its-path-file-name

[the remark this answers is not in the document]
tomsilver: haha I just saw your comment, GMTA! yeah let's see...

src/envs/behavior.py
```diff
@@ -110,8 +112,9 @@ def __init__(self) -> None:
                 self._options.add(option)
 
     def simulate(self, state: State, action: Action) -> State:
-        assert state.simulator_state is not None
-        self.igibson_behavior_env.task.reset_scene(state.simulator_state)
+        if not state.allclose(self.current_ig_state_to_state()):
```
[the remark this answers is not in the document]
NishanthJKumar: I feel like there's not too much point in this, since it'll change 2 lines into 1 line, but we'll have to add a new class function that itself calls `load_checkpoint_states`, which is now its own separate function. I'm more inclined to keep things this way, but take a look at the new refactor and let me know if you still feel we should implement this change and I'm happy to do it!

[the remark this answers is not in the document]
tomsilver: got it, makes sense. I'll defer to your judgement!

src/envs/behavior.py
```diff
@@ -21,9 +22,14 @@
     from igibson.robots.behavior_robot import BRBody
     from igibson.activity.bddl_backend import SUPPORTED_PREDICATES, \
         ObjectStateUnaryPredicate,ObjectStateBinaryPredicate
+    from igibson.utils.checkpoint_utils import save_checkpoint, load_checkpoint
 
     _BEHAVIOR_IMPORTED = True
     bddl.set_backend("iGibson")  # pylint: disable=no-member
+    if not os.path.exists("tmp_behavior_states/"):
```
ronuchit, line 29: Use "exist_ok" argument instead of this if (see other usages in this repo for examples)

src/envs/behavior.py
```diff
@@ -460,6 +469,14 @@ def _create_type_combo_name(original_name: str,
         return f"{original_name}-{type_names}"
 
 
+def load_checkpoint_state(s: State, env: "behavior_env.BehaviorEnv") -> None:
+    """Sets the underlying iGibson environment to a particular saved state."""
+    assert s.simulator_state is not None
+    load_checkpoint(env.simulator, "tmp_behavior_states/", s.simulator_state)
+    # We step the environment to update the visuals of where the robot is!
+    env.step(np.zeros(17))
```
ronuchit, line 477: can we do `env.step(np.zeros_like(env.action_space.shape))`?

src/envs/behavior_options.py
```diff
@@ -249,7 +243,8 @@ def navigateToOptionPolicy(_state: State,
 
 
 def create_navigate_option_model(
-    plan: List[List[float]], _original_orientation: List[List[float]]
+        plan: List[List[float]], _original_orientation: List[List[float]],
+        _obj_to_nav_to: "URDFObject"
```
ronuchit, line 247: just curious, how come this argument isn't used? how can we make a navigation option without knowing what we're navigating to?

src/envs/behavior_options.py
```diff
@@ -514,48 +509,102 @@ def graspObjectOptionPolicy(_state: State,
...
+                    env,
+                    np.array(current_pos),
+                    np.array(current_orn),
+                    np.array(plan[hand_i][0:3]),
+                    np.array(plan[hand_i][3:]),
+                )
+                # if the corrective action is 0, move on
+                if np.allclose(
+                        low_level_action,
+                        np.zeros((17, 1)),
```
ronuchit, line 564: as in another comment i made, can we somehow use env.action_space here?

ronuchit: just trying to reduce the number of magic numbers where possible

src/envs/behavior_options.py
```diff
@@ -602,15 +651,15 @@ def grasp_obj_at_pos(
...
+    if not (volume < 0.3 * 0.3 * 0.3 and
             not obj.main_body_is_fixed):  # say we can only grasp small objects
         print(f"PRIMITIVE: grasp {obj.name} fail, too big or fixed")
         return None
 
     # If the object is too far away, fail and return None
     if (np.linalg.norm(  # type: ignore
             np.array(obj.get_position()) -
-            np.array(env.robots[0].get_position())) < 2):
+            np.array(env.robots[0].get_position())) > 2):
```
ronuchit, line 662: lol was this a bug before?

src/envs/behavior_options.py
```diff
@@ -602,15 +651,15 @@ def grasp_obj_at_pos(
 
     # If the object is too big to be grasped, or bolted to its surface,
     # fail and return None
-    if (volume < 0.3 * 0.3 * 0.3 and
+    if not (volume < 0.3 * 0.3 * 0.3 and
```
ronuchit, line 654: lol was this a bug before?

ronuchit (approved):
looks fine to me, minor comments but also wait for @tomsilver 's approval

src/envs/behavior_options.py
NishanthJKumar: because we already have the plan that was computed by RRT!

src/envs/behavior_options.py
NishanthJKumar: yup, and a pretty serious one lol

src/envs/behavior_options.py
NishanthJKumar: yup, and it's why grasping wasn't working

src/envs/behavior_options.py
```diff
@@ -514,49 +509,103 @@ def graspObjectOptionPolicy(_state: State,
...
+                env.step(low_level_action)
+            # Else, move the hand_i pointer to make the expected position
+            # the next position in the plan.
+            else:
+                hand_i += 1
+            ec_loop_counter += 1
 
-        # 2 Simulate Grasp
+        # 3. Close hand and simulate grasp
         a = np.zeros(17, dtype=float)
```
tomsilver, line 583: here's another place where we could use `action_space` instead -- check all occurrences of 17

tomsilver (approved):
one tiny thing, otherwise lgtm, merge when ready!

src/envs/behavior_options.py
ronuchit: `git grep 17` :D

PR #311, opened 2022-01-19
fix bug in run_policy_until

tomsilver:


ronuchit (approved):
agh sorry my bad! wonder why this wasn't caught by the previous test you had put in...

tomsilver:
>agh sorry my bad! wonder why this wasn't caught by the previous test you had put in...
no worries. I think the previous test got stuck on the first state, which the old implementation did correctly break out of

PR #313, opened 2022-01-20
BEHAVIOR: Save RRT plans in option memory

NishanthJKumar:
Refactors behavior env and options such that an option's `memory` saves the RRT plan and then passes this to either an option model or an option policy, thereby avoiding having to run RRT twice (which is inefficient + creates a number of weird issues).


src/envs/behavior_options.py
```diff
@@ -177,11 +177,13 @@ def navigate_to_param_sampler(rng: Generator,
...
 
 def create_navigate_policy(
-    plan: List[Tuple[float, float,
-                     float]], original_orientation: Tuple[float, float, float]
+    plan: List[Tuple[float, float, float]]
 ) -> Callable[[State, "BehaviorEnv"], Tuple[Array, bool]]:
     """Instantiates and returns a navigation option policy given an RRT
     plan."""
+    # We remove the last element from the input policy since this is the
+    # original orientation
```
tomsilver, line 185: so is the plan in reverse order?

src/envs/behavior.py
```diff
@@ -75,7 +81,11 @@ def __init__(self) -> None:
                     params_space=Box(parameter_limits[0], parameter_limits[1],
                                      (param_dim, )),
                     env=self.igibson_behavior_env,
-                    controller_fn=controller_fn,  # type: ignore
+                    planner_fn=planner_fn,
+                    # NOTE: type ignores below are necessary because these
+                    # functions take inputs of slightly different types.
```
tomsilver, line 86: hmm, where does the difference arise? it looks to me like the policy and option model functions now both just take in plans, is that not the case?

tomsilver (commented):
love this refactor! great stuff

src/envs/behavior.py
```diff
@@ -28,7 +28,10 @@
     _BEHAVIOR_IMPORTED = False
 from gym.spaces import Box
 from predicators.src.envs.behavior_options import navigate_to_obj_pos, \
-        grasp_obj_at_pos,place_ontop_obj_pos
+        grasp_obj_at_pos,place_ontop_obj_pos, create_navigate_policy, \
```
ronuchit, line 31: add space after comma

src/envs/behavior.py
```diff
@@ -58,14 +61,17 @@ def __init__(self) -> None:
 
         # name, controller_fn, param_dim, arity, parameter upper and
```
ronuchit, line 62: update comment

src/envs/behavior.py
```diff
@@ -433,8 +443,8 @@ def _create_type_combo_name(original_name: str,
 
 
 def make_behavior_option(name: str, types: Sequence[Type], params_space: Box,
-                         env: "behavior_env.BehaviorEnv",
-                         controller_fn: Callable,
+                         env: "behavior_env.BehaviorEnv", planner_fn: Callable,
+                         policy_fn: Callable, option_model_fn: Callable,
```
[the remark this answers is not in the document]
ronuchit: +1. Actually it would be really good to get rid of these generic `Callable` types altogether... running `git grep "Callable" | grep -v "Callable\[" | grep -v "from typing"` gives me:
```
src/envs/behavior.py:                         controller_fn: Callable,
src/envs/behavior.py:                         object_to_ig_object: Callable,
```
on master right now. Ideally, all the Callables should have clearly specified parameter and return types. If that's not able to be done in this case, perhaps you need a base class with a well-defined API for what a policy_fn and option_model_fn should do.

src/envs/behavior.py
```diff
@@ -454,26 +465,19 @@ def _initiable(state: State, memory: Dict, objects: Sequence[Object],
                    params: Array) -> bool:
         igo = [object_to_ig_object(o) for o in objects]
         assert len(igo) == 1
-        if memory.get("policy_controller") is None:
+        if memory.get("rrt_plan") is None:
```
ronuchit, line 468: The logic is written in a bit of a weird way here with the hanging `return True` at the end. Instead can we do:
```
if memory.get("rrt_plan") is not None:
    # TODO: add a comment here explaining what exactly it means if we are here
    return True
# We want to reset the state of the environment to
etc etc etc
```

src/envs/behavior_options.py
ronuchit: maybe clarify what the inputs mean in the docstring? What exactly is an element of `plan`?

NishanthJKumar: The plan is in normal, forward order, an each element is a 3-tuple containing `(x, y, rot)` and is interpreted as a waypoint for the robot to pass thru. I needed the original orientation of the robot, which isn't directly accessible from the first element of the plan (since `rot` is only a yaw, and I need roll and pitch as well), so I just tacked it onto the end of the plan.

NishanthJKumar: will update docstring to clarigy this

src/envs/behavior_options.py
```diff
@@ -385,14 +374,12 @@ def sample_fn(env: "BehaviorEnv",
     p.restoreState(state)
     p.removeState(state)
 
-    if not ret_option_model:
-        print(f"PRIMITIVE: navigate to {obj.name} success! Plan found with " +
-              f"continuous params {pos_offset}. Executing now.")
-        return create_navigate_policy(plan, original_orientation)
-
+    # NOTE: we need the original_orientation in our option policy, so we
```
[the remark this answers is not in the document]
NishanthJKumar: see above comment

src/envs/behavior.py
NishanthJKumar: yes, but in the case of navigation, a plan is a list of tuples, and otherwise it is a list of lists. I think i can standardize on a list of lists to fix this though

src/envs/behavior.py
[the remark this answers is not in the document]
NishanthJKumar: I updated this and got rid of the `type: ignore` comments, but there's a weird mypy error now that I can't figure out (perhaps this is because of the way we call `make_behavior_options` by iterating through the elements of a list?)

src/envs/behavior.py
```diff
@@ -53,19 +57,40 @@ def __init__(self) -> None:
...
+                                               create_grasp_policy,
+                                               create_place_policy
+                                           ]
+        option_model_fns: List[Callable[[List[List[float]]], Callable[
+            [State, "behavior_env.BehaviorEnv"], None]]] = [
+                create_navigate_option_model, create_grasp_option_model,
+                create_place_option_model
+            ]
+
+        # name, planning_fn, option_policy_fn, option_model_fn,
```
ronuchit, line 81: planner_fn*

src/envs/behavior.py
```diff
@@ -53,19 +57,40 @@ def __init__(self) -> None:
...
-            ("NavigateTo", navigate_to_obj_pos, 2, 1, (-5.0, 5.0)),
-            ("Grasp", grasp_obj_at_pos, 3, 1, (-np.pi, np.pi)),
-            ("PlaceOnTop", place_ontop_obj_pos, 3, 1, (-1.0, 1.0)),
+        planner_fns: List[Callable[[
+            "behavior_env.BehaviorEnv", Union[
+                "URDFObject", "RoomFloor"], Array, Optional[Generator]
+        ], Optional[List[List[float]]]]] = [
+            navigate_to_obj_pos, grasp_obj_at_pos, place_ontop_obj_pos
+        ]
+        policy_fns: List[Callable[[List[List[float]]],
```
ronuchit, line 68: rename to option_policy_fns to match comment below

src/envs/behavior.py
```diff
@@ -53,19 +57,40 @@ def __init__(self) -> None:
...
         self.igibson_behavior_env.robots[0].initial_z_offset = 0.7
-
         self._type_name_to_type: Dict[str, Type] = {}
 
-        # name, controller_fn, param_dim, arity, parameter upper and
-        # lower bounds
-        controllers = [
-            ("NavigateTo", navigate_to_obj_pos, 2, 1, (-5.0, 5.0)),
-            ("Grasp", grasp_obj_at_pos, 3, 1, (-np.pi, np.pi)),
-            ("PlaceOnTop", place_ontop_obj_pos, 3, 1, (-1.0, 1.0)),
```
ronuchit, line 64: Add comments to describe the I/O of each of these 3 things

src/envs/behavior.py
```diff
@@ -454,27 +491,26 @@ def _initiable(state: State, memory: Dict, objects: Sequence[Object],
...
-            memory["has_terminated"] = False
-            return policy_controller is not None
-        return True
+        if memory.get("rrt_plan") is not None:
+            # In this case, an rrt_plan has already been found for this
+            # option (most likely, this will occur when executing a
+            # series of options after having planned).
+            return True
+
+        # In this case, we want to reset the state of the environment
```
ronuchit, line 500: move this comment into the `if`

ronuchit (commented):
LGTM, just minor things. Let @tomsilver do the final approval!

src/envs/behavior_options.py
```diff
@@ -1123,11 +1088,9 @@ def place_ontop_obj_pos(
               " to find plan to continuous params" + f" {place_rel_pos}")
         return None
 
-    if not ret_option_model:
-        print(f"PRIMITIVE: placeOnTop {obj.name} success! Plan found with " +
-              f"continuous params {place_rel_pos}. Executing now.")
-        return create_place_policy(plan)
-
+    original_orientation = list(
```
tomsilver, line 1091: if env gets changed within `place_obj_plan`, I guess it's possible that this is not the "original" orientation with respect to `state`. is that matter / does that matter?

tomsilver (approved):
one question but lgtm!

src/envs/behavior_options.py
NishanthJKumar: good question! Two things:
1. `place_obj_plan` resets the state at the end, so we should guaranteeably be back to the robot state consistent with `state`
2. Barring the above, we only use `original_orientation` for navigation; it's vestigial for grasping and placing

tomsilver: cool cool. merge when ready!

PR #314, opened 2022-01-20
rename some things in grammar search

ronuchit:
Names were getting a bit out of date:
* grammar_search_size_weight renamed to
grammar_search_operator_size_weight
* Both energy-based and count-based perform one-step lookahead, so it
didn't make sense that the energy-based one was called "Lookahead"
everywhere. Renamed to "Energy" for better distinction.
* "depth" in the grammar_search_score_function was ambiguous, especially
due to the addition of grammar_search_hill_climbing_depth. Renamed to
"lookahead_depth".


tomsilver (approved):
thanks this is good

PR #317, opened 2022-01-20
fix nondeterminism in parallelized hillclimbing 

tomsilver:


ronuchit (approved):
some confusion with naming (this PR also captures #314) but whatever

PR #318, opened 2022-01-20
Interactive learning: filter out static predicates

amburger66:
Filters out static predicates so they aren't considered during glib sampling.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -830,6 +830,21 @@ def extract_preds_and_types(
     return preds, types
 
 
+def get_static_preds(ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
+                     predicates: Set[Predicate]) -> Set[Predicate]:
+    """Get predicates that are static with respect to the operators."""
+    static_preds = set()
+    for pred in predicates:
+        # This predicate is not static if it appears in any op's effects.
+        if any(
+                any(atom.predicate == pred for atom in op.add_effects) or any(
+                    atom.predicate == pred for atom in op.delete_effects)
+                for op in ground_ops):
+            continue
+        static_preds.add(pred)
+    return static_preds
+
+
 def get_static_preds_atoms(
         ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
         atoms: Collection[GroundAtom]
--- a/src/utils.py
+++ b/src/utils.py
@@ -830,7 +830,7 @@ def extract_preds_and_types(
     return preds, types
 
 
-def get_static_preds(ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
+def get_static_preds(ops: Collection[NSRTOrSTRIPSOperator],
                      predicates: Set[Predicate]) -> Set[Predicate]:
     """Get predicates that are static with respect to the operators."""
     static_preds = set()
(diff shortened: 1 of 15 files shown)


tomsilver:
sorry I meant to mention this on our call, but I got so distracted with the crazy bug...

instead of manually specifying the static predicates, can we detect them automatically? see `utils.get_static_preds_atoms`

src/utils.py
```diff
@@ -830,6 +830,21 @@ def extract_preds_and_types(
     return preds, types
 
 
+def get_static_preds(ground_ops: Collection[GroundNSRTOrSTRIPSOperator],
```
tomsilver, line 833: so this utility is useful when we don't have just one initial state, is that the idea?

that makes sense. can we actually then do this on the basis of lifted NSRTs/operators instead of ground ones, to avoid the potentially expensive grounding? I believe we'll be able to identify the same predicates from the lifted operators alone, but let me know if I'm wrong.

tomsilver: oh but actually,, we have a `state` here. so why do we need another utility?

amburger66: oh i did not see this. hmmm

amburger66: wait so do we want to avoid the grounding or should I just use the existing utility?

ronuchit: i'll try to do a light refactor that might help here. will let you know if it works...

ronuchit: see #320

tomsilver: >wait so do we want to avoid the grounding or should I just use the existing utility?
let's use the existing utility for now, unless grounding takes more than ~1 second, in which case we can use rohan's adaption of your new utility

ronuchit: why not just use the utility i just made? i made it for this reason lol

tomsilver: because the one that uses ground nsrts is strictly better, efficiency aside, I think. it will be able to filter out more static things.

ronuchit: you just have to use the utils.py from master, and then Amber's change to src/approaches/interactive_learning_approach.py should be all you need

tomsilver: @amburger66 sorry for the back and forth here, rohan and I touched base and we're on the same page now. so here's what we think:
1. if grounding is not a noticeable slowdown, then you can use `utils.get_static_atoms`. in this case, you will need to refactor `glib_sample`, perhaps doing the static atoms calculation inside of glib_sample instead (because right now you're just filtering out predicates rather than static atoms).
2. if grounding is causing a slowdown, then you can use `utils.get_static_predicates`. here you wouldn't need to change the `glib_sample` code.

let me know if that's clear. thanks again!

tomsilver: (minor caveat: we realized we actually just made a bug in the new `utils.get_static_atoms`, which we'll fix shortly in a new PR 😬 )

ronuchit: fixing that bug in #321. science is hard!

amburger66: makes sense!

amburger66: keeping it as is since grounding is ~2.5s slower (playroom has maaaany static atoms)

tomsilver (approved):
yay such a small change! :)

PR #321, opened 2022-01-20
fix get_static_atoms()

ronuchit:
we realized that static_atoms can be more than just the atoms whose predicate is static under the operators. see updated test for an example


tomsilver (approved):
I checked carefully and lgtm

PR #322, opened 2022-01-21
minor renaming predicates -> candidate_predicates

ronuchit:
to allow better distinguishability from self._initial_predicates


tomsilver (approved):
thanks

PR #323, opened 2022-01-21
Interactive approach: modified teacher dataset creation

amburger66:
a better way for the teacher to pick which ground atoms to comment on while adhering to `teacher_dataset_label_ratio` as closely as possible. now ensures the ratio for each predicate even when multiple predicates are being learned


tomsilver (commented):
Looks great! Can you add a test to ensure that the ratio is actually approximately observed?

tomsilver (approved):
lgtm!

PR #324, opened 2022-01-21
Fix metrics `min_skeletons_optimized`

amburger66:
initialization needs to be inside `BaseApproach.reset_metrics()` which is called before every test run

### DIFF
--- a/src/approaches/tamp_approach.py
+++ b/src/approaches/tamp_approach.py
@@ -58,6 +58,14 @@ class TAMPApproach(BaseApproach):
 
         return _policy
 
+    def reset_metrics(self) -> None:
+        """Reset the metrics dictionary."""
+        super().reset_metrics()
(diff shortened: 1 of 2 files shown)


ronuchit (approved):
sorry, i should have caught this before

ronuchit:
oh wait actually can you make tamp_approach.py do this (in an overriden method) rather than base_approach? not all approaches need to have this field in metrics

ronuchit (changes requested):
(see comment)

src/approaches/tamp_approach.py
```diff
@@ -62,6 +58,14 @@ def _policy(s: State) -> Action:
 
         return _policy
 
+    def reset_metrics(self) -> None:
+        """Reset the metrics dictionary."""
```
ronuchit, line 62: remove docstring (linter won't complain)

ronuchit (approved):

PR #325, opened 2022-01-22
Behavior task arguments

wmcclinton:
Enables automatic switching between different houses and tasks for BEHAVIOR via command line arguments


src/envs/behavior.py
```diff
@@ -63,6 +67,7 @@ def __init__(self) -> None:
         )
         self.igibson_behavior_env.robots[0].initial_z_offset = 0.7
         self._type_name_to_type: Dict[str, Type] = {}
+        self.env_config_file = config_file
```
tomsilver, line 70: is this used anywhere?

tomsilver (approved):
looks good! just a small question

src/envs/behavior.py
NishanthJKumar: +1

src/settings.py
```diff
@@ -49,11 +49,13 @@ class GlobalSettings:
...
         "configs",
-        # "njk_re-shelving_library_books_full_obs.yaml",
-        "njk_sorting_books_full_obs.yaml")
+        "wbm3_modifiable_full_obs.yaml",
+    )
     behavior_mode = "headless"  # headless, pbgui, iggui
     behavior_action_timestep = 1.0 / 10.0
     behavior_physics_timestep = 1.0 / 120.0
+    behavior_task_name = "re-shelving_library_books"
+    behavior_scene_id = "Pomaria_1_int"
```
NishanthJKumar, line 58: Small thing, but could we change the name of this to something other than `behavior_scene_id`? Really, this is a house/environment/scene name, so could we change it to '<something something> name` instead of 'id` so users don't get confused and think this should be a number?

src/envs/behavior.py
ronuchit: +1

src/settings.py
ronuchit: +1

ronuchit: behavior_task_name and behavior_scene_name would be good i think?

ronuchit (changes requested):
no new comments, but echoing others' comments

src/settings.py
wmcclinton: Sure, but originally I kept it as behavior_scene_id because that's what they decided to call it in the .yaml file.

ronuchit: oh then maybe we should keep this as is. could you also add a comment describing what are the allowable values for these settings? maybe a link to a yaml file with all the options spelled out?

src/envs/behavior.py
wmcclinton: No it's not anymore, good catch!

src/settings.py
```diff
@@ -54,11 +54,13 @@ class GlobalSettings:
...
         "examples",
         "configs",
-        # "njk_re-shelving_library_books_full_obs.yaml",
-        "njk_sorting_books_full_obs.yaml")
+        "wbm3_modifiable_full_obs.yaml",
+    )
     behavior_mode = "headless"  # headless, pbgui, iggui
     behavior_action_timestep = 1.0 / 10.0
     behavior_physics_timestep = 1.0 / 120.0
+    behavior_task_name = "re-shelving_library_books"
```
ronuchit, line 62: Would it be easy/possible to add a comment describing what the options are for both of these settings? Maybe a link to a yaml file which defines all the options?

ronuchit (approved):
lgtm, just one small question

PR #326, opened 2022-01-24
adds error-correction to grasp option policy

### LINKED ISSUE #242: BEHAVIOR: Include error-correcting closed-loop code for grasp option
The grasp option is the only one that doesn't have an error-correction mechanism built into the controller. Probably a useful thing to do, especially given how much grasp fails in practice...

### LINKED ISSUE #273: BEHAVIOR: Make the `get_delta_low_level_hand_action` second argument take in something more specific than an `env`
We could probably make it take in `env.robots[0].parts["right_hand"].parent.parts["body"]`, but need to test before!

NishanthJKumar:
* minor change to place option policy
* also makes `get_delta_low_level_hand_action` 2nd argument more specific than just an env
* Fixes issue that was breaking grasp/place option policy execution
Resolves #242 and resolves #273


src/envs/behavior.py
```diff
@@ -89,7 +89,7 @@ def __init__(self) -> None:
         option_elems = [
             ("NavigateTo", planner_fns[0], option_policy_fns[0],
              option_model_fns[0], 2, 1, (-5.0, 5.0)),
-            ("Grasp", planner_fns[1], option_policy_fns[2],
+            ("Grasp", planner_fns[1], option_policy_fns[1],
```
ronuchit, line 92: lol oh gosh, was that a bug?

src/envs/behavior_options.py
```diff
@@ -25,6 +25,8 @@
         get_aabb_extent,
     )
     from igibson.robots.robot_base import BaseRobot  # pylint: disable=unused-import
+    from igibson.robots.behavior_robot import BRBody  # pylint: disable=unused-import
```
ronuchit, line 28: what's the purpose of this import?

src/envs/behavior_options.py
```diff
@@ -461,46 +462,160 @@ def create_grasp_policy(
...
+                                    dtype=np.float32), done_bit
+                low_level_action = (get_delta_low_level_hand_action(
+                    env.robots[0].parts["body"],
+                    np.array(current_pos),
+                    np.array(current_orn),
+                    np.array(plan[0][0:3]),
+                    np.array(plan[0][3:]),
+                ))
+
+                # But if the corrective action is 0
```
ronuchit, line 580: is this an incomplete comment?

src/envs/behavior_options.py
```diff
@@ -461,46 +462,160 @@ def create_grasp_policy(
...
+            low_level_action = np.zeros(env.action_space.shape, dtype=float)
+            done_bit = True
+            print("PRIMITIVE: grasp policy completed execution!")
+
+        else:
+            # Placing Phase 3: getting the hand back to
             # resting position near the robot.
-            ret_action = get_delta_low_level_hand_action(
-                env,
+            low_level_action = get_delta_low_level_hand_action(
```
ronuchit, line 605: add comment describing what's going on with all this indexing into reversed_plan?

ronuchit (approved):
lgtm, minor questions

src/envs/behavior_options.py
```diff
@@ -382,7 +384,7 @@ def grasp_obj_param_sampler(rng: Generator) -> Array:
 
 
 def get_delta_low_level_hand_action(
-    env: "BehaviorEnv",
+    body: "BehaviorEnv",
```
tomsilver, line 387: is the type here wrong now?

src/envs/behavior_options.py
```diff
@@ -461,46 +462,160 @@ def create_grasp_policy(
 
     def graspObjectOptionPolicy(_state: State,
                                 env: "BehaviorEnv") -> Tuple[Array, bool]:
+        nonlocal plan
+        nonlocal reversed_plan
```
tomsilver, line 466: can we keep track of just `plan` or `reversed_plan` rather than both?

src/envs/behavior_options.py
```diff
@@ -461,46 +462,160 @@ def create_grasp_policy(
...
+        # 1. Get current position and orientation
+        current_pos, current_orn_quat = p.multiplyTransforms(
+            env.robots[0].parts["right_hand"].parent.parts["body"].new_pos,
+            env.robots[0].parts["right_hand"].parent.parts["body"].new_orn,
+            env.robots[0].parts["right_hand"].local_pos,
+            env.robots[0].parts["right_hand"].local_orn,
+        )
+        current_orn = p.getEulerFromQuaternion(current_orn_quat)
+
+        if (  # pylint:disable=no-else-return
```
tomsilver, line 484: what's going on with this pylint?

src/envs/behavior.py
NishanthJKumar: yup - it was executing place instead of grasp and I was confused for about an hour...

src/envs/behavior_options.py
NishanthJKumar: woops - I forgot to use this. It's for the type-spec of the second argument of `get_delta_low_level_hand_action`

src/envs/behavior_options.py
NishanthJKumar: yup! Thanks for pointing out; fixed!

src/envs/behavior_options.py
NishanthJKumar: yup, fixed!

src/envs/behavior_options.py
NishanthJKumar: pylint doesn't like there being an 'elif' after an 'if' with a return statement, but i'm not sure if there's a better way to organize the logic (lmk if you see one!). Will add a comment explaining this though

src/envs/behavior_options.py
NishanthJKumar: it's possible but would require the use of a pointer instead of `pop()` like we're currently doing. My sense is that this will complicate the logic here a bit more, but tbh it's quite messy as is. I'll try to see if I can refactor to make it cleaner/easier to understand!

src/envs/behavior_options.py
ronuchit: just change the elif to an if (and un-indent any else if you have one)

ronuchit: pylint is just saying that there's no point using `elif` bc if the `if` statement is run, it's guaranteed that the function will `return` somewhere inside it

src/envs/behavior_options.py
tomsilver: when I wrote this I didn't understand that the policy first does the plan, then does the reversed_plan, so I think it's fine to keep both. thanks for clearing it up!

tomsilver (approved):

PR #327, opened 2022-01-24
cover: allow some chance of holding a block in the initial state

tomsilver:


src/envs/cover.py
```diff
@@ -251,7 +251,15 @@ def _create_initial_state(self, rng: np.random.Generator) -> State:
             data[target] = np.array([0.0, 1.0, width, pose])
         # [hand]
         data[self._robot] = np.array([0.0])
-        return State(data)
+        state = State(data)
+        # Allow some chance of holding a block in the initial state.
+        if rng.uniform() < 0.5:
```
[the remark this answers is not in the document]
tomsilver: I'll make it a setting, but I think we'll want to do a number of experiments with it, so I do think it should be on master

src/envs/cover.py
```diff
@@ -251,7 +251,16 @@ def _create_initial_state(self, rng: np.random.Generator) -> State:
             data[target] = np.array([0.0, 1.0, width, pose])
         # [hand]
         data[self._robot] = np.array([0.0])
-        return State(data)
+        state = State(data)
+        # Allow some chance of holding a block in the initial state.
+        if CFG.cover_initial_holding_prob > 0 and \
```
ronuchit, line 256: we don't need the first clause here, just the second is sufficient

tests/envs/test_cover.py
```diff
@@ -85,6 +85,8 @@ def test_cover():
                                       max_num_steps=100)
     assert len(traj.states) == 2
     assert traj.states[0].allclose(traj.states[1])
+    # Leave the default to 0.0 for other tests.
```
ronuchit, line 88: grammar... "Revert to 0.0 for other tests." ?

tests/envs/test_cover.py
```diff
@@ -11,7 +11,7 @@
 
 def test_cover():
     """Tests for CoverEnv class."""
-    utils.update_config({"env": "cover"})
+    utils.update_config({"env": "cover", "cover_initial_holding_prob": 0.5})
```
ronuchit, line 14: This is a sort of weird test. It would be better to avoid randomness in the tests, instead of doing this, just add a new test that sets this prob to 1.0, and check that all the train/tests have some object being held...
`assert sum(state.get(obj, "held") for obj in state if obj.type.name == "block") == 1`
and you can add this same assert but with `0` at the end for the existing test.

ronuchit (approved):
lgtm thanks!!

PR #328, opened 2022-01-24
Painting initial state updates

tomsilver:


src/envs/painting.py
```diff
@@ -246,30 +246,35 @@ def _transition_place(self, state: State, action: Action) -> State:
...
+        if self.table_lb < y < self.table_ub:
+            receptacle = "table"
+        elif self.shelf_lb < y < self.shelf_ub:
+            receptacle = "shelf"
         elif self.box_lb < y < self.box_ub:
-            shelf_or_box = "box"
+            receptacle = "box"
         else:
-            # Cannot place outside of shelf or box
+            # Cannot place outside of table, shelf or box
```
ronuchit, line 257: to be or not to be oxford comma users, that is the question 😛

src/envs/painting.py
```diff
@@ -246,30 +246,35 @@ def _transition_place(self, state: State, action: Action) -> State:
...
             # Cannot place in box if lid is not open
             raise EnvironmentFailure("box lid is closed", {self._lid})
         # Detect top grasp vs side grasp
         rot = state.get(self._robot, "gripper_rot")
+        top_or_side = "neither"
         if rot > self.top_grasp_thresh:
             top_or_side = "top"
         elif rot < self.side_grasp_thresh:
             top_or_side = "side"
-        else:
```
ronuchit, line 266: why not keep this `else` as is? it's impossible for top_or_side to be neither. since you're grasping an object currently, you must have one of the two grasp poses on it by definition.

src/ground_truth_nsrts.py
```diff
@@ -749,6 +749,42 @@ def openlid_sampler(state: State, rng: np.random.Generator,
...
+            LiftedAtom(GripperOpen, [robot]),
+        }
+        delete_effects = {
+            LiftedAtom(Holding, [obj]),
+            LiftedAtom(holding_rot_pred, [robot]),
+        }
+
+        def placeontable_sampler(state: State, rng: np.random.Generator,
+                                 objs: Sequence[Object]) -> Array:
+            # Always release the object where it is, to avoid the possibility
```
ronuchit, line 774: I see. So you're avoiding the EnvironmentFailure above. I guess we're hoping that a learned sampler would have this same behavior?

src/envs/painting.py
```diff
@@ -490,15 +497,25 @@ def _get_tasks(self,
...
                     goal.add(GroundAtom(self._IsBoxColor, [obj, self._box]))
                 else:
                     goal.add(GroundAtom(self._InShelf, [obj, self._shelf]))
                     goal.add(GroundAtom(self._IsShelfColor,
                                         [obj, self._shelf]))
-            tasks.append(Task(State(data), goal))
+            state = State(data)
+            # Sometimes start out holding an object, possibly with the wrong
+            # grip, so that we'll have to put it on the table and regrasp
+            if rng.uniform() < 0.5:
```
ronuchit, line 511: same as in the Cover PR #327 , we should make this a setting and add a test specifically for it

src/envs/painting.py
tomsilver: that's not true -- if the rotation is 0.5 or close to it, it's neither a top nor side grasp

src/envs/painting.py
```diff
@@ -490,15 +497,25 @@ def _get_tasks(self,
...
+                rot = rng.choice([0.0, 0.1])
```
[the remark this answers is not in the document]
tomsilver: 🤦

src/ground_truth_nsrts.py
tomsilver: yeah

src/envs/painting.py
```diff
@@ -246,30 +246,35 @@ def _transition_place(self, state: State, action: Action) -> State:
...
-        if (shelf_or_box, top_or_side) not in [("shelf", "side"),
-                                               ("box", "top")]:
+        if receptacle == "shelf" and top_or_side != "side":
+            return next_state
+        if receptacle == "box" and top_or_side != "top":
             return next_state
+        # Detect collisions
+        collider = self._get_object_at_xyz(state, x, y, z)
+        if collider is not None and collider != held_obj:
+            raise EnvironmentFailure("Collision during placing.", {collider})
```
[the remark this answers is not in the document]
tomsilver: (resolved on slack, decided to keep this)

src/envs/painting.py
ronuchit: that's true. i guess it can't happen with our current samplers but it could happen in general. and then are we saying here that to place back onto the table, it doesn't matter what grasp we have? technically we'd need a ground truth NSRT for that too.... or we can just accept that our ground truth NSRTs for PlaceOnTable are wrong

src/settings.py
```diff
@@ -41,6 +41,9 @@ class GlobalSettings:
     repeated_nextto_num_dots = 25
 
     # painting env parameters
+    painting_initial_holding_prob = 0.0
+    painting_num_objs_train = [3, 4]
```
ronuchit, line 45: note we won't be able to override this with command line args lol, oh well

tests/envs/test_painting.py
```diff
@@ -122,7 +123,8 @@ def test_painting_failure_cases():
     assert OnTable([obj0]) in atoms
     assert OnTable([obj1]) in atoms
     assert OnTable([obj2]) in atoms
-    # In the first initial state, we are holding an object
+    # In the first initial state, we are holding an object, because
+    # painting_initial_holding_prob = 1.0
```
ronuchit, line 127: lol wait how was this test here before?!

src/envs/painting.py
tomsilver: yeah they've always been "partial" and will continue to be

src/settings.py
tomsilver: good point... yeah lol oh well is all I have to say too

tests/envs/test_painting.py
tomsilver: this is just a diff on the last commit right?

ronuchit (approved):

PR #329, opened 2022-01-24
remove policy stuck check

tomsilver:
this check does not make sense for all policies; for example, policies with certain kinds of memory may be able to get out of "stuck" states.


NishanthJKumar (approved):

PR #331, opened 2022-01-24
handle more argument types for cmd line

### LINKED ISSUE #330: command line float overrides treated as strings
we should use one of the solutions here:
https://stackoverflow.com/questions/736043/checking-if-a-string-can-be-converted-to-float-in-python

and add this check here:
https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/utils.py#L1229

ronuchit:
closes #330

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1668,7 +1668,7 @@ def test_run_hill_climbing():
 
 
 def test_ops_and_specs_to_dummy_nsrts():
-    """Tests fo ops_and_specs_to_dummy_nsrts()."""
+    """Tests for ops_and_specs_to_dummy_nsrts()."""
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -1668,7 +1668,7 @@ def _local_minimum_grid_heuristic_fn(state: S) -> float:
 
 
 def test_ops_and_specs_to_dummy_nsrts():
-    """Tests fo ops_and_specs_to_dummy_nsrts()."""
+    """Tests for ops_and_specs_to_dummy_nsrts()."""
```
tomsilver, line 1671: i prefer fo

tomsilver (approved):
thank you!!

PR #336, opened 2022-01-24
add option to use gbfs in grammar search instead of hill climbing

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -1097,13 +1097,21 @@ def _get_successors(
...
+    if CFG.grammar_search_search_algorithm == "hill_climbing":
+        path, _ = utils.run_hill_climbing(
+            init,
+            _check_goal,
+            _get_successors,
+            score_function.evaluate,
+            enforced_depth=CFG.grammar_search_hill_climbing_depth,
+            parallelize=CFG.grammar_search_parallelize_hill_climbing)
+    else:
+        assert CFG.grammar_search_search_algorithm == "gbfs"
```
ronuchit, line 1109: could we actually do this as an `elif` and have an `else` with an exception raised? that would make it easier to add new `elif` clauses in the future. you'll have to cover the `else` but that should be easy with a short test:
```
    # Test approach with unrecognized search algorithm.
    utils.update_config({
        "grammar_search_search_algorithm": "not a real search algorithm",
        "grammar_search_gbfs_num_evals": 10,
    })
    with pytest.raises(Exception) as e:
        _test_approach(env_name="cover",
                       approach_name="grammar_search_invention",
                       excluded_predicates="Holding",
                       try_solving=False,
                       sampler_learner="random")
    assert "Unrecognized grammar_search_search_algorithm" in str(e.value)
```

tomsilver: done, thanks for writing the test :)

ronuchit: i knew you wouldn't do it if i didn't ;)

src/approaches/grammar_search_invention_approach.py
```diff
@@ -1097,13 +1097,24 @@ def _get_successors(
...
+            parallelize=CFG.grammar_search_parallelize_hill_climbing)
+    elif CFG.grammar_search_search_algorithm == "gbfs":
+        path, _ = utils.run_gbfs(init,
+                                 _check_goal,
+                                 _get_successors,
+                                 score_function.evaluate,
+                                 max_evals=CFG.grammar_search_gbfs_num_evals)
+    else:
+        raise NotImplementedError(
+            "Unrecognized grammar_search_search_algorithm "
```
ronuchit, line 1116: nit: add a colon for consistency?

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -21,7 +21,8 @@ def _test_approach(env_name,
     utils.update_config({
         "env": env_name,
         "approach": approach_name,
-        "seed": 12345
+        "seed": 12345,
+        "experiment_id": "",
```
ronuchit, line 25: let's leave this out of this PR and wait till we fix the utils thing?

ronuchit (approved):
approving to unblock, but as commented, we should wait

PR #337, opened 2022-01-24
make task planning a generator like skeleton generator

tomsilver:


src/planning.py
```diff
@@ -165,10 +165,10 @@ def task_plan(
...
     dummy_task = Task(State({}), goal)
     metrics: Metrics = defaultdict(float)
-    generator = _skeleton_generator(dummy_task, ground_nsrts, init_atoms,
-                                    heuristic, seed, timeout, metrics)
-    skeleton, atoms_sequence = next(generator)  # get the first one
-    return skeleton, atoms_sequence, metrics
+    for skeleton, atoms_sequence in _skeleton_generator(
+            dummy_task, ground_nsrts, init_atoms, heuristic, seed, timeout,
+            metrics):
+        yield skeleton, atoms_sequence, metrics.copy()
```
ronuchit, line 171: nice copy!

tests/test_planning.py
```diff
@@ -54,24 +57,34 @@ def test_task_plan():
...
+                                    seed=123)
+    skeleton, _, metrics = next(task_plan_generator)
     assert len(skeleton) == 2
     assert isinstance(skeleton[0], _GroundNSRT)
     assert isinstance(skeleton[1], _GroundNSRT)
+    total_num_skeletons = 1
+    with pytest.raises(ApproachFailure) as e:
+        for _, _, metrics in task_plan_generator:
+            total_num_skeletons += 1
+            assert metrics["num_skeletons_optimized"] == total_num_skeletons
```
ronuchit, line 75: this is testing that the metrics copying works as intended, right? just curious if you remove the `metrics.copy()` would this test break?

tests/test_planning.py
```diff
@@ -54,24 +57,34 @@ def test_task_plan():
...
+    assert "Planning reached max_skeletons_optimized!" in str(e.value)
```
ronuchit, line 76: i'm confused, why is this assert getting hit? is it because we don't pass a max_num into task_plan? should we be?

tests/test_planning.py
```diff
@@ -54,24 +57,34 @@ def test_task_plan():
...
+                                    reachable_atoms,
+                                    heuristic,
+                                    timeout=1,
+                                    seed=123)
+    skeleton, _, metrics = next(task_plan_generator)
     assert len(skeleton) == 2
     assert isinstance(skeleton[0], _GroundNSRT)
     assert isinstance(skeleton[1], _GroundNSRT)
+    total_num_skeletons = 1
+    with pytest.raises(ApproachFailure) as e:
```
ronuchit, line 72: just curious, do you have any idea what these 3 skeletons are?

tests/test_planning.py
tomsilver: it doesn't break, but i'll add another check for this

tests/test_planning.py
tomsilver: well the first one is what we'd expect (a length 2 optimal plan), and then the second two are length 4 plans that pick and place unnecessarily.

ronuchit (approved):
lgtm

PR #338, opened 2022-01-24
update_config() updates

ronuchit:
* don't need to pass in duplicated env/approach/seed
* experiment_id is optional

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -154,10 +154,6 @@ class GlobalSettings:
     def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
         """A workaround for global settings that are derived from the
         experiment-specific args."""
-        if "env" not in args:
-            args["env"] = ""
-        if "approach" not in args:
-            args["approach"] = ""
         return dict(
             # Task planning heuristic to use in SeSamE.
             task_planning_heuristic=defaultdict(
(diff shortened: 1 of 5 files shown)


src/utils.py
```diff
@@ -1184,15 +1184,22 @@ def update_config(args: Dict[str, Any], default_seed: int = 123) -> None:
     for k in args:
         if k not in allowed_args:
             raise ValueError(f"Unrecognized arg: {k}")
+    for k in ("env", "approach", "seed"):
+        if not args.get(k, None) and hasattr(CFG, k):
+            # For env, approach, and seed, if we don't pass in a value
```
ronuchit, line 1189: only want to do this for env/approach/seed bc for instance if we set some boolean flag to False, and then in a later test don't pass anything in, we don't want it to retain False value

src/settings.py
```diff
@@ -166,7 +162,7 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
                 {
                     # In the playroom domain, HFF works better.
                     "playroom": "hff",
-                })[args["env"]],
+                })[args.get("env", "")],
```
ronuchit, line 165: annoying that we can't get rid of this altogether, it's just here to make single unit tests work (both unit test suite and regular execution will always have env and approach set)

src/utils.py
tomsilver: isn't `experiment_id` another one of this kind?

tomsilver (approved):
good job threading this needle, one question

src/utils.py
ronuchit: fixed

PR #340, opened 2022-01-25
add inheritances from abc.ABC

ronuchit:
* this lets us remove some pylint disables, yay!
* also removed some tests that unnecessarily check that abstract methods
are abstract
* rename OptionModel to OptionModelBase for consistency


tomsilver (approved):
Yay!

PR #341, opened 2022-01-25
speed up tests ~3.3x

ronuchit:
Remove some tests which have outlived their usefulness, and generally
decrease hyperparameters for some of the slowest tests.

Before this PR, unit tests took about 200 seconds locally. After this
PR, unit tests take about 60 seconds locally.

PS: when invoking pytest, a useful parameter is `--durations=N`, which
will print out the names+times of the `N` slowest tests. `N = 20` is a
reasonable choice.


tomsilver (approved):
Love it! And good tip about pytest.

PR #342, opened 2022-01-25
expected node count grammar search score function 

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -667,6 +670,101 @@ def _evaluate_with_operators(self,
         return score
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class _ExpectedNodesScoreFunction(_OperatorLearningBasedScoreFunction):
+    """Score a predicate set by learning operators and planning in the training
+    tasks.
+
+    The score corresponds to expected number of nodes that would need to be
```
ronuchit, line 678: the expected*

analysis/grammar_search_analysis.py
```diff
@@ -105,6 +105,7 @@ def _run_proxy_analysis(env_names: List[str], score_function_names: List[str],
             all_predicates - {IsWet, IsDry},
             all_predicates - {IsClean, IsDirty},
             all_predicates - {OnTable},
+            all_predicates - {HoldingTop, HoldingSide},
```
ronuchit, line 108: we could also try ablating each individually to ensure hill climbing would work with depth=0

src/approaches/grammar_search_invention_approach.py
```diff
@@ -667,6 +670,101 @@ def _evaluate_with_operators(self,
...
+    created before a low-level plan is found. This calculation requires
+    estimating the probability that each goal-reaching skeleton is refinable.
+    To estimate this, we assume a prior on how optimal the demonstrations are,
```
ronuchit, line 681: it's sorta Bayesian! right? right?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -667,6 +670,101 @@ def _evaluate_with_operators(self,
...
+    and say that if a skeleton is found by planning that is a different length
+    than the demos, then the likelihood of the predicates/operators goes down
```
ronuchit, line 683: is one direction handled differently from the other?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -667,6 +670,101 @@ def _evaluate_with_operators(self,
...
+                    assert goal.issubset(plan_atoms_sequence[-1])
+                    # Estimate the probability that this skeleton is refinable.
+                    refinement_prob = self._get_refinement_prob(
+                        atoms_sequence, plan_atoms_sequence)
+                    # Get the number of nodes that have been created so far.
+                    num_nodes = metrics["num_nodes_created"]
+                    # This contribution to the expected number of nodes is for
+                    # the event that the current skeleton is refinable, but no
+                    # previous skeleton has been refinable.
+                    p = refinable_skeleton_not_found_prob * refinement_prob
```
ronuchit, line 736: i don't understand why we're doing this more complicated thing instead of just expected_num_nodes += refinement_prob * num_nodes (maybe we can discuss on slack)

src/approaches/grammar_search_invention_approach.py
```diff
@@ -667,6 +670,101 @@ def _evaluate_with_operators(self,
...
+                    expected_num_nodes += p * num_nodes
+                    # Update the probability that no skeleton yet is refinable.
+                    refinable_skeleton_not_found_prob *= (1 - refinement_prob)
+            except (ApproachTimeout, ApproachFailure):
+                # Note if we failed to find any skeleton, the next lines add
+                # the upper bound with refinable_skeleton_not_found_prob = 1.0,
+                # so no special action is required.
+                pass
+            # After exhausting the skeleton budget or timeout, we use this
```
ronuchit, line 745: the simpler thing i mentioned above is equivalent to a pessimistic assumption that we are screwed after exhausting the budget?

src/approaches/grammar_search_invention_approach.py
tomsilver: but of course!

src/approaches/grammar_search_invention_approach.py
tomsilver: not right now. but I think we're almost always (maybe always always) going to be in the situation where the demo len is >= the plan len.

src/approaches/grammar_search_invention_approach.py
tomsilver: yeah that's right

src/approaches/grammar_search_invention_approach.py
tomsilver: well I'm not sure how that would work. what would we add to the expected node count? what probability would we use?

ronuchit: that's true, we'd need to score based on the tuple (refinable_skeleton_not_found_prob, expected_nodes), where the latter tiebreaks for the former

src/approaches/grammar_search_invention_approach.py
```diff
@@ -724,6 +727,101 @@ def _evaluate_with_operators(self,
...
+            score += expected_num_nodes
+        return score
+
+    @staticmethod
+    def _get_refinement_prob(
+            demo_atoms_sequence: Sequence[Set[GroundAtom]],
+            plan_atoms_sequence: Sequence[Collection[GroundAtom]]) -> float:
+        """Estimate the probability that plan_atoms_sequence is refinable using
+        the demonstration demo_atoms_sequence."""
+        # Make a soft assumption that the demonstrations are optimal.
```
ronuchit, line 818: can we say ", using a geometric distribution" here?

src/settings.py
```diff
@@ -136,14 +136,14 @@ class GlobalSettings:
     grammar_search_true_pos_weight = 10
     grammar_search_false_pos_weight = 1
     grammar_search_bf_weight = 1
-    grammar_search_operator_size_weight = 1e-2
-    grammar_search_pred_complexity_weight = 1
+    grammar_search_operator_size_weight = 0.0
```
ronuchit, line 139: do we need to pull this into a separate PR? in case it screws something else up?

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -14,7 +14,8 @@
     _RelaxationHeuristicEnergyBasedScoreFunction, _TaskPlanningScoreFunction,
     _ExactHeuristicEnergyBasedScoreFunction,
     _RelaxationHeuristicCountBasedScoreFunction,
-    _ExactHeuristicCountBasedScoreFunction, _BranchingFactorScoreFunction)
```
ronuchit, line 17: can you make sure that your tests aren't too slow? local should take ~60 seconds total and you can also just try `pytest -s tests/approaches/test_grammar_search_invention_approach.py --durations=0` to see the full speed printout for this file

ronuchit (approved):
lgtm just minor stuff

tests/approaches/test_grammar_search_invention_approach.py
tomsilver: looks okay, here's the printout:
```
================================================================== slowest durations ==================================================================
11.09s call     tests/approaches/test_grammar_search_invention_approach.py::test_relaxation_energy_score_function
9.59s call     tests/approaches/test_grammar_search_invention_approach.py::test_prediction_error_score_function
5.22s call     tests/approaches/test_grammar_search_invention_approach.py::test_hadd_match_score_function
5.14s call     tests/approaches/test_grammar_search_invention_approach.py::test_exact_energy_score_function
2.66s call     tests/approaches/test_grammar_search_invention_approach.py::test_expected_nodes_score_function
2.03s call     tests/approaches/test_grammar_search_invention_approach.py::test_task_planning_score_function
1.10s call     tests/approaches/test_grammar_search_invention_approach.py::test_count_score_functions
0.82s call     tests/approaches/test_grammar_search_invention_approach.py::test_branching_factor_score_function
0.06s call     tests/approaches/test_grammar_search_invention_approach.py::test_predicate_grammar
0.02s call     tests/approaches/test_grammar_search_invention_approach.py::test_predicate_search_heuristic_base_classes
```

src/settings.py
tomsilver: see https://github.com/Learning-and-Intelligent-Systems/predicators/pull/351

PR #344, opened 2022-01-25
Saving videos of not-fully-solved plans 

ashayathalye:
Adds a flag that enables saving videos of plans that did not solve the goal to help debug planning failures. For each NSRT skeleton, it saves the refined plan that made the most progress towards the goal as measured by number of options in the plan (not number of goal conditions fulfilled– maybe we should change it to this.). 

Still need to write tests.


tomsilver:
moved this code to #371

PR #345, opened 2022-01-25
some updates to painting and grammar search settings

ronuchit:


tomsilver (approved):
picasso-level painting

PR #351, opened 2022-01-25
weight settings changes for grammar search

tomsilver:


tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -713,6 +713,7 @@ def test_task_planning_score_function():
         "env": "cover",
         "offline_data_method": "demo+replay",
         "seed": 0,
+        "num_train_tasks": 15,
```
ronuchit, line 716: lol is this gonna be a problem in a lot of places if we switch to 50

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -604,7 +604,7 @@ def test_exact_energy_score_function():
     utils.update_config({"task_planning_heuristic": old_heur})
     old_hbmd = CFG.grammar_search_heuristic_based_max_demos
     utils.update_config({"grammar_search_heuristic_based_max_demos": 0})
-    assert abs(score_function.evaluate(set()) - 0.39) < 0.11  # only op penalty
+    score_function.evaluate(set())
```
ronuchit, line 607: won't this score just be 0.0 now?

src/settings.py
```diff
@@ -136,8 +136,8 @@ class GlobalSettings:
     grammar_search_true_pos_weight = 10
     grammar_search_false_pos_weight = 1
     grammar_search_bf_weight = 1
-    grammar_search_operator_size_weight = 1e-2
-    grammar_search_pred_complexity_weight = 1
+    grammar_search_operator_size_weight = 0.0
+    grammar_search_pred_complexity_weight = 1e-2
```
ronuchit, line 140: i thought we were gonna do 1e-4, why change back?

ronuchit (commented):
questions/suggestions

tests/approaches/test_grammar_search_invention_approach.py
tomsilver: possibly, we'll see 😬

src/settings.py
tomsilver: oops thank you

tests/approaches/test_grammar_search_invention_approach.py
tomsilver: yup thanks

ronuchit (approved):
lgtm

PR #356, opened 2022-01-26
changes to num_train_tasks, data loading, and score function settings

tomsilver:


analysis/run_supercloud_experiments.sh
```diff
@@ -5,7 +5,7 @@ NUM_SEEDS=10
 FILE="analysis/submit.py"
 
 for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
-    COMMON_ARGS="--remake_data --seed $SEED"
+    COMMON_ARGS="--seed $SEED"
```
ronuchit, line 8: nice

src/settings.py
```diff
@@ -14,6 +14,7 @@
 class GlobalSettings:
     """Unchanging settings."""
     # parameters for all envs
+    num_train_tasks = 50
```
ronuchit, line 17: do we need to make an announcement about this? it affects amber/ashay

ronuchit (approved):
thank you 🙏

src/settings.py
tomsilver: yeah

PR #358, opened 2022-01-26
Behavior initial state sampling

### LINKED ISSUE #243: BEHAVIOR: Incorporate an `rng` into the placeOnTop Sampler
The `placeOnTop` sampler internally doesn't use an rng, which means we can't control it's randomness. Fix this.

NishanthJKumar:
Enables us to sample new initial states for scenes in a seeded fashion so that we can run our pipeline on multiple instances of a task!
Also seeds the sampler for the placeOnTop option and thereby resolves  #243


src/envs/behavior.py
```diff
@@ -164,12 +167,20 @@ def _get_tasks(self, num: int, rng: np.random.Generator) -> List[Task]:
         for _ in range(num):
             # Behavior uses np.random everywhere. This is a somewhat
             # hacky workaround for that.
-            np.random.seed(rng.integers(0, (2**32) - 1))
-            self.igibson_behavior_env.reset()
+            curr_env_seed = rng.integers(0, (2**32) - 1)
+            if CFG.behavior_randomize_init_state:
+                self.set_igibson_behavior_env(curr_env_seed)
+                self.igibson_behavior_env.reset()
```
ronuchit, line 173: Can you pull this line out of the `if` and delete the else?

src/envs/behavior.py
```diff
@@ -301,6 +312,44 @@ def render(self,
...
+                    physics_timestep=CFG.behavior_physics_timestep,
+                    action_filter="mobile_manipulation",
+                    rng=self._rng,
+                )
+                self.igibson_behavior_env.step(
+                    np.zeros(self.igibson_behavior_env.action_space.shape))
+                ig_objs_bddl_scope = [
+                    self._ig_object_name(obj)
+                    for obj in self._get_task_relevant_objects()
+                ]
```
ronuchit, line 351: No need to duplicate the code in this loop with the code above. A better pattern is:
```
while True:
    self.igibson_behavior_env = blah
    self.igibson_behavior_env.step()
    ig_objs_bddl_scope = [blah]
    if None not in ig_objs_bddl_scope:
        break
```

ronuchit: One concern is, how likely is it that this loop might run forever? Should we impose a max number of iterations before killing it or something?

ronuchit (commented):
aside from my comments, rest lgtm, but let @tomsilver look too

src/envs/behavior.py
NishanthJKumar: from our tests, there's a very low probability this will run forever, but I'll add a counter just in case!

tomsilver (approved):
great job!

PR #362, opened 2022-01-27
add option for suspicious multistep effects score in expected_nodes score function

ronuchit:
It's a bit hard to test this right now, but I was able to see some improvement from it. Specifically, what I did was run the analysis script on Cover with the following change:
`expected_planning_time += p * num_nodes` -> `expected_planning_time += p`
I made this change because in our current envs, the number of nodes is helping the score function a lot, making suspicious scores unnecessary.

Results without suspicious scoring:
```
Evaluating predicates: frozenset(), with total cost 0
	Total score: 534026.0908040795 computed in 0.072 seconds
Evaluating predicates: frozenset({HandEmpty}), with total cost 1.0
	Total score: 534026.0906951863 computed in 0.114 seconds
Evaluating predicates: frozenset({Holding}), with total cost 1.0
	Total score: 13071.029276505998 computed in 0.139 seconds
*** Evaluating predicates: frozenset({HandEmpty, Holding}), with total cost 2.0
	Total score: 82.17972948588978 computed in 0.160 seconds
Evaluating predicates: frozenset({HandEmpty, Clear, Holding}), with total cost 3.0
	Total score: 82.17985818755962 computed in 0.135 seconds ***
Evaluating predicates: frozenset({NOT-HandEmpty}), with total cost 1.0
	Total score: 13071.159739185763 computed in 0.127 seconds
Evaluating predicates: frozenset({NOT-HandEmpty, HandEmpty}), with total cost 2.0
	Total score: 13071.159947090002 computed in 0.139 seconds
```
In particular, note that there is no improvement from adding in the Clear predicate to {HandEmpty, Holding}, because you get plans of the same length as the demos even without Clear (you just might stupidly place what you're holding onto a target that needs to be kept free for later).

Results with suspicious scoring:
```
Evaluating predicates: frozenset(), with total cost 0
	Total score: 2316024.9189511305 computed in 0.089 seconds
Evaluating predicates: frozenset({HandEmpty}), with total cost 1.0
	Total score: 2909930.812771448 computed in 0.152 seconds
Evaluating predicates: frozenset({Holding}), with total cost 1.0
	Total score: 722045.9403480866 computed in 0.218 seconds
*** Evaluating predicates: frozenset({HandEmpty, Holding}), with total cost 2.0
	Total score: 6094.929471400891 computed in 0.289 seconds
Evaluating predicates: frozenset({HandEmpty, Clear, Holding}), with total cost 3.0
	Total score: 95.049599498305 computed in 0.269 seconds ***
Evaluating predicates: frozenset({NOT-HandEmpty}), with total cost 1.0
	Total score: 13084.029559696703 computed in 0.168 seconds
Evaluating predicates: frozenset({NOT-HandEmpty, HandEmpty}), with total cost 2.0
	Total score: 13084.029559700068 computed in 0.218 seconds
```
Here, we can see that almost nothing looks suspicious (for the first refinable skeleton at least) under the optimal predicate set, but transitions that make two blocks be on the same target look suspicious under the predicate set that's missing Clear.

### DIFF
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -8,7 +8,7 @@ import os
 from typing import Dict, DefaultDict, Set, List, Tuple, Sequence
 import pandas as pd
 from predicators.src.datasets import create_dataset
-from predicators.src.envs import create_env, BaseEnv
+from predicators.src.envs import create_env, BaseEnv, CoverEnv
 from predicators.src.approaches import create_approach
 from predicators.src.approaches.grammar_search_invention_approach import \
     _create_score_function, _ForallClassifier, _SingleAttributeCompareClassifier
@@ -34,14 +34,7 @@ def _run_proxy_analysis(env_names: List[str], score_function_names: List[str],
             for block in state:
                 if block.type.name != "block":
                     continue
-                block_pose = state.get(block, "pose")
-                block_width = state.get(block, "width")
-                target_pose = state.get(target, "pose")
-                target_width = state.get(target, "width")
-                if (block_pose-block_width/2 <=
-                    target_pose-target_width/2) and \
-                    (block_pose+block_width/2 >=
-                     target_pose+target_width/2):
+                if CoverEnv._Covers_holds(state, [block, target]):  # pylint:disable=protected-access
                     return False
             return True
 
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 from operator import le
 import glob
 import os
-from typing import Dict, DefaultDict, Set, List, Tuple
+from typing import Dict, DefaultDict, Set, List, Tuple, Sequence
 import pandas as pd
 from predicators.src.datasets import create_dataset
 from predicators.src.envs import create_env, BaseEnv
@@ -15,7 +15,7 @@ from predicators.src.approaches.grammar_search_invention_approach import \
 from predicators.src.ground_truth_nsrts import _get_predicates_by_names
 from predicators.src.main import _run_testing
 from predicators.src import utils
-from predicators.src.structs import Predicate, Dataset
+from predicators.src.structs import Predicate, Dataset, State, Object
 from predicators.src.settings import CFG
 
 
@@ -29,7 +29,7 @@ def _run_proxy_analysis(env_names: List[str], score_function_names: List[str],
         targ_type = Covers.types[1]
         NotHandEmpty = HandEmpty.get_negation()
 
-        def Clear_classifier(state, objects):
+        def _Clear_holds(state: State, objects: Sequence[Object]) -> bool:
             target, = objects
             for block in state:
                 if block.type.name != "block":
@@ -45,7 +45,7 @@ def _run_proxy_analysis(env_names: List[str], score_function_names: List[str],
                     return False
             return True
 
-        Clear = Predicate("Clear", [targ_type], Clear_classifier)
+        Clear = Predicate("Clear", [targ_type], _Clear_holds)
         covers_pred_sets: List[Set[Predicate]] = [
             set(),
             {HandEmpty},
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -24,14 +24,34 @@ def _run_proxy_analysis(env_names: List[str], score_function_names: List[str],
     utils.update_config({"seed": 0})
     if "cover" in env_names:
         env_name = "cover"
-        HandEmpty, Holding = _get_predicates_by_names(env_name,
-                                                      ["HandEmpty", "Holding"])
+        HandEmpty, Holding, Covers = _get_predicates_by_names(
+            env_name, ["HandEmpty", "Holding", "Covers"])
+        targ_type = Covers.types[1]
         NotHandEmpty = HandEmpty.get_negation()
+
+        def Clear_classifier(state, objects):
+            target, = objects
+            for block in state:
+                if block.type.name != "block":
+                    continue
+                block_pose = state.get(block, "pose")
+                block_width = state.get(block, "width")
+                target_pose = state.get(target, "pose")
+                target_width = state.get(target, "width")
+                if (block_pose-block_width/2 <=
+                    target_pose-target_width/2) and \
+                    (block_pose+block_width/2 >=
+                     target_pose+target_width/2):
+                    return False
+            return True
+
+        Clear = Predicate("Clear", [targ_type], Clear_classifier)
         covers_pred_sets: List[Set[Predicate]] = [
             set(),
             {HandEmpty},
             {Holding},
             {HandEmpty, Holding},
+            {HandEmpty, Holding, Clear},
             {NotHandEmpty},
             {NotHandEmpty, HandEmpty},
         ]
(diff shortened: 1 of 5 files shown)


analysis/grammar_search_analysis.py
```diff
@@ -24,14 +24,34 @@ def _run_proxy_analysis(env_names: List[str], score_function_names: List[str],
...
     if "cover" in env_names:
         env_name = "cover"
-        HandEmpty, Holding = _get_predicates_by_names(env_name,
-                                                      ["HandEmpty", "Holding"])
+        HandEmpty, Holding, Covers = _get_predicates_by_names(
+            env_name, ["HandEmpty", "Holding", "Covers"])
+        targ_type = Covers.types[1]
         NotHandEmpty = HandEmpty.get_negation()
+
+        def _Clear_holds(state: State, objects: Sequence[Object]) -> bool:
```
tomsilver, line 32: any reason to implement this this way instead of using Covers? i'm guessing it's equivalent to the feature that will be learned by the grammar, but good to be sure

src/approaches/grammar_search_invention_approach.py
```diff
@@ -820,14 +827,56 @@ def _evaluate_with_operators(self,
...
+            for i in range(len(plan_atoms_sequence) - 1):
+                for j in range(i + 1, len(plan_atoms_sequence)):
+                    atoms_i = plan_atoms_sequence[i]
+                    atoms_j = plan_atoms_sequence[j]
+                    for plan_eff in [
+                            frozenset(atoms_i - atoms_j),
+                            frozenset(atoms_j - atoms_i)
+                    ]:
+                        if not any(
+                                utils.unify(demo_eff, plan_eff)[0]
```
tomsilver, line 852: how much slower does the overall score function get when using suspicious transitions?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -820,14 +827,56 @@ def _evaluate_with_operators(self,
     @staticmethod
     def _get_refinement_prob(
             demo_atoms_sequence: Sequence[Set[GroundAtom]],
-            plan_atoms_sequence: Sequence[Collection[GroundAtom]]) -> float:
+            plan_atoms_sequence: Sequence[Set[GroundAtom]],
+            demo_multistep_effects: Set[FrozenSet[GroundAtom]]) -> float:
         """Estimate the probability that plan_atoms_sequence is refinable using
         the demonstration demo_atoms_sequence."""
-        # Make a soft assumption that the demonstrations are optimal.
+        # Make a soft assumption that the demonstrations are optimal,
```
tomsilver, line 834: oops sorry I thought that I added this comment

src/approaches/grammar_search_invention_approach.py
```diff
@@ -820,14 +827,56 @@ def _evaluate_with_operators(self,
...
+        # using a geometric distribution.
         demo_len = len(demo_atoms_sequence)
         plan_len = len(plan_atoms_sequence)
+        # The exponent is the difference in plan lengths.
+        exponent = abs(demo_len - plan_len)
+        if CFG.grammar_search_expected_nodes_include_suspicious_score:
+            # Handle suspicious effect scoring via unification.
+            num_suspicious_eff = 0
+            for i in range(len(plan_atoms_sequence) - 1):
+                for j in range(i + 1, len(plan_atoms_sequence)):
```
tomsilver, line 844: interesting, this is a bit different from what I had in mind. I was thinking that we would specify a hyperparameter for what interval we want to use for the effects. for example, we'll just look at two-step effects.

this brings up a few questions:
1. how much slower is this way? one concern is that the effects could be quite large as you go further into the sequence, so unification could get slow.
2. are there examples that we know about where >2 step effects will help?
3. if we stick with this way, should we still compare 2-step demo effects with 2-step plan effects, then 3-step demo effects with 3-step plan effects, and so on? maybe not? i'm not sure.
4. should we skip 1-step effects because they should never be suspicious (because of our operator learning) except for min_data_for_nsrts>0?

src/approaches/grammar_search_invention_approach.py
```diff
@@ -820,14 +827,56 @@ def _evaluate_with_operators(self,
...
+                    for plan_eff in [
+                            frozenset(atoms_i - atoms_j),
+                            frozenset(atoms_j - atoms_i)
+                    ]:
+                        if not any(
+                                utils.unify(demo_eff, plan_eff)[0]
+                                for demo_eff in demo_multistep_effects):
+                            num_suspicious_eff += 1
+            p = CFG.grammar_search_expected_nodes_optimal_demo_prob
+            # Add the number of suspicious effects to the exponent.
```
tomsilver, line 856: hmm interesting... i thought this would enter in through a different probability. but this way is probably fine. at some point we should actually write down the equations that we'd use to explain this.

src/approaches/grammar_search_invention_approach.py
```diff
@@ -820,14 +827,56 @@ def _evaluate_with_operators(self,
...
+        if CFG.grammar_search_expected_nodes_include_suspicious_score:
+            # Handle suspicious effect scoring via unification.
+            num_suspicious_eff = 0
+            for i in range(len(plan_atoms_sequence) - 1):
+                for j in range(i + 1, len(plan_atoms_sequence)):
+                    atoms_i = plan_atoms_sequence[i]
+                    atoms_j = plan_atoms_sequence[j]
+                    for plan_eff in [
+                            frozenset(atoms_i - atoms_j),
+                            frozenset(atoms_j - atoms_i)
```
tomsilver, line 849: hmm it seems wrong to me to treat add and delete effects separately like this. I think we should do it like we do in `unify_preconds_effects_options` -- wrap in ADD- and DEL- and unify them both together.

tomsilver (commented):
thanks! some questions

analysis/grammar_search_analysis.py
ronuchit: i wanted to avoid the pylint issue of `CoverEnv._Covers_holds` but on second thought, it's cleaner to do it that way, so will fix

src/approaches/grammar_search_invention_approach.py
ronuchit: Speed:
without suspicious score:
* cover: 2.040 sec
* blocks: 21.633 sec
* painting: 25.072 sec

with suspicious score:
* cover: 2.341 sec
* blocks: 31.537 sec
* painting: 85.906 sec

So yeah as expected it's a lot slower when the demos are longer, but there's lots of things we can do to speed it up like only doing some multistep effects.

tomsilver: thanks for checking that. what are those times exactly? hopefully not a single evaluation...

src/approaches/grammar_search_invention_approach.py
ronuchit: There's a lot of variants we can try. We just don't have the ability to test anything meaningfully so I just wanted to put this code in, but when we actually have a domain where these suspicious transitions matter, we can definitely try these variants.
1. See below comment for general speed, but yeah it would be a lot faster if you just do the hyperparameter because your loops would be O(n) not O(n^2) which matters a lot for painting where the demos are long
2. We don't even have examples where any of this helps right now, so TBD
3. Same, no idea, it comes down to experimentation
4. No, we shouldn't skip them because of min_data_for_nsrts, but I did check that nothing gets counted as suspicious if you limit to 1-step effects, likely bc min_data_for_nsrts isn't filtering anything out right now

src/approaches/grammar_search_invention_approach.py
ronuchit: the analysis pipeline which evaluates 7 pred sets for cover, 10 for blocks, 12 for painting

ronuchit: so it's probably unusably slow in the current form for painting, but that's okay...

src/approaches/grammar_search_invention_approach.py
ronuchit: fixed

ronuchit:
@tomsilver ready for another look

tomsilver (approved):
thanks lgtm

PR #369, opened 2022-01-27
purge self._dataset

ronuchit:
The whole self._dataset thing was put in by me to handle the train_tasks_generator stuff. It's not useful anymore since now, learn_from_offline_dataset is called exactly once. This should have been done as part of #366, but I didn't think of it then.


amburger66 (approved):
looks functionally the same to me

tomsilver (approved):
if @amburger66 's happy i'm happy!

PR #370, opened 2022-01-27
basic API

ronuchit:


src/approaches/base_approach.py
```diff
@@ -64,16 +66,27 @@ def seed(self, seed: int) -> None:
...
-        Note: this is not an abc.abstractmethod because it does
-        not need to be defined by the subclasses. (mypy complains
-        if you try to instantiate a subclass with an undefined abc).
         """
 
     def load(self) -> None:
         """Load anything from CFG.get_approach_save_path_str().
 
-        Only called if self.is_learning_based.
+        Only called if self.is_learning_based, BEFORE any online learning.
```
tomsilver, line 74: what do you mean by this exactly?

src/interaction/agent_env_interaction.py
```diff
@@ -0,0 +1,26 @@
...
+@dataclass(frozen=True, eq=False, repr=False)
+class InteractionRequest:
+    """A request for interacting with a training task. Contains an index
+    for that training task, an acting policy, a query policy, and a
+    termination function.
+    """
+    train_task_idx: int
+    act_policy: Callable[[State], Action]
+    query_policy: Callable[[State], Optional[Query]]  # query can be None
+    termination_function: Callable[[State], bool]
```
tomsilver, line 21: this is good

src/interaction/agent_env_interaction.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+# A response to an InteractionRequest. This response is a sequence
+# of (state, action, response to query if provided, next_state) tuples.
+InteractionResponse = List[Tuple[State, Action, Optional[QueryResponse], State]]
```
tomsilver, line 26: hmm. I know I proposed basically this, but looking at it now, I think it should be allowed to just query about the initial state without taking any actions... so maybe instead, we should mirror the pattern we have elsewhere (e.g. LowLevelTrajectory) and have separate lists?

src/interaction/teacher.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Teacher:
+    """The teacher can respond to queries of various types.
+    """
+    def __init__(self) -> None:
+        pass  # TODO: instantiate an oracle approach
+
+    def ask(self, state: State, query: Query) -> QueryResponse:
+        """The key method that a teacher defines.
```
tomsilver, line 18: *the* teacher?

src/interaction/agent_env_interaction.py
```diff
@@ -0,0 +1,26 @@
+"""Contains classes for handling the interaction between an agent and
```
tomsilver, line 1: is this all this file is ever going to be? if so, move to structs?

src/main.py
```diff
@@ -93,44 +97,101 @@ def main() -> None:
                 "Can't exclude a goal predicate!"
     else:
         preds = env.predicates
+    # Create the agent (approach).
     approach = create_approach(CFG.approach, preds, env.options, env.types,
                                env.action_space)
-    # If approach is learning-based, get training datasets and do learning,
-    # testing after each learning call. Otherwise, just do testing.
+    # If agent is learning-based, generate offline dataset, allow the agent
```
tomsilver, line 103: "an" offline dataset

src/main.py
```diff
@@ -93,44 +97,101 @@ def main() -> None:
...
+    # to learn from it, and then proceed with the online learning loop. Test
+    # after each learning call. If agent is not learning-based, just test once.
     if approach.is_learning_based:
-        if CFG.load_approach:
+        train_tasks = env.get_train_tasks()
+        dataset, total_num_transitions = _generate_or_load_offline_dataset(
```
tomsilver, line 108: is `total_num_transitions` not something that we can just easily get from `dataset`?

src/main.py
```diff
@@ -93,44 +97,101 @@ def main() -> None:
...
+            env, train_tasks)
+        learning_start = time.time()
+        if CFG.load_approach:  # we only save/load for initial offline learning
```
tomsilver, line 111: why?

src/main.py
```diff
@@ -93,44 +97,101 @@ def main() -> None:
...
-                    pkl.dump(dataset, f)
-            learning_start = time.time()
             approach.learn_from_offline_dataset(dataset)
-            learning_time = time.time() - learning_start
+        teacher = Teacher()
+        for _ in range(CFG.num_online_learning_cycles):
             results = _run_testing(env, approach)
-            _save_test_results(results, learning_time=learning_time)
+            _save_test_results(results, num_transitions=total_num_transitions,
+                               learning_time=(time.time()-learning_start))
```
tomsilver, line 119: num transitions and learning time are probably just two of several metrics that we'll want to save. should we get ahead of that and make a Metrics thing here?

src/interaction/agent_env_interaction.py
tomsilver: I kinda wanna call this `Episode`... one reason I don't like `InteractionResponse` is because it sounds like it's just talking about interacting with a teacher, when in reality it's environment + teacher. I would also be okay with `InteractionResult`.

src/main.py
```diff
@@ -93,44 +97,101 @@ def main() -> None:
...
+    interaction_responses = []
+    num_transitions = 0
+    for request in requests:
+        # First, roll out the acting policy.
+        task = train_tasks[request.train_task_idx]
+        traj = utils.run_policy_until(
+            request.act_policy, simulator, task.init,
+            request.termination_function,
+            max_num_steps=CFG.max_num_steps_interaction_request)
+        # Now, go through the trajectory and handle queries
```
tomsilver, line 175: can you add a note "apologizing" for the two passes through the states (because of the disconnect with reality)?

src/main.py
```diff
@@ -93,44 +97,101 @@ def main() -> None:
...
+        # while assembling response objects.
+        interaction_response: InteractionResponse = []
+        for i in range(len(traj.states)-1):
+            num_transitions += 1
+            state = traj.states[i]
+            action = traj.actions[i]
```
tomsilver, line 181: after my suggested refactor of InteractionResponse, you won't need actions here

tomsilver (commented):
looking good!

src/approaches/base_approach.py
ronuchit: right now only load_from_offline_dataset() will do any saving, and then load() will load the result of that, before starting the learning loop. i'll clarify the comment. later on, we can fix this so that saving/loading can happen online too

ronuchit: updated the doc

src/interaction/agent_env_interaction.py
ronuchit: I'll change to separate lists and rename to InteractionResult.

src/approaches/base_approach.py
tomsilver: okay sounds good. can we make an issue so we don't forget? :)

src/interaction/agent_env_interaction.py
ronuchit: i thought it would be fancier when i started.... done

src/main.py
ronuchit: thanks Rohan

src/main.py
ronuchit: lazy; todo is in doc

src/main.py
ronuchit: it's not bc of any disconnect from reality, it's just bc i wanted to use `utils.run_policy_until` without modifying it lol. you could always make a new function that steps the policy through and does the query handling at the same time.

src/approaches/base_approach.py
ronuchit: okay once we merge this PR

ronuchit:
@tomsilver ready for another look

src/structs.py
```diff
@@ -1020,6 +1022,36 @@ def __str__(self) -> str:
         return repr(self)
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class InteractionRequest:
+    """A request for interacting with a training task during online learning.
+    Contains the index for that training task, an acting policy, a query policy,
+    and a termination function.
+
+    Note: the act_policy will not be called on the state where the
```
ronuchit, line 1031: @tomsilver a bit strange but the only thing i could think of that makes sense, esp. since we'll want to allow queries about just the initial state without any actions.

tomsilver: it doesn't seem that strange to me

tests/test_structs.py
```diff
@@ -739,7 +740,7 @@ def test_segment():
 
 
 def test_pnad():
-    """Tests for PartialNSRTAndDatastoreclass."""
+    """Tests for PartialNSRTAndDatastore class."""
```
tomsilver, line 743: lol

tomsilver (approved):
yay!!

PR #371, opened 2022-01-28
Make failure videos

tomsilver:
make videos of failed planning attempts, for debugging etc

this is heavily heavily based on @ashayathalye 's code (#344 ) with changes mostly to get checks passing

### DIFF
--- a/src/approaches/base_approach.py
+++ b/src/approaches/base_approach.py
@@ -8,6 +8,7 @@ from gym.spaces import Box
 from predicators.src.structs import State, Task, Predicate, Type, \
     ParameterizedOption, Action, Dataset, Metrics
 from predicators.src.settings import CFG
+from predicators.src.utils import ExceptionWithInfo
 
 
 class BaseApproach(abc.ABC):
@@ -87,9 +88,9 @@ class BaseApproach(abc.ABC):
         self._metrics = defaultdict(float)
 
 
-class ApproachTimeout(Exception):
+class ApproachTimeout(ExceptionWithInfo):
     """Exception raised when approach.solve() times out."""
 
 
-class ApproachFailure(Exception):
+class ApproachFailure(ExceptionWithInfo):
     """Exception raised when approach.solve() fails to compute a policy."""
--- a/src/main.py
+++ b/src/main.py
@@ -138,6 +138,7 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
     approach.reset_metrics()
     total_suc_time = 0.0
     total_num_execution_failures = 0
+    video_prefix = utils.get_config_path_str()
     for i, task in enumerate(test_tasks):
         start = time.time()
         print(end="", flush=True)
@@ -146,6 +147,12 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
         except (ApproachTimeout, ApproachFailure) as e:
             print(f"Task {i+1} / {len(test_tasks)}: Approach failed to "
                   f"solve with error: {e}")
+            if CFG.make_failure_videos and e.info.get("partial_refinements"):
+                video = utils.create_video_from_partial_refinements(
+                    task, env.simulate, env.render,
+                    e.info["partial_refinements"])
+                outfile = f"{video_prefix}__task{i+1}_failure.mp4"
+                utils.save_video(outfile, video)
             continue
         num_found_policy += 1
         try:
@@ -168,7 +175,7 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
         else:
             print(f"Task {i+1} / {len(test_tasks)}: Policy failed")
         if CFG.make_videos:
-            outfile = f"{utils.get_config_path_str()}__task{i}.mp4"
+            outfile = f"{video_prefix}__task{i+1}.mp4"
             utils.save_video(outfile, video)
     metrics: Metrics = defaultdict(float)
     metrics["num_solved"] = num_solved
--- a/src/main.py
+++ b/src/main.py
@@ -39,7 +39,7 @@ import subprocess
 import time
 import dill as pkl
 from predicators.src.settings import CFG
-from predicators.src.envs import create_env, EnvironmentFailure, BaseEnv
+from predicators.src.envs import create_env, BaseEnv
 from predicators.src.approaches import create_approach, ApproachTimeout, \
     ApproachFailure, BaseApproach
 from predicators.src.datasets import create_dataset
@@ -222,7 +222,7 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
             _, video, solved = utils.run_policy_on_task(
                 policy, task, env.simulate, CFG.max_num_steps_check_policy,
                 env.render if CFG.make_videos else None)
-        except EnvironmentFailure as e:
+        except utils.EnvironmentFailure as e:
             print(f"Task {i+1} / {len(test_tasks)}: Environment failed "
                   f"with error: {e}")
             continue
--- a/src/planning.py
+++ b/src/planning.py
@@ -16,8 +16,8 @@ from predicators.src.structs import State, Task, NSRT, Predicate, \
     GroundAtom, _GroundNSRT, DummyOption, DefaultState, _Option, \
     Metrics, STRIPSOperator, OptionSpec, Object
 from predicators.src import utils
-from predicators.src.utils import _TaskPlanningHeuristic, ExceptionWithInfo
-from predicators.src.envs import EnvironmentFailure
+from predicators.src.utils import _TaskPlanningHeuristic, ExceptionWithInfo, \
+    EnvironmentFailure
 from predicators.src.option_model import _OptionModelBase
 from predicators.src.settings import CFG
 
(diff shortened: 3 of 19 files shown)


ashayathalye (commented):
Looks good!

src/approaches/base_approach.py
```diff
@@ -9,6 +9,7 @@
     ParameterizedOption, Action, Dataset, Metrics, InteractionRequest, \
```
ronuchit, line 9: Could you add an example command line and attach the resulting video to this PR?

src/approaches/base_approach.py
```diff
@@ -9,6 +9,7 @@
...
     InteractionResult
 from predicators.src.settings import CFG
+from predicators.src.utils import ExceptionWithInfo
```
ronuchit, line 12: Let's make `EnvironmentFailure` in base_env.py also inherit from ExceptionWithInfo, with dictionary key "offending_objects"?

src/main.py
```diff
@@ -209,6 +210,12 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
         except (ApproachTimeout, ApproachFailure) as e:
             print(f"Task {i+1} / {len(test_tasks)}: Approach failed to "
                   f"solve with error: {e}")
+            if CFG.make_failure_videos and e.info.get("partial_refinements"):
```
ronuchit, line 213: Is the idea of this `get` here that "partial_refinements" might not be in the info dict sometimes? When would this happen?

src/planning.py
```diff
@@ -265,15 +287,18 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
                 failure = _DiscoveredFailure(e, nsrt)
                 # Remember only the most recent failure.
                 discovered_failures[cur_idx] = failure
                 # If we're immediately propagating failures, raise it now.
                 if CFG.sesame_propagate_failures == "immediately":
-                    raise _DiscoveredFailureException("Discovered a failure",
-                                                      failure)
+                    raise _DiscoveredFailureException(
+                        "Discovered a failure", failure,
+                        {"longest_refinement": longest_refinement})
```
ronuchit, line 301: Confused by the logic here a bit. We'll get into this `except EnvironmentFailure as e:` clause if the call to `option_model.get_next_state(state, option)` above raises that EnvironmentFailure. In that case, `longest_refinement = list(plan[:cur_idx + 1])` won't be run, right?

src/planning.py
```diff
@@ -224,16 +238,23 @@ def _skeleton_generator(
...
-                          skeleton: List[_GroundNSRT],
-                          atoms_sequence: List[Set[GroundAtom]], seed: int,
-                          timeout: float) -> Optional[List[_Option]]:
-    """Backtracking search over continuous values."""
+def _run_low_level_search(
+        task: Task, option_model: _OptionModelBase,
+        skeleton: List[_GroundNSRT], atoms_sequence: List[Set[GroundAtom]],
+        seed: int,
+        timeout: float) -> Tuple[Optional[List[_Option]], List[_Option]]:
+    """Backtracking search over continuous values.
```
ronuchit, line 251: Wouldn't it make more sense to collapse these return values into one, and just have the caller check whether the length of the returned plan equals the length of the skeleton? (If so, that's a success.)

src/utils.py
```diff
@@ -1152,6 +1163,35 @@ def create_pddl_problem(objects: Collection[Object],
...
 
+def create_video_from_partial_refinements(
+    task: Task, simulator: Callable[[State, Action], State],
+    render: Callable[[State, Task, Optional[Action]], List[Image]],
+    partial_refinements: Sequence[Tuple[Sequence[_GroundNSRT],
+                                        Sequence[_Option]]]
+) -> Video:
+    """Create a video from a list of skeletons and partial refinements."""
+    # Right now, the video is created by finding the longest partial refinement.
+    # We imagine one day also implementing an "all_skeletons" mode that would
```
ronuchit, line 1174: Let's not make a promise like this in the comments...unclear if we'll ever actually do this :P Instead, maybe "One could also implement..."?

tests/test_main.py
```diff
@@ -63,6 +63,14 @@ def test_main():
         "--results_dir", results_dir
     ]
     main()
```
ronuchit, line 65: how slow are these two tests you added?

ronuchit:
Command line:
python src/main.py --env painting --approach oracle --seed 123 --num_test_tasks 1 --video_dir videos/ --results_dir results/ --max_skeletons_optimized 1 --painting_lid_open_prob 0.0 --make_failure_videos

Video:

https://user-images.githubusercontent.com/3521150/151569954-9600882f-440a-4418-8932-69f1bb381b8c.mp4

src/approaches/base_approach.py
ronuchit: done

src/utils.py
ronuchit: done

tests/test_main.py
ronuchit: A: very fast lol

src/planning.py
ronuchit: i'm pretty sure this was a logic error, addressed this by making a bunch of changes

src/planning.py
ronuchit: done

src/main.py
ronuchit: @tomsilver need your help understanding this one

ronuchit:
@tomsilver ready for you to look

src/planning.py
```diff
@@ -272,29 +295,33 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
-                # predicates).
+                # This "if all" statement is equivalent to, but faster than,
+                # checking whether expected_atoms is a subset of
+                # utils.abstract(traj[cur_idx], predicates).
                 if all(atom.holds(traj[cur_idx]) for atom in expected_atoms):
                     can_continue_on = True
-                    if cur_idx == len(skeleton):  # success!
-                        result = plan
-                        return result
+                    # Update the longest_refinement found so far.
```
ronuchit, line 319: This logic was up higher earlier, which I think was wrong. We only want to update the longest_refinement after you pass the expected atoms check at any step.

src/planning.py
```diff
@@ -272,29 +295,33 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
+                    if cur_idx > len(longest_refinement):
+                        longest_refinement = list(options[:cur_idx])
+                    # If cur_idx is the length of the skeleton, success!
+                    if cur_idx == len(skeleton):
+                        return longest_refinement
```
ronuchit, line 324: note that at this point (and only this point), `longest_refinement` will be the same as `options`

src/main.py
tomsilver: it could happen if we time out during grounding, or just never get to the point where we start refining, I think?

src/planning.py
tomsilver: i'm not sure if this is what we want. i think it is helpful to see where the refinement is getting stuck. for example, in the option learning stuff, it is informative to see the option trying to do something but failing (e.g. trying to pick or trying to place). we can rename or make clearer somehow, but I do think we want the original implementation.

src/planning.py
tomsilver: we'll probably have to undo this change if you accept my comment above

src/planning.py
```diff
@@ -264,46 +288,52 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
                 else:
                     can_continue_on = False
-            else:
-                cur_idx += 1  # it's about to be decremented again
         else:
             # If the option is not initiable, need to resample / backtrack.
             can_continue_on = False
             cur_idx += 1  # it's about to be decremented again
         if not can_continue_on:
+            # Update the longest_failed_refinement found so far.
```
ronuchit, line 324: @tomsilver i think this is the right place to do this, finally...

ronuchit: note that this will only update longest_failed_refinement if we're at a step where we failed (for any reason: non-initiable option, expected_atoms check failed, or an EnvironmentFailure occurred). i think this is alright & exactly what we want -- that's what makes it a "failed" refinement

src/planning.py
```diff
@@ -264,46 +288,52 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
         if option.initiable(state):
             try:
                 next_state = option_model.get_next_state(state, option)
```
ronuchit, line 290: shuffled around some of the lines in this part to enhance readability (impossible to understand from diffs, but the actual code is clearer than before)

src/planning.py
```diff
@@ -264,46 +288,52 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
             can_continue_on = False
             cur_idx += 1  # it's about to be decremented again
         if not can_continue_on:
+            # Update the longest_failed_refinement found so far.
+            if cur_idx > len(longest_failed_refinement):
+                longest_failed_refinement = list(plan[:cur_idx])
+            # If we're immediately propagating failures, and we got a failure,
+            # raise it now. We don't do this right after catching the
+            # EnvironmentFailure because we want to make sure to update
+            # the longest_failed_refinement first.
```
ronuchit, line 330: PS this was the reason painting wasn't working! due to the "immediately", which i now moved down here.

ronuchit:
Command line:
`python src/main.py --env painting --approach oracle --seed 123 --num_test_tasks 1 --video_dir videos/ --results_dir results/ --max_skeletons_optimized 1 --painting_lid_open_prob 0.0 --make_failure_videos`

Video:

https://user-images.githubusercontent.com/3521150/151594160-eeedb85c-ecc8-4b61-b97d-51aa64762205.mp4



Command line:
`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --offline_data_method demo --option_learner neural --make_failure_videos --sampler_learner neural --num_test_tasks 5`

Video:

https://user-images.githubusercontent.com/3521150/151594191-9492af53-fd5f-4de8-961d-c6bf1c392347.mp4

src/planning.py
tomsilver: yeah, this looks good

tomsilver (commented):
@ronuchit thanks! lgtm. did you remember to eat your yapfies? if so, feel free to approve and merge. (I can't approve b/c I opened the PR.)

ronuchit (approved):

PR #372, opened 2022-01-28
Dataset refactor

tomsilver:
change Dataset from List[LowLevelTrajectory] to a dataclass. this is in preparation for refactoring the dataset creation for interactive_learning so that it happens in datasets/ rather than in the approach itself. that change isn't in this PR though; that will happen next.

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -909,6 +909,34 @@ class LowLevelTrajectory:
         return self._goal
 
 
+@dataclass(frozen=True, repr=False, eq=False)
+class Dataset:
+    """A collection of LowLevelTrajectory objects, and optionally, lists of
+    annotations, one per trajectory.
+
+    For example, in interactive learning, an annotation for an offline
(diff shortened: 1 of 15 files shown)


amburger66 (approved):
lgtm

src/structs.py
```diff
@@ -909,6 +909,34 @@ def goal(self) -> Set[GroundAtom]:
         return self._goal
 
 
+@dataclass(frozen=True, repr=False, eq=False)
+class Dataset:
+    """A collection of LowLevelTrajectory objects, and optionally, lists of
+    annotations, one per trajectory.
+
+    For example, in interactive learning, an annotation for an offline
+    learning Dataset would be of type List[Set[GroundAtom]] (with
```
ronuchit, line 918: Note you might have to do some mypy casting stuff.

ronuchit (approved):
thanks for taking the bullet on this annoying but important change!

PR #373, opened 2022-01-28
initialize approaches with train tasks 

tomsilver:
and replace goals with train task idxs in lowleveltrajectory


ronuchit (approved):
idk what you mean, this looks super fun!

PR #376, opened 2022-01-28
saving/loading approaches with online learning

ronuchit:
includes a non-working version for interactive_learning_approach.py,
because predicates aren't being saved & loaded properly, but we will
address this after the refactoring of that file


tomsilver (approved):
after changing -1 to None everywhere as discussed, lgtm

PR #377, opened 2022-01-28
fix bug in construction of negative examples for interactive learning

tomsilver:


ronuchit (approved):
lgtm, just make sure it works correctly with `ratio=0` and such

PR #379, opened 2022-01-29
refactor interactive learning to use new online infrastructure

ronuchit:
wip for interactive online stuff
* now that we're doing online learning without any hacks, we can completely get rid of `interactive_known_predicates` and should use the existing `excluded_predicates` infrastructure instead. i've done that in this PR
* removed hacks surrounding predicate stripping from interactive_learning_approach, now there's no way it can cheat. the initial predicates that are passed into `__init__` are the known predicates. the ones to be learned are extracted from the offline dataset at the start of `learn_from_offline_dataset` (note their classifiers were stripped during dataset construction, so there's no way to cheat)
* there's one functional difference that i wrote a "TODO" comment for, we should discuss and make a decision

@tomsilver @amburger66 read this before looking at the diffs!

### DIFF
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -74,8 +74,6 @@ class InteractiveLearningApproach(NSRTLearningApproach):
                 if score > self._best_score:
                     atoms_to_query.add(atom)
                     self._best_score = score
-            if not atoms_to_query:
-                return None
             return GroundAtomsHoldQuery(atoms_to_query)
 
         def _termination_function(s: State) -> bool:
@@ -95,7 +93,6 @@ class InteractiveLearningApproach(NSRTLearningApproach):
             if response is None:
                 continue  # we didn't ask a query on this timestep
             assert isinstance(response, GroundAtomsHoldResponse)
-            assert isinstance(response.query, GroundAtomsHoldQuery)
             for query_atom, atom_holds in response.holds.items():
                 # Still need a way to use negative examples.
                 if not atom_holds:
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -90,8 +90,6 @@ class InteractiveLearningApproach(NSRTLearningApproach):
         assert len(results) == 1
         result = results[0]
         for state, response in zip(result.states, result.responses):
-            if response is None:
-                continue  # we didn't ask a query on this timestep
             assert isinstance(response, GroundAtomsHoldResponse)
             for query_atom, atom_holds in response.holds.items():
                 # Still need a way to use negative examples.
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -101,8 +101,12 @@ class InteractiveLearningApproach(NSRTLearningApproach):
                 if not atom_holds:
                     continue
                 # Add this atom because it's a positive example.
-                traj = LowLevelTrajectory([state], [])
-                self._dataset.append(traj, [{query_atom}])
+                # Note: these pragma no covers are very temporary; we will
+                # remove them in a future PR where we change the score function
+                # to use >= instead of >, among other things. But for the sake
+                # of a pure refactor, we're leaving it alone for now.
+                traj = LowLevelTrajectory([state], [])  # pragma: no cover
+                self._dataset.append(traj, [{query_atom}])  # pragma: no cover
         self._relearn_predicates_and_nsrts(
             self._dataset, online_learning_cycle=self._online_learning_cycle)
         self._online_learning_cycle += 1
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -13,7 +13,8 @@ from predicators.src.torch_models import LearnedPredicateClassifier, \
     MLPClassifier
 from predicators.src.option_model import _OracleOptionModel
 from predicators.src.utils import get_object_combinations, strip_predicate
-from predicators.src.teacher import GroundAtomHoldsQuery
+from predicators.src.teacher import GroundAtomHoldsQuery, \
+    GroundAtomHoldsResponse
 from predicators.src.settings import CFG
 
 
@@ -29,7 +30,7 @@ class InteractiveLearningApproach(NSRTLearningApproach):
         self._best_score = 0.0
         # Initialize things that will be set correctly in offline learning
         self._dataset = Dataset([])
-        self._predicates_to_learn = set()
+        self._predicates_to_learn: Set[Predicate] = set()
         self._online_learning_cycle = 0
 
     def _get_current_predicates(self) -> Set[Predicate]:
@@ -97,6 +98,8 @@ class InteractiveLearningApproach(NSRTLearningApproach):
         for state, response in zip(result.states, result.responses):
             if response is None:
                 continue  # we didn't ask a query on this timestep
+            assert isinstance(response, GroundAtomHoldsResponse)
+            assert isinstance(response.query, GroundAtomHoldsQuery)
             if response.holds:
                 # Add this atom if it's a positive example
                 traj = LowLevelTrajectory([state], [])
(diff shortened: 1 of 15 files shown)


src/main.py
```diff
@@ -76,26 +76,7 @@ def main() -> None:
     for option in env.options:
         option.params_space.seed(CFG.seed)
     assert env.goal_predicates.issubset(env.predicates)
```
ronuchit, line 78: @tomsilver This refactor moves this code to utils.py, since dataset `__init__.py` needs it too. Should we refactor this change into a separate PR? (Added benefit: test_main.py will get a nice speedup bc we don't need to run the whole pipeline just to test predicate exclusion lol)

tomsilver: sure sounds good to me

tomsilver: #380

src/approaches/interactive_learning_approach.py
```diff
@@ -21,86 +24,106 @@ class InteractiveLearningApproach(NSRTLearningApproach):
...
+                    # TODO: the previous line is a > on master, changed to >=
+                    # because otherwise i wasn't seeing any new data get added
+                    # (due mostly to the next TODO)
+                    self._best_score = score
+                    # TODO: one difference from current master is that
+                    # we can now only ask one Query per state, whereas
+                    # before you could ask the teacher multiple things
+                    # about the same state. hopefully this is still okay...
+                    # otherwise we can change the InteractionRequest object's
+                    # query object to allow a set of Querys per timestep
```
tomsilver, line 83: good point. I think we do want to allow the agent to ask about multiple atoms per state. we could do what you suggest here, or we could change the Query/Response for this project to be about sets of ground atoms instead of one. I don't have a preference.

src/approaches/interactive_learning_approach.py
```diff
@@ -21,86 +24,106 @@ class InteractiveLearningApproach(NSRTLearningApproach):
...
+        assert task.init is init
+        def _query_policy(s: State) -> Optional[GroundAtomHoldsQuery]:
+            # Decide whether to ask about each possible atom
+            ground_atoms = utils.all_possible_ground_atoms(
+                s, self._predicates_to_learn)
+            for atom in ground_atoms:
+                score = score_atom(self._dataset, atom)
+                # Ask about this atom if it is the best seen so far
+                if score >= self._best_score:
+                    # TODO: the previous line is a > on master, changed to >=
```
tomsilver, line 74: this is a significant functional change, but since this score function is just a placeholder anyway, I think this change is okay. it will certainly lead to more interesting and testable behavior, where queries continue after the first state. @amburger66 let us know what you think though.

tomsilver:
TODO: we realized that ground-truth predicate classifiers are leaking into the interactive learning approach for goal predicates, through train tasks. we'll need to fix this either here or in a separate PR.

src/approaches/interactive_learning_approach.py
amburger66: sounds good!

tomsilver: cool. we'll still do this in a separate PR if we do it, to keep this one a pure refactor.

src/approaches/interactive_learning_approach.py
tomsilver: I decided to go with the latter: #381

tomsilver:
There are a number of functional changes I'd like to make after we merge this PR, I'm putting them here so I don't forget:
* Change the annotations from positive only to positive/negative, and put the negative results from active learning into negative
* Implement an option to use a different kind of trivial score function for selecting atoms to ask about, one that just asks about everything. (Maybe change to >= instead of > in this step)
* Allow falling back to random policy if no goal can be found
* Figure out if we want the agent to be able to ask a different kind of query like "tell me everything in this state" or "tell me something in this state", which we would need to pure active learning, because otherwise it doesn't know what the universe of predicates is
* Test that active learning alone works

tests/approaches/test_interactive_approach.py
```diff
@@ -52,38 +44,3 @@ def test_interactive_learning_approach():
...
-
-    # Test teacher
-    traj = dataset.trajectories[0]
-    for s in traj.states:
-        ground_atoms = sorted(utils.abstract(s, env.predicates))
-        for g in ground_atoms:
-            assert approach.ask_teacher(s, g)
-
-
-def test_interactive_learning_approach_no_ground_atoms():
```
tomsilver, line 64: I'm removing this test because the expected behavior is different now; the agent will not attempt to learn about a predicate that it has no data about, because it won't even know that this predicate exists.

tomsilver:
@amburger66 @ronuchit I think this is ready for review now.

After we merge, we should launch supercloud experiments to identify if we've introduced any major regressions. (@amburger66 , could you do this? thanks!)

tests/approaches/test_interactive_approach.py
```diff
@@ -1,89 +1,53 @@
...
-        "num_test_tasks": 5,
-    })
-    env = CoverEnv()
-    train_tasks = env.get_train_tasks()
-    approach = _DummyInteractiveLearningApproach(env.predicates, env.options,
-                                                 env.types, env.action_space,
-                                                 train_tasks)
-    dataset = create_dataset(env, train_tasks)
-    assert approach.is_learning_based
-    # MLP training fails since there are 0 positive examples
```
ronuchit, line 87: oh lol so we knew about this issue before. we should def fix it!

src/approaches/interactive_learning_approach.py
```diff
@@ -21,95 +23,116 @@ class InteractiveLearningApproach(NSRTLearningApproach):
...
+                # Ask about this atom if it is the best seen so far.
+                if score > self._best_score:
+                    atoms_to_query.add(atom)
+                    self._best_score = score
+            if not atoms_to_query:
+                return None
+            return GroundAtomsHoldQuery(atoms_to_query)
+
+        def _termination_function(s: State) -> bool:
+            # Stop the episode if we reach the goal that we babbled.
```
ronuchit, line 82: should we change "if" to "when"? oh nvm, because the option model could be learned/wrong in the final version of the system (just keeping this comment for clarity's sake, no action needed)

src/approaches/interactive_learning_approach.py
```diff
@@ -21,95 +23,116 @@ class InteractiveLearningApproach(NSRTLearningApproach):
...
+            ground_atoms = utils.all_possible_ground_atoms(
+                s, self._predicates_to_learn)
+            atoms_to_query = set()
+            for atom in ground_atoms:
+                score = score_atom(self._dataset, atom)
+                # Ask about this atom if it is the best seen so far.
+                if score > self._best_score:
+                    atoms_to_query.add(atom)
+                    self._best_score = score
+            if not atoms_to_query:
```
ronuchit, line 77: note that we don't really need this... can always make the Query with an empty set. not sure if you'd prefer that idk

src/approaches/interactive_learning_approach.py
```diff
@@ -21,95 +23,116 @@ class InteractiveLearningApproach(NSRTLearningApproach):
...
+
+    def learn_from_interaction_results(
+            self, results: Sequence[InteractionResult]) -> None:
+        assert len(results) == 1
+        result = results[0]
+        for state, response in zip(result.states, result.responses):
+            if response is None:
+                continue  # we didn't ask a query on this timestep
+            assert isinstance(response, GroundAtomsHoldResponse)
+            assert isinstance(response.query, GroundAtomsHoldQuery)
```
ronuchit, line 98: remove this line, we no longer need it (it was initially there to appease mypy when we were using `response.query`)

ronuchit (commented):
woohoo! i can't approve as the author, minor comments

amburger66:
> After we merge, we should launch supercloud experiments to identify if we've introduced any major regressions. (@amburger66 , could you do this? thanks!)

yup

amburger66:
lgtm

tomsilver (approved):

PR #381, opened 2022-01-29
change ground atoms query and response to allow multiple atoms 

tomsilver:
also change to GroundAtoms themselves instead of (str, objects) for consistency with data+ground_atoms.


amburger66 (commented):
looks good

ronuchit (approved):
lgtm thanks

PR #383, opened 2022-01-29
strip task of excluded goal predicate classifiers

tomsilver:


ronuchit (approved):
thanks, can you make sure this works as expected in interactive approach?

amburger66 (approved):

PR #384, opened 2022-01-30
take a shortcut for classification with only one class in data

### LINKED ISSUE #378: fix ratio 0.0 MLP learning
interactive learning approach with ratio 0.0 is crashing right now

tomsilver:
closes #378

### DIFF
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -294,6 +294,8 @@ class MLPClassifier(nn.Module):
         self._input_shift = np.zeros(1, dtype=np.float32)
         self._input_scale = np.zeros(1, dtype=np.float32)
         self._max_itr = max_itr
+        self._do_single_class_prediction = False
+        self._single_class = False
 
     def fit(self, X: Array, y: Array) -> None:
         """Train classifier on the given data.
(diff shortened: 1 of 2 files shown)


src/torch_models.py
```diff
@@ -303,6 +305,17 @@ def fit(self, X: Array, y: Array) -> None:
         torch.manual_seed(CFG.seed)
         assert X.ndim == 2
         assert y.ndim == 1
+        # If there is only one class in the data, then there's no point in
+        # learning a NN, since any predictions other than that one class
+        # could only be strange generalization issues.
+        if np.all(y == 0):
+            self._do_single_class_prediction = True
+            self._single_class = False
```
ronuchit, line 313: Can we change this to `predicted_single_class`?

src/torch_models.py
```diff
@@ -337,6 +351,8 @@ def classify(self, x: Array) -> bool:
 
         x is single-dimensional.
         """
+        if self._do_single_class_prediction:
```
ronuchit, line 354: Let's move this into `_classify` so that the asserts later in this method are still run?

tests/test_models.py
```diff
@@ -83,3 +83,19 @@ def test_mlp_classifier():
     model = MLPClassifier(input_size, 10000)
     model.fit(X, y)
     assert time.time() - start_time < 3, "Didn't early stop"
+    # Test with no positive examples.
+    y = np.zeros(len(X))
+    model = MLPClassifier(input_size, 100)
```
ronuchit, line 88: Can we make these 100s bigger, and do a time check like above? (`assert time.time() - start_time < 3, "Didn't early stop"`) The time should be nearly instantaneous for the calls to `fit`.

ronuchit (commented):
some suggestions

src/torch_models.py
tomsilver: I want to bypass the normalization, but I refactored so the asserts are still run.

ronuchit (approved):
lgtm after comments

PR #385, opened 2022-01-30
main pipeline evaluation schedule fix

tomsilver:
our current code doesn't evaluate learning-based approaches if `num_online_learning_cycles` is 0. also, it doesn't evaluate approaches after the final learning cycle. this should fix both issues

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -105,13 +105,14 @@ def _run_pipeline(env: BaseEnv, approach: BaseApproach,
             approach.load(online_learning_cycle=None)
         else:
             approach.learn_from_offline_dataset(dataset)
+        # Run evaluation once before online learning starts.
+        results = _run_testing(env, approach)
+        results["num_transitions"] = total_num_transitions
+        results["learning_time"] = time.time() - learning_start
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -105,13 +105,14 @@ def _run_pipeline(env: BaseEnv, approach: BaseApproach,
...
+        results["num_transitions"] = total_num_transitions
+        results["learning_time"] = time.time() - learning_start
         teacher = Teacher()
         # The online learning loop.
         for i in range(CFG.num_online_learning_cycles):
             print(f"\n\nONLINE LEARNING CYCLE {i}\n\n")
-            results = _run_testing(env, approach)
-            results["num_transitions"] = total_num_transitions
-            results["learning_time"] = time.time() - learning_start
             _save_test_results(results)
```
ronuchit, line 116: don't we need to move this too?

tomsilver: whoops thanks

ronuchit (approved):
good eye, thanks for the fix

amburger66 (approved):

PR #386, opened 2022-01-30
switch interactive dataset to positive/unlabeled/negative 

tomsilver:
this is a functional change to the interactive approach, and should not be merged until we confirm that we didn't break anything (else) during refactoring.

the functional changes are in active learning:
(1) previously, we would ignore any "negative" responses from the teacher, rather than treating them as known negatives. this actually didn't make a difference though because...
(2) previously, all of the ground atoms that we did not ask about would be treated as negative examples (because they're unlabelled). that's bad, especially because we generally ask about one or a few ground atoms out of many possible, so there would be a whole lot of false negative ground atoms collected during active learning, and active learning could ultimately be harmful. so this change is somewhat of a bug fix.

### DIFF
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -1,6 +1,6 @@
 """An approach that learns predicates from a teacher."""
 
-from typing import Set, List, Optional, Tuple, Callable, Sequence
+from typing import Set, List, Optional, Tuple, Callable, Sequence, Dict
 import numpy as np
 from gym.spaces import Box
 from predicators.src import utils
@@ -11,7 +11,6 @@ from predicators.src.structs import State, Predicate, ParameterizedOption, \
     InteractionResult, Action
 from predicators.src.torch_models import LearnedPredicateClassifier, \
     MLPClassifier
-from predicators.src.utils import get_object_combinations
 from predicators.src.teacher import GroundAtomsHoldQuery, \
     GroundAtomsHoldResponse
 from predicators.src.settings import CFG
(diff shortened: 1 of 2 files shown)


ronuchit (approved):
code changes lgtm. but i don't understand what the functional change is, exactly -- it looks like a cosmetic change to me. could you expand the comment at the top of the PR?

src/approaches/interactive_learning_approach.py
```diff
@@ -91,17 +109,16 @@ def learn_from_interaction_results(
         result = results[0]
         for state, response in zip(result.states, result.responses):
             assert isinstance(response, GroundAtomsHoldResponse)
+            state_annotation: Dict[str, Set[GroundAtom]] = {
```
ronuchit, line 112: random question, what if this `state` already appeared somewhere else in the dataset? then wouldn't you need to update its annotation, otherwise you could give bad data to the learning?

ronuchit: e.g. if the teacher left out an atom due to `ratio < 1.0` but then during online you ask about that state and the teacher says it's positive

tomsilver:
> code changes lgtm. but i don't understand what the functional change is, exactly -- it looks like a cosmetic change to me. could you expand the comment at the top of the PR?

thanks, updated

src/approaches/interactive_learning_approach.py
tomsilver: hmm that's true... that is annoying... I'll have to think more about this.

ronuchit:
I see. The functional change is that you're not looping over all possible ground atoms during the setup of the classifier anymore. It's only happening once during offline learning. Which makes the difference occur in active learning. Right?

src/approaches/interactive_learning_approach.py
tomsilver: well actually it might be fine. the only case where we would want to update an annotation is if we previously marked something as unlabelled, but now we know that it's actually positive or negative. right?

so i'm going to file this concern under the comment in the code:
```
                    # Here we make the (wrong in general!) assumption that
                    # unlabeled ground atoms are negative. In the future, we
                    # may want to modify this, e.g., downweight or remove
                    # the unlabeled examples once we collect enough negatives.
```

tomsilver:
> I see. The functional change is that you're not looping over all possible ground atoms during the setup of the classifier anymore. It's only happening once during offline learning. Which makes the difference occur in active learning. Right?

yeah exactly

src/approaches/interactive_learning_approach.py
ronuchit: Note that active learning can never create an "unlabeled" label. So active learning should always overwrite any unlabeled label that was created during offline learning. But unfortunately this means you have to loop through the dataset to find matching states to the one you queried?

ronuchit: lol yeah we said the same thing

ronuchit: the only problem is when we now know that it's actually positive, that will lead to contradictory data

amburger66 (commented):
looks good so far! i am also up-to-date on the comments

PR #387, opened 2022-01-30
fix crash in data generation when goal predicates excluded

tomsilver:
another bug created by me and found by @amburger66 😬


amburger66 (approved):
thanks :D

ronuchit (approved):

PR #389, opened 2022-01-30
more refactoring of interactive approach

tomsilver:
A number of refactors to the active learning part of interactive learning, in preparation for implementing new choices of different subparts.

* `CFG.interactive_action_strategy` controls how we want to set up the exploration policy and termination function
* `CFG.interactive_query_policy` controls how we want to decide what atoms to query in each state
* `CFG.interactive_score_function` controls the score function used in the query policy to determine what atoms are worth querying about

Each of these only has one default choice right now, but I expect we'll implement more options for each in the not too distant future.

Two other minor refactors:
* Combined the two previously separate score functions into one (`_score_atom_set`)
* Refactored `glib_sample` into a generic `sample_subsets` utility and the other parts. One minor functional change here: previously we had two separate parameters, which controlled how many goals to babble, and then how many of those goals to return before trying out each one with task planning. I removed the second parameter, so now we just specify how many goals we want to babble, and then we try all of them in the worst case. (This is how we did it in GLIB iirc.)

### DIFF
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -13,7 +13,7 @@ from predicators.src.torch_models import LearnedPredicateClassifier, \
     MLPClassifier
 from predicators.src.utils import get_object_combinations
 from predicators.src.teacher import GroundAtomsHoldQuery, \
-    GroundAtomsHoldResponse
+    GroundAtomsHoldResponse, Query
 from predicators.src.settings import CFG
 
 
@@ -35,6 +35,8 @@ class InteractiveLearningApproach(NSRTLearningApproach):
     def _get_current_predicates(self) -> Set[Predicate]:
         return self._initial_predicates | self._predicates_to_learn
 
+    ###################### Semi-supervised learning #########################
+
     def learn_from_offline_dataset(self, dataset: Dataset) -> None:
         # First, go through the dataset's annotations and figure out the
         # set of predicates to learn. Note that their classifiers were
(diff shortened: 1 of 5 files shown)


src/approaches/interactive_learning_approach.py
```diff
@@ -67,41 +69,165 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
         # Save dataset, to be used for online interaction.
         self._dataset = dataset
 
+    def _relearn_predicates_and_nsrts(
```
tomsilver, line 72: this function should be unchanged, I just moved it up in the file.

src/approaches/interactive_learning_approach.py
```diff
@@ -67,41 +69,165 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
+            return self._score_atom_set_frequency(atom_set)
+        raise NotImplementedError("Unrecognized interactive_score_function:"
+                                  f" {CFG.interactive_score_function}.")
+
+    def _select_interaction_train_task_idx(self) -> int:
+        # At the moment, we only have one way to select a train task idx:
+        # choose one uniformly at random. In the future, we may want to
+        # try other strategies. But one nice thing about random selection
+        # is that we're not making a hard commitment to the agent having
+        # control over which train task it gets to use.
```
ronuchit, line 156: good point, it's like sampling from a uniform initial state distribution

src/approaches/interactive_learning_approach.py
```diff
@@ -67,41 +69,165 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
-        # Sample a train task.
-        train_task_idx = self._rng.choice(len(self._train_tasks))
+        # We will create a single interaction request.
+        # Determine the train task that we will be using.
+        train_task_idx = self._select_interaction_train_task_idx()
+        # Determine the action policy and termination function.
+        act_policy, termination_function = \
+            self._create_interaction_action_strategy(train_task_idx)
+        # Determine the query policy.
+        query_policy = self._create_query_policy(train_task_idx)
```
ronuchit, line 132: rename to "_create_interaction_query_policy" to match the above function name?

src/approaches/interactive_learning_approach.py
```diff
@@ -67,41 +69,165 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
+        """Find the most interesting reachable ground goal and plan to it."""
         init = self._train_tasks[train_task_idx].init
         # Detect and filter out static predicates.
         static_preds = utils.get_static_preds(self._nsrts,
                                               self._predicates_to_learn)
         preds = self._predicates_to_learn - static_preds
-        # Find acting policy for the request.
-        task_list = glib_sample(init, preds, self._dataset)
-        assert task_list
+        # Sample possible goals to plan toward.
```
ronuchit, line 186: You may want to make this fall back to random by making this sample_subsets function return a generator, and then `find_first_solvable` loops over that generator and if none of the goals are reachable, fall back. Can be done in  a different PR too.

src/utils.py
```diff
@@ -844,6 +844,25 @@ def all_possible_ground_atoms(state: State,
...
 
+_T = TypeVar("_T")  # element of a set
+
+
+def sample_subsets(universe: Sequence[_T], num_samples: int, min_set_size: int,
+                   max_set_size: int,
+                   rng: np.random.Generator) -> List[Set[_T]]:
+    """Sample multiple subsets from a universe."""
+    samples = []
+    assert min_set_size <= len(universe), "Not enough elements in universe"
```
ronuchit, line 855: do we also need to assert that `min_set_size <= max_set_size`?

ronuchit, line 855: shouldn't this be `max_set_size`?

src/utils.py
```diff
@@ -844,6 +844,25 @@ def all_possible_ground_atoms(state: State,
...
+    for _ in range(num_samples):
+        set_size = rng.integers(min_set_size, max_set_size + 1)
+        idxs = rng.choice(np.arange(len(universe)),
+                          size=set_size,
+                          replace=False)
+        sample = {universe[i] for i in idxs}
+        samples.append(sample)
+    return samples
```
ronuchit, line 863: as mentioned above, we may want to turn this into a generator

tests/test_utils.py
```diff
@@ -359,6 +359,20 @@ def test_strip_task():
...
+    universe = list(range(10))
+    num_samples = 5
+    min_set_size = 1
+    max_set_size = 2
+    rng = np.random.default_rng(0)
+    samples = utils.sample_subsets(universe, num_samples, min_set_size,
+                                   max_set_size, rng)
+    assert len(samples) == 5
+    assert {len(s) for s in samples} == {1, 2}
+    assert all(s.issubset(set(universe)) for s in samples)
```
ronuchit, line 373: some tests to add:
```
assert not utils.sample_subsets(universe, 0, min_set_size, max_set_size, rng)
assert utils.sample_subsets([], num_samples, 0, 0, rng) == [set() for _ in range(num_samples)]
with pytest.raises(AssertionError):
  utils.sample_subsets(universe, num_samples, min_set_size, 0, rng)
with pytest.raises(AssertionError):
  utils.sample_subsets([], num_samples, min_set_size, max_set_size, rng)
```

src/utils.py
ronuchit: actually, regardless of above, this function makes more sense as a generator right

src/approaches/interactive_learning_approach.py
tomsilver: I'll make it a generator, but we need to sort these, so it won't make a difference here.

ronuchit (approved):
LGTM. My general feeling is that you'll need to first come up with a situation where active learning is useful, before any of this new experimentation can be done. Hopefully playroom has everything you need!

amburger66 (approved):

PR #391, opened 2022-01-31
reset_config instead of update_config in tests

tomsilver:
this will hopefully put an end to the nightmare of unit tests interfering with each other. we should now use `reset_config` instead of `update_config` in all tests.

I went through all of the tests that previously used `update_config` and removed superfluous things like `seed: 123` (which happens automatically), multiple consecutive calls to `update_config` to deal with the `get_arg_specific_settings` stuff, and calls to `update_config` at the end of unit tests.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -1250,31 +1250,37 @@ def save_video(outfile: str, video: Video) -> None:
     print(f"Wrote out to {outpath}")
 
 
-def update_config(args: Dict[str, Any], default_seed: int = 123) -> None:
+def update_config(args: Dict[str, Any]) -> None:
     """Args is a dictionary of new arguments to add to the config CFG."""
-    # Only override attributes, don't create new ones
-    allowed_args = set(CFG.__dict__)
+    arg_specific_settings = GlobalSettings.get_arg_specific_settings(args)
+    # Only override attributes, don't create new ones.
+    allowed_args = set(CFG.__dict__) | set(arg_specific_settings)
     parser = create_arg_parser()
-    # Unfortunately, can't figure out any other way to do this
+    # Unfortunately, can't figure out any other way to do this.
     for parser_action in parser._actions:  # pylint: disable=protected-access
         allowed_args.add(parser_action.dest)
     for k in args:
         if k not in allowed_args:
             raise ValueError(f"Unrecognized arg: {k}")
-    for k in ("env", "approach", "seed", "experiment_id"):
-        if k not in args and hasattr(CFG, k):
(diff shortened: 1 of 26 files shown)


src/utils.py
```diff
@@ -1267,16 +1268,34 @@ def update_config(args: Dict[str, Any], default_seed: int = 123) -> None:
             # pass in a value and this key is already in the
             # configuration dict, add the current value to args.
             args[k] = getattr(CFG, k)
-    # Maintain the invariant that CFG has some seed and some
-    # experiment_id set. This is very useful in unit tests, where
-    # there are often no command line args being passed.
-    args["seed"] = args.get("seed", default_seed)
-    args["experiment_id"] = args.get("experiment_id", "")
```
ronuchit, line 1274: Where is the default setting of experiment_id happening now?

src/utils.py
```diff
@@ -1267,16 +1268,34 @@ def update_config(args: Dict[str, Any], default_seed: int = 123) -> None:
...
+        "--env",
+        "default env placeholder",
+        "--seed",
+        str(default_seed),
+        "--approach",
+        "default approach placeholder",
+    ])
+    arg_dict = {
+        k: v
+        for k, v in GlobalSettings.__dict__.items() if not k.startswith("_")
```
ronuchit, line 1292: Will this work correctly with arg-specific settings?

tests/approaches/test_base_approach.py
```diff
@@ -28,9 +28,7 @@ def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
 
 def test_base_approach():
     """Tests for BaseApproach class."""
-    utils.update_config({
-        "seed": 123,
-    })
+    utils.reset_config({})
```
ronuchit, line 31: Why do we need calls like this? Does it just set you back to the default CFG that's in settings.py? If so, maybe we should make the argument be optional (default None) and turn calls like this into `utils.reset_config()`.

ronuchit (approved):
thank you so much 🙏 🙏 🙏  this is amazing

tests/approaches/test_base_approach.py
ronuchit: should be quick to do with a grep

src/utils.py
tomsilver: it defaults to an empty str in args.py

src/utils.py
tomsilver: yeah, the arg-specific settings are added in the call to `update_config` at the end of this function

PR #392, opened 2022-01-31
don't use num_transitions in results filenames

### LINKED ISSUE #390: saved results might get overwritten during active learning
`num_transitions` is supposed to be the distinguishing factor between filenames corresponding to different episodes of active learning, but it's possible that the number of transitions doesn’t increase during an episode, so the previous result would get overwritten

ronuchit:
@amburger66 sorry for all the fluctuations/wrinkles in the code recently, hope this is what you had in mind for fixing your issue

also updated the results script to not crash, and gave it a better name

closes #390


tomsilver (approved):
thanks!

PR #394, opened 2022-01-31
fix saving/loading for interactive_learning_approach

### LINKED ISSUE #382: test if saving/loading work with interactive learning approach (esp the online learning part)
in previous experiments, i noticed that they weren't working because the predicate classifiers weren't being saved correctly. need to dig into this

ronuchit:
@tomsilver this completes all the TODOs on the predicates refactoring doc! woohoo!!

Usage: after, e.g.:
`python src/main.py --env cover --approach interactive_learning --seed 0 --excluded_predicates HandEmpty,Holding --num_test_tasks 5`
you can do:
`python src/main.py --env cover --approach interactive_learning --seed 0 --excluded_predicates HandEmpty,Holding --num_test_tasks 5 --load_approach --load_data`
to feel like the flash

closes #382

### DIFF
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 from operator import le
 import glob
 import os
-from typing import Dict, DefaultDict, Set, List, Tuple, Sequence
+from typing import Dict, DefaultDict, Set, List, Tuple, Sequence, Any
 import pandas as pd
 from predicators.src.datasets import create_dataset
 from predicators.src.envs import create_env, BaseEnv, CoverEnv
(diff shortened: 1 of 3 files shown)


src/approaches/interactive_learning_approach.py
```diff
@@ -117,7 +121,13 @@ def _relearn_predicates_and_nsrts(
                 (self._predicates_to_learn - {pred}) | {new_pred}
 
         # Learn NSRTs via superclass
-        self._learn_nsrts(dataset.trajectories, online_learning_cycle)
+        self._learn_nsrts(self._dataset.trajectories, online_learning_cycle)
+
+        # Save the things we need other than the NSRTs, which were already
+        # saved in the above call to self._learn_nsrts()
+        save_path = utils.get_approach_save_path_str()
+        with open(f"{save_path}_{online_learning_cycle}.DATA", "wb") as f:
```
tomsilver, line 129: just to check, the reason this doesn't overwrite the file that's created by the parent class is that the suffixes are different (DATA here vs NSRTs there). right?

src/approaches/interactive_learning_approach.py
```diff
@@ -117,7 +121,13 @@ def _relearn_predicates_and_nsrts(
                 (self._predicates_to_learn - {pred}) | {new_pred}
 
         # Learn NSRTs via superclass
-        self._learn_nsrts(dataset.trajectories, online_learning_cycle)
+        self._learn_nsrts(self._dataset.trajectories, online_learning_cycle)
+
+        # Save the things we need other than the NSRTs, which were already
```
tomsilver, line 126: I think we need one other thing too -- `self._best_score`. and in general, we'll need to save/load any class attributes that can change after initialization. (cc @amburger66 )

tomsilver: actually, best_score might get set correctly through the re-running of the interactions? but regardless, I think we should add it to the things we're saving and loading.

src/approaches/interactive_learning_approach.py
```diff
@@ -117,7 +121,13 @@ def _relearn_predicates_and_nsrts(
...
+        # saved in the above call to self._learn_nsrts()
+        save_path = utils.get_approach_save_path_str()
+        with open(f"{save_path}_{online_learning_cycle}.DATA", "wb") as f:
+            pkl.dump((self._dataset, self._predicates_to_learn), f)
```
tomsilver, line 130: instead of a tuple, can we make this a dict with keys `dataset, predicates_to_learn`?

tomsilver (commented):
thank you! yay for being on the last TODO

src/approaches/interactive_learning_approach.py
ronuchit: right

tomsilver (approved):
lgtm thanks!

amburger66 (approved):
thanks!

PR #395, opened 2022-01-31
make grammar search analysis script mirror main pipeline

tomsilver:
the idea is that it should now be easy to just replace `src/main.py` with `analysis/grammar_search_analysis.py` and be on your way.

for example:
```
python analysis/grammar_search_analysis.py --experiment_id painting_none_excluded_demo_large --env painting --approach grammar_search_invention --seed 456 --offline_data_method demo --grammar_search_expected_nodes_allow_noops True --grammar_search_max_predicates 200
```

### DIFF
--- a/analysis/grammar_search_analysis.py
+++ b/analysis/grammar_search_analysis.py
@@ -18,6 +18,20 @@ from predicators.src import utils
 from predicators.src.structs import Predicate, Dataset, State, Object, Task
 from predicators.src.settings import CFG
 
+DEFAULT_ENV_NAMES = [
+    "cover",
+    "blocks",
+    "painting",
+]
+
+DEFAULT_SCORE_FUNCTION_NAMES = [
+    "expected_nodes",
+]
+
+DEFAULT_SEED = 0
+
+RUN_PLANNING = False
+
 
 def _run_proxy_analysis(args: Dict[str, Any], env_names: List[str],
                         score_function_names: List[str], run_planning: bool,
@@ -102,7 +116,9 @@ def _run_proxy_analysis(args: Dict[str, Any], env_names: List[str],
              "IsWet", "IsDry", "IsDirty", "IsClean"
          ])
 
-        def all_lids_open_classifier(state, objects):
+        def all_lids_open_classifier(state: State,
+                                     objects: Sequence[Object]) -> bool:
+            del objects  # unused
             for o in state:
                 if o.type.name == "lid" and state.get(o, "is_open") < 0.5:
                     return False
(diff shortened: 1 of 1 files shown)


analysis/grammar_search_analysis.py
```diff
@@ -242,29 +242,33 @@ def _make_proxy_analysis_results(outdir: str) -> None:
...
-    ]
-    run_planning = False
+    args = utils.parse_args(env_required=False,
+                            approach_required=False,
+                            seed_required=False)
+    assert args["excluded_predicates"] == ""  # gets ignored
+    if args["env"] is not None:
+        env_names = [args["env"]]
+    else:
+        del args["env"]
```
ronuchit, line 252: Is there any reason to `del` this?

analysis/grammar_search_analysis.py
```diff
@@ -242,29 +242,33 @@ def _make_proxy_analysis_results(outdir: str) -> None:
...
+        env_names = [
```
ronuchit, line 253: Let's do `env_names = DEFAULT_ENV_NAMES` here and define that constant at top of file?

analysis/grammar_search_analysis.py
```diff
@@ -242,29 +242,33 @@ def _make_proxy_analysis_results(outdir: str) -> None:
...
+            "cover",
+            "blocks",
+            "painting",
+        ]
+    if args["seed"] is None:
+        args["seed"] = 0
+    if "grammar_search_score_function" in args:
+        score_function_names = [args["grammar_search_score_function"]]
+    else:
+        score_function_names = ["expected_nodes"]
```
ronuchit, line 263: same, `DEFAULT_SCORE_FUNCTION_NAMES`

analysis/grammar_search_analysis.py
```diff
@@ -242,29 +242,33 @@ def _make_proxy_analysis_results(outdir: str) -> None:
...
+    run_planning = True
```
ronuchit, line 264: same, `RUN_PLANNING` can be a constant at top of file, and also can we set it False unless we have a good reason to have it True by default?

analysis/grammar_search_analysis.py
```diff
@@ -242,29 +242,33 @@ def _make_proxy_analysis_results(outdir: str) -> None:
...
-        # "lmcut_count_lookaheaddepth0",
-        # "hadd_count_lookaheaddepth0",
-        # "exact_count",
-        "expected_nodes",
-    ]
-    run_planning = False
+    args = utils.parse_args(env_required=False,
+                            approach_required=False,
+                            seed_required=False)
+    assert args["excluded_predicates"] == ""  # gets ignored
```
ronuchit, line 248: the comment confused me for a while, can we say `This script ignores excluded predicates, so we disallow them.`

ronuchit (approved):
great idea!

analysis/grammar_search_analysis.py
tomsilver: yeah if we don't, `None` will overwrite `env_name` in
```
utils.reset_config({
        "env": env_name,
        **args,
    })
```

ronuchit (approved):
lgtm. if you care, you can just use RUN_PLANNING directly everywhere instead of passing it around as a parameter

tomsilver:
> lgtm. if you care, you can just use RUN_PLANNING directly everywhere instead of passing it around as a parameter

eh

PR #399, opened 2022-02-01
demo generation is allowed to fail, emits a warning

ronuchit:
prevents occasional crashes in cluttered_table, where our oracle can fail due to loops

### DIFF
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -1,5 +1,6 @@
 """Create offline datasets by collecting demonstrations."""
 
+import sys
 from typing import List
 from predicators.src.approaches import create_approach, ApproachTimeout, \
     ApproachFailure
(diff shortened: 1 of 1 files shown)


src/datasets/demo_only.py
```diff
@@ -14,8 +15,12 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
                                       env.types, env.action_space, train_tasks)
     trajectories = []
     for idx, task in enumerate(train_tasks):
-        policy = oracle_approach.solve(
-            task, timeout=CFG.offline_data_planning_timeout)
+        try:
+            policy = oracle_approach.solve(
+                task, timeout=CFG.offline_data_planning_timeout)
+        except (ApproachTimeout, ApproachFailure) as e:
+            print(f"WARNING: Approach failed to solve with error: {e}")
```
tomsilver, line 22: can you make the warning more aggressive? e.g. new lines, all caps, exclamation points... this should be super rare, and we don't want to miss it when it's not. (worried more about people other than us, who aren't used to staring at the logs)

tomsilver (approved):

src/datasets/demo_only.py
ronuchit: changing to supercloud only

tomsilver (approved):
lgtm, feel free to pragma no cover...

PR #401, opened 2022-02-01
nonrefinable variant of cover environment

tomsilver:
https://user-images.githubusercontent.com/2816502/151999138-3a07764b-1006-497e-b8d6-012dc78a671b.mp4

### DIFF
--- a/src/envs/__init__.py
+++ b/src/envs/__init__.py
@@ -3,7 +3,7 @@
 from predicators.src.envs.base_env import BaseEnv
 from predicators.src.envs.cover import CoverEnv, CoverEnvTypedOptions, \
     CoverEnvHierarchicalTypes, CoverMultistepOptions, \
-    CoverMultistepOptionsFixedTasks, CoverEnvNonrefinable
+    CoverMultistepOptionsFixedTasks, CoverEnvRegrasp
 from predicators.src.envs.behavior import BehaviorEnv
 from predicators.src.envs.cluttered_table import ClutteredTableEnv, \
     ClutteredTablePlaceEnv
@@ -17,7 +17,7 @@ __all__ = [
     "CoverEnv",
     "CoverEnvTypedOptions",
     "CoverEnvHierarchicalTypes",
-    "CoverEnvNonrefinable",
+    "CoverEnvRegrasp",
     "CoverMultistepOptions",
     "CoverMultistepOptionsFixedTasks",
     "ClutteredTableEnv",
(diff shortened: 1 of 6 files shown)


src/envs/__init__.py
```diff
@@ -17,6 +17,7 @@
     "CoverEnv",
     "CoverEnvTypedOptions",
     "CoverEnvHierarchicalTypes",
+    "CoverEnvNonrefinable",
```
ronuchit, line 20: annoying suggestion: this env is not actually non-refinable, what we mean is non-downward-refinable. Can we rename to CoverEnvNondownwardref or maybe just CoverEnvRegrasp?

src/ground_truth_nsrts.py
```diff
@@ -282,6 +283,34 @@ def place_sampler(state: State, rng: np.random.Generator,
...
     nsrts.add(place_nsrt)
 
+    # Place (not on any target)
+    if CFG.env == "cover_nonrefinable":
+        parameters = [block]
+        preconditions = {
+            LiftedAtom(IsBlock, [block]),
+            LiftedAtom(Holding, [block])
+        }
+        add_effects = {
```
ronuchit, line 293: thank god this isn't actually a no-op, otherwise we'd have to disable the filtering in planner :P speaking of which, we might want to discuss on slack whether we should purge that whole thing now that we wiped replays

src/ground_truth_nsrts.py
```diff
@@ -282,6 +283,34 @@ def place_sampler(state: State, rng: np.random.Generator,
...
+        option = PickPlace
+        option_vars = []
+
+        def place_on_table_sampler(state: State, rng: np.random.Generator,
+                                   objs: Sequence[Object]) -> Array:
+            del state  # unused
+            assert len(objs) == 1
+            lb = 0.0
+            ub = 1.0
+            return np.array(rng.uniform(lb, ub, size=(1, )), dtype=np.float32)
```
ronuchit, line 306: so this isn't a perfect sampler, then? will that have any bad implications? i guess not as long as we're planning to keep max_samples_per_step = 10. but it would be a problem if we change that to 1 to allow ourselves a higher skeleton budget. don't think we need the higher skeleton budget for cover though.

ronuchit (commented):
thanks for doing this!

tomsilver:
@ronuchit thanks ready

ronuchit (approved):
lgtm

PR #403, opened 2022-02-01
ExpectedNodesScoreFunction can use either nodes_created or nodes_expanded

ronuchit:
after merging this, will update supercloud-vary-score-function branch


tomsilver (approved):
great!

PR #404, opened 2022-02-01
hack: filter out invented predicates not appearing in any precondition

ronuchit:
tested by adding `Forall[1:box].[InBox(0,1)]` into the debug grammar and verifying that it gets pruned out by the new code


tomsilver (approved):
Thank you!

PR #405, opened 2022-02-01
change sesame to keep no-op operators by default

ronuchit:
tested by ensuring that OpenLid operator is kept even on the first run
of A* when doing `python src/main.py --env painting --approach oracle
--seed 0 --sesame_allow_noops True`

`python src/main.py --env painting --approach oracle --seed 0
--sesame_allow_noops True`
-> Main script terminated in 32.23630 seconds

`python src/main.py --env painting --approach oracle --seed 0 --sesame_allow_noops False`
-> Main script terminated in 31.31945 seconds


tomsilver (approved):
Lgtm

PR #406, opened 2022-02-01
new score function and query policy for "ask always" strategy

tomsilver:


tests/approaches/test_interactive_approach.py
```diff
@@ -55,6 +55,14 @@ def test_interactive_learning_approach():
             pass
         # We won't check the policy here because we don't want unit tests to
         # have to train very good models, since that would be slow.
+    # Test with a query policy that always queries about every atom.
+    utils.update_config({
+        "interactive_query_policy": "nonstrict_best_seen",
```
ronuchit, line 60: so nonstrict + trivial = ask about every atom in every state along the path to the glibbed goal. right?

ronuchit (approved):

tests/approaches/test_interactive_approach.py
tomsilver: right!

ronuchit: dumb robot!

amburger66 (approved):

PR #407, opened 2022-02-02
random explore policy and glib fallback

tomsilver:


tests/approaches/test_interactive_approach.py
```diff
@@ -55,6 +55,18 @@ def test_interactive_learning_approach():
...
+        "interactive_action_strategy": "random",
+    })
+    interaction_requests = approach.get_interaction_requests()
+    _generate_interaction_results(env.simulate, teacher, train_tasks,
+                                  interaction_requests)
+    # Test that glib falls back to random if no solvable task can be found.
+    utils.update_config({
+        "interactive_action_strategy": "glib",
+        "timeout": 0.0,
+    })
```
amburger66, line 69: do we need a call to `approach.get_interaction_requests()` here?

tomsilver: thank you yes! got lost in merging

amburger66 (approved):

src/approaches/interactive_learning_approach.py
```diff
@@ -223,7 +225,12 @@ def _create_glib_interaction_strategy(
                            key=self._score_atom_set_frequency,
                            reverse=True)  # largest to smallest
         task_list = [Task(init, goal) for goal in goal_list]
-        task, act_policy = self._find_first_solvable(task_list)
+        try:
+            task, act_policy = self._find_first_solvable(task_list)
+        except ApproachFailure:
```
ronuchit, line 230: what about approachtimeout?

src/approaches/interactive_learning_approach.py
```diff
@@ -232,6 +239,25 @@ def _termination_function(s: State) -> bool:
 
         return act_policy, _termination_function
 
+    def _create_random_interaction_strategy(
+        self, train_task_idx: int
+    ) -> Tuple[Callable[[State], Action], Callable[[State], bool]]:
+        """Sample and execute random initiable options until timeout."""
+
+        random_options_approach = RandomOptionsApproach(
```
ronuchit, line 247: initialize this in init? we don't need a new object each time

ronuchit: oh, the get_current_predicates() might change though

tests/approaches/test_interactive_approach.py
```diff
@@ -55,6 +55,21 @@ def test_interactive_learning_approach():
...
+    utils.update_config({
+        "interactive_action_strategy": "random",
+    })
+    interaction_requests = approach.get_interaction_requests()
+    _generate_interaction_results(env.simulate, teacher, train_tasks,
+                                  interaction_requests)
+    # Test that glib falls back to random if no solvable task can be found.
+    utils.update_config({
+        "interactive_action_strategy": "glib",
+        "timeout": 0.0,
```
ronuchit, line 68: how did this work if you're just catching approachfailure above, not approachtimeout? (see previous comment)

ronuchit (approved):

src/approaches/interactive_learning_approach.py
tomsilver: that gets caught within `_find_first_solvable`. this is only catching the last line of that method

src/approaches/interactive_learning_approach.py
tomsilver: right

tests/approaches/test_interactive_approach.py
tomsilver: what will happen is it times out for every attempted task, leading to a final approach failure

PR #408, opened 2022-02-02
empty dataset

tomsilver:


src/approaches/interactive_learning_approach.py
```diff
@@ -38,6 +38,9 @@ def _get_current_predicates(self) -> Set[Predicate]:
     ######################## Semi-supervised learning #########################
 
     def learn_from_offline_dataset(self, dataset: Dataset) -> None:
+        # Special case: empty offline dataset. Has no annotations.
+        if not dataset.trajectories:
+            return
```
ronuchit, line 43: what breaks if you don't do this?

tests/datasets/test_datasets.py
```diff
@@ -221,3 +221,15 @@ def test_ground_atom_dataset():
...
+def test_empty_dataset():
+    """Test creation of empty dataset."""
+    utils.reset_config({
+        "env": "cover",
+        "offline_data_method": "empty",
+    })
+    env = CoverEnv()
+    train_tasks = env.get_train_tasks()
+    dataset = create_dataset(env, train_tasks)
+    assert len(dataset.trajectories) == 0
```
ronuchit, line 235: also assert something about annotations?

src/approaches/interactive_learning_approach.py
ronuchit: oh, is it that dataset.annotations will be None?

ronuchit (approved):

amburger66 (approved):

src/approaches/interactive_learning_approach.py
tomsilver: yeah (see the comment)

ronuchit: oh i see, it's not None it'll actually just crash upon access. nvm

PR #412, opened 2022-02-02
specify max skeletons in expected node score function separately

tomsilver:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -733,7 +737,7 @@ def _evaluate_with_operators(self,
                 _, _, metrics = next(
                     task_plan(init_atoms, traj_goal, ground_nsrts,
                               reachable_atoms, heuristic, CFG.seed,
-                              CFG.grammar_search_task_planning_timeout))
+                              CFG.grammar_search_task_planning_timeout, 1))
```
ronuchit, line 740: do `max_skeletons=1` for clarity

src/approaches/grammar_search_invention_approach.py
```diff
@@ -1212,7 +1221,7 @@ def _task_planning_h(atoms: Set[GroundAtom]) -> float:
                 skeleton, atoms_sequence, _ = next(
                     task_plan(atoms, goal, ground_nsrts, reachable_atoms,
                               heuristic, CFG.seed,
-                              CFG.grammar_search_task_planning_timeout))
+                              CFG.grammar_search_task_planning_timeout, 1))
```
ronuchit, line 1224: `max_skeletons=1`

src/settings.py
```diff
@@ -153,6 +153,8 @@ class GlobalSettings:
     grammar_search_expected_nodes_backtracking_cost = 1e3
     grammar_search_expected_nodes_include_suspicious_score = False
     grammar_search_expected_nodes_allow_noops = True
+    # If None, defaults to CFG.max_skeletons_optimized.
+    grammar_search_max_skeletons_optimized = None
```
ronuchit, line 157: instead of None should we just do `grammar_search_max_skeletons_optimized = max_skeletons_optimized`? btw our parse_args stuff might not work with `--grammar_search_max_skeletons_optimized None`, you may want to test that (update `string_to_python_object`)

tomsilver: (resolved on slack: keeping it this way because otherwise, command line override of `max_skeletons_optimized` alone would not affect `grammar_search_max_skeletons_optimized`, which is confusing)

ronuchit (approved):
thx

PR #415, opened 2022-02-02
add assertion + test

amburger66:
Assert that all the predicates we want to learn about (aka the annotating predicates) are all "seen" (commented on by the teacher) in the demo+ground_atoms dataset, in preparation for almost-active-learning-only experiments where only a few training tasks are annotated but we make sure that the robot knows about all the predicates

### DIFF
--- a/tests/approaches/test_interactive_approach.py
+++ b/tests/approaches/test_interactive_approach.py
@@ -119,3 +119,10 @@ def test_interactive_learning_approach():
     with pytest.raises(NotImplementedError) as e:
         approach._score_atom_set(set(), train_tasks[0].init)  # pylint:disable=protected-access
     assert "Unrecognized interactive_score_function" in str(e)
+    # Test assertion that all predicates are seen in the data
+    utils.update_config({
+        "approach": "interactive_learning",
+        "teacher_dataset_label_ratio": 0.0,
+    })
+    with pytest.raises(AssertionError):
+        create_dataset(env, train_tasks)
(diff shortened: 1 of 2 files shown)


tests/approaches/test_interactive_approach.py
```diff
@@ -119,3 +119,10 @@ def test_interactive_learning_approach():
     with pytest.raises(NotImplementedError) as e:
         approach._score_atom_set(set(), train_tasks[0].init)  # pylint:disable=protected-access
     assert "Unrecognized interactive_score_function" in str(e)
+    # Test assertion that all predicates are seen in the data
+    utils.update_config({
+        "approach": "interactive_learning",
```
amburger66, line 124: idk if my brain's just not working, but this test wouldn't work without this line and i have no idea why

ronuchit:
general question, how are you all planning to handle the active-learning-only case? wouldn't the agent have to be given knowledge of the predicates from somewhere? if you do a new query type of the form "tell me all the atoms that hold in this state", then you don't actually need this assertion, right?

amburger66:
instead of making a new query type, we're planning to only annotate a very small number of the trajectories in the training tasks (enough to get a positive example of each predicate) using ratio=1.0. the rest of the tasks are ignored until active learning begins

tests/approaches/test_interactive_approach.py
tomsilver: yeah this is our fault I think 😬 I can explain what's going on if you're curious. but TLDR, this is good, and in general, `reset_config` is much preferred over `update_config` for tests now

PR #416, opened 2022-02-02
skeleton length analysis 

tomsilver:
[skeleton_analysis_results.zip](https://github.com/Learning-and-Intelligent-Systems/predicators/files/7995803/skeleton_analysis_results.zip)

examples:
expected_nodes_cover_regrasp_plot
expected_nodes_cover_regrasp_heat
num_nodes_cover_regrasp_plot
num_nodes_cover_regrasp_heat
skeleton_len_cover_regrasp_plot
skeleton_len_cover_regrasp_heat


analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+    outdir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                          "results")
+    os.makedirs(outdir, exist_ok=True)
+
+    for env_name in ENV_NAMES:
+        env_results = []
+        for seed in SEEDS:
+            seed_results = _run_analysis_for_env(env_name, seed)
+            env_results.append(seed_results)
+        env_results_arr = np.array(env_results, dtype=np.int32)
```
ronuchit, line 209: glad there's no weirdness when np.mean()'ing an int array... i would have thought it truncates & makes a new int array...

analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+        max_skeletons=max_skeletons,
+        timeout=timeout,
+        error_upper_bound=error_upper_bound)
+    all_results.append(initial_predicate_set_result)
+    for _ in range(num_candidates):
+        # Evaluate each possible next candidate predicate set and keep the
+        # best one.
+        best_next_predicate_set = set()
+        # This is just for type checking; it will get overwritten.
+        best_next_predicate_set_result = initial_predicate_set_result
```
ronuchit, line 65: is all this info really just for making the heatmap? for the graphs you just need the scores right?

analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+    plt.savefig(outfile)
+    print(f"Wrote out to {outfile}.")
+
+
+def _create_plot(env_results: NDArray[np.int32], env_name: str,
+                 outfile: str) -> None:
+    # Env results shape is (seed, predicate set, task, skeleton idx).
+    # Reorganize into array of shape (predicate set,) by minning over skeleton
+    # idx and averaging out seed and task.
+    arr = np.mean(np.min(env_results, axis=3), axis=(0, 2))  # type: ignore
```
ronuchit, line 184: i'm confused how score_result isn't needed here.... are you duplicating its functionality?

analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+    ax.set_ylabel("Skeleton Index")
+    fig.tight_layout()
+    plt.savefig(outfile)
+    print(f"Wrote out to {outfile}.")
+
+
+def _create_plot(env_results: NDArray[np.int32], env_name: str,
+                 outfile: str) -> None:
+    # Env results shape is (seed, predicate set, task, skeleton idx).
+    # Reorganize into array of shape (predicate set,) by minning over skeleton
```
ronuchit, line 182: i don't agree that this is a valid way to say "taking a min"

analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+                          max_skeletons: int = 8,
+                          timeout: float = 10,
+                          error_upper_bound: int = 100) -> NDArray[np.int32]:
+    # Create data for this environment and seed.
+    utils.reset_config({"seed": seed, "env": env_name})
+    env = create_env(env_name)
+    env.seed(seed)
+    train_tasks = env.get_train_tasks()
+    dataset = create_dataset(env, train_tasks)
+    assert [t.train_task_idx for t in dataset.trajectories] == \
```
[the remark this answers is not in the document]
tomsilver: I'm assuming that the dataset trajectories are in order of the train tasks when computing the demo lengths

analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+    strips_ops = [pnad.op for pnad in pnads]
+    option_specs = [pnad.option_spec for pnad in pnads]
+    # For each train task / demo, run task planning, and measure the error
+    # in skeleton length relative to the demos.
+    skeleton_length_errors = []  # shape (num tasks, max skeletons)
+    for train_task, demo_len in zip(train_tasks, demo_skeleton_lengths):
+        # Run task planning.
+        init_atoms = utils.abstract(train_task.init, current_predicate_set)
+        objects = set(train_task.init)
+        ground_nsrts, reachable_atoms = task_plan_grounding(
```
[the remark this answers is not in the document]
tomsilver: yeah lol sigh

analysis/skeleton_length_analysis.py
tomsilver: these are just scores, they're just not averaged / min'd out yet. you're right that if we just made the graphs we could average / min immediately

analysis/skeleton_length_analysis.py
tomsilver: you're right, minning is what minnie mouse does

analysis/skeleton_length_analysis.py
tomsilver: yeah good point, will fix

analysis/skeleton_length_analysis.py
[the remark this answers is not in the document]
ronuchit: but why? wouldn't it be better to just make demo_skeleton_lengths a map from train_task_idx to the length value? in case we ever change the mapping

analysis/skeleton_length_analysis.py
```diff
@@ -0,0 +1,219 @@
...
+    ax.set_ylabel("Skeleton Index")
+    fig.tight_layout()
+    plt.savefig(outfile)
+    print(f"Wrote out to {outfile}.")
+
+
+def _create_plot(env_results: NDArray[np.int32], env_name: str,
+                 outfile: str) -> None:
+    # Env results shape is (seed, predicate set, task, skeleton idx).
+    # Reorganize into array of shape (predicate set,) by taking a min over
```
ronuchit, line 182: "by scoring each seed's result and then averaging out seed."

analysis/skeleton_length_analysis.py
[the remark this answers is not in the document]
tomsilver: yeah nvm I don't need this assumption

ronuchit (approved):

tomsilver:
TODO: save raw results before plotting (e.g. to fix the cutoff in the painting heatmap y axis...)

tomsilver:
@ronuchit do you want to look at this again before i merge? if you prefer to avert your eyes, that could be better, even...

ronuchit:
i can avert my eyes if you promise i never have to modify any code in this monstrosity

tomsilver:
okay it's a deal

PR #419, opened 2022-02-03
fixes to cover_regrasp

### LINKED ISSUE #417: cover_regrasp sometimes times out in data generation
To reproduce:
`python src/main.py --env cover_regrasp --approach nsrt_learning --seed 3`

tomsilver:
closes #417 

I found two separate issues:
1. we were unwittingly checking collisions between a held block and itself when placing on table. that means that the sampler had to find a free space that was far enough away from the held block's original location, which was sometimes hard, leading to sampler-related failures in planning. I fixed the collision check, and also made the oracle placeontable sampler trivial so it just drops off the object whereever it starts out. (this always works by construction.)
2. in some of the harder problems requiring a plan of length 5 (place, pick, place, pick, place), the skeleton search was getting distracted by skeletons like: place the first object onto the second object's target, then place the second object onto that same target. in these cases, skeleton search never considered placeontable within the allotted max skeletons budget. this only happened rarely -- it required the problem being hard, and the ground_nsrts being sorted in a way that disfavored placeontable first. I addressed this by adding the clear predicate which we know gets invented anyway.

I tested by running oracle on 100 seeds and checking that they all got 50/50.

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -93,7 +93,8 @@ class CoverEnv(BaseEnv):
             if self._any_intersection(new_pose,
                                       state.get(held_block, "width"),
                                       state.data,
-                                      block_only=True):
+                                      block_only=True,
+                                      excluded_object=held_block):
                 return next_state
             # Only place if free space placing is allowed, or if we're
             # placing onto some target.
@@ -320,10 +321,13 @@ class CoverEnv(BaseEnv):
                           width: float,
                           data: Dict[Object, Array],
                           block_only: bool = False,
-                          larger_gap: bool = False) -> bool:
+                          larger_gap: bool = False,
+                          excluded_object: Optional[Object] = None) -> bool:
         mult = 1.5 if larger_gap else 0.5
         for other in data:
             if block_only and other not in self._blocks:
+                continue
+            if other == excluded_object:
                 continue
             other_feats = data[other]
             distance = abs(other_feats[3] - pose)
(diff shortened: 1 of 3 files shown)


src/envs/cover.py
```diff
@@ -401,13 +405,27 @@ class CoverEnvRegrasp(CoverEnv):
...
     Finally, to allow placing on the table, we need to change the allowed
     hand regions. We implement it so that there is a relatively small hand
     region centered at each target, but then everywhere else is allowed.
     """
     _allow_free_space_placing = True
     _initial_pick_offsets = [-0.95, 0.0, 0.95]
 
+    def __init__(self) -> None:
+        super().__init__()
+        # Add a Clear predicate to prevent attempts at placing on already
```
ronuchit, line 420: That's interesting. The lack of a Clear predicate was the reason why the base Cover env is not downward refinable right? (In combination with the fact that initial_holding_prob > 0).

src/ground_truth_nsrts.py
```diff
@@ -299,11 +303,12 @@ def place_sampler(state: State, rng: np.random.Generator,
...
                                    objs: Sequence[Object]) -> Array:
-            del state  # unused
+            # Always at the current location.
+            del rng  # this sampler is deterministic
             assert len(objs) == 1
-            lb = 0.0
-            ub = 1.0
-            return np.array(rng.uniform(lb, ub, size=(1, )), dtype=np.float32)
+            held_obj = objs[0]
+            x = state.get(held_obj, "pose") + state.get(held_obj, "grasp")
```
ronuchit, line 310: just checking that this shouldn't be a minus? Not sure but I always get them mixed up

ronuchit (approved):

src/envs/cover.py
tomsilver: hmm you're right that the base Cover env is not downward refinable, in that there are skeletons that able to be refinable but are not. however, unlike in CoverRegrasp, we would never actually try to refine a nonrefinable skeleton in the base env, because all of those skeletons would just place the respective object onto the respect target immediately.

src/ground_truth_nsrts.py
tomsilver: yeah thanks I checked

src/envs/cover.py
ronuchit: note: we realized on slack that ^ is not true. suppose your goal is just b0 on t0, and you start out holding b1. a nonrefinable skeleton is [place b1 on t0, place b0 on t0]. so, base Cover env is not downward refinable!

PR #420, opened 2022-02-03
tools environment

ronuchit:
tested that both oracle and nsrt_learning work well on supercloud

### DIFF
--- a/analysis/run_supercloud_experiments.sh
+++ b/analysis/run_supercloud_experiments.sh
@@ -1,14 +1,44 @@
 #!/bin/bash
 
 START_SEED=456
-NUM_SEEDS=50
+NUM_SEEDS=10
 FILE="analysis/submit.py"
 
 for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     COMMON_ARGS="--seed $SEED"
 
-    # tools
-    python $FILE --experiment_id oracleall --env tools --approach oracle $COMMON_ARGS
-    python $FILE --experiment_id learnall --env tools --approach nsrt_learning $COMMON_ARGS
-    python $FILE --experiment_id oraclesamplers --env tools --approach nsrt_learning $COMMON_ARGS --sampler_learner oracle
+    # cover
+    python $FILE --experiment_id cover_oracle --env cover --approach oracle $COMMON_ARGS
+    python $FILE --experiment_id cover_noinvent_noexclude --env cover --approach nsrt_learning $COMMON_ARGS
+    python $FILE --experiment_id cover_noinvent_allexclude --env cover --approach nsrt_learning --excluded_predicates all $COMMON_ARGS
+    python $FILE --experiment_id cover_invent_noexclude --env cover --approach grammar_search_invention $COMMON_ARGS
+    python $FILE --experiment_id cover_invent_allexclude --env cover --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
+
+    # blocks
+    python $FILE --experiment_id blocks_oracle --env blocks --approach oracle $COMMON_ARGS
+    python $FILE --experiment_id blocks_noinvent_noexclude --env blocks --approach nsrt_learning $COMMON_ARGS
+    python $FILE --experiment_id blocks_noinvent_allexclude --env blocks --approach nsrt_learning --excluded_predicates all $COMMON_ARGS
+    python $FILE --experiment_id blocks_invent_noexclude --env blocks --approach grammar_search_invention $COMMON_ARGS
+    python $FILE --experiment_id blocks_invent_allexclude --env blocks --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
+
+    # painting
+    python $FILE --experiment_id painting_oracle --env painting --approach oracle $COMMON_ARGS
+    python $FILE --experiment_id painting_noinvent_noexclude --env painting --approach nsrt_learning $COMMON_ARGS
+    python $FILE --experiment_id painting_noinvent_allexclude --env painting --approach nsrt_learning --excluded_predicates all $COMMON_ARGS
+    python $FILE --experiment_id painting_invent_noexclude --env painting --approach grammar_search_invention $COMMON_ARGS
+    python $FILE --experiment_id painting_invent_allexclude --env painting --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
+
+    # painting with lid always open
+    python $FILE --experiment_id painting_always_open_oracle --painting_lid_open_prob 1.0 --env painting --approach oracle $COMMON_ARGS
+    python $FILE --experiment_id painting_always_open_noinvent_noexclude --painting_lid_open_prob 1.0 --env painting --approach nsrt_learning $COMMON_ARGS
+    python $FILE --experiment_id painting_always_open_noinvent_allexclude --painting_lid_open_prob 1.0 --env painting --approach nsrt_learning --excluded_predicates all $COMMON_ARGS
+    python $FILE --experiment_id painting_always_open_invent_noexclude --painting_lid_open_prob 1.0 --env painting --approach grammar_search_invention $COMMON_ARGS
+    python $FILE --experiment_id painting_always_open_invent_allexclude --painting_lid_open_prob 1.0 --env painting --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
+
+    # cluttered_table
+    python $FILE --experiment_id ctable_oracle --env cluttered_table --approach oracle $COMMON_ARGS
+    python $FILE --experiment_id ctable_noinvent_noexclude --env cluttered_table --approach nsrt_learning $COMMON_ARGS
+    python $FILE --experiment_id ctable_noinvent_allexclude --env cluttered_table --approach nsrt_learning --excluded_predicates all $COMMON_ARGS
+    python $FILE --experiment_id ctable_invent_noexclude --env cluttered_table --approach grammar_search_invention $COMMON_ARGS
+    python $FILE --experiment_id ctable_invent_allexclude --env cluttered_table --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
 done
--- a/analysis/run_supercloud_experiments.sh
+++ b/analysis/run_supercloud_experiments.sh
@@ -8,7 +8,7 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     COMMON_ARGS="--seed $SEED"
 
     # tools
-    python $FILE --env tools --approach oracle $COMMON_ARGS
-    python $FILE --env tools --approach nsrt_learning $COMMON_ARGS
-    python $FILE --env tools --approach nsrt_learning $COMMON_ARGS --sampler_learner oracle
+    python $FILE --experiment_id oracleall --env tools --approach oracle $COMMON_ARGS
+    python $FILE --experiment_id learnall --env tools --approach nsrt_learning $COMMON_ARGS
+    python $FILE --experiment_id oraclesamplers --env tools --approach nsrt_learning $COMMON_ARGS --sampler_learner oracle
 done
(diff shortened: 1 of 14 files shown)


src/ground_truth_nsrts.py
```diff
@@ -816,6 +818,341 @@ def placeontable_sampler(state: State, rng: np.random.Generator,
...
+
+    PickScrew, PickScrewdriver, PickNail, PickHammer, PickBolt, PickWrench, \
+        Place, FastenScrewWithScrewdriver, FastenScrewByHand, \
+        FastenNailWithHammer, FastenBoltWithWrench = _get_options_by_names(
+            "tools", ["PickScrew", "PickScrewdriver", "PickNail", "PickHammer",
+                      "PickBolt", "PickWrench", "Place",
+                      "FastenScrewWithScrewdriver", "FastenScrewByHand",
+                      "FastenNailWithHammer", "FastenBoltWithWrench"])
+
+    def null_sampler(state: State, rng: np.random.Generator,
```
tomsilver, line 849: move to utils?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+        self._PickWrench = ParameterizedOption(
+            # variables: [robot, wrench to pick]
+            # params: []
+            "PickWrench",
+            types=[self._robot_type, self._wrench_type],
+            params_space=Box(0, 1, (0, )),  # no parameters
+            _policy=self._Pick_policy,
+            _initiable=utils.always_initiable,
+            _terminal=utils.onestep_terminal)
+        self._Place = ParameterizedOption(
```
tomsilver, line 156: why do we need a million pick options but only one place option?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                return next_state
+            on_table = self._get_contraption_pose_is_on(state, x, y) is None
+            if self._is_tool(held) and not on_table:
+                # Failure: cannot place a tool on a contraption
+                return next_state
+            next_state.set(held, "is_held", 0.0)
+            next_state.set(held, "pose_x", x)
+            next_state.set(held, "pose_y", y)
+            if self._is_item(held):
+                next_state.set(held, "on_table", float(int(on_table)))
```
tomsilver, line 233: is just `float(on_table)` not sufficient?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+            _policy=self._Fasten_policy,
+            _initiable=utils.always_initiable,
+            _terminal=utils.onestep_terminal)
+        # Objects
+        self._robot = Object("robby", self._robot_type)
+
+    def simulate(self, state: State, action: Action) -> State:
+        assert self.action_space.contains(action.arr)
+        next_state = state.copy()
+        x, y, is_pick, is_place = action.arr
```
tomsilver, line 218: so this is going to do something interesting only when is_pick is True, is_place is True, or both are false, which corresponds to a fasten. right? do you want to `assert not (is_pick > 0.5 and is_place > 0.5)` for clarity?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                # Failure: cannot place a tool on a contraption
+                return next_state
+            next_state.set(held, "is_held", 0.0)
+            next_state.set(held, "pose_x", x)
+            next_state.set(held, "pose_y", y)
+            if self._is_item(held):
+                next_state.set(held, "on_table", float(int(on_table)))
+            next_state.set(self._robot, "fingers", 1.0)
+            return next_state
+        target = self._get_closest_object(state, x, y)
```
tomsilver, line 236: can we rename this to make it more clear somehow that a contraption is not eligible to be returned?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+        # Note: we don't update the pose of an object when it is
+        # picked. This is a sort of hack that provides "memory",
+        # so that when placing tools down, there is always the
+        # easy choice of placing it back where you got it from.
+        # The oracle sampler makes use of this.
+        next_state.set(target, "is_held", 1.0)
+        if self._is_item(target):
+            next_state.set(target, "on_table", 0.0)
+        next_state.set(self._robot, "fingers", 0.0)
+        return next_state
```
tomsilver, line 286: okay I am now feeling confident that we should _not_ try to pybullet this lol

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                           abs(data[i][1] - pose_y) < self.close_thresh
+                           for i in items):
+                        continue
+                    # Otherwise, we found a valid pose
+                    break
+                on_table = 1.0  # always start off on table
+                is_fastened = 0.0  # always start off not fastened
+                is_held = 0.0  # always start off not held
+                choices = ["screw", "nail", "bolt"]
+                if screw_created:
```
tomsilver, line 409: can't we just check `screw_cnt`?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                   num_contraptions_lst: List[int],
+                   rng: np.random.Generator) -> List[Task]:
+        tasks = []
+        for i in range(num_tasks):
+            num_items = num_items_lst[i % len(num_items_lst)]
+            num_contraptions = num_contraptions_lst[i %
+                                                    len(num_contraptions_lst)]
+            data = {}
+            # Initialize robot
+            data[self._robot] = np.array([1.0])  # fingers start off open
```
tomsilver, line 361: add dtype=np.float32? idk if necessary but it's used everywhere else

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                    bolt_cnt += 1
+                    feats = [pose_x, pose_y, on_table, is_fastened, is_held]
+                    goal.add(GroundAtom(self._BoltFastened, [item]))
+                    goal.add(
+                        GroundAtom(self._BoltPlaced, [item, goal_contraption]))
+                items.append(item)
+                data[item] = np.array(feats, dtype=np.float32)
+            # Initialize tools (screwdrivers, hammers, wrenches)
+            # We will always generate the same number of tools:
+            # 3 screwdrivers (two graspable, one not)
```
tomsilver, line 444: convert these to constants to avoid magic numbers?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                    pose_y = rng.uniform(self.table_ly, self.table_uy)
+                    # If collision with any contraption, try again
+                    if any(data[c][0] < pose_x < data[c][2] and \
+                           data[c][1] < pose_y < data[c][3]
+                           for c in contraptions):
+                        continue
+                    # If collision with any item, try again
+                    if any(abs(data[i][0] - pose_x) < self.close_thresh and \
+                           abs(data[i][1] - pose_y) < self.close_thresh
+                           for i in items):
```
tomsilver, line 466: combine this check and the tool one?

src/envs/tools.py
```diff
@@ -0,0 +1,643 @@
...
+                        continue
+                    # If collision with any tool, try again
+                    if any(abs(data[t][0] - pose_x) < self.close_thresh and \
+                           abs(data[t][1] - pose_y) < self.close_thresh
+                           for t in tools):
+                        continue
+                    # Otherwise, we found a valid pose
+                    break
+                is_held = 0.0  # always start off not held
+                if j < 3:
```
tomsilver, line 476: these magic numbers especially would be good to get rid of

tomsilver (commented):
great stuff, i made it about halfway through, will continue later

src/envs/tools.py
ronuchit: because the pick option's policy has to know the object so that we can use the object's pose in the action arguments, but then the objects all have different types. alternative solution is hierarchical typing

src/envs/tools.py
ronuchit: i'll make it a no-op instead of an assert, since an agent could plausibly pass in such an action (e.g., `random_actions_approach`). in general, asserts in simulate() should only occur when there's a literally impossible situation under the dynamics

src/envs/tools.py
ronuchit: yes thanks, the code went through a lot of revisions...

src/envs/tools.py
ronuchit: actually i copied this from blocks/painting where it's not used there lol. but i just fixed those too

src/envs/tools.py
tomsilver: got it, i assumed it was for a hacky operator-learning reason but forgot there could be a legitimate reason lol

src/envs/tools.py
```diff
@@ -0,0 +1,646 @@
...
+        # Works for any item
+        item, = objects
+        return state.get(item, "on_table") > 0.5
+
+    def _Placed_holds(self, state: State, objects: Sequence[Object]) -> bool:
+        # Works for any item
+        item, contraption = objects
+        pose_x = state.get(item, "pose_x")
+        pose_y = state.get(item, "pose_y")
+        return self._is_pose_on_contraption(state, pose_x, pose_y, contraption)
```
tomsilver, line 517: we should be able to learn this as two different predicates in our current grammar, right?

src/envs/tools.py
```diff
@@ -0,0 +1,646 @@
...
+
+    def _is_item(self, obj: Object) -> bool:
+        return obj.type in (self._screw_type, self._nail_type, self._bolt_type)
+
+    def _is_screwdriver_or_hammer(self, obj: Object) -> bool:
+        return obj.type in (self._screwdriver_type, self._hammer_type)
+
+    def _get_best_screwdriver_or_none(self, state: State,
+                                      screw: Object) -> Optional[Object]:
+        """Use the shape of the given screw to figure out the best graspable
```
tomsilver, line 626: no action required, just want to say this so i remember: it's counterintuitive that whether a screw is fastenable by hand depends on properties of the other screwdrivers

tomsilver (commented):
I find no other complaints at this time

src/envs/tools.py
ronuchit: yeah it's just a way of ensuring we don't have tons of low-level features lol. you can think of it like "if i have the right screwdriver for the job, i'll use that, but otherwise i'll use my hands like an ape"

src/envs/tools.py
ronuchit: 4 because of the lower and upper bounds, but yeah theoretically... it's a goal predicate though so we don't have to worry about this

analysis/run_supercloud_experiments.sh
```diff
@@ -1,44 +1,14 @@
...
-    python $FILE --experiment_id painting_always_open_invent_noexclude --painting_lid_open_prob 1.0 --env painting --approach grammar_search_invention $COMMON_ARGS
-    python $FILE --experiment_id painting_always_open_invent_allexclude --painting_lid_open_prob 1.0 --env painting --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
-
-    # cluttered_table
-    python $FILE --experiment_id ctable_oracle --env cluttered_table --approach oracle $COMMON_ARGS
-    python $FILE --experiment_id ctable_noinvent_noexclude --env cluttered_table --approach nsrt_learning $COMMON_ARGS
-    python $FILE --experiment_id ctable_noinvent_allexclude --env cluttered_table --approach nsrt_learning --excluded_predicates all $COMMON_ARGS
-    python $FILE --experiment_id ctable_invent_noexclude --env cluttered_table --approach grammar_search_invention $COMMON_ARGS
-    python $FILE --experiment_id ctable_invent_allexclude --env cluttered_table --approach grammar_search_invention --excluded_predicates all $COMMON_ARGS
+    # tools
```
tomsilver, line 10: Are you intending to commit this file?

tomsilver (approved):
Fantastimo

analysis/run_supercloud_experiments.sh
ronuchit: nope, will revert it now

PR #423, opened 2022-02-04
fix bug where placing can could collide with its past position

czeng6:
Basically ignore the can we're placing when checking for collisions.


tomsilver (approved):
LGTM thank you!

PR #424, opened 2022-02-04
simplify tools env and add in DebugGrammar

ronuchit:
simplifications to env:
* remove OnTable predicates and on_table low-level features
* remove contraption upper bound features, since size is fixed

confirmed that DebugGrammar predicates are all included in our full
grammar with `grammar_search_max_predicates 200`

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -157,8 +157,6 @@ class GlobalSettings:
     grammar_search_expected_nodes_backtracking_cost = 1e3
     grammar_search_expected_nodes_include_suspicious_score = False
     grammar_search_expected_nodes_allow_noops = True
-    # If None, defaults to CFG.max_skeletons_optimized.
-    grammar_search_expected_nodes_max_skeletons = None
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


src/settings.py
```diff
@@ -223,6 +221,15 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
...
                     "tools": 1,
                 })[args.get("env", "")],
+
+            # Maximum number of skeletons used by ExpectedNodesScoreFunction.
+            # If -1, defaults to CFG.max_skeletons_optimized.
+            grammar_search_expected_nodes_max_skeletons=defaultdict(
+                lambda: -1,
+                {
+                    # For the tools environment, keep it much lower.
+                    "tools": 1,
```
tomsilver, line 231: should we also try 8 for consistency?

tomsilver (approved):

src/settings.py
ronuchit: currently experimenting with this

PR #425, opened 2022-02-04
goal-conditioned sampler API update

### LINKED ISSUE #422: allow samplers to be goal-conditioned

tomsilver:
closes #422

### DIFF
--- a/src/datasets/demo_replay.py
+++ b/src/datasets/demo_replay.py
@@ -55,6 +55,7 @@ def create_demo_replay_data(env: BaseEnv,
         # Sample a trajectory
         traj_idx = rng.choice(len(demo_dataset.trajectories), p=weights)
         traj = demo_dataset.trajectories[traj_idx]
+        goal = train_tasks[traj.train_task_idx].goal
         # Sample a state
         # We don't allow sampling the final state in the trajectory here,
         # because there's no guarantee that an initiable option exists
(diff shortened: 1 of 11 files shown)


src/datasets/demo_replay.py
```diff
@@ -55,6 +55,7 @@ def create_demo_replay_data(env: BaseEnv,
         # Sample a trajectory
         traj_idx = rng.choice(len(demo_dataset.trajectories), p=weights)
         traj = demo_dataset.trajectories[traj_idx]
+        goal = train_tasks[traj.train_task_idx].goal
```
ronuchit, line 58: lol i would have definitely messed up here and used `train_tasks[traj_idx]` which would have worked for now but possibly caused horrible bugs later in life

src/ground_truth_nsrts.py
```diff
@@ -400,7 +411,7 @@ def grasp_sampler(state: State, rng: np.random.Generator,
         }
         add_effects = {LiftedAtom(HandEmpty, [])}
         delete_effects = {LiftedAtom(Holding, [can])}
-        place_sampler = lambda s, r, o: np.array([
+        place_sampler = lambda s, g, r, o: np.array([
```
ronuchit, line 414: wasn't the goal of this PR to make a goal-conditioned version of the place sampler? is that coming in a separate PR?

ronuchit (approved):
whew, this is an awesome and large change! great stuff.

src/ground_truth_nsrts.py
tomsilver: yeah that will happen separately

PR #431, opened 2022-02-06
Constrained cluttered place

czeng6:
Further constraining cluttered place environment to make task more difficult + goal-conditioned action sampling that works. 

python src/main.py --env cluttered_table_place --approach oracle --max_skeletons_optimized 2 --max_samples_per_step 1 --num_test_tasks 50 --cluttered_table_num_cans_test 2 --cluttered_table_place_goal_conditioned_sampling False  --seed 0
 
solves 9/50  tasks compared to 

 python src/main.py --env cluttered_table_place --approach oracle --max_skeletons_optimized 2 --max_samples_per_step 1 --num_test_tasks 50 --cluttered_table_num_cans_test 2 --cluttered_table_place_goal_conditioned_sampling True  --seed 0 

solves 50/50 tasks

### DIFF
--- a/src/ground_truth_nsrts.py
+++ b/src/ground_truth_nsrts.py
@@ -414,6 +414,7 @@ def _get_cluttered_table_gt_nsrts(with_place: bool = False) -> Set[NSRT]:
         def place_sampler(state: State, goal: Set[GroundAtom],
                           rng: np.random.Generator,
                           objs: Sequence[Object]) -> Array:
+            print("sample in place sampler")
             start_x, start_y = 0.2, 0.0
             # Goal-conditioned sampling
             if CFG.cluttered_table_place_goal_conditioned_sampling:
@@ -438,7 +439,7 @@ def _get_cluttered_table_gt_nsrts(with_place: bool = False) -> Set[NSRT]:
             return np.array(
                 [start_x, start_y,
                  rng.uniform(0, 0.4),
-                 rng.uniform(0, 0.4)],
+                 rng.uniform(0, 1.0)],
                 dtype=np.float32)
 
         place_nsrt = NSRT("Place", parameters, preconditions, add_effects,
--- a/src/ground_truth_nsrts.py
+++ b/src/ground_truth_nsrts.py
@@ -367,8 +367,7 @@ def _get_cluttered_table_gt_nsrts(with_place: bool = False) -> Set[NSRT]:
         end_x = max(0.0, state.get(can, "pose_x"))
         end_y = max(0.0, state.get(can, "pose_y"))
         if with_place:
-            start_x, start_y = rng.uniform(0.0, 0.2,
-                                           size=2)  # start from 0.2x0.2 corner
+            start_x, start_y = 0.2, 0
         else:
             start_x, start_y = rng.uniform(0.0, 1.0,
                                            size=2)  # start from anywhere
--- a/src/ground_truth_nsrts.py
+++ b/src/ground_truth_nsrts.py
@@ -414,7 +414,6 @@ def _get_cluttered_table_gt_nsrts(with_place: bool = False) -> Set[NSRT]:
         def place_sampler(state: State, goal: Set[GroundAtom],
                           rng: np.random.Generator,
                           objs: Sequence[Object]) -> Array:
-            print("sample in place sampler")
             start_x, start_y = 0.2, 0.0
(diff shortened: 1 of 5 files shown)


src/ground_truth_nsrts.py
```diff
@@ -411,12 +410,38 @@ def grasp_sampler(state: State, goal: Set[GroundAtom],
...
-            r.uniform(0, 0.2),
-            r.uniform(0, 0.2),
-            r.uniform(0, 1.),
-            r.uniform(0, 1.)
-        ])
+
+        def place_sampler(state: State, goal: Set[GroundAtom],
+                          rng: np.random.Generator,
+                          objs: Sequence[Object]) -> Array:
+            print("sample in place sampler")
```
tomsilver, line 417: let's remove this print

tests/approaches/test_oracle_approach.py
```diff
@@ -388,15 +399,19 @@ def test_cluttered_table_get_gt_nsrts(place_version=False):
             assert env.action_space.contains(dump_action.arr)
             env.simulate(state, dump_action)  # never raises EnvironmentFailure
         else:
-            place3_nsrt = place_nsrt.ground([can3])
+            print("hihi again")
```
tomsilver, line 402: hihi! plz remove :)

tests/approaches/test_oracle_approach.py
```diff
@@ -388,15 +399,19 @@ def test_cluttered_table_get_gt_nsrts(place_version=False):
...
+
+            print("hihi again 2!")
+            place_option = place1_nsrt.sample_option(state, task.goal, rng)
+            print("hihi again 3!")
             place_action = place_option.policy(state)
             assert env.action_space.contains(place_action.arr)
             try:
                 env.simulate(state, place_action)
-            except utils.EnvironmentFailure as e:
+            except utils.EnvironmentFailure as e:  # pragma: no cover
```
tomsilver, line 414: let's try to see if we can cover this case (and remove the pragma: no cover). the EnvironmentFailure will get raised when there's a collision upon placing. if you want, you can manually create the `place_action` to force such a collision (by extracting the objects and state from `state`, seeing where the goal object's pose is). or, you might be able to just change the code above to something like:
```
train_tasks = env.get_train_tasks()
...
if i < len(train_tasks) / 2:
            utils.reset_config(
                {"cluttered_table_place_goal_conditioned_sampling": False})
        else:
            utils.reset_config(
                {"cluttered_table_place_goal_conditioned_sampling": True})
```
because I imagine that collisions are more likely with the non-goal conditioned sampler.

tests/envs/test_cluttered_table.py
```diff
@@ -44,6 +44,7 @@ def test_cluttered_table(place_version=False):
     # Test init state and simulate()
     for i, task in enumerate(env.get_test_tasks()):
         state = task.init
+        print("hihi")
```
tomsilver, line 47: make sure you catch all of these!

tomsilver (commented):
looking good! a few tiny things and one minor thing. can you also update the PR comment to include the two command line prompts, one for the goal-conditioned sampler and the other for non-goal-conditioned sampler, and report the expected outputs (just the number of test tasks solved)? thanks!

ronuchit:
no further comments beyond @tomsilver's! but make sure you catch all the print statements, an easy way is to go to https://github.com/Learning-and-Intelligent-Systems/predicators/pull/431/files and ctrl-f "print"

tests/approaches/test_oracle_approach.py
czeng6: ok, I was following some other patterns in the code where I noticed we didn't necessarily cover exceptions. through some careful if statements this case  is now covered...

tests/approaches/test_oracle_approach.py
czeng6: oops :C

src/ground_truth_nsrts.py
czeng6: +1

tests/envs/test_cluttered_table.py
czeng6: my bad

czeng6:
ok, I believe all comments have been addressed! Please take another look when you get the chance :)

tomsilver (approved):
LGTM, great work!

PR #432, opened 2022-02-06
online learning curve script

tomsilver:
@amburger66 I'm hoping this can be useful for you too!

example outputs (don't worry about what the content means, probably garbage):

blocks_cycle_avg_nodes_created
cover_(regrasp)_cycle_avg_nodes_created
blocks_learning_time_num_solved
cover_(regrasp)_learning_time_num_solved

### DIFF
--- a/analysis/analyze_results_directory.py
+++ b/analysis/analyze_results_directory.py
@@ -1,6 +1,7 @@
 """Script to analyze experiments resulting from running the script
 analysis/run_supercloud_experiments.sh."""
 
+from typing import Tuple
 import glob
 import dill as pkl
 import numpy as np
@@ -8,8 +9,8 @@ import pandas as pd
 from predicators.src.settings import CFG
 
 
-def _main() -> None:
-    # Gather data.
+def create_dataframes() -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+    # Returns means standard deviations, and sizes.
     all_data = []
     column_names = [
         "ENV", "APPROACH", "EXCLUDED_PREDICATES", "EXPERIMENT_ID", "SEED",
@@ -87,6 +88,11 @@ def _main() -> None:
     means = grouped.mean()
     stds = grouped.std(ddof=0)
     sizes = grouped.size()
+    return means, stds, sizes
+
+
+def _main() -> None:
+    means, stds, sizes = create_dataframes()
     # Add standard deviations to the printout.
     for col in means:
         for row in means[col].keys():
--- a/analysis/analyze_results_directory.py
+++ b/analysis/analyze_results_directory.py
@@ -10,7 +10,7 @@ from predicators.src.settings import CFG
 
 
 def create_dataframes() -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
-    # Returns means standard deviations, and sizes.
+    """Returns means standard deviations, and sizes."""
     all_data = []
     column_names = [
         "ENV", "APPROACH", "EXCLUDED_PREDICATES", "EXPERIMENT_ID", "SEED",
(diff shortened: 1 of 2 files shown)


analysis/analyze_results_directory.py
```diff
@@ -1,15 +1,16 @@
...
 import dill as pkl
 import numpy as np
 import pandas as pd
 from predicators.src.settings import CFG
 
 
-def _main() -> None:
-    # Gather data.
+def create_dataframes() -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+    """Returns means standard deviations, and sizes."""
```
ronuchit, line 13: add comma

analysis/create_online_learning_curves.py
```diff
@@ -0,0 +1,94 @@
...
+from predicators.analysis.analyze_results_directory import create_dataframes
+
+pd.options.mode.chained_assignment = None  # default='warn'
+
+############################ Change below here ################################
+
+# X_KEY is the name of the metric that will be plotted on the x axis. See
+# analyze_results_directory.py for all available metrics.
+# X_LABEL is used to label the x axis.
+X_KEY = "CYCLE"  # common: CYCLE, NUM_TRANSITIONS, LEARNING_TIME
```
ronuchit, line 17: is there any hope of making this script work for our offline learning experiments where we want to vary num_demos? it would be nice to have a single script since it's really the same plot... (performance vs data)

analysis/create_online_learning_curves.py
```diff
@@ -0,0 +1,94 @@
...
+X_LABEL = "Num online learning episodes"
+
+# Y_KEY is the name of the metric that will be plotted on the y axes. See
+# analyze_results_directory.py for all available metrics.
+# Y_LABEL is used to label the y axis.
+Y_KEY = "AVG_NODES_CREATED"  # common: NUM_SOLVED, AVG_NODES_CREATED
```
ronuchit, line 23: could you refactor these key/label things to be tuples? and instead of writing "common" like this, just make a list where we can comment out lines, and loop over all elements of that list. so for instance...
```
X_KEY_AND_LABEL = [
  ("CYCLE", "Num online learning episodes"),
  ("NUM_TRANSITIONS", "Num transitions"),
  # ("LEARNING_TIME", "Learning time in seconds"),
]
```

which should generate plots for both of the first two rows (but not the third because it's commented out)

analysis/create_online_learning_curves.py
```diff
@@ -0,0 +1,94 @@
...
+Y_LABEL = "Average nodes created"  #"Test tasks solved"
```
ronuchit, line 24: add space after `#` (may be irrelevant in light of other comments)

analysis/create_online_learning_curves.py
```diff
@@ -0,0 +1,94 @@
...
+#################### Should not need to change below here #####################
+
+
+def _get_df_for_experiment_id(df: pd.DataFrame,
+                              experiment_id: str) -> pd.DataFrame:
+    df = df[df["EXPERIMENT_ID"] == experiment_id]
+    # Handle CYCLE as a special case, since the offline learning phase is
+    # logged as None. Note that we shift everything by 1 so the first data
+    # point is 0, meaning 0 online learning cycles have happened so far.
+    df["CYCLE"].replace("None", "-1", inplace=True)
```
ronuchit, line 51: aha, so you would have liked -1 all along!! :P

ronuchit (commented):
some suggestions

amburger66 (approved):

analysis/create_online_learning_curves.py
tomsilver: I think that if we include NUM_DEMOs as a metric then the script should just work? If it doesn't, I will fix it once we're there.

ronuchit (approved):

PR #433, opened 2022-02-06
mark demo responses as demos with train task idxs

tomsilver:
demonstration responses should be marked as demos. we also will need to know at least the goal that the trajectories are for. one thing to be aware of in general is: now, if a LowLevelTrajectory has train_task_idx 5, it does _not_ mean that the trajectory is a full demonstration of the entire train_task_idx 5; it just means that the trajectory ends up in the goal for that train task, starting at a potentially arbitrary state. if this is too confusing, we may need to make `is_demo`, `train_task_idx`, and `goal` three different properties for `LowLevelTrajectory`. @ronuchit lmk what you think -- I am neutral about whether we should do that.

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -1124,8 +1124,7 @@ class GroundAtomsHoldResponse(Response):
 
 @dataclass(frozen=True, eq=False, repr=False)
 class DemonstrationQuery(Query):
-    """A query requesting a demonstration to get from the state to a goal."""
-    goal: Set[GroundAtom]
+    """A query requesting a demonstration to finish a train task."""
     train_task_idx: int
 
 
--- a/src/structs.py
+++ b/src/structs.py
@@ -1126,6 +1126,7 @@ class GroundAtomsHoldResponse(Response):
 class DemonstrationQuery(Query):
     """A query requesting a demonstration to get from the state to a goal."""
     goal: Set[GroundAtom]
+    train_task_idx: int
 
 
 @dataclass(frozen=True, eq=False, repr=False)
--- a/src/teacher.py
+++ b/src/teacher.py
@@ -4,7 +4,7 @@ information to assist an agent during online learning."""
 from __future__ import annotations
 from predicators.src.structs import State, Task, Query, Response, \
     GroundAtomsHoldQuery, GroundAtomsHoldResponse, DemonstrationQuery, \
-    DemonstrationResponse
+    DemonstrationResponse, LowLevelTrajectory
 from predicators.src.settings import CFG, get_allowed_query_type_names
 from predicators.src.envs import create_env
 from predicators.src.approaches import ApproachTimeout, ApproachFailure
@@ -51,7 +51,11 @@ class Teacher:
         except (ApproachTimeout, ApproachFailure):
             return DemonstrationResponse(query, teacher_traj=None)
 
-        teacher_traj, _, goal_reached = utils.run_policy_on_task(
+        traj, _, goal_reached = utils.run_policy_on_task(
             policy, task, self._simulator, CFG.max_num_steps_option_rollout)
         assert goal_reached
+        teacher_traj = LowLevelTrajectory(traj.states,
+                                          traj.actions,
+                                          _is_demo=True,
+                                          _train_task_idx=query.train_task_idx)
         return DemonstrationResponse(query, teacher_traj)
--- a/src/teacher.py
+++ b/src/teacher.py
@@ -2,6 +2,7 @@
 information to assist an agent during online learning."""
 
 from __future__ import annotations
+from typing import Sequence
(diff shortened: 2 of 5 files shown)


tests/test_teacher.py
```diff
@@ -56,17 +56,20 @@ def test_DemonstrationQuery():
...
     # Test normal usage
-    query = DemonstrationQuery(goal)
+    query = DemonstrationQuery(goal, train_task_idx)
     response = teacher.answer_query(state, query)
     assert isinstance(response, DemonstrationResponse)
     assert response.query is query
     assert isinstance(response.teacher_traj, LowLevelTrajectory)
     assert len(response.teacher_traj.actions) == 2
     assert all(atom.holds(response.teacher_traj.states[-1]) for atom in goal)
+    assert response.teacher_traj.is_demo
```
tomsilver, line 71: This is the assertion that matters -- the other changes in this PR are just to make this assertion pass

NishanthJKumar (commented):
LGTM! I'll let @ronuchit give the actual approval.

I agree that it's important to make clear that `DemonstrationResponses` are not the same as demonstrations for a task. Given the paradigm that @ashayathalye wants to use these in (i.e, interactively), I think it should be fine to not have a separate bit like `is_full_demo_for_task` or something, since I don't see why we might ever mix full demos with `DemonstrationResponses`.

src/structs.py
```diff
@@ -1126,6 +1126,7 @@ class GroundAtomsHoldResponse(Response):
 class DemonstrationQuery(Query):
     """A query requesting a demonstration to get from the state to a goal."""
     goal: Set[GroundAtom]
+    train_task_idx: int
```
ronuchit, line 1129: should we initialize the teacher with the training tasks instead, so that you only have to give `train_task_idx` here? i guess that's less explicit so maybe not?

src/teacher.py
```diff
@@ -51,7 +51,11 @@ def _answer_Demonstration_query(
         except (ApproachTimeout, ApproachFailure):
             return DemonstrationResponse(query, teacher_traj=None)
 
-        teacher_traj, _, goal_reached = utils.run_policy_on_task(
+        traj, _, goal_reached = utils.run_policy_on_task(
             policy, task, self._simulator, CFG.max_num_steps_option_rollout)
         assert goal_reached
+        teacher_traj = LowLevelTrajectory(traj.states,
+                                          traj.actions,
+                                          _is_demo=True,
```
ronuchit, line 59: to answer your comment on the PR, i think this is fine. anyone bold enough to use DemonstrationQuery/Response must be willing to accept the semantic consequences.

ronuchit (approved):
approving to unblock, but let's decide on my question before merging

src/structs.py
ronuchit: OTOH, it's slightly weird to give both of these fields...

tomsilver: yeah I think the weirdness of having both fields outweighs the explicitness concern. I'll try it that way

tomsilver:
@ronuchit @NishanthJKumar ready for another look

src/teacher.py
```diff
@@ -15,7 +16,8 @@
 class Teacher:
     """The teacher can respond to queries of various types."""
 
-    def __init__(self) -> None:
+    def __init__(self, train_tasks: Sequence[Task]) -> None:
+        self._train_tasks = train_tasks
```
ronuchit, line 20: it's funny how we've decided what the teacher gets in the init versus has to create for itself via `create_env` and the ensuing lines...

ronuchit: (not sure if any action item required, but it is weird)

tests/test_teacher.py
```diff
@@ -54,33 +55,40 @@ def test_GroundAtomsHold():
...
     response = teacher.answer_query(response.teacher_traj.states[-1], query)
     assert isinstance(response, DemonstrationResponse)
     assert response.query is query
     assert isinstance(response.teacher_traj, LowLevelTrajectory)
     assert len(response.teacher_traj.actions) == 0
     # Test usage when achieving goal is impossible
-    block_type = [t for t in env.types if t.name == "block"][0]
-    block = block_type("block0")
+    state = train_tasks[0].init
+    block = [b for b in state if b.name == "block1"][0]
```
ronuchit, line 83: does it matter that this was block0 before but block1 now?

ronuchit (approved):

src/teacher.py
tomsilver: I think it's okay. It's not "cheating" for the teacher to know the train tasks, vs knowing env.simulate, since the agent has train tasks as well.

tests/test_teacher.py
tomsilver: no, but i changed it back to reduce entropy in the cosmos

PR #434, opened 2022-02-06
max initial demos

tomsilver:
for online learning methods, it makes sense to have the number of initial demos and the number of train tasks be two different things, because the method can collect its own data in train tasks that it didn't receive a demo for.


ronuchit (approved):
lgtm. just to be clear, we're thinking of this as something the designer sets, not the agent, right?

tomsilver:
>lgtm. just to be clear, we're thinking of this as something the designer sets, not the agent, right?

right

PR #436, opened 2022-02-06
Semi-supervised is dead

amburger66:
so the focus is on active learning now! for the offline dataset, there is just one hyperparameter that controls the # of pos examples and # of neg examples for each predicate (e.g. `teacher_dataset_num_examples=1` means there is 1 positive and 1 negative annotated example for each excluded predicate)

### DIFF
--- a/src/datasets/ground_atom_data.py
+++ b/src/datasets/ground_atom_data.py
@@ -1,7 +1,8 @@
 """Create offline data by annotating low-level trajectories with ground atoms
 that hold in the respective states."""
 
-from typing import List, Set
+import numpy as np
+from typing import DefaultDict, List, Set, Dict, Tuple
 from predicators.src.envs import BaseEnv
 from predicators.src.structs import Dataset, Predicate, GroundAtom
 from predicators.src import utils
@@ -9,7 +10,7 @@ from predicators.src import utils
 
 def create_ground_atom_data(env: BaseEnv, base_dataset: Dataset,
                             annotating_predicates: Set[Predicate],
-                            annotating_ratio: float) -> Dataset:
+                            num_examples: int) -> Dataset:
     """Create offline data by annotating low-level trajectories with ground
     atoms. The ground atoms must hold in the respective states. Only the
     predicates in annotating_predicates are used in the annotations.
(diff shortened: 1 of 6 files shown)


src/datasets/ground_atom_data.py
```diff
@@ -24,31 +25,35 @@ def create_ground_atom_data(env: BaseEnv, base_dataset: Dataset,
...
+                    pos_examples[ga.predicate].append((i, j, ga))
+                else:
+                    neg_examples[ga.predicate].append((i, j, ga))
+    # Sample `num_examples` from each list
+    pos_picks: List[Tuple] = []
+    neg_picks: List[Tuple] = []
+    for p in annotating_predicates:
+        for examples, picks in zip((pos_examples, neg_examples),
+                                   (pos_picks, neg_picks)):
+            idxs = np.random.choice(len(examples[p]),
```
amburger66, line 47: do i need to seed this?

ronuchit: yeah, sorry. you can copy what we do in datasets/demo_replay.py (ctrl-f "rng")

src/settings.py
```diff
@@ -89,7 +89,8 @@ class GlobalSettings:
     # data collection timeout for planning
 
     # teacher dataset parameters
-    teacher_dataset_label_ratio = 1.0
+    teacher_dataset_num_examples = 1  # num of positive examples and negative
```
ronuchit, line 92: yapf hates comments like this, which makes me hate yapf. instead:
```
# Number of positive examples and negative examples per predicate.
teacher_dataset_num_examples = 1
```

src/datasets/ground_atom_data.py
```diff
@@ -24,31 +25,35 @@ def create_ground_atom_data(env: BaseEnv, base_dataset: Dataset,
...
-                        annotated_preds.add(pred)
-                    totals[pred] += 1
-            ground_atoms_traj.append(subset_atoms)
-        assert len(traj.states) == len(ground_atoms_traj)
-        annotations.append(ground_atoms_traj)
-    assert len(annotations) == len(base_dataset.trajectories)
-    assert annotated_preds == annotating_predicates, \
-        "Not all predicates are seen in the dataset"
+    # Generate all positive and negative examples
+    pos_examples = DefaultDict(list)  # predicate: [(i, j, ground_atom)]
```
tomsilver, line 29: I think we generally want to use things imported from `typing` for annotations only. (I don't have a good reason other than convention.) so here, use `from collections import defaultdict`

src/datasets/ground_atom_data.py
```diff
@@ -24,31 +25,35 @@ def create_ground_atom_data(env: BaseEnv, base_dataset: Dataset,
...
+            for ga in ground_atom_universe:
+                if ga in ground_atoms:  # positive example
+                    pos_examples[ga.predicate].append((i, j, ga))
+                else:
+                    neg_examples[ga.predicate].append((i, j, ga))
+    # Sample `num_examples` from each list
+    pos_picks: List[Tuple] = []
+    neg_picks: List[Tuple] = []
+    for p in annotating_predicates:
+        for examples, picks in zip((pos_examples, neg_examples),
```
tomsilver, line 45: swap the order here so that everywhere we have a tuple corresponding to negative/positive things, negative comes first

src/datasets/ground_atom_data.py
```diff
@@ -24,31 +25,35 @@ def create_ground_atom_data(env: BaseEnv, base_dataset: Dataset,
...
+                                   (pos_picks, neg_picks)):
+            idxs = np.random.choice(len(examples[p]),
+                                    size=num_examples,
```
tomsilver, line 48: can you raise an informative error when `num_examples > len(examples[p])` and add a test case for this?

tomsilver (commented):
🔥 🔥 🔥 🔥 🔥

src/datasets/ground_atom_data.py
amburger66: is this ok?

tomsilver: Yes perf thanks

tomsilver (approved):
Lgtm

PR #441, opened 2022-02-07
updates to analysis scripts to enable plotting vs CFG values

tomsilver:
like num train tasks

example:
performance_vs_number_of_demos_num_train_tasks_num_solved


analysis/analyze_results_directory.py
```diff
@@ -1,81 +1,88 @@
...
+    # "APPROACH",
+    # "EXCLUDED_PREDICATES",
+    "EXPERIMENT_ID",
+    # "NUM_TRAIN_TASKS",
+    # "CYCLE"
+]
 
-def create_dataframes() -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
-    """Returns means, standard deviations, and sizes."""
+COLUMN_NAMES_AND_KEYS = [
```
ronuchit, line 20: this is just for generating the pandas table right? can we leave most of this out and only keep what we care about (experiment_id, seed, num_solved, avg_num_preds, avg_test_time, avg_nodes_created, learning_time)

analysis/analyze_results_directory.py
```diff
@@ -1,81 +1,88 @@
...
         with open(filepath, "rb") as f:
-            run_data_defaultdict = pkl.load(f)
+            outdata = pkl.load(f)
+        if "config" in outdata:
+            config = outdata["config"].__dict__.copy()
+            run_data_defaultdict = outdata["results"]
+            assert not set(config.keys()) & set(run_data_defaultdict.keys())
+            run_data_defaultdict.update(config)
+        else:
+            print("WARNING: results are old format. Support ending soon.")
```
ronuchit, line 66: are we really planning to end support for this? if not, let's not say that?

analysis/create_online_learning_curves.py
```diff
@@ -29,34 +61,28 @@
 # PLOT_GROUPS is a nested dict where each outer dict corresponds to one plot,
 # and each inner entry corresponds to one line on the plot.
 # The keys of the outer dict are are plot titles.
```
ronuchit, line 63: "are are"

analysis/create_online_learning_curves.py
```diff
@@ -66,24 +92,25 @@ def _main() -> None:
...
+                                                      select_value)
+                    exp_stds = _get_df_for_plot_line(x_key, stds, select_key,
+                                                     select_value)
                     ax.errorbar(exp_means[x_key],
                                 exp_means[y_key],
                                 yerr=exp_stds[y_key],
                                 label=label)
                 # Automatically make x ticks integers for certain X KEYS.
-                if y_key in ("CYCLE", "NUM_TRANSITIONS"):
+                if x_key in ("CYCLE", "NUM_TRANSITIONS", "NUM_TRAIN_TASKS"):
```
ronuchit, line 113: lol was this a bug before?

analysis/analyze_results_directory.py
ronuchit: for example we want to be able to download old results files and still run this script

analysis/create_online_learning_curves.py
tomsilver: what's your your point

analysis/create_online_learning_curves.py
tomsilver: yerp

PR #444, opened 2022-02-07
allow different planning params for dataset creation 

tomsilver:
Example for collecting optimal data, but testing with different heuristic / max skeletons
```
python src/main.py --env cover --approach nsrt_learning --seed 0 --sesame_task_planning_heuristic hadd --sesame_max_skeletons_optimized 2 --offline_data_task_planning_heuristic lmcut --offline_data_max_skeletons_optimized 8
```


src/settings.py
```diff
@@ -92,6 +92,10 @@ class GlobalSettings:
     # dataset parameters
     # For learning-based approaches, the data collection timeout for planning.
     offline_data_planning_timeout = 3
+    # If default, defaults to CFG.task_planning_heuristic.
```
ronuchit, line 95: put the word "default" in quotes

src/teacher.py
```diff
@@ -21,9 +21,13 @@ def __init__(self, train_tasks: Sequence[Task]) -> None:
         env = create_env(CFG.env)
         self._pred_name_to_pred = {pred.name: pred for pred in env.predicates}
         self._allowed_query_type_names = get_allowed_query_type_names()
-        self._oracle_approach = create_approach("oracle", env.predicates,
-                                                env.options, env.types,
-                                                env.action_space, [])
+        self._oracle_approach = OracleApproach(
```
ronuchit, line 24: should we remove the "oracle" `if` statement from create_approach entirely, so we don't have this issue again in the future?

ronuchit: oh shoot but that would make `--oracle` crash so probably a bad idea lol

ronuchit (approved):
this looks like it was annoying, thanks!

PR #448, opened 2022-02-08
improve API for side predicate learning

ronuchit:
theoretically, a person in the future may want to work on side predicate
stuff. if such a person were to exist, they may appreciate the API
providing the things they would need. this is an attempt to remedy that.

* learn_pnad_side_predicates now takes in PNADs, ground atom data,
train_tasks, predicates, and segments
* added a helper function `op.effect_to_side_predicate()` with tests


tomsilver (approved):
you have strange hobbies at 10pm

PR #450, opened 2022-02-08
generate latex tables from results directory

tomsilver:
😮‍💨


ronuchit (approved):
you're the only person who will ever understand this and i'm fine with that

PR #453, opened 2022-02-09
repeated nextto learning

### LINKED ISSUE #175: get RepeatedNextToEnv working with nsrt_learning

ronuchit:
Tried a super simple thing for repeated nextto learning, which ended up working well. Results:
```
              NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED    LEARNING_TIME AVG_NUM_NSRTS  NUM_SEEDS
EXPERIMENT_ID
noside           50.00 (0.00)   8.88 (6.55)   3.00 (0.00)   4.62 (3.17)  1341.49 (1172.79)  623.02 (368.52)   6.10 (0.81)         24
oracle           50.00 (0.00)  50.00 (0.00)   3.00 (0.00)   0.01 (0.00)       47.62 (0.77)      0.00 (0.00)   2.00 (0.00)         46
withside         50.00 (0.00)  48.98 (1.22)   3.00 (0.00)   0.43 (0.28)       50.28 (5.99)    75.34 (43.42)   4.11 (0.44)         44
```
Note: I was too lazy to wait for the rest of the results so I killed them. The difference between noside and withside is the flag `--learn_side_predicates True`. Both noside and withside use `--approach nsrt_learning`

The approach is to just do hill climbing with our prediction error score function, on side-ifications of effects one at a time. I think the reason it's working is when you sideline a predicate for a move operator, you're just increasing the number of true positives this operator has, since in your data, you'll see way more instances of moving to 0 or 1 dots than 2 or 3, say. This therefore improves the prediction error objective. So, one way to understand what's happening here is that you’re making the operator more general, which is the whole idea of this side predicate stuff in the first place.

I tried our other score functions, but they were all either too slow or didn't work well.

Example command: `python src/main.py --env repeated_nextto --approach nsrt_learning --seed 0 --learn_side_predicates True`

Note: There's still a TON more research to be done in this space. The solution in this PR is very hacky and brittle; in other environments it definitely won't do the right thing. But it's hopefully a good starting point. And also my favorite part -- closes #175 

EDIT: after updating the function to also allow removal of operators within hill climbing, it turns out that the system always prefers to remove operators rather than side-ifying any predicate.... not sure what to do about that for now. We just need a better environment for testing side predicates, I guess.


analysis/run_supercloud_experiments.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 START_SEED=456
```
ronuchit, line 3: @tomsilver will revert changes to this file

src/nsrt_learning.py
```diff
@@ -248,8 +244,102 @@ def _learn_pnad_side_predicates(
...
+                    if not ground_op.preconditions.issubset(
+                            segment.init_atoms):
+                        continue
+                    # Check if effects match. Note that we're using the side
+                    # predicates semantics here!
+                    atoms = utils.apply_operator(ground_op, segment.init_atoms)
+                    if not atoms.issubset(segment.final_atoms):
+                        continue
+                    # This segment belongs in this datastore, so add it.
+                    # TODO: this is an edge case we need to fix by changing
```
ronuchit, line 334: @tomsilver this was your TODO, i never understood it. do you remember?

src/settings.py
```diff
@@ -47,7 +47,7 @@ class GlobalSettings:
     cluttered_table_place_goal_conditioned_sampling = True
 
     # repeated nextto env parameters
-    repeated_nextto_num_dots = 25
+    repeated_nextto_num_dots = 15
```
ronuchit, line 50: @tomsilver made this change for speed, everything is way too slow without it (even prediction error)

src/nsrt_learning.py
```diff
@@ -248,8 +244,102 @@ def _learn_pnad_side_predicates(
...
+                    # ObjToVarSubs in the partition to VarToObjSubs.
+                    if len(set(ground_op.objects)) != len(ground_op.objects):
+                        continue
+                    sub = dict(zip(ground_op.objects, pnad.op.parameters))
+                    pnad.add_to_datastore((segment, sub),
+                                          check_consistency=False)
+                    all_indices[pnad_idx].add(identifier)
+    return all_indices
```
ronuchit, line 342: @tomsilver reminder for myself to get rid of this return value, we aren't pruning anymore (it hurts, bc you might prune an op that's good for planning even if it fits all the same data)

src/nsrt_learning.py
```diff
@@ -248,8 +244,102 @@ def _learn_pnad_side_predicates(
...
+                for effect in effect_set:
+                    new_pnad = PartialNSRTAndDatastore(
+                        pnad.op.effect_to_side_predicate(
+                            effect, option_vars, add_or_delete),
+                        pnad.datastore, pnad.option_spec)
+                    sprime = list(s)
+                    sprime[i] = new_pnad
+                    yield (None, tuple(sprime), 1.0)
+
+    # Hard to avoid this circular import, since approaches need nsrt_learning.
```
tomsilver, line 272: let's try to resolve this before merging

src/nsrt_learning.py
```diff
@@ -248,8 +244,102 @@ def _learn_pnad_side_predicates(
...
+    from predicators.src.approaches.grammar_search_invention_approach import \
```
tomsilver, line 273: let's try to resolve this before merging

src/nsrt_learning.py
```diff
@@ -248,8 +244,102 @@ def _learn_pnad_side_predicates(
...
-        verbose: bool) -> List[PartialNSRTAndDatastore]:
-    raise NotImplementedError
+        segmented_trajs: List[List[Segment]]) -> List[PartialNSRTAndDatastore]:
+    # There are no goal states for this search; run until exhausted.
+    def _check_goal(s: Tuple[PartialNSRTAndDatastore, ...]) -> bool:
+        del s  # unused
+        return False
+
+    # Consider changing each effect to a side predicate.
+    def _get_successors(
```
tomsilver, line 254: so this will be a search over PNADs and data stores? why are we doing it that way? i would've thought the search would just be over the operators themselves. is this an efficiency thing?

src/nsrt_learning.py
```diff
@@ -248,8 +244,102 @@ def _learn_pnad_side_predicates(
...
+    # Final PNADs are the last ones in the path.
+    pnads = list(path[-1])
+    # Recompute the datastores in the PNADs. We need to do this
+    # because now that we have side predicates, each transition may be
+    # assigned to *multiple* datastores.
+    _recompute_datastores_from_segments(segmented_trajs, pnads)
+    return pnads
+
+
+def _recompute_datastores_from_segments(
```
tomsilver, line 298: why do we need to do this if we're reusing the machinery from grammar search? is the idea that the "pruned atom dataset" thing doesn't work here because it's not just about pruning atoms?

src/nsrt_learning.py
tomsilver: oh gosh ummm... I don't remember, but looking at it now, I think this TODO was wrong? I probably thought there would be an issue with two different objects mapping to the same variable (because of the dict getting created on line 338), but the issue would instead be if there were two different variables mapping to the same object. so maybe just remove it and see if anything breaks... sorry!

src/nsrt_learning.py
tomsilver: interesting... but we still want to allow for pruning at the end of the day, I think. we should just do it in the same way that we're sidelining?

tomsilver: i.e., test whether removing the operator improves the score function

src/settings.py
tomsilver: because we're trying to ground large operators in the beginning?

tomsilver (commented):
nice work!

src/nsrt_learning.py
ronuchit: it is over just the operators, i just keep them in the PNAD data structure because it's easier to write the code. it shouldn't matter from an efficiency perspective, it's not like the datastores themselves are getting copied or anything (there's no deep copying)

src/nsrt_learning.py
ronuchit: i don't have the slightest idea how. any suggestions? even if we move this side predicate stuff into a new file it won't help, bc approaches -> nsrt_learning -> side_predicate_learning will be a cycle

src/nsrt_learning.py
ronuchit: this was your code! :P i think the comment above, which you wrote long ago, is clear enough:
```
    # Recompute the datastores in the PNADs. We need to do this
    # because now that we have side predicates, each transition may be
    # assigned to *multiple* datastores.
```

src/nsrt_learning.py
ronuchit: okay will do

src/nsrt_learning.py
ronuchit: i can try that

src/settings.py
ronuchit: yup

src/nsrt_learning.py
tomsilver: I know, but I was wondering now that we're using the score function stuff, whether we could just reuse the existing code in the score function. but I think probably not, for the reason that I said above

ronuchit: i think you are right

src/nsrt_learning.py
ronuchit: resolved in #454

src/nsrt_learning.py
ronuchit: it crashed; ignoring for now

analysis/run_supercloud_experiments.sh
tomsilver: don't forget!

ronuchit: lets see if we want to relaunch supercloud after whatever changes you make

tomsilver (approved):

PR #455, opened 2022-02-10
a few minor updates to the analysis scripts

tomsilver:


analysis/skeleton_score_analysis.py
```diff
@@ -252,7 +253,10 @@ def get_default_result(_skeleton_idx: int) -> float:
 
 def _create_predicate_labels(
         predicate_set_order: Tuple[FrozenSet[Predicate], ...]) -> List[str]:
-    labels = [", ".join(p.name for p in predicate_set_order[0])]
+    if len(predicate_set_order[0]) > 5:
```
ronuchit, line 256: is this a hack?

analysis/create_latex_tables.py
```diff
@@ -188,8 +226,8 @@ def _main() -> None:
                     inner_label_to_best_mean_std[inner_label] = (mean, std)
         for (outer_label, inner_label), (mean, _,
                                          _) in entry_to_mean_std_size.items():
-            # Special case: exclude oracle
-            if outer_label == "Oracle":
+            # Special case: exclude Handwritten
+            if outer_label == "Handwritten":
```
ronuchit, line 230: ah, thank you for remembering to do this!

ronuchit (approved):
thank you!!

analysis/skeleton_score_analysis.py
tomsilver: yeah just for making the axis labels not crazy

PR #456, opened 2022-02-10
change objtovar dicts to maps in datastores 

tomsilver:
this is (maybe) needed to avoid a crash in #453 . we should look at this with fresh eyes tomorrow before learning. one thing I'm especially not sure about is whether the logic in `_KnownOptionsOptionLearner` for setting up `option_vars` is correct.

I'm also not 100% sure that there isn't a simpler workaround to the issue in #453.


tomsilver:
in favor of #457

PR #457, opened 2022-02-10
switch datastore subs to vartoobj

tomsilver:
this is an alternative to #456 , which should also resolve the crash in #453 , and is probably preferable.

a third option is to do neither of these things, assume that duplicate arguments in ground operators are disallowed, and just remove the TODO comment.

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -348,7 +348,7 @@ class GroundAtom(_Atom):
         return (str(self.predicate) + "(" + ", ".join(map(str, self.objects)) +
                 ")")
 
-    def lift(self, sub: dict[Object, Variable]) -> LiftedAtom:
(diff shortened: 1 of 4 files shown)


src/structs.py
```diff
@@ -348,7 +348,7 @@ def _str(self) -> str:
         return (str(self.predicate) + "(" + ", ".join(map(str, self.objects)) +
                 ")")
 
-    def lift(self, sub: dict[Object, Variable]) -> LiftedAtom:
+    def lift(self, sub: ObjToVarSub) -> LiftedAtom:
```
ronuchit, line 351: While we're at it, let's fix the `def ground` definition above in the same way?

ronuchit (approved):

PR #462, opened 2022-02-13
Metric for counting num queries made

amburger66:
it's a little weird because we decided to allow the robot to ask about multiple ground atoms in one state, so in the case of `GroundAtomsHoldQuery` I'm taking "number of queries" to mean "number of ground atoms asked about."

also added some printouts for during the exploration part of online learning

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -1135,12 +1135,10 @@ class InteractionResult:
 class Query(abc.ABC):
     """Base class for a Query."""
 
-    def __len__(self) -> int:
-        """The number of queries in this Query.
-
-        Defaults to 1.
-        """
-        return 1
+    @property
+    def cost(self) -> float:
+        """The cost of making this Query."""
+        raise NotImplementedError("Override me")
 
 
 @dataclass(frozen=True, eq=False, repr=False)
@@ -1157,7 +1155,8 @@ class GroundAtomsHoldQuery(Query):
     """A query for whether ground atoms hold in the state."""
     ground_atoms: Collection[GroundAtom]
 
-    def __len__(self) -> int:
+    @property
+    def cost(self) -> float:
         return len(self.ground_atoms)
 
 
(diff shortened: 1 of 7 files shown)


tomsilver (commented):
@amburger66 thank you, this looks great! good point about there being sort of multiple queries per query. to avoid confusion, I wonder if we should call the metric something like "cumulative query cost", and have `Query` implement a property `cost` which returns a float? in addition to avoiding confusion, this cost idea would have the benefit of supporting multiple kinds of queries with different costs. (e.g. if we were to implement a different type of query like "ask for all true ground atoms in the state", we could make that query more expensive than the ground atom queries.)

let me know what you think. also will want to hear @ronuchit 's thoughts about this.

amburger66:
I like that idea!

src/structs.py
```diff
@@ -1133,10 +1133,14 @@ def __post_init__(self) -> None:
 
 @dataclass(frozen=True, eq=False, repr=False)
 class Query(abc.ABC):
-    """Base class for a Query.
+    """Base class for a Query."""
 
-    Has no API.
-    """
+    def __len__(self) -> int:
+        """The number of queries in this Query.
```
ronuchit, line 1139: we should def do what tom suggested instead of this, since it doesn't make much sense for an arbitrary query to have a thing called length

tomsilver:
great, let's proceed then!

tomsilver:
Looking good! Do you wanna also go ahead and update the script(s) that you'll use to generate the results tables and plots? Might as well test that it runs all the way through to plot generation now with some small hyper parameters before we scale up.

amburger66:
> i feel ur pain

:")

> Looking good! Do you wanna also go ahead and update the script(s) that you'll use to generate the results tables and plots? Might as well test that it runs all the way through to plot generation now with some small hyper parameters before we scale up.

yup!

amburger66:
example plot lol

active_learning_baseline_cycle_perc_solved

tomsilver:
@amburger66 thanks for the example plot! don't we want cumulative number of queries on the x axis though?

amburger66:
right!
image

active_learning_baseline_cum_query_cost_perc_solved

tomsilver (approved):
excellent work!

ronuchit (approved):
nice change!!

PR #464, opened 2022-02-13
respond in same channel

ronuchit:


launch_slack_bot.py
```diff
@@ -353,12 +343,12 @@ def _callback(ack, body):
     event = body["event"]
     query = event["text"]
     inquirer = event["user"]
+    channel_id = event["channel"]
```
ronuchit, line 346: respond in same channel as the @ occurred

tomsilver (approved):

PR #470, opened 2022-02-14
some plotting and table updates 

tomsilver:


ronuchit (approved):
i mentioned this in overleaf, but we'll have to change "skel" in 1skelscore/eval since we purged the word skeleton. up to you whether to make a separate PR for that

PR #477, opened 2022-02-17
pretty str for predicates and NSRTs

ronuchit:


src/structs.py
```diff
@@ -710,6 +730,25 @@ def pddl_str(self) -> str:
...
+                            ("Delete Effects", self.delete_effects)]:
+            out += f"\n\t{name}:"
+            for atom in atoms:
+                pretty_pred = atom.predicate.pretty_str()[1]
+                new_name = (name_map[pretty_pred] if pretty_pred in name_map
+                            else str(atom.predicate))
+                var_str = ", ".join(map(str, atom.variables))
+                out += f"\n\t\t{new_name}({var_str})"
+        option_var_strs = [str(v) for v in self.option_vars]
+        out += f"\n\tOption Spec: ({self.option.name}, {option_var_strs})"
```
ronuchit, line 749: for paper, we'll want to switch to Controller, but keeping this consistent with codebase here

src/structs.py
```diff
@@ -239,6 +240,25 @@ def __str__(self) -> str:
...
         return str(self)
 
+    def pretty_str(self) -> Tuple[str, str]:
+        """Display the predicate in a nice human-readable format.
+
+        Returns a tuple of (variables string, body string).
+        """
+        if hasattr(self._classifier, "pretty_str"):
+            # This is an invented predicate, from the predicate grammar.
+            pretty_str_f = getattr(self._classifier, "pretty_str")
```
ronuchit, line 250: hack around mypy which was complaining if you just do self._classifier.pretty_str() bc self._classifier might be a function. tried other ways to get around this, but this was easiest

PR #478, opened 2022-02-17
Flexible initial state

ashayathalye:
Creates initial state by (1) placing targets and blocks in random locations such that each target has enough space on either side to ensure no covering placement will cause a collision (note that this is not necessary to make the task solveable; but we can do this and instead sufficiently tune the difficulty through hand region specification), and (2) choosing hand region intervals on the targets and blocks such that the problem is solveable.

Note that the hand regions could be made attributes of the target and block objects. This was not done just because it would require editing more code all over the codebase, and creating new "hand region objects" didn't seem like an issue. But, can make this change if preferred.  

There are two kinds of difficulties we want to be able to tune: 
1. Measure of backtracking difficulty (need to resample the pick()): what percent of samples in the pick() option will lead to a place() that is impossible to cover the target with, assuming you try placing somewhere along the target's hand_region?
2. Measure of resampling difficulty (need to resample the place()): what percent of samples in the place() option will lead to a place() that is infeasible due to not hitting the target's hand_region?
Will follow up with some code that simulates to figure out what parameters to choose for CFG.cover_multistep_bhr_percent and CFG.cover_multistep_thr_percent to get the kind of difficulty we want (explicitly measuring the difficulty) while keeping the problem solveable in reasonable time. 

Linter is failing on code I didn't touch, and the tests are saying missing coverage for lines I didn't touch (still looking into this). Also, I will add a test that hard-codes solving one task with the new create_initial_state function.

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -794,10 +794,10 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
         """Creates initial state by (1) placing targets and blocks in random
         locations such that each target has enough space on either side to
         ensure no covering placement will cause a collision (note that this is
-        not necessary to make the task solveable; but we can do this and instead
+        not necessary to make the task solvable; but we can do this and instead
         sufficiently tune the difficulty through hand region specification), and
         (2) choosing hand region intervals on the targets and blocks such that
-        the problem is solveable.
+        the problem is solvable.
         """
         data: Dict[Object, Array] = {}
 
@@ -807,7 +807,7 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
             overlap = False
             counter += 1
             if counter > CFG.cover_multistep_max_placements:
-                raise MaxPlacementsFailure("Reached maximum number of " \
+                raise RuntimeError("Reached maximum number of " \
                 "placements of targets and blocks.")
             block_placements = []
             for block, bw in zip(self._blocks, CFG.cover_block_widths):
@@ -851,7 +851,7 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
         data[self._robot] = np.array([0.0, self.initial_robot_y, -1.0, -1.0])
 
         # Make the hand regions
-        # sample a hand region interval in each target
+        # Sample a hand region interval in each target
         target_hand_regions = []
         for i, target in enumerate(self._targets):
             target_hr = self._target_hand_regions[i]
@@ -863,7 +863,7 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
             data[target_hr] = np.array(region)
             target_hand_regions.append(region)
 
-        # sample a hand region interval in each block
+        # Sample a hand region interval in each block
         for i, block in enumerate(self._blocks):
             block_hr = self._block_hand_regions[i]
             thr_left, thr_right = target_hand_regions[i]
@@ -885,7 +885,7 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
             # one placement of the block which covers the target, and in
             # which the block's hand region and the target's hand region
             # have nonzero overlap.
-            # A proxy for this operation is to check:
+            # A proxy for this operation, which we do below, is to check:
             # (1) That in the block's rightmost covering placement, its
             # interval IS NOT completely to the left of the target's
             # hand region, and
@@ -896,16 +896,16 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
             while True:
                 counter += 1
                 if counter > CFG.cover_multistep_max_placements:
-                    raise MaxPlacementsFailure("Reached maximum number of " \
+                    raise RuntimeError("Reached maximum number of " \
                     "placements of hand regions.")
-                # sample hand region
+                # Sample hand region
                 left_pt = rng.uniform(bx - bw/2, bx + bw/2 - region_length)
                 region = [left_pt, left_pt + region_length]
-                # need to make hand region relative to center of block for
+                # Need to make hand region relative to center of block for
                 # the hand region to move with the block for use by
                 # _get_hand_regions()
                 relative_region = [region[0]-bx, region[1]-bx]
-                # perform the valid interval check
+                # Perform the valid interval check
                 relative_r = region[1] - (bx - bw/2)  # for (1)
                 relative_l = bx + bw/2 - region[0]  # for (2)
                 if relative_l >= (tx + tw/2 - thr_right) and \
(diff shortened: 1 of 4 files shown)


src/envs/cover.py
```diff
@@ -489,6 +489,8 @@ def __init__(self) -> None:
                                  ["is_block", "is_target", "width", "x"])
         # Also removing "hand" because that's ambiguous.
         self._robot_type = Type("robot", ["x", "y", "grip", "holding"])
+        self._hand_region_type = Type("hand_region", ["lb", "ub"])
```
tomsilver, line 492: nice! this is a clever way to do it.

src/envs/cover.py
```diff
@@ -775,34 +791,148 @@ def render(self,
...
 
     def _create_initial_state(self, rng: np.random.Generator) -> State:
-        # Need to override to account for new low-level state features.
-        # Note: I originally tried to use super(), but ran into issues
-        # because the parent class types and this class types are
-        # actually different (in terms of equality checking).
+        """Creates initial state by (1) placing targets and blocks in random
+        locations such that each target has enough space on either side to
+        ensure no covering placement will cause a collision (note that this is
+        not necessary to make the task solveable; but we can do this and instead
```
tomsilver, line 797: nit: solvable*

src/envs/cover.py
```diff
@@ -775,34 +791,148 @@ def render(self,
...
+        sufficiently tune the difficulty through hand region specification), and
+        (2) choosing hand region intervals on the targets and blocks such that
+        the problem is solveable.
```
tomsilver, line 800: ditto

src/envs/cover.py
```diff
@@ -1089,3 +1206,6 @@ def _create_initial_state(self, rng: np.random.Generator) -> State:
         del rng
         zero_rng = np.random.default_rng(0)
         return super()._create_initial_state(zero_rng)
+
+class MaxPlacementsFailure(Exception):
```
tomsilver, line 1210: if we aren't planning to catch this kind of exception specifically, I think we can just remove it and use `Exception` or `RuntimeError`

src/envs/cover.py
```diff
@@ -775,34 +791,148 @@ def render(self,
...
+            # "Place" the block in the leftmost position that still covers
+            # the target. Move the block to the right until it reaches the
+            # rightmost position that still covers the target. During this,
+            # check that there is nonzero overlap between the interval
+            # spanned by the moving block's hand region, and the target's
+            # hand region. In other words, make sure that there is at least
+            # one placement of the block which covers the target, and in
+            # which the block's hand region and the target's hand region
+            # have nonzero overlap.
+            # A proxy for this operation is to check:
```
tomsilver, line 888: my understanding of the code is that we're using this proxy alone, not the operation described above. is that right? if so, let's rephrase the comment above a bit to make that clear.

tests/nsrt_learning/test_option_learning.py
```diff
@@ -180,8 +180,10 @@ def test_learned_neural_parameterized_option():
     # Get an initial state where picking should be possible.
     env.seed(123)
     task = env.get_test_tasks()[0]
+
     state = task.init.copy()
-    block0, block1, robot, _, _ = sorted(state)
+    print("SORTED STATE: ", sorted(state))
```
tomsilver, line 185: remove print

src/envs/cover.py
```diff
@@ -775,34 +791,148 @@ def render(self,
...
+            # interval IS NOT completely to the left of the target's
+            # hand region, and
+            # (2) That in the block's leftmost covering placement, its
+            # interval IS NOT completely to the right of the target's
+            # hand region.
+            counter = 0
+            while True:
+                counter += 1
+                if counter > CFG.cover_multistep_max_placements:
+                    raise MaxPlacementsFailure("Reached maximum number of " \
```
tomsilver, line 899: FYI, it looks like this is the line that is missing unit test coverage

tomsilver (commented):
very nice, thank you!

as I mentioned on slack, it looks like one of the github check failures is due to autoformatting -- see README for instructions on installing and using that. the other check failure looks like one line of missing coverage; I marked the line in my review.

my comments are all minor. before we merge, though, could you report example command lines and outputs that demonstrate the key things you wanted to enable here? for example, you can artificially limit `sesame_max_samples_per_step` and `sesame_max_skeletons_optimized` and show that they fail with certain values of your new hyperparameters, but succeed with more forgiving values.

ashayathalye:
`python3 src/main.py --env cover_multistep_options --approach oracle --seed 0 --timeout 10 --sesame_max_samples_per_step 10 --cover_multistep_thr_percent 0.5 cover_multistep_bhr_percent 0.5` solves 50/50. 

`python3 src/main.py --env cover_multistep_options --approach oracle --seed 0 --timeout 10 --sesame_max_samples_per_step 10 --cover_multistep_thr_percent 0.1 cover_multistep_bhr_percent 0.5` solves 33/50, where all failures are due to "Planning timed out in backtracking!". 

`python3 src/main.py --env cover_multistep_options --approach oracle --seed 0 --timeout 30 --sesame_max_samples_per_step 10 --cover_multistep_thr_percent 0.1 --cover_multistep_bhr_percent 0.5` solves 48/50, where the two failures are due to "Planning timed out in backtracking!".

`python src/main.py --env cover_multistep_options --approach oracle --seed 0 --timeout 40 --sesame_max_samples_per_step 20 --cover_multistep_thr_percent 0.1 --cover_multistep_bhr_percent 0.5` solves 50/50. Note that as we increase `sesame_max_samples_per_step`, we also have to increase timeout, since it takes time to do more samples.

src/envs/cover.py
```diff
@@ -775,34 +791,150 @@ def render(self,
...
-                if not self._any_intersection(x, width, data):
-                    break
+
+        # Place targets and blocks
+        counter = 0
+        while True:
+            overlap = False
+            counter += 1
+            if counter > CFG.cover_multistep_max_tb_placements:
+                print("Reached maximum number of " \
```
tomsilver, line 809: this looks redundant with the error message, can remove

src/envs/cover.py
```diff
@@ -775,34 +791,150 @@ def render(self,
...
+            # interval IS NOT completely to the left of the target's
+            # hand region, and
+            # (2) That in the block's leftmost covering placement, its
+            # interval IS NOT completely to the right of the target's
+            # hand region.
+            counter = 0
+            while True:
+                counter += 1
+                if counter > CFG.cover_multistep_max_hr_placements:
+                    print("Reached maximum number of " \
```
tomsilver, line 900: ditto

tomsilver (approved):
two tiny things, otherwise lgtm. great work! feel free to merge after addressing the comments.

PR #479, opened 2022-02-17
Added "repeated_nextto_painting" Environment

wmcclinton:
I added an environment to make learning side predicates even harder.


ronuchit:
We generally don't want to have large chunks of copy-pasted code whenever possible. Could you describe, perhaps by writing out in a comment here or in a doc/slides, what exactly you're trying to do in this environment, and in what ways it's different from the base painting env? Then, @tomsilver and I can work with you to figure out how to best set it up as a subclass.

(If you are just setting this up to hack on research ideas as quickly as possible, it's fine to leave it in as a draft PR and build new branches off of it, but of course it's always best to get things into master so that the code stays up to date.)

wmcclinton:
Sorry @ronuchit it's fixed now. I just realized that and subclassed it.

ronuchit:
@wmcclinton could you nevertheless describe what's going on in the environment, at a high level? it'll be hard to review without having that context first

NishanthJKumar (commented):
lgtm. and ready for review from @ronuchit and @tomsilver

src/envs/painting.py
```diff
@@ -432,6 +438,26 @@ def render(self,
             "blue = wet+clean, green = dry+dirty, cyan = dry+clean;\n"
             "yellow border = side grasp, orange border = top grasp",
             fontsize=12)
+        # List of NextTo objects to render
+        # Added this to display what objects we are nextto
+        # during video rendering
+        if isinstance(self, envs.RepeatedNextToPaintingEnv):
```
tomsilver, line 444: can we refactor this in the following way: implement a function called something like `_render_matplotlib` that does everything up until `utils.fig2data(fig)` and returns `fig`. Then, rather than this `isinstance` check here, we can do:

in painting.py:
```
def render(...):
    fig = self._render_matplotlib(...)
    img = utils.fig2data(fig)
    plt.close()
    return [img]
```

and in repeated_nextto_painting.py:
```
def render(...):
    fig = self._render_matplotlib(...)
    nextto_objs = []
    for obj in state:
        ...
    img = utils.fig2data(fig)
    plt.close()
    return [img]
```

one reason for this change would be so that all of the repeated nextto specific code is centralized in that one file.

src/envs/painting.py
```diff
@@ -501,7 +531,15 @@ def _get_tasks(self, num_tasks: int, num_objs_lst: List[int],
                 state.set(self._robot, "fingers", 0.0)
                 state.set(target_obj, "grasp", grasp)
                 state.set(target_obj, "held", 1.0)
-                assert self._OnTable_holds(state, [target_obj])
+                if isinstance(self, envs.RepeatedNextToPaintingEnv):
```
tomsilver, line 534: why does this need to be specific to RepeatedNextToPaintingEnv? could we just remove the if and keep the change below?

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+        self._MoveToShelf = ParameterizedOption(
+            "MoveToShelf",
+            types=[self._robot_type, self._shelf_type],
+            params_space=Box(self.env_lb, self.env_ub, (1, )),
+            _policy=self._Move_policy,
+            _initiable=utils.always_initiable,
+            _terminal=utils.onestep_terminal)
+
+    def simulate(self, state: State, action: Action) -> State:
+        assert self.action_space.contains(action.arr)
```
tomsilver, line 58: I'm wondering if we can make this a little more decoupled from the parent class, and also if we can reduce the duplicated lines here. In both cases, I'm thinking that if (god forbid) we ever want to change the parent class, it would be easy to forget to change the child class correspondingly.

Would the following work? At the beginning of this simulate, compute just the `move_affinity`, and if it's below some threshold, automatically use `self._transition_move`. Otherwise, call `super().simulate(state, action)`.

I see that there are changes to `transition_pick` and `transition_place` too --- but I think we can address those in another way, which I'll comment on separately.

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+            (wash_affinity, self._transition_wash),
+            (dry_affinity, self._transition_dry),
+            (paint_affinity, self._transition_paint),
+            (abs(-1 - arr[4]), self._transition_nextto_place),
+            (move_affinity, self._transition_move),
+        ]
+        _, transition_fn = min(affinities, key=lambda item: item[0])
+        return transition_fn(state, action)
+
+    def _transition_nextto_pick_or_openlid(self, state: State,
```
tomsilver, line 78: can we use the same name for this method as the parent class? then to use the parent class's implementation, we can do `next_state = super()._transition_pick_or_openlid(state, action)`.

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+                target_obj, "held") == 0.0 and next_state.get(
+                    target_obj, "held") == 1.0:
+            # Added cannot pick if obj is too far
+            if abs(state.get(self._robot, "y") \
+                - state.get(target_obj, "pose_y")) \
+                >= self.nextto_thresh:
+                return state
+        return next_state
+
+    def _transition_nextto_place(self, state: State, action: Action) -> State:
```
tomsilver, line 94: ditto on the method name

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+        ]
+        _, transition_fn = min(affinities, key=lambda item: item[0])
+        return transition_fn(state, action)
+
+    def _transition_nextto_pick_or_openlid(self, state: State,
+                                           action: Action) -> State:
+        x, y, z, _ = action.arr[:4]
+        next_state = self._transition_pick_or_openlid(state, action)
+        target_obj = self._get_object_at_xyz(state, x, y, z)
+        # if PaintingEnv pick was sucessful check if the object was next to
```
tomsilver, line 83: nit: successful*

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+            (abs(-1 - arr[4]), self._transition_nextto_place),
+            (move_affinity, self._transition_move),
+        ]
+        _, transition_fn = min(affinities, key=lambda item: item[0])
+        return transition_fn(state, action)
+
+    def _transition_nextto_pick_or_openlid(self, state: State,
+                                           action: Action) -> State:
+        x, y, z, _ = action.arr[:4]
+        next_state = self._transition_pick_or_openlid(state, action)
```
tomsilver, line 81: I would have expected this to happen in the opposite order: first check next to, then call the parent class method. this is probably fine too, but just wondering, is there any reason that it needs to happen this way?

tomsilver: oh, is it because this way saves you the trouble of figuring out which object is being picked? that makes sense.

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+            # Added cannot pick if obj is too far
+            if abs(state.get(self._robot, "y") \
+                - state.get(target_obj, "pose_y")) \
+                >= self.nextto_thresh:
+                return state
+        return next_state
+
+    def _transition_nextto_place(self, state: State, action: Action) -> State:
+        # Action args are target pose for held obj
+        _, y, _ = action.arr[:3]
```
tomsilver, line 96: can just do `y = action.arr[1]` here, like in transition_move below

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+        next_state = self._transition_place(state, action)
+        #  Added restriction on place if not close enough to location
+        if abs(state.get(self._robot, "y") - y) \
+            >= self.nextto_thresh:
```
tomsilver, line 100: does this fit on one line?

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+        # Execute move
+        next_state.set(self._robot, "y", y)
+        held_obj = self._get_held_object(state)
+        if held_obj is not None:
+            next_state.set(held_obj, "pose_y", y)
+        return next_state
+
+    @property
+    def predicates(self) -> Set[Predicate]:
+        return {
```
tomsilver, line 117: can we call the parent method to make obvious what new predicates are being added? so:
```return super().predicates | {self._NextTo, self._NextToBox, self._NextToShelf, self._NextToNothing}```

File: src/envs/repeated_nextto_painting.py [hunk 11]
```diff
@@ -0,0 +1,182 @@
...
+            self._InBox, self._InShelf, self._IsBoxColor, self._IsShelfColor,
+            self._GripperOpen, self._OnTable, self._HoldingTop,
+            self._HoldingSide, self._Holding, self._IsWet, self._IsDry,
+            self._IsDirty, self._IsClean, self._NextTo, self._NextToBox,
+            self._NextToShelf, self._NextToNothing
+        }
+
+    @property
+    def options(self) -> Set[ParameterizedOption]:
+        return {
```
tomsilver, line 127: ditto on calling `super().options` here

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+            1.0, 1.0
+        ],
+                          dtype=np.float32)
+        return Box(lowers, uppers)
+
+    def _OnTable_holds(self, state: State, objects: Sequence[Object]) -> bool:
+        obj, = objects
+        obj_y = state.get(obj, "pose_y")
+        return self.table_lb < obj_y < self.table_ub and \
+            np.allclose(state.get(obj, "pose_z"), self.table_height\
```
tomsilver, line 154: nit: add a space between `self.table_height` and `\`

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+            [self.obj_x - 1e-2, self.env_lb, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0],
+            dtype=np.float32)
+        uppers = np.array([
+            self.obj_x + 1e-2, self.env_ub, self.obj_z + 1e-2, 1.0, 1.0, 1.0,
+            1.0, 1.0
+        ],
+                          dtype=np.float32)
+        return Box(lowers, uppers)
+
+    def _OnTable_holds(self, state: State, objects: Sequence[Object]) -> bool:
```
tomsilver, line 150: I can see how this definition is more correct in some sense, but does something break if we use the original definition?

src/envs/painting.py
```diff
@@ -52,10 +58,10 @@ def __init__(self) -> None:
             "pose_x", "pose_y", "pose_z", "dirtiness", "wetness", "color",
             "grasp", "held"
         ])
-        self._box_type = Type("box", ["color"])
+        self._box_type = Type("box", ["pose_x", "pose_y", "color"])
         self._lid_type = Type("lid", ["is_open"])
-        self._shelf_type = Type("shelf", ["color"])
-        self._robot_type = Type("robot", ["fingers"])
+        self._shelf_type = Type("shelf", ["pose_x", "pose_y", "color"])
+        self._robot_type = Type("robot", ["x", "y", "fingers"])
```
tomsilver, line 64: can we make this `pose_x` and `pose_y` for consistency?

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,182 @@
...
+                   state.get(obj, "pose_y")) < self.nextto_thresh
+
+    def _NextToNothing_holds(self, state: State,
+                             objects: Sequence[Object]) -> bool:
+        robot, = objects
+        for typed_obj in state:
+            if typed_obj.type in \
+                [self._obj_type, self._box_type, self._shelf_type] and \
+                self._NextTo_holds(state, [robot, typed_obj]) and \
+                typed_obj is not self._get_held_object(state):
```
tomsilver, line 180: can we check first at the beginning of this function once whether anything is held?

src/ground_truth_nsrts.py
```diff
@@ -862,6 +864,375 @@ def placeontable_sampler(state: State, goal: Set[GroundAtom],
     return nsrts
 
 
+def _get_repeated_nextto_painting_gt_nsrts() -> Set[NSRT]:
```
tomsilver, line 867: most of this should be the same as the painting gt nsrts right? can you refactor to remove the duplicate code? I'll take another look at this part after that

tests/approaches/test_oracle_approach.py
```diff
@@ -514,6 +516,29 @@ def test_oracle_approach_painting():
...
+def test_oracle_approach_repeated_nextto_painting():
+    """Tests for OracleApproach class with RepeatedNextToPaintingEnv."""
+    utils.reset_config({
+        "env": "repeated_nextto_painting",
+        "num_train_tasks": 5,
+        "num_test_tasks": 5,
+        "sesame_task_planning_heuristic": "hff"
+    })
+    env = RepeatedNextToPaintingEnv()
+    env.seed(3210)  # This random seed necessary to get full coverage.
```
tomsilver, line 528: 🤨 what's going on here? :)

tests/envs/test_repeated_nextto_painting.py
```diff
@@ -0,0 +1,337 @@
+"""Test cases for the painting environment."""
+
+import pytest
+import numpy as np
+from predicators.src.envs import RepeatedNextToPaintingEnv
+from predicators.src import utils
+
+
+def test_painting():
```
tomsilver, line 9: rename to `test_repeated_nextto_painting()`

tests/envs/test_repeated_nextto_painting.py
```diff
@@ -0,0 +1,337 @@
+"""Test cases for the painting environment."""
```
tomsilver, line 1: update docstring

tests/envs/test_repeated_nextto_painting.py
```diff
@@ -0,0 +1,337 @@
...
+        {"InBox", "IsBoxColor", "InShelf", "IsShelfColor"}
+    assert len(env.options) == 9
+    assert len(env.types) == 5
+    obj_type = [t for t in env.types if t.name == "obj"][0]
+    robot_type = [t for t in env.types if t.name == "robot"][0]
+    box_type = [t for t in env.types if t.name == "box"][0]
+    shelf_type = [t for t in env.types if t.name == "shelf"][0]
+    lid_type = [t for t in env.types if t.name == "lid"][0]
+    assert env.action_space.shape == (8, )
+    for i, task in enumerate(env.get_test_tasks()):
```
tomsilver, line 33: so this part of the test is looking to see whether one of each type exists in the initial state? could we instead make this shorter then by doing something like:
```assert {box_type, shelf_type, obj_type, robot_type, lid_type} == {item.type for item in state}```

**Review by tomsilver — COMMENTED** [review]
looking really good! thank you! :)

File: src/envs/painting.py [hunk 21]
```diff
@@ -54,10 +60,10 @@ def __init__(self) -> None:
             "pose_x", "pose_y", "pose_z", "dirtiness", "wetness", "color",
             "grasp", "held"
         ])
-        self._box_type = Type("box", ["color"])
+        self._box_type = Type("box", ["pose_x", "pose_y", "color"])
         self._lid_type = Type("lid", ["is_open"])
-        self._shelf_type = Type("shelf", ["color"])
-        self._robot_type = Type("robot", ["fingers"])
+        self._shelf_type = Type("shelf", ["pose_x", "pose_y", "color"])
+        self._robot_type = Type("robot", ["pose_x", "pose_y", "fingers"])
```
ronuchit, line 66: Looks like the robot's pose_y field here is not constant, which means its presence is going to affect the predicate invention pipeline. So definitely we should run the nightly before merging this.

src/envs/painting.py
```diff
@@ -328,11 +334,11 @@ def action_space(self) -> Box:
         # [x, y, z, grasp, pickplace, water level, heat level, color]
         # Note that pickplace is 1 for pick, -1 for place, and 0 otherwise,
         # while grasp, water level, heat level, and color are in [0, 1].
-        lowers = np.array([
-            self.obj_x - 1e-2, self.env_lb, self.obj_z - 1e-2, 0.0, -1.0, 0.0,
-            0.0, 0.0
-        ],
-                          dtype=np.float32)
+        # Changed lower bound for z to 0.0 for RepeatedNextToPainting
```
ronuchit, line 337: Preferable to write comments describing things in absolutes, rather than saying you changed something. Because in a few months, no one's going to remember what the previous value was, and this comment is going to be confusing to anyone reading the code. Instead, perhaps:
"We set the lower bound for z to 0.0, rather than self.obj_z - 1e-2, because in RepeatedNextToPainting, we use this dimension to check affinity of the move action."

src/envs/painting.py
```diff
@@ -500,7 +510,11 @@ def _get_tasks(self, num_tasks: int, num_objs_lst: List[int],
                 state.set(self._robot, "fingers", 0.0)
                 state.set(target_obj, "grasp", grasp)
                 state.set(target_obj, "held", 1.0)
-                assert self._OnTable_holds(state, [target_obj])
+                if CFG.env == "repeated_nextto_painting":
```
ronuchit, line 513: why are we only doing this if repeated nextto painting? seems like a natural thing we'd want to do even in the normal painting?

src/envs/repeated_nextto.py
```diff
@@ -52,6 +52,8 @@ def __init__(self) -> None:
             types=[self._robot_type, self._dot_type])
         # Static objects (always exist no matter the settings).
         self._robot = Object("robby", self._robot_type)
+        # Override this boolean to be true
+        self._is_repeated_nextto_painting = False
```
ronuchit, line 56: unused; remove?

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,169 @@
...
+    @classmethod
+    def get_name(cls) -> str:
+        return "repeated_nextto_painting"
+
+    def simulate(self, state: State, action: Action) -> State:
+        assert self.action_space.contains(action.arr)
+        arr = action.arr
+        # NOTE: added move_affinity
+        move_affinity = sum(abs(val) for val in arr[2:])
+        if move_affinity <= 0.0001:
```
ronuchit, line 62: use `1e-4` instead of `0.0001`

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,169 @@
...
+            return self._transition_move(state, action)
+        return super().simulate(state, action)
+
+    def _transition_pick_or_openlid(self, state: State,
+                                    action: Action) -> State:
+        x, y, z, _ = action.arr[:4]
+        next_state = super()._transition_pick_or_openlid(state, action)
+        target_obj = self._get_object_at_xyz(state, x, y, z)
+        # if PaintingEnv pick was successful check if the object was next to
+        if target_obj is not None and state.get(
```
ronuchit, line 72: formatting nit, preferable to do:
```
        if target_obj is not None and \
           state.get(target_obj, "held") == 0.0 and \
           next_state.get(target_obj, "held") == 1.0:
```

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,169 @@
...
+            self._MoveToObj, self._MoveToBox, self._MoveToShelf
+        }
+
+    def render_state(self,
+                     state: State,
+                     task: Task,
+                     action: Optional[Action] = None,
+                     caption: Optional[str] = None) -> List[Image]:
+        # List of NextTo objects to render
+        # Added this to display what objects we are nextto
```
ronuchit, line 121: NextTo*

src/envs/repeated_nextto_painting.py
```diff
@@ -0,0 +1,169 @@
...
+            if obj.is_instance(self._obj_type) or \
+                obj.is_instance(self._box_type) or \
+                obj.is_instance(self._shelf_type):
+                if abs(
+                        state.get(self._robot, "pose_y") -
+                        state.get(obj, "pose_y")) < self.nextto_thresh:
+                    nextto_objs.append(obj)
+        # Added this to display what objects we are nextto
+        # during video rendering, as a caption
+        return super().render_state(state, task, caption="NextTo: " + \
```
ronuchit, line 134: nice use of the caption stuff :)

src/ground_truth_nsrts.py
```diff
@@ -33,7 +35,7 @@ def get_gt_nsrts(predicates: Set[Predicate],
         nsrts = _get_blocks_gt_nsrts()
     elif CFG.env == "behavior":
         nsrts = _get_behavior_gt_nsrts()  # pragma: no cover
-    elif CFG.env == "painting":
+    elif CFG.env in ("painting", "repeated_nextto_painting"):
```
ronuchit, line 38: while we're at it, just for consistency can you change the above line:
`elif CFG.env in ["blocks", "pybullet_blocks"]:`
to:
`elif CFG.env in ("blocks", "pybullet_blocks"):`
sorry and thanks!

tests/approaches/test_oracle_approach.py
ronuchit: indeed... what is going on here lol

tests/approaches/test_oracle_approach.py
```diff
@@ -596,6 +600,35 @@ def test_oracle_approach_painting():
         assert policy_solves_task(policy, test_task, env.simulate)
 
 
+def test_oracle_approach_repeated_nextto_painting():
+    """Tests for OracleApproach class with RepeatedNextToPaintingEnv."""
+    utils.reset_config({
+        "env": "repeated_nextto_painting",
+        "num_train_tasks": 3,
+        "num_test_tasks": 3,
+        "sesame_task_planning_heuristic": "hff"
```
ronuchit, line 609: interesting, it doesn't work unless you use hff instead of lmcut?

tests/approaches/test_oracle_approach.py
```diff
@@ -596,6 +600,35 @@ def test_oracle_approach_painting():
...
+                              env.action_space, train_tasks)
+    assert not approach.is_learning_based
+    for train_task in train_tasks:
+        try:
+            policy = approach.solve(train_task, timeout=100)
+            assert policy_solves_task(policy, train_task, env.simulate)
+        except ApproachTimeout:  # pragma: no cover
+            pass  # pragma: no cover
+
+    for test_task in env.get_test_tasks():
```
ronuchit, line 624: what's the speed of this test?

tests/envs/test_repeated_nextto_painting.py
```diff
@@ -0,0 +1,320 @@
...
+        state = task.init
+        assert {box_type, shelf_type, obj_type, robot_type,
+                lid_type} == {item.type
+                              for item in state}
+        if i < 3:
+            # Test rendering
+            env.render_state(state, task)
+
+
+def test_repeated_nextto_painting_failure_cases():
```
ronuchit, line 45: I don't understand why this function has to be so big. Is this a relic from a previous era, where you had copy-pasted the all the transition functions into your new file? Now that you guys are using `super()` calls really nicely, you shouldn't have to work this hard for coverage.

tests/envs/test_repeated_nextto_painting.py
```diff
@@ -0,0 +1,320 @@
...
+    box_type = [t for t in env.types if t.name == "box"][0]
+    shelf_type = [t for t in env.types if t.name == "shelf"][0]
+    lid_type = [t for t in env.types if t.name == "lid"][0]
+    assert env.action_space.shape == (8, )
+    for i, task in enumerate(env.get_test_tasks()):
+        state = task.init
+        assert {box_type, shelf_type, obj_type, robot_type,
+                lid_type} == {item.type
+                              for item in state}
+        if i < 3:
```
ronuchit, line 40: How fast are the tests in this file? If they're slow, maybe you can render fewer states?

ronuchit (commented):
lots of small comments! and great stuff!

src/envs/repeated_nextto_painting.py
wmcclinton: yapf did this but we will fix

tests/approaches/test_oracle_approach.py
wmcclinton: It takes 144.59s for 3 tests

src/envs/painting.py
wmcclinton: In the painting environment the held objects remained on the table. I had to add a change in z (to simulate lifting the object), because in the repeated_nextto_painting environment there are move options and if you move to another block the locations of the blocks would be the same.

src/envs/repeated_nextto_painting.py
wmcclinton: Since, in the painting environment the held objects remained on the table. I had to add a change in z (to simulate lifting the object), because in the repeated_nextto_painting environment there are move options and if you move to another block the locations of the blocks would be the same. Nothing really breaks, I believe I could've defined the nsrts with the object being On_Table the whole time, but I didn't like objects having the same location after movement.

tests/envs/test_repeated_nextto_painting.py
wmcclinton: Good suggestion

tests/approaches/test_oracle_approach.py
wmcclinton: It does also work with lmcut. We switched it.

tests/envs/test_repeated_nextto_painting.py
wmcclinton: Smart, your right!

src/ground_truth_nsrts.py
```diff
@@ -853,36 +946,126 @@ def placeinshelf_sampler(state: State, goal: Set[GroundAtom],
...
             x = state.get(objs[0], "pose_x")
-            y = state.get(objs[0], "pose_y")
-            z = state.get(objs[0], "pose_z")
+            if CFG.env == "painting":
+                # Always release the object where it is, to avoid the
+                # possibility of collisions with other objects.
+                y = state.get(objs[0], "pose_y")
+                z = state.get(objs[0], "pose_z")
+            elif CFG.env == "repeated_nextto_painting":
+                # Release the object at a random position on the table.
```
ronuchit, line 988: note this is now out of date due to your change @NishanthJKumar

src/ground_truth_nsrts.py
```diff
@@ -853,36 +946,127 @@ def placeinshelf_sampler(state: State, goal: Set[GroundAtom],
...
+                                 option_vars, placeontable_sampler)
         nsrts.add(placeontable_nsrt)
 
+    if CFG.env == "repeated_nextto_painting":
+
+        def moveto_sampler(state: State, goal: Set[GroundAtom],
+                           _rng: np.random.Generator,
+                           objs: Sequence[Object]) -> Array:
+            del goal  # unused
+            y = state.get(objs[1], "pose_y")  # + rng.uniform() * 0.4
```
ronuchit, line 1015: what's going on with the commented out thing here? let's either remove or add a comment describing when we might want to re-enable it in the future?

ronuchit (approved):
lgtm, one minor comment!!

src/ground_truth_nsrts.py
NishanthJKumar: oh good question; i think @wmcclinton left that in and forgot to delete; will remove!

tomsilver (approved):
awesome work!

PR #481, opened 2022-02-18
disable disabling the sampler classifier

ronuchit:
* because we want to be able to say in papers that we are robust to non unimodal stuff


tomsilver (approved):
t h x

PR #485, opened 2022-02-22
Classifier ABC

amburger66:
- adds the Classifier class
- makes MLPClassifier inherit from it
- replaces MLPClassifier in type definitions that can be more general

this is in preparation for other types of classifiers, like an MLPClassifierEnsemble I implemented


src/torch_models.py
```diff
@@ -278,7 +278,29 @@ def _normalize_data(data: Tensor) -> Tuple[Tensor, Tensor, Tensor]:
...
+
+    def fit(self, X: Array, y: Array) -> None:
+        """Train classifier on the given data.
+
+        X is multi-dimensional, y is single-dimensional.
+        """
+        raise NotImplementedError("Override me")
+
+    def forward(self, inputs: Array) -> Tensor:
+        """Pytorch forward method."""
```
tomsilver, line 292: Can probably remove this method, since not classifiers need one, intuitively.

tomsilver (commented):
Looking good thank you! I left some comments suggesting a more general implementation. If you think it's better to stick with PyTorch, though, another option is to rename Classifer to PytorchClassifier and keep what you have. I'll leave the judgement call to you!

tomsilver (approved):
Lgtm thank you!

src/torch_models.py
```diff
@@ -278,7 +279,27 @@ def _normalize_data(data: Tensor) -> Tuple[Tensor, Tensor, Tensor]:
         return (data - shift) / scale, shift, scale
 
 
-class MLPClassifier(nn.Module):
+class Classifier(abc.ABC):
+    """ABC for classifier types."""
+
+    @abc.abstractmethod
+    def fit(self, X: Array, y: Array) -> None:
+        """Train classifier on the given data.
```
ronuchit, line 287: Sorry to revive a merged PR! @amburger66 could you remove the duplicate docstrings on the fit() and classify() in MLPClassifier below? Linter won't complain bc you don't need a docstring for overriden methods, and so it's better to have fewer duplicate lines when possible.

amburger66: my bad, will do!

PR #486, opened 2022-02-22
MLPClassifierEnsemble

amburger66:
- adds the MLPClassifierEnsemble and makes it the default model in the interactive learning approach


src/torch_models.py
```diff
@@ -302,10 +302,14 @@ def classify(self, x: Array) -> bool:
 class MLPClassifier(Classifier, nn.Module):
     """MLPClassifier definition."""
 
-    def __init__(self, in_size: int, max_itr: int) -> None:
+    def __init__(self, in_size: int, max_itr: int, seed: int = None) -> None:
```
tomsilver, line 305: I'm surprised mypy doesn't complain -- shouldn't this type be `Optional[int]`?

src/torch_models.py
```diff
@@ -437,6 +435,30 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+        for i, member in enumerate(self._members):
+            print(f"Fitting member {i} of ensemble...")
+            member.fit(X, y)
+
+    def classify(self, x: Array) -> bool:
+        member_vals = []
+        for member in self._members:
+            x_normalized = member._normalize(x)  # pylint: disable=protected-access
+            member_vals.append(member(x_normalized).item())
+        avg = np.mean(member_vals)
```
tomsilver, line 456: it probably won't make a difference here, because I'm guessing most outputs are extremely close to 0 or 1, but I'm wondering whether it's preferable to aggregate the ensemble outputs before the `> 0.5`, or after. can you look for some reputable source / other code that uses ensembles for classification and see what they do? maybe scikit learn?

tomsilver (commented):
nice! a few minor things.

src/torch_models.py
amburger66: from what i can tell in the scikit source code, they have both implementations
image
aggregating before is what it does if the individual members are probabilistic, while aggregating after is called "voting". also while i was trying to figure out why the ensemble wasn't working, i printed out some values and surprisingly, not all of them were always very close to 0 or 1 (i saw like a 0.7), so i would like to keep it like this?

tomsilver: huh good to know! thanks for looking into it, sounds good.

tomsilver (approved):
lgtm! let's see if @ronuchit has any additional comments

src/torch_models.py
ronuchit: what was the resolution?

src/torch_models.py
```diff
@@ -437,6 +435,30 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+
+    def fit(self, X: Array, y: Array) -> None:
+        for i, member in enumerate(self._members):
+            print(f"Fitting member {i} of ensemble...")
+            member.fit(X, y)
+
+    def classify(self, x: Array) -> bool:
+        member_vals = []
+        for member in self._members:
+            x_normalized = member._normalize(x)  # pylint: disable=protected-access
```
ronuchit, line 454: any reason we don't want to make this public?

ronuchit (approved):
lgtm, left some questions

src/torch_models.py
amburger66: oh sorry i made that change and hadn't pushed yet

src/torch_models.py
amburger66: changed it to `Optional[int]`

PR #488, opened 2022-02-22
clip regressor outputs when classifier is disabled

tomsilver:
I don't think this matters for anything anytime soon, but was in this part of the code, and thought this implementation makes more sense.

### DIFF
--- a/src/nsrt_learning/sampler_learning.py
+++ b/src/nsrt_learning/sampler_learning.py
@@ -287,7 +287,6 @@ class _LearnedSampler:
             low = self._param_option.params_space.low
             high = self._param_option.params_space.high
             params = np.clip(params, low, high)
-            assert self._param_option.params_space.contains(params)
             return params
         while num_rejections <= CFG.max_rejection_sampling_tries:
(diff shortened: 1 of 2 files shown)


src/nsrt_learning/sampler_learning.py
```diff
@@ -281,6 +281,14 @@ def sampler(self, state: State, goal: Set[GroundAtom],
             x_lst.extend(state[goal_obj])  # add goal state
         x = np.array(x_lst)
         num_rejections = 0
+        if CFG.sampler_disable_classifier:
+            params = np.array(self._regressor.predict_sample(x, rng),
+                              dtype=self._param_option.params_space.dtype)
+            low = self._param_option.params_space.low
+            high = self._param_option.params_space.high
+            params = np.clip(params, low, high)
+            assert self._param_option.params_space.contains(params)
```
ronuchit, line 290: You don't need to assert this, it's asserted in ParameterizedOption.ground in structs.py

ronuchit (approved):
thank you!

PR #491, opened 2022-02-23
Entropy

amburger66:
- added "threshold" `interactive_query_policy`
- added "entropy" and "BALD" `interactive_score_function`s
- `MLPClassifierEnsemble` implements `logits()`
- a couple of scripts for possible experiments to run?

### DIFF
--- a/analysis/run_score_function_experiments.sh
+++ b/analysis/run_score_function_experiments.sh
@@ -11,5 +11,5 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     # ours
     python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy threshold --interactive_score_threshold 0.5 --interactive_score_function entropy --num_online_learning_cycles $CYCLES
     # silent kid
-    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy strict_best_seen --interactive_score_function trivial --num_online_learning_cycles $CYCLES
+    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy threshold --interactive_score_threshold 1.0 --interactive_score_function entropy --num_online_learning_cycles $CYCLES
 done
--- a/analysis/run_score_function_experiments.sh
+++ b/analysis/run_score_function_experiments.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+START_SEED=456
+NUM_SEEDS=10
+CYCLES=100
+FILE="analysis/submit.py"
+
+for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
+    # section kid
+    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy nonstrict_best_seen --interactive_score_function trivial --num_online_learning_cycles $CYCLES
+    # ours
+    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy threshold --interactive_score_threshold 0.5 --interactive_score_function entropy --num_online_learning_cycles $CYCLES
+    # silent kid
+    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy strict_best_seen --interactive_score_function trivial --num_online_learning_cycles $CYCLES
+done
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -150,7 +150,7 @@ class InteractiveLearningApproach(NSRTLearningApproach):
             return 0.0  # always return the same score
         if CFG.interactive_score_function == "entropy":
             return self._score_atom_set_entropy(atom_set, state)
-        if CFG.interactive_score_function == "bald":
+        if CFG.interactive_score_function == "BALD":
             return self._score_atom_set_bald(atom_set, state)
         raise NotImplementedError("Unrecognized interactive_score_function:"
                                   f" {CFG.interactive_score_function}.")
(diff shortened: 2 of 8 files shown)


analysis/run_score_function_experiments.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+START_SEED=456
+NUM_SEEDS=10
+CYCLES=100
+FILE="analysis/submit.py"
+
+for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
+    # section kid
```
tomsilver, line 9: 🤣

analysis/run_score_function_experiments.sh
```diff
@@ -0,0 +1,15 @@
...
+    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy nonstrict_best_seen --interactive_score_function trivial --num_online_learning_cycles $CYCLES
+    # ours
+    python $FILE --env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers --interactive_query_policy threshold --interactive_score_threshold 0.5 --interactive_score_function entropy --num_online_learning_cycles $CYCLES
+    # silent kid
```
tomsilver, line 13: a truly silent kid would never query about anything, but this one may query one or a few times, right? can we change this to be truly silent, to make the eventual explanation of this baseline simpler?

also just to check -- this silent kid will still be learning things and changing over the course of online interactions, even though it's not querying, because we're adding all of the collected data into the set that's used for NSRT learning. right?

src/approaches/interactive_learning_approach.py
```diff
@@ -264,6 +272,23 @@ def _query_policy(s: State) -> Optional[GroundAtomsHoldQuery]:
...
+
+        def _query_policy(s: State) -> Optional[GroundAtomsHoldQuery]:
+            ground_atoms = utils.all_possible_ground_atoms(
+                s, self._predicates_to_learn)
+            atoms_to_query = set()
+            for atom in ground_atoms:
+                score = self._score_atom_set({atom}, s)
+                if score > CFG.interactive_score_threshold:
+                    atoms_to_query.add(atom)
+                    self._best_score = score
```
tomsilver, line 287: is this needed?

src/torch_models.py
```diff
@@ -453,15 +453,20 @@ def fit(self, X: Array, y: Array) -> None:
             member.fit(X, y)
 
     def classify(self, x: Array) -> bool:
-        member_vals = []
-        for member in self._members:
-            x_normalized = member.normalize(x)
-            member_vals.append(member(x_normalized).item())
-        avg = np.mean(member_vals)
+        avg = np.mean(self.logits(x))
```
tomsilver, line 456: nice refactor!

src/torch_models.py
```diff
@@ -453,15 +453,20 @@ def fit(self, X: Array, y: Array) -> None:
...
         assert classification in [True, False]
         return classification
 
+    def logits(self, x: Array) -> Array:
+        """Return logits calculated by each member."""
+        assert x.ndim == 1
+        logits = []
+        for member in self._members:
+            x_normalized = member.normalize(x)
+            logits.append(member(x_normalized).item())
```
tomsilver, line 467: I'm getting turned around about whether the output of `member(*)` is a logit or just a probability. I think I stated the definition of logit wrong when we were talking about it earlier. A logit should be `log(p / (1- p))`, right? But I don't think that's what we want it to be here, nor do I think that's what it actually is -- I think this is really just the probability for class True. And for the calculation of entropy and the BALD objective, that's what we want it to be, too. If you agree, can we rename this function to `predict_proba`, in line with sklearn? Sorry for my confusion!

src/approaches/interactive_learning_approach.py
```diff
@@ -303,3 +328,27 @@ def _score_atom_set_frequency(self, atom_set: Set[GroundAtom]) -> float:
...
+    def _score_atom_set_bald(self, atom_set: Set[GroundAtom],
+                             state: State) -> float:
+        """Score an atom set as the sum of the BALD objectives of each atom's
+        predicate."""
+        objective = 0.0
+        for atom in atom_set:
+            x = state.vec(atom.objects)
+            logits = self._pred_to_ensemble[atom.predicate.name].logits(x)
+            entropy = utils.entropy(np.mean(logits))
+            objective += entropy - np.mean(
```
tomsilver, line 352: looks good. btw, did you ever figure out what they were talking about w.r.t. computational advantages here?

tomsilver (commented):
looking good! excited for these experiments! :)

src/approaches/interactive_learning_approach.py
amburger66: oops nope

src/torch_models.py
amburger66: oh yes, that's my bad! i forgot that's the actual definition of logit. i will rename the function

src/approaches/interactive_learning_approach.py
amburger66: hm so the original objective looks to maximize the decrease in expected posterior entropy (given the new data point x):
image
but the problems include: posteriors are high dimensional, calculating entropy is intractable, possibly infinite dimensional parameter spaces, expensive posterior updates.
however, that objective can be rewritten into the BALD (Bayesian Active Learning by Disagreement) objective (which one of these days i will verify for myself :P)
image
which calculates entropy in low-dim output space and only requires O(1) posterior updates

tomsilver: thank you! I'm not sure if I fully understand the details, but the high level makes sense -- let's talk through this in person. it looks good to me for now.

analysis/run_score_function_experiments.sh
amburger66: yep that makes sense, i'll make the silent kid truly silent :P

and yes, that's right

tomsilver (approved):
Great work!

PR #495, opened 2022-03-02
fixes to cover multistep and tests

### LINKED ISSUE #493: CoverMultistepOptions environment appears to find invalid solutions
I changed the hand regions in the environment, but solutions seem to allow the robot to place outside of the changed hand regions:

https://user-images.githubusercontent.com/8839164/155862505-3d29193f-0ba6-49c3-b161-7fa18212ec18.mp4

tomsilver:
closes #493 and should resolve the unit test failures in #479 

the issue in #493 was an edge case where the hand is placing outside of both the allowed hand region _and_ outside of the target region. before this PR, the code would think that this placement was a "place on table", which is allowed. but that's not always the case, like in this video:

https://user-images.githubusercontent.com/2816502/156375704-ed0135e9-1f9e-467a-9080-e93cc85f74df.mp4

the unit test failures in #479 were related to default hyperparameter settings being challenging for oracle planning. we want to leave the defaults alone because we want the environment to be challenging, but I made them easier in the relevant unit tests.

### DIFF
--- a/tests/envs/test_cover.py
+++ b/tests/envs/test_cover.py
@@ -530,8 +530,7 @@ def test_cover_multistep_options():
     robot = [r for r in traj.states[0] if r.name == "robby"][0]
     assert np.array_equal(traj.states[-1][robot], traj.states[-2][robot])
 
-    # Cover the case where a place is attempted outside of a hand region
-    # Check collision of held block with the floor.
+    # Cover the case where a place is attempted outside of a hand region.
     action_arrs = [
         np.array([0.05, 0., 0.], dtype=np.float32),
         np.array([0.05, 0., 0.], dtype=np.float32),
--- a/tests/envs/test_cover.py
+++ b/tests/envs/test_cover.py
@@ -530,6 +530,49 @@ def test_cover_multistep_options():
     robot = [r for r in traj.states[0] if r.name == "robby"][0]
     assert np.array_equal(traj.states[-1][robot], traj.states[-2][robot])
 
+    # Cover the case where a place is attempted outside of a hand region
+    # Check collision of held block with the floor.
+    action_arrs = [
+        np.array([0.05, 0., 0.], dtype=np.float32),
+        np.array([0.05, 0., 0.], dtype=np.float32),
+        np.array([0.05, 0., 0.], dtype=np.float32),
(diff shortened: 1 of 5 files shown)


ronuchit (commented):
lgtm, should wait for @ashayathalye 's approval

tests/envs/test_cover.py
```diff
@@ -530,6 +530,48 @@ def policy(s: State) -> Action:
...
+        np.array([0., -0.1, 0.1], dtype=np.float32),
+        np.array([0., -0.1, 0.1], dtype=np.float32),
+        np.array([0., 0., -0.1], dtype=np.float32),
+        np.array([0., 0.1, 0.], dtype=np.float32),
+    ]
+    make_video = False  # Can toggle to true for debugging
+    traj, video, _ = utils.run_policy_on_task(
+        policy, task, env.simulate, len(action_arrs),
+        env.render if make_video else None)
+    if make_video:
```
ronuchit, line 569: if keeping this, you can move the two below `pragma: no cover` to just being on this `if` statement, so there's just one

tests/envs/test_cover.py
```diff
@@ -530,6 +530,48 @@ def policy(s: State) -> Action:
...
+        np.array([0.05, 0., 0.1], dtype=np.float32),
+        np.array([0.05, 0., 0.1], dtype=np.float32),
+        np.array([0.05, 0., 0.1], dtype=np.float32),
+        np.array([0., -0.1, 0.1], dtype=np.float32),
+        np.array([0., -0.1, 0.1], dtype=np.float32),
+        np.array([0., -0.1, 0.1], dtype=np.float32),
+        np.array([0., 0., -0.1], dtype=np.float32),
+        np.array([0., 0.1, 0.], dtype=np.float32),
+    ]
+    make_video = False  # Can toggle to true for debugging
```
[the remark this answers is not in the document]
tomsilver: they are useful, for example the way that I made the new unit test was by first making a video for one of the existing ones to remember what it does, and the modifying the action sequence and rendering videos along the way

tests/envs/test_cover.py
tomsilver: did that

NishanthJKumar (commented):
LGTM!

ronuchit (approved):

PR #496, opened 2022-03-02
deprecate nonoptimal replay data

tomsilver:


ronuchit (approved):
i love when code gets deleted

PR #497, opened 2022-03-02
Add Goal Count Heuristic

williamshen-nz:
Duplicate of https://github.com/Learning-and-Intelligent-Systems/predicators/pull/494 but on a branch on the main repository

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -1050,6 +1050,8 @@ def create_task_planning_heuristic(
     if heuristic_name in _PYPERPLAN_HEURISTICS:
         return _create_pyperplan_heuristic(heuristic_name, init_atoms, goal,
                                            ground_ops, predicates, objects)
+    if heuristic_name == GoalCountHeuristic.HEURISTIC_NAME:
+        return GoalCountHeuristic(heuristic_name, init_atoms, goal, ground_ops)
     raise ValueError(f"Unrecognized heuristic name: {heuristic_name}.")
 
 
@@ -1063,6 +1065,19 @@ class _TaskPlanningHeuristic:
 
     def __call__(self, atoms: Collection[GroundAtom]) -> float:
         raise NotImplementedError("Override me!")
+
+
+class GoalCountHeuristic(_TaskPlanningHeuristic):
+    """The number of goal atoms that are not in the current state."""
+    HEURISTIC_NAME: str = "goal_count"
+
+    @functools.cached_property
+    def _goal_set(self) -> Set[GroundAtom]:
+        """Convert the goal atom Collection into an explicit Set."""
+        return set(self.goal)
+
+    def __call__(self, atoms: Collection[GroundAtom]) -> float:
+        return len(self._goal_set.difference(atoms))
 
 
 ############################### Pyperplan Glue ###############################
--- a/src/utils.py
+++ b/src/utils.py
@@ -1037,6 +1037,10 @@ def ops_and_specs_to_dummy_nsrts(
     return nsrts
 
 
+# Note: create separate `heuristics.py` module if we need to add new
+#  heuristics in the future.
+
+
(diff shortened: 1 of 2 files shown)


ronuchit (approved):
LGTM! If you want the full experience, you can try a supercloud run (see `analysis/run_supercloud_experiments.sh`). Would be fun to see the results.

src/utils.py
```diff
@@ -1065,6 +1067,19 @@ def __call__(self, atoms: Collection[GroundAtom]) -> float:
         raise NotImplementedError("Override me!")
 
 
+class GoalCountHeuristic(_TaskPlanningHeuristic):
+    """The number of goal atoms that are not in the current state."""
+    HEURISTIC_NAME: str = "goal_count"
+
+    @functools.cached_property
+    def _goal_set(self) -> Set[GroundAtom]:
+        """Convert the goal atom Collection into an explicit Set."""
```
tomsilver, line 1076: goals should already be `Set[GroundAtom]` (see the definition of `Task` in `structs.py` for example). did you run into a type checking issue without this? if so, we should probably change whichever Collection is causing the issue.

tests/test_utils.py
```diff
@@ -1339,37 +1339,86 @@ def test_operator_application():
     assert next_atoms == {pred1([cup1, plate1]), pred2([cup1, plate1])}
 
 
-def test_create_task_planning_heuristic():
+@pytest.mark.parametrize("heuristic_name, expected_heuristic_cls", [
```
tomsilver, line 1342: this is awesome, thank you for introducing the pattern!

tests/test_utils.py
```diff
@@ -1339,37 +1339,86 @@ def test_operator_application():
...
+        GroundAtom(on, [block3, table2]),
+    }) == 2
+    assert heuristic({
+        GroundAtom(on, [block1, table2]),
+        GroundAtom(on, [block2, table2]),
+        GroundAtom(on, [block3, table1]),
+    }) == 3
+
+    # Some edge cases
+    assert heuristic({}) == 3
```
tomsilver, line 1416: should this be `set()` instead?

tomsilver (commented):
awesome work! a few questions

src/utils.py
williamshen-nz:
```
@dataclass(frozen=True)
class _TaskPlanningHeuristic:
    """A task planning heuristic."""
    name: str
    init_atoms: Collection[GroundAtom]
    goal: Collection[GroundAtom]
    ...
```

It's still a `Collection[GroundAtom]` in the parent class which is why I added this here. That being said, `goal: Set[GroundAtom]` in `create_task_planning_heuristic` so we could change the typing to be a `Set` rather than a `Collection`

tests/test_utils.py
williamshen-nz: Yes! Good catch

src/utils.py
tomsilver: ah I see, yeah that was our mistake. would you mind changing it there to be a Set? thanks!

ronuchit: yeah, there are similarly two instances of `git grep "init_atoms: Colle"`, we should fix both as well!

src/utils.py
```diff
@@ -1037,6 +1037,10 @@ def ops_and_specs_to_dummy_nsrts(
     return nsrts
 
 
+# Note: create separate `heuristics.py` module if we need to add new
+#  heuristics in the future.
```
tomsilver, line 1041: nit: extra space here

tomsilver (approved):
LGTM thanks again!

src/utils.py
williamshen-nz: pycharm only highlights multiline notes when there's two spaces 😢

PR #498, opened 2022-03-02
Interaction loop refactor

tomsilver:
@ronuchit I fell down a bit of a rabbit hole and now I'm not sure if this was necessary for the other changes I'm making, but I still think this is better, because it keeps the act_policy and query_policy in lockstep.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -189,7 +189,7 @@ def _generate_interaction_results(
         task = train_tasks[request.train_task_idx]
         # While transitioning through the environment, when each state is
         # encountered, query the teacher and record the response.
-        responses: List[Optional[Response]] = []
+        request_responses: List[Optional[Response]] = []
 
         # Note that query_policy and responses need to be arguments because
         # request and responses are changing in the loop.
@@ -206,7 +206,7 @@ def _generate_interaction_results(
                 query_cost += query.cost
 
         callback = functools.partial(_process_state, request.query_policy,
-                                     responses)
+                                     request_responsese)
         policy = utils.policy_with_callback(request.act_policy, callback)
 
         traj = utils.run_policy_until(
--- a/src/main.py
+++ b/src/main.py
@@ -192,7 +192,7 @@ def _generate_interaction_results(
         request_responses: List[Optional[Response]] = []
 
         # Note that query_policy and responses need to be arguments because
-        # request and responses are changing in the loop.
+        # request and request_responses are changing in the loop.
         def _process_state(query_policy: Callable[[State], Optional[Query]],
                            responses: List[Optional[Response]],
                            s: State) -> None:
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -185,27 +186,40 @@ def _generate_interaction_results(
...
+        traj = utils.run_policy_until(
+            policy,
+            simulator,
+            task.init,
+            request.termination_function,
+            max_num_steps=CFG.max_num_steps_interaction_request)
+
+        # The environment is now in the last state, but _policy was not called
+        # yet, so we need to process the query once more.
+        callback(traj.states[-1])
```
ronuchit, line 221: This is a bit sad, but unavoidable I think...

src/main.py
```diff
@@ -185,27 +186,40 @@ def _generate_interaction_results(
...
-        # the state only when the environment is in that state.
+        # While transitioning through the environment, when each state is
+        # encountered, query the teacher and record the response.
         responses: List[Optional[Response]] = []
-        for state in traj.states:
-            query = request.query_policy(state)
+
+        # Note that query_policy and responses need to be arguments because
+        # request and responses are changing in the loop.
+        def _process_state(query_policy: Callable[[State], Optional[Query]],
```
ronuchit, line 196: I'm confused why responses needs to be an argument. Wouldn't it work fine if you just remove this argument, and let the `.append` append directly to the responses variable initialized a few lines ago?

src/main.py
```diff
@@ -185,27 +186,40 @@ def _generate_interaction_results(
     results = []
     query_cost = 0.0
```
ronuchit, line 187: Our query cost seems to be summed over all requests. This is what we want, right?

src/main.py
tomsilver: I think you're right that this would be logically fine, but pylint complains with `cell-var-in-loop`. so probably leave it, unless you want to add a pylint disable

src/main.py
tomsilver: yeah

src/main.py
ronuchit: I see. Could we at least rename this variable then? The two `responses` are really different variables.

tomsilver: done

src/main.py
```diff
@@ -185,29 +186,43 @@ def _generate_interaction_results(
...
     for request in requests:
-        # First, roll out the acting policy.
         task = train_tasks[request.train_task_idx]
+        # While transitioning through the environment, when each state is
+        # encountered, query the teacher and record the response.
+        request_responses: List[Optional[Response]] = []
+
+        # Note that query_policy and responses need to be arguments because
+        # request and responses are changing in the loop.
```
ronuchit, line 195: change to "because request_responses is changing in the loop."?

ronuchit (approved):

src/main.py
ronuchit: sorry i mean "because request and request_responses are changing in the loop."

PR #499, opened 2022-03-02
updates to environment API

tomsilver:
transitioning more toward a gym-style environment API, where the environment maintains a current internal state, and stepping with an action advances that state.


tomsilver:
@ronuchit do you think we should also change render, so that you can only render the current state/task (as in gym)?

src/envs/base_env.py
```diff
@@ -81,3 +115,24 @@ def seed(self, seed: int) -> None:
         self._train_rng = np.random.default_rng(self._seed)
         self._test_rng = np.random.default_rng(self._seed +
                                                CFG.test_env_seed_offset)
+
+    def reset(self, train_or_test: str, task_idx: int) -> State:
+        """Resets the current state to the train or test task initial state."""
+        task = self.get_task(train_or_test, task_idx)
+        self._current_state = task.init
```
ronuchit, line 122: should we also do task.init.copy() here? I know that Tasks are immutable, but didn't we see before that we could inadvertently change attributes, e.g. if we do `task.init.set(...)`. OTOH, we use this pattern in a lot of places without making copies, so maybe fine as is?

src/envs/base_env.py
```diff
@@ -81,3 +115,24 @@ def seed(self, seed: int) -> None:
...
+        # Copy to prevent external changes to the environment's state.
+        return self._current_state.copy()
+
+    def step(self, action: Action) -> State:
+        """Apply the action, and update and return the current state.
+
+        Note that this action is a low-level action (i.e., its array
```
ronuchit, line 129: "its array representation" -> "action.arr"

src/envs/base_env.py
```diff
@@ -81,3 +115,24 @@ def seed(self, seed: int) -> None:
...
+        representation is a member of self.action_space), NOT an option.
+
+        By default, this function just calls self.simulate. However,
+        environments that maintain a more complicated internal state may
```
ronuchit, line 133: that maintain a more complicated internal state, or that don't implement simulate()*

src/main.py
```diff
@@ -245,16 +244,20 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
                   f"solve with error: {e}")
             if CFG.make_failure_videos and e.info.get("partial_refinements"):
                 video = utils.create_video_from_partial_refinements(
-                    task, env.simulate, env.render,
-                    e.info["partial_refinements"])
+                    e.info["partial_refinements"], env, "test", i,
```
ronuchit, line 247: let's rename `i` in this loop to `test_task_idx`?

src/teacher.py
```diff
@@ -60,9 +60,13 @@ def _answer_Demonstration_query(
         except (ApproachTimeout, ApproachFailure):
             return DemonstrationResponse(query, teacher_traj=None)
 
-        traj, _, goal_reached = utils.run_policy_on_task(
```
ronuchit, line 63: can we just purge `run_policy_on_task` everywhere? where is it even used now? isn't it just a super light wrapper around run_policy_until?

src/utils.py
```diff
@@ -267,13 +267,47 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+
+
+def simulate_policy_until(policy: Callable[[State], Action],
+                          simulator: Callable[[State, Action],
+                                              State], init_state: State,
+                          termination_function: Callable[[State], bool],
+                          max_num_steps: int) -> LowLevelTrajectory:
     """Execute a policy from an initial state, using a simulator.
 
+    This is very similar to run_policy_until, with the difference that a
```
ronuchit, line 306: with the difference being*

src/utils.py
```diff
@@ -267,13 +267,47 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+    simulator (function that takes state as input) is assumed. This function
+    should not be used with any core code, because it violates the environment
```
ronuchit, line 308: this isn't true -- the environment API does require a simulator to be provided (it's an abstract method).

to remedy this, we could make it not abstract (and update the docstring to explain why). or, we can change this comment to say something else.

src/teacher.py
ronuchit: i looked through the code, it seems like the things that use `run_policy_on_task` are:
* demo creation
* main.py for evaluation and video creation

I feel like we should just use `run_policy_until` in both those places. All these utility methods are getting too confusing! Would be nice to just have `run_policy_until`, `simulate_policy_until`, and `option_to_trajectory` (the latter two would not be used in the final system).

(Of course we'd have to keep `run_policy_on_task` in tests/utils.py)

src/utils.py
```diff
@@ -297,51 +331,47 @@ def run_policy_until(policy: Callable[[State], Action],
...
-    return solved
-
-
 def option_to_trajectory(init_state: State,
                          simulator: Callable[[State, Action],
                                              State], option: _Option,
                          max_num_steps: int) -> LowLevelTrajectory:
     """A light wrapper around run_policy_until that takes in an option and uses
-    achieving its terminal() condition as the termination_function."""
+    achieving its terminal() condition as the termination_function.
```
ronuchit, line 367: Similar to `run_policy_on_task` -- we should really just get rid of this method and substitute in these two lines. That way it's easy to remember: `run_policy_until` = good, `simulate_policy_until` = bad.

src/teacher.py
ronuchit: after reading below, i also feel like we should just wipe out `option_to_trajectory` too

src/utils.py
```diff
@@ -267,13 +267,47 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+                          simulator: Callable[[State, Action],
+                                              State], init_state: State,
+                          termination_function: Callable[[State], bool],
+                          max_num_steps: int) -> LowLevelTrajectory:
     """Execute a policy from an initial state, using a simulator.
 
+    This is very similar to run_policy_until, with the difference that a
+    simulator (function that takes state as input) is assumed. This function
+    should not be used with any core code, because it violates the environment
+    API (which does not assume that a simulator is available).
```
ronuchit, line 309: add: Note that the environment internal state is NOT updated.

src/utils.py
```diff
@@ -267,13 +267,47 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
 
-def run_policy_until(policy: Callable[[State], Action],
-                     simulator: Callable[[State, Action], State],
-                     init_state: State, termination_function: Callable[[State],
-                                                                       bool],
+def run_policy_until(policy: Callable[[State], Action], env: BaseEnv,
+                     train_or_test: str, task_idx: int,
+                     termination_function: Callable[[State], bool],
                      max_num_steps: int) -> LowLevelTrajectory:
+    """Execute a policy on a train or test task in the environment.
```
ronuchit, line 274: "on a train or test task" -> "starting from the initial state of a train or test task"

ronuchit: clarify that the task's goal is not used

src/utils.py
```diff
@@ -267,13 +267,47 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+
+def simulate_policy_until(policy: Callable[[State], Action],
+                          simulator: Callable[[State, Action],
+                                              State], init_state: State,
+                          termination_function: Callable[[State], bool],
+                          max_num_steps: int) -> LowLevelTrajectory:
     """Execute a policy from an initial state, using a simulator.
 
+    This is very similar to run_policy_until, with the difference that a
+    simulator (function that takes state as input) is assumed. This function
```
ronuchit, line 307: there's one other critical difference: the `init_state` argument can be anything, it doesn't have to be one of the task initial states. i'd say we should highlight both these differences clearly

tests/envs/test_base_env.py
```diff
@@ -26,5 +26,19 @@ def test_create_env():
         assert isinstance(env, BaseEnv)
         other_env = get_cached_env_instance(name)
         assert env is other_env
+        train_tasks = env.get_train_tasks()
+        for idx, train_task in enumerate(train_tasks):
+            task = env.get_task("train", idx)
+            assert train_task.init.allclose(task.init)
+            task = env.get_task("train", idx)
```
ronuchit, line 33: do you need this line?

tests/envs/test_base_env.py
```diff
@@ -26,5 +26,19 @@ def test_create_env():
...
+            assert train_task.goal == task.goal
+        test_tasks = env.get_test_tasks()
+        for idx, test_task in enumerate(test_tasks):
+            task = env.get_task("test", idx)
+            assert test_task.init.allclose(task.init)
+            task = env.get_task("test", idx)
```
ronuchit, line 39: do you need this line?

tests/test_utils.py
```diff
@@ -174,8 +174,8 @@ def test_get_static_atoms():
     assert [obj.name for obj in added_atom.objects] == ["block0", "target0"]
 
 
-def test_run_policy_until():
```
ronuchit, line 177: Let's throw in some tests for `run_policy_until()`? They will be similar to these ones in `test_simulate_policy_until()`.

ronuchit (commented):
sorry about all the comments 🤕

src/utils.py
```diff
@@ -1264,12 +1294,17 @@ def create_pddl_problem(objects: Collection[Object],
...
+    env: BaseEnv,
+    train_or_test: str,
+    task_idx: int,
+    max_num_steps: int,
 ) -> Video:
-    """Create a video from a list of skeletons and partial refinements."""
+    """Create a video from a list of skeletons and partial refinements.
+
+    Note that the environment internal state is updated.
+    """
```
ronuchit, line 1307: it won't cause any problems that this method updates the internal state right? since it's only called from main.py which would always reset to a new initial state afterward?

src/envs/base_env.py
tomsilver: either way is okay with me

src/teacher.py
tomsilver: works for me!

src/utils.py
tomsilver: just change the comment

tests/envs/test_base_env.py
tomsilver: no, remove

tests/envs/test_base_env.py
tomsilver: ditto

src/utils.py
tomsilver: right

src/envs/base_env.py
```diff
@@ -13,25 +13,36 @@ class BaseEnv(abc.ABC):
...
     @abc.abstractmethod
     def simulate(self, state: State, action: Action) -> State:
         """Get the next state, given a state and an action.
 
         Note that this action is a low-level action (i.e., its array
         representation is a member of self.action_space), NOT an option.
+
+        This function is primarily used in the default option model, and
+        for implementing the default self.step(action). It is not meant to
+        be part of the "final system", where the environment is the real world.
```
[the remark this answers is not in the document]
ronuchit: ignoring this

src/envs/base_env.py
ronuchit: ignoring. we should be careful to always copy states before doing any `.set` stuff, which we already are in all our simulate functions

tomsilver (commented):
MTGL

ronuchit (approved):

PR #500, opened 2022-03-02
allow lists and tuples on the commandline

tomsilver:
examples:
* `python src/main.py --env cover --approach oracle --seed 0 --cover_target_widths "[0.03, 0.05]"`
* `python src/main.py --env cover --approach oracle --seed 0 --cover_target_widths [0.03,0.05]`

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -1417,6 +1417,14 @@ def string_to_python_object(value: str) -> Any:
         return float(value)
     except ValueError:
         pass
+    if value.startswith("["):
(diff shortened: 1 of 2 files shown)


ronuchit (approved):
Lgtm. how is it working with the tests with spaces? does float(string) in python automatically handle stripping whitespace?

tomsilver:
> Lgtm. how is it working with the tests with spaces? does float(string) in python automatically handle stripping whitespace?

Yeah I think so

PR #502, opened 2022-03-03
purge IterativeInventionApproach, a relic from times past. we will miss you, RIP

ronuchit:


tomsilver (approved):
6 months overdue, ye shall not be missed

PR #505, opened 2022-03-03
implements Monitors for rendering and teacher interaction

tomsilver:


src/envs/base_env.py
```diff
@@ -14,6 +14,7 @@ class BaseEnv(abc.ABC):
 
     def __init__(self) -> None:
         self._current_state = DefaultState  # set in reset
+        self._current_task = DefaultTask  # set in reset
```
ronuchit, line 17: should we clarify that this is only used for rendering? are there other potential use cases down the line?

src/main.py
```diff
@@ -251,19 +226,19 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
             continue
         num_found_policy += 1
         try:
+            if CFG.make_videos:
```
ronuchit, line 229: Great change! Let's change this in `tests/envs/test_cover.py` too, inside `run_policy_with_simulator_on_task`

src/structs.py
```diff
@@ -1240,6 +1243,19 @@ class DemonstrationResponse(Response):
     teacher_traj: Optional[LowLevelTrajectory]
 
 
+class Monitor(abc.ABC):
```
ronuchit, line 1246: i think this should go in utils.py? it's really not a core struct, it's more of a helpful utility?

src/teacher.py
```diff
@@ -72,3 +74,32 @@ def _answer_Demonstration_query(
                                           _is_demo=True,
                                           _train_task_idx=query.train_task_idx)
         return DemonstrationResponse(query, teacher_traj)
+
+
+@dataclass
+class TeacherInteractionMonitor(Monitor):
+    """Wraps the interaction between agent and teacher."""
```
ronuchit, line 81: expand this docstring, e.g.
"""Wraps the interaction between agent and teacher to include generating and answering queries."""

src/teacher.py
```diff
@@ -72,3 +74,32 @@ def _answer_Demonstration_query(
...
+    _request: InteractionRequest
+    _teacher: Teacher
+    _responses: List[Optional[Response]] = field(init=False)
+    _query_cost: float = field(init=False, default=0.0)
+
+    def __post_init__(self) -> None:
+        self._responses = []
```
ronuchit, line 88: instead, use the pattern in the question here:
https://stackoverflow.com/questions/53632152/why-cant-dataclasses-have-mutable-defaults-in-their-class-attributes-declaratio

while we can't do:
_responses: List[Optional[Response]] = field(init=False, default=[])
we can do:
_responses: List[Optional[Response]] = field(init=False, default_factory=list)

src/utils.py
```diff
@@ -1250,6 +1263,29 @@ def create_pddl_problem(objects: Collection[Object],
...
+
+    The render_fn is generally env.render. Note that the state is unused
+    because the environment should use its current internal state to
+    render.
+    """
+    _render_fn: Callable[[Optional[Action]], List[Image]]
+    _video: Video = field(init=False)
+
+    def __post_init__(self) -> None:
+        self._video = []
```
ronuchit, line 1278: same here as with above regarding default_factory

tests/envs/test_cover.py
```diff
@@ -339,22 +314,32 @@ def test_cover_multistep_options():
...
         np.array([0., 0., -0.1], dtype=np.float32),
     ]
-    make_video = False  # Can toggle to true for debugging
 
-    def policy(s: State) -> Action:
-        del s  # unused
-        return Action(action_arrs.pop(0))
+    policy = utils.action_arrs_to_policy(action_arrs)
+
+    # Here's an example of how to make a video within this test.
```
ronuchit, line 320: So we're making it harder to generate the videos here right? If someone wants to generate videos for one of the other tests in this file, they would have to copy paste this example, and figure out what arguments to change. What's the reason to move away from what we had before, where we could just use monitor = either None or utils.VideoMonitor(env.render)?

tests/envs/test_cover.py
```diff
@@ -339,22 +314,32 @@ def test_cover_multistep_options():
...
+    # env.reset("train", 0)
```
ronuchit, line 321: Why is this needed? (I think I know the answer, and it's related to the difference between `env.render` and `env.render_state`...)

tests/test_utils.py
```diff
@@ -1485,6 +1501,26 @@ def test_create_pddl():
 """
 
 
+def test_VideoMonitor():
+    """Tests for VideoMonitor()."""
+    env = CoverEnv()
```
ronuchit, line 1506: Videos are typically quite slow to generate. How slow is this test?

src/utils.py
```diff
@@ -392,6 +392,19 @@ def _policy(state: State) -> Action:
     return _policy
 
 
+def action_arrs_to_policy(
```
ronuchit, line 395: are there tests for this function?

tests/envs/test_cover.py
```diff
@@ -339,22 +314,32 @@ def test_cover_multistep_options():
...
+    # monitor = utils.VideoMonitor(env.render)
+    # traj = utils.run_policy_with_simulator(policy,
+    #                                        env.simulate,
+    #                                        task.init,
+    #                                        lambda _: False,
+    #                                        max_num_steps=len(action_arrs),
+    #                                        monitor=monitor)
+    # Uncomment to save the video.
```
ronuchit, line 329: do we still need this comment if we're already wrapping this whole block in a "Here's an example"?

src/main.py
ronuchit: oh, you were way ahead of me haha

src/envs/base_env.py
tomsilver: I don't have any use cases in mind right now, but it seems plausible that there would be other ones down the line, and I think that's fine API-wise. I'm going to leave it alone

src/teacher.py
tomsilver: coolio

tests/envs/test_cover.py
tomsilver: what we had before would get messy because we would have a bunch of commented out lines for actually saving the video to a file. I think this is a fine compromise, I don't think we're going to be making new videos in unit tests too frequently, but when we do, we can refer to this example to start.

tests/envs/test_cover.py
tomsilver: yeah right. since the video monitor uses render, and since render relies on the current state and current task in the env being correct, we need to reset the environment. a bit unfortunate, but since this is only for tests I think okay

tests/test_utils.py
tomsilver: 0.77 seconds (note it's only 5 frames)

tomsilver: I changed to 2 frames and it's now 0.43 seconds

tomsilver:

src/utils.py
```diff
@@ -267,10 +268,26 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
-               task_idx: int, termination_function: Callable[[State], bool],
-               max_num_steps: int) -> LowLevelTrajectory:
+class Monitor(abc.ABC):
+    """Observes states and actions during environment interaction."""
+
+    @abc.abstractmethod
+    def observe(self, state: State, action: Optional[Action]) -> None:
+        """Record a state and the action that is about to be taken.
+
+        On the last time step of a trajectory, no action is taken, so
```
ronuchit, line 278:
```
(igibson) rohan@Rohans-MacBook-Pro:~/Documents/predicators$ git grep "timestep" | wc -l
       8
(igibson) rohan@Rohans-MacBook-Pro:~/Documents/predicators$ git grep "time step" | wc -l
       0
```

ronuchit (approved):
lgtm, minor. great work!

PR #507, opened 2022-03-03
change create_env to get_cached_env_instance in ground truth nsrts

tomsilver:
now there are two methods: `create_new_env` and `get_cached_env`

in the main code path, main.py will be the only thing that calls
`create_new_env` (with `do_cache = True`), and everything else just uses
that same single env instance.

one complication is that for unit testing, we don't want to do caching
at all, so that's why we have a new setting "allow_env_caching"


ronuchit (approved):
should we check the other uses of `create_env` too?

PR #510, opened 2022-03-03
make parameterized option initiable, policy, and terminal public 

tomsilver:


ronuchit (approved):
lgtm, but maybe missing some in tests/approaches/test_random_options_approach.py and tests/test_utils.py? i see some `def _terminal` there

tomsilver:
> lgtm, but maybe missing some in tests/approaches/test_random_options_approach.py and tests/test_utils.py? i see some `def _terminal` there

Yeah I saw those and changed them but then reverted, since it's not important for this change.

PR #511, opened 2022-03-03
linear chain parameterized options

tomsilver:


src/utils.py
```diff
@@ -268,6 +268,76 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
 
 
+class LinearChainParameterizedOption(ParameterizedOption):
+    """A parameterized option implemented via a sequence of "child"
+    parameterized options.
+
+    This class is meant to help ParameterizedOption manual design.
+
+    The children are executed in order starting with the first in the sequence
+    and transitioning when the terminal function of the respect child is hit.
```
ronuchit, line 278: respective?

src/utils.py
```diff
@@ -268,6 +268,76 @@ def wrap_atom_predicates(atoms: Collection[LiftedOrGroundAtom],
...
+
+    The children are assumed to chain together, so the initiable of the next
+    child should always be True when the previous child terminates. If this
+    is not the case, an AssertionError is raised.
+
+    The children must all have the same types and params_space, which in turn
+    become the types and params_space for this ParameterizedOption.
+
+    The options have memory, which stores the current child index.
```
ronuchit, line 287: What do you mean by "The options"? Do you mean "A LinearChainParameterizedOption has memory"?

tests/test_utils.py
```diff
@@ -868,6 +868,65 @@ def test_find_substitution():
     assert assignment == {var3: plate0, var0: cup0}
 
 
+def test_LinearChainParameterizedOption():
```
ronuchit, line 871: also test what happens if empty list is passed in for children, and what happens if the options don't chain correctly

ronuchit (approved):
approving to unblock, left some comments

PR #512, opened 2022-03-04
Goal conditioned cover

czeng6:
Non-goal conditioned command

python src/main.py --env cover_multistep_options --approach oracle --seed 0 --make_videos --num_train_tasks 1 --num_test_tasks 50 --cover_num_blocks 1 --cover_num_targets 1 --cover_multistep_bimodal_goal True --cover_multistep_thr_percent 0.3 --cover_multistep_bhr_percent 0.99 --cover_block_widths "[0.12]" --cover_target_widths "[0.1]" --sesame_max_skeletons_optimized 1 --sesame_max_samples_per_step 1 --cover_multistep_goal_conditioned_sampling False

solves 5/50 tasks and goal-conditioned

python src/main.py --env cover_multistep_options --approach oracle --seed 0 --make_videos --num_train_tasks 1 --num_test_tasks 50 --cover_num_blocks 1 --cover_num_targets 1 --cover_multistep_bimodal_goal True --cover_multistep_thr_percent 0.3 --cover_multistep_bhr_percent 0.99 --cover_block_widths "[0.12]" --cover_target_widths "[0.1]" --sesame_max_skeletons_optimized 1 --sesame_max_samples_per_step 1 --cover_multistep_goal_conditioned_sampling True

solves 50/50 tasks.

PR #513, opened 2022-03-04
Purge unintentional env states

tomsilver:
A few of our environments are maintaining objects internally, even though the number of objects can change according to `CFG` settings. This is now an issue because of the change we made to `create_env`. For example, in #512, a new unit test is failing because that test changes the number of objects in cover using `reset_config`, but then oracle approach calls option model, which loads a cached cover environment, which had the wrong number of objects.

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -518,7 +518,7 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
         self._Holding = Predicate("Holding",
                                   [self._block_type, self._robot_type],
                                   self._Holding_holds)
-        # Need to override object creation because the types are now
+        # Need to override static object creation because the types are now
         # different (in terms of equality).
         self._robot = Object("robby", self._robot_type)
         # Override the original options to make them multi-step.
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -47,13 +47,7 @@ class CoverEnv(BaseEnv):
                                               policy=self._PickPlace_policy,
                                               initiable=utils.always_initiable,
                                               terminal=utils.onestep_terminal)
-        # Objects
-        self._blocks = []
-        self._targets = []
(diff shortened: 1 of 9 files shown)


tomsilver:
@ashayathalye could you spot check the changes to your environment and make sure I'm not breaking anything here? there should be just one semi-meaningful change: I'm now storing the correspondence between block and block hand regions in the block hand region state itself. I also split up the hand region object types into block and target hand region object types, but hopefully that doesn't have any practical difference.

src/envs/cover.py
```diff
@@ -505,19 +520,6 @@ def __init__(self) -> None:
                                   self._Holding_holds)
         # Need to override object creation because the types are now
```
ronuchit, line 521: does this comment need updating?

src/structs.py
```diff
@@ -128,6 +128,10 @@ def set(self, obj: Object, feature_name: str, feature_val: Any) -> None:
         idx = obj.type.feature_names.index(feature_name)
         self.data[obj][idx] = feature_val
 
+    def get_objects(self, object_type: Type) -> List[Object]:
+        """Return objects of the given type in the order of __iter__()."""
+        return [o for o in self if o.type == object_type]
```
ronuchit, line 133: This would be useful in a lot of places right? I guess it's too hard to figure out those places though. But we no longer need patterns like `for obj in state: if obj.type == blah: process(obj)`

ronuchit (approved):
ahhhh very good catch and smart change! this create_env thing is a mess (but i still think these are steps in the right direction, esp in light of upcoming physics simulator stuff)

src/structs.py
tomsilver: yeah... I'm too lazy to go hunting, do you want to? 😛

ronuchit: no

PR #518, opened 2022-03-04
create a local copy of options in option models

tomsilver:


tests/test_planning.py
```diff
@@ -245,23 +245,33 @@ def test_planning_determinism():
...
-            ns[robby][1] = 1
-        return ns
-
     goal = {asleep([robby]), asleep([robin]), cried([robby]), cried([robin])}
     task1 = Task(State({robby: [0, 0], robin: [0, 0]}), goal)
     task2 = Task(State({robin: [0, 0], robby: [0, 0]}), goal)
-    option_model = create_option_model("oracle")
-    option_model._simulator = _simulator  # pylint:disable=protected-access
+
+    class _MockEnv(BaseEnv):
```
tomsilver, line 252: @ronuchit if you don't like this pattern, we could also create one BaseMockEnv somewhere that is just BaseEnv with all of the abstractmethods not abstract, and then inherit. but I'm not sure where that would live.

src/option_model.py
```diff
@@ -3,17 +3,19 @@
...
-from predicators.src.envs import get_or_create_env
+from predicators.src.envs import BaseEnv, get_or_create_env
 from predicators.src.envs.behavior import BehaviorEnv
 
 
 def create_option_model(name: str) -> _OptionModelBase:
     """Create an option model given its name."""
     if name == "oracle":
-        return _OracleOptionModel(CFG.env)
+        env = get_or_create_env(CFG.env)
```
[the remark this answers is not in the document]
ronuchit: oh, to make testing easier?

tests/test_option_model.py
```diff
@@ -18,32 +26,46 @@ def test_default_option_model():
...
+            """A mock simulate method."""
+            next_state = state.copy()
+            obj = list(state)[0]
+            next_state.set(obj, "feat3",
+                           next_state.get(obj, "feat3") + action.arr[1])
+            return next_state
 
-    params_space = Box(-10, 10, (2, ))
+        @property
+        def options(self) -> Set[ParameterizedOption]:
```
ronuchit, line 41: we don't typically do mypy stuff in tests, and you don't do it on policy/initiable/etc below, so might as well remove here and in simulate() too

tests/test_option_model.py
```diff
@@ -18,32 +26,46 @@ def test_default_option_model():
...
+                return s[obj][0] > 50 and not s.allclose(m["start_state"])
+
+            parameterized_option = ParameterizedOption("Pick", [],
+                                                       params_space, policy,
+                                                       initiable, terminal)
+
+            return {parameterized_option}
+
+    env = _MockEnv()
+    parameterized_option = env.options.pop()
```
ronuchit, line 67: how come this pop doesn't mess up the env's options? is it because `env.options` creates a new set object every time?

tests/test_option_model.py
```diff
@@ -57,9 +79,13 @@ def terminal(s, m, o, p):
         obj4: [8, 9, 10],
         obj9: [11, 12, 13]
     })
-    model = create_option_model("oracle")
-    model._simulator = _simulate  # pylint:disable=protected-access
```
ronuchit, line 61: yay

ronuchit (approved):

tests/test_option_model.py
ronuchit: (and in test_planning.py too)

src/option_model.py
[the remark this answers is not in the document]
tomsilver: yeah

tests/test_option_model.py
tomsilver: yeah

PR #519, opened 2022-03-05
state-based demonstration query and response 

tomsilver:


src/settings.py
```diff
@@ -259,11 +259,16 @@ def get_allowed_query_type_names() -> Set[str]:
     """Get the set of names of query types that the teacher is allowed to
     answer, computed based on the configuration CFG."""
     if CFG.option_learner == "neural":
-        return {"DemonstrationQuery"}
+        return {"DemonstrationQuery", "StateBasedDemonstrationQuery"}
```
ronuchit, line 262: shouldn't we disallow DemonstrationQuery based on what we discussed?

src/structs.py
```diff
@@ -1247,6 +1247,16 @@ class DemonstrationResponse(Response):
     teacher_traj: Optional[LowLevelTrajectory]
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class StateBasedDemonstrationQuery(Query):
```
ronuchit, line 1251: i think we should rename `DemonstrationQuery` since both `DemonstrationQuery` and `StateBasedDemonstrationQuery` are currently using `DemonstrationResponse`

src/teacher.py
```diff
@@ -75,6 +80,81 @@ def _answer_Demonstration_query(
...
+    def _answer_StateBasedDemonstration_query(
+            self, state: State,
+            query: StateBasedDemonstrationQuery) -> DemonstrationResponse:
+        # The query is asking for a demonstration from the current state to
+        # the goal state. Planning to a low-level state is hard. This
+        # implementation currently assumes that only one option is required
+        # to get from the state to the goal state. Furthermore, the option
+        # is identified in an environment-specific way. This is all a stand-in
+        # for a human teacher. Note that although these trajectories are good,
+        # they are not demonstrations per se, because they do not reach task
```
ronuchit, line 92: huh? how are they not demonstrations? then this class is named incorrectly (but i do think they are demonstrations. why do demonstrations necessarily have to be associated with a task?)

src/teacher.py
```diff
@@ -75,6 +80,81 @@ def _answer_Demonstration_query(
...
+                # involved in the definition of the Place policy. The only
+                # reason that it's included in the ParameterizedOption is
+                # that there is a change in the symbolic effects for the
+                # target (it becomes covered).
+                target = targets[0]
+                arguments = [block, robot, target]
+                changing_objs = [block, robot]
+            # Case 3: invalid or unsupported request.
+            else:
+                return null_response
```
ronuchit, line 133: can this ever happen?

src/teacher.py
```diff
@@ -75,6 +80,81 @@ def _answer_Demonstration_query(
...
+                self._simulator,
+                state,
+                termination_function,
+                max_num_steps=CFG.max_num_steps_option_rollout)
+        else:
+            raise NotImplementedError("State-based demonstration queries not"
+                                      f"supported for env {CFG.env}.")
+        # Validate. If the goal state was not reached, the query is assumed
+        # to be invalid, and a None trajectory is returned.
+        if trajectory is None or not trajectory.states[-1].allclose(
```
ronuchit, line 149: what code path could possibly cause `trajectory` to be None here?

src/teacher.py
```diff
@@ -75,6 +80,81 @@ def _answer_Demonstration_query(
...
+                goal_state):
+            return null_response
+        # Strip the trajectory of options to prevent cheating.
+        for action in trajectory.actions:
```
ronuchit, line 153: smart, i would have definitely forgotten this...

src/teacher.py
```diff
@@ -75,6 +80,81 @@ def _answer_Demonstration_query(
...
+        # goals (and are not necessarily associated with any particular task).
+        trajectory = None
+        null_response = DemonstrationResponse(query, teacher_traj=None)
+        goal_state = query.goal_state
+        if CFG.env == "cover_multistep_options":
+            assert CFG.cover_multistep_use_learned_equivalents
+            # Setup.
+            block_type, target_type, robot_type = _get_types_by_names(
+                CFG.env, ["block", "target", "robot"])
+            Pick, Place = _get_options_by_names(
```
ronuchit, line 102: so you're using the Learned Equivalents rather than the basic Pick/Place options? i'm not sure i fully understand the significance of using one versus the other...

src/structs.py
ronuchit: also move this class up above the associated response?

src/teacher.py
tomsilver: yeah, one of the test cases covers it

src/teacher.py
tomsilver: good point, none, will remove

src/teacher.py
tomsilver: the learned equivalent ones have objects and parameter spaces that should exactly match those of the learned options. that's not true for Pick/Place, especially re: the parameter spaces.

tomsilver:

src/structs.py
```diff
@@ -1247,6 +1247,23 @@ class DemonstrationResponse(Response):
     teacher_traj: Optional[LowLevelTrajectory]
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class PathToStateQuery(Query):
+    """A query requesting a demonstration to reach a specific state."""
```
ronuchit, line 1252: let's update this comment to remove the word demonstration

src/teacher.py
```diff
@@ -75,6 +80,80 @@ def _answer_Demonstration_query(
...
+            trajectory = utils.run_policy_with_simulator(
+                policy,
+                self._simulator,
+                state,
+                termination_function,
+                max_num_steps=CFG.max_num_steps_option_rollout)
+        else:
+            raise NotImplementedError("State-based demonstration queries not"
+                                      f"supported for env {CFG.env}.")
+        # Validate. If the goal state was not reached, the query is assumed
```
ronuchit, line 147: how is this possible, when this function is hand-coded to do the right thing for the cover multistep env?

ronuchit (approved):

src/teacher.py
tomsilver: it shouldn't happen in practice, but it would be possible if the student erroneously asks the teacher to get to a state that's multiple options away

src/teacher.py
```diff
@@ -75,6 +80,80 @@ def _answer_Demonstration_query(
...
+                b for b in blocks if abs(goal_state.get(b, "grasp") - 1) < 1e-6
+            ]
+            # Case 1: Pick.
+            if len(goal_blocks_held) == 1:
+                parameterized_option = Pick
+                block = goal_blocks_held[0]
+                arguments = [block, robot]
+                changing_objs = [block, robot]
+            # Case 2: Place.
+            elif len(state_blocks_held) == 1:
```
NishanthJKumar, line 120: Just a question: how do we know that if we're holding a block mentioned in the goal, then the agent must be trying to pick? Same for the place case. Could you add a comment to this code explaining the logic/idea here (unless this is somewhere else) to clarify for anyone that tries to use this in the future?

NishanthJKumar (commented):
LGTM - though one minor confusion (see comment in review)

src/teacher.py
tomsilver: thanks for the review! I added a comment, let me know if that clarifies.

NishanthJKumar (commented):
This clarifies things quite a bit, thank you! Would be nice to also have a small comment in `Case 1` mentioning why the condition being checked implies that the option must be a pick.

tomsilver:
ah yeah, sorry for missing that part of your feedback the first time. added!

tomsilver:
@ashayathalye signed off on slack, so merging!

PR #520, opened 2022-03-05
singleton options and actions to option

tomsilver:
* implements `SingletonParameterizedOption` and refactors a lot of code to use it
* implements `action_sequence_to_parameterized_option`, which is not used yet but I want to use it in pybullet blocks
* fixes a bug in `LinearChainParameterizedOption` regarding memory collisions

### DIFF
--- a/src/envs/painting.py
+++ b/src/envs/painting.py
@@ -85,65 +85,59 @@ class PaintingEnv(BaseEnv):
         self._IsClean = Predicate("IsClean", [self._obj_type],
                                   self._IsClean_holds)
         # Options
-        self._Pick = ParameterizedOption(
+        self._Pick = utils.SingletonParameterizedOption(
             # variables: [robot, object to pick]
             # params: [grasp]
             "Pick",
+            self._Pick_policy,
             types=[self._robot_type, self._obj_type],
             params_space=Box(np.array([-0.01], dtype=np.float32),
                              np.array([1.01], dtype=np.float32)),
-            policy=self._Pick_policy,
-            initiable=self._handempty_initiable,
-            terminal=utils.onestep_terminal)
-        self._Wash = ParameterizedOption(
+            initiable=self._handempty_initiable)
+        self._Wash = utils.SingletonParameterizedOption(
             # variables: [robot]
             # params: []
             "Wash",
+            self._Wash_policy,
             types=[self._robot_type],
             params_space=Box(0, 1, (0, )),  # no parameters
-            policy=self._Wash_policy,
-            initiable=self._holding_initiable,
-            terminal=utils.onestep_terminal)
-        self._Dry = ParameterizedOption(
+            initiable=self._holding_initiable)
+        self._Dry = utils.SingletonParameterizedOption(
             # variables: [robot]
             # params: []
             "Dry",
+            self._Dry_policy,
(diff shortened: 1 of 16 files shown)


tomsilver:
it turns out I probably won't use `action_sequence_to_parameterized_option` after all, so we can remove it, unless it seems generally useful

src/utils.py
```diff
@@ -311,23 +288,32 @@ def __init__(self, name: str,
...
     def _initiable(self, state: State, memory: Dict, objects: Sequence[Object],
                    params: Array) -> bool:
-        # Reset the current child to the first one.
+        # Initialize the current child to the first one.
         memory["current_child_index"] = 0
+        # Create memory dicts for each child to avoid key collisions. One
+        # example of a failure that arises without this is when using
+        # action_sequence_to_parameterized_option, each of the singleton
+        # options would be referencing the same start_state in memory.
+        memory["child_memory"] = [{} for _ in self._children]
```
ronuchit, line 297: good catch. there's never going to be anything else in `memory` that we need to copy over into the child memories, right?

src/envs/painting.py
```diff
@@ -85,65 +85,59 @@ def __init__(self) -> None:
...
             # params: []
             "OpenLid",
+            self._OpenLid_policy,
             types=[self._robot_type, self._lid_type],
             params_space=Box(0, 1, (0, )),  # no parameters
-            policy=self._OpenLid_policy,
-            initiable=self._handempty_initiable,
-            terminal=utils.onestep_terminal)
+            initiable=self._handempty_initiable)
         # Static objects (always exist no matter the settings).
```
ronuchit, line 141: in this file (painting.py) can't we also wipe out the appearances of the `Always update the memory dict` stuff, like you did in playroom.py?

ronuchit: also, wipe out appearances of `params_space=Box(0, 1, (0, )),  # no parameters` since it's default?

src/utils.py
```diff
@@ -336,7 +322,90 @@ def _terminal(self, state: State, memory: Dict, objects: Sequence[Object],
...
+
+        super().__init__(name,
+                         types,
+                         params_space,
+                         policy=policy,
+                         initiable=_initiable,
+                         terminal=_terminal)
+
+
+def action_sequence_to_parameterized_option(
```
ronuchit, line 380: let's remove this if it's unused. it's in the commit history if we want to come back to it

src/utils.py
```diff
@@ -311,23 +288,32 @@ def __init__(self, name: str,
 
     def _initiable(self, state: State, memory: Dict, objects: Sequence[Object],
                    params: Array) -> bool:
-        # Reset the current child to the first one.
+        # Initialize the current child to the first one.
         memory["current_child_index"] = 0
+        # Create memory dicts for each child to avoid key collisions. One
+        # example of a failure that arises without this is when using
+        # action_sequence_to_parameterized_option, each of the singleton
```
ronuchit, line 295: if we agree on removing `action_sequence_to_parameterized_option` then update this comment too

ronuchit (commented):
awesome idea/change!! i think we should remove the unused thing

tomsilver:

PR #521, opened 2022-03-06
Flag for stopping active learning when num transitions collected

amburger66:


tomsilver (approved):
Lgtm after unit test!

ronuchit (commented):
rename to something more descrptive like online_learning_max_transitions?

tests/test_main.py
```diff
@@ -110,6 +110,17 @@ def test_main():
         "--num_test_tasks", "3", "--predicate_mlp_classifier_max_itr", "100"
     ]
     main()
+    # Cover the case where online learning stops after collecting max number of
+    # transitions.
+    sys.argv = [
```
tomsilver, line 115: how long does this test take to run?

amburger66: on my computer, `test_main` went from 6.4s to 12.6s :/

ronuchit: why not just lump this in with the test on the previous line?

tomsilver (approved):

PR #522, opened 2022-03-06
updates to slack bot

ronuchit:
* I think we don't actually want scheduled nightlies because they might
conflict with other experiments we want to run, and it would be annoying
to set up a system that pauses them (we'd have to remember to do that).
I think this is a fine compromise: a new command `launch <username>`,
which conveniently handles switching to master, pulling, wiping previous
results, and launching.
* Added git commit hash into `results/` files to help us sanity check that
results are on the commit we expect.


tomsilver (approved):
great!

PR #524, opened 2022-03-06
Fix load interactive learning approach

amburger66:
it wasn't working because `self._pred_to_ensemble` wasn't getting saved/loaded


tomsilver (approved):
good catch!

PR #526, opened 2022-03-06
fix nsrt learning with cover-multistep-options

### LINKED ISSUE #525: make nsrt_learning with cover_multistep_options work
to reproduce:

`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0`

there are at least two issues:
1. data collection times out. this is apparently resolved with `--offline_data_planning_timeout 10`, but we should verify this over many seeds and then commit it to settings.py.
2. the assertions in `_Pick_learned_equivalent_terminal` and `_Place_learned_equivalent_terminal` are broken. they should be turned off at test time, if not removed completely; their purpose was only to validate the _oracle_ samplers.

tomsilver:
closes #525

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -1073,18 +1073,8 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
                                           o: Sequence[Object],
                                           p: Array) -> bool:
         assert np.allclose(p, m["params"])
-        del p
-        absolute_params = m["absolute_params"]
-        block, robot = o
         # Pick is done when we're holding the desired object.
-        terminal = self._Holding_holds(s, [block, robot])
-        if terminal and CFG.sampler_learner == "neural":
-            # Ensure terminal state matches parameterization.
-            param_from_terminal = np.hstack((s[block], s[robot]))
-            assert np.allclose(absolute_params,
-                               param_from_terminal,
-                               atol=1e-05)
-        return terminal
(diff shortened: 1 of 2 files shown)


src/settings.py
```diff
@@ -188,6 +186,16 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
         """A workaround for global settings that are derived from the
         experiment-specific args."""
         return dict(
+            # For learning-based approaches, the data collection timeout for
+            # planning.
+            offline_data_planning_timeout=defaultdict(
```
ronuchit, line 191: why can't we just make this default across the board? none of our stuff really cares about if training might take a little bit longer right, and usually it'll succeed very fast anyway

src/envs/cover.py
```diff
@@ -1073,18 +1073,8 @@ def _Pick_learned_equivalent_terminal(self, s: State, m: Dict,
                                           o: Sequence[Object],
                                           p: Array) -> bool:
```
ronuchit, line 1074: how is this function different from Pick_terminal above? i see that `robot` is an extra argument but i'm unsure why

tomsilver: this is related to your question in the other PR. this parameterized option should match the one that we anticipate learning exactly.

tomsilver:

ronuchit (approved):

PR #527, opened 2022-03-07
script for finding a seed that fails on the first test task

tomsilver:
the idea is that you can let this run for a while, and then once you find a seed that's failing, you can dig into that one without having to wait around for a later failing test within a single seed.

### DIFF
--- a/analysis/seed_scanner.sh
+++ b/analysis/seed_scanner.sh
@@ -6,11 +6,12 @@ START_SEED=0
 NUM_SEEDS=100
 
 for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
-    if python src/main.py $FLAGS --num_test_tasks 1 --seed $SEED | grep -q 'Tasks solved: 0'
+    CMD="python src/main.py ${FLAGS} --num_test_tasks 1 --seed ${SEED}"
+    if $CMD | grep -q 'Tasks solved: 0'
     then
         echo "Found failing seed: ${SEED}."
         echo "Command to reproduce:"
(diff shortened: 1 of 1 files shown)


tomsilver:
example output:
```
$ ./analysis/seed_scanner.sh 
pybullet build time: Mar  3 2022 08:25:10
Seed 0 did not fail.
pybullet build time: Mar  3 2022 08:25:10
Seed 1 did not fail.
pybullet build time: Mar  3 2022 08:25:10
Found failing seed: 2.
Command to reproduce:
python src/main.py --approach oracle --env blocks --timeout 0.1 --num_test_tasks 1 --seed 2
```

analysis/seed_scanner.sh
```diff
@@ -0,0 +1,18 @@
...
+FLAGS="--approach oracle --env blocks --timeout 0.1"
+START_SEED=0
+NUM_SEEDS=100
+
+for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
+    if python src/main.py $FLAGS --num_test_tasks 1 --seed $SEED | grep -q 'Tasks solved: 0'
+    then
+        echo "Found failing seed: ${SEED}."
+        echo "Command to reproduce:"
+        echo "python src/main.py ${FLAGS} --num_test_tasks 1 --seed ${SEED}"
```
ronuchit, line 13: to ensure this doesn't diverge from line 9 in the future, let's make a variable `CMD="python src/main.py ${FLAGS} --num_test_tasks 1 --seed ${SEED}"` at the start of each for loop iteration, then use that both in line 9 and here

ronuchit (approved):

PR #528, opened 2022-03-07
bot sets PYTHONHASHSEED

ronuchit:
we have no idea why you can't just do `source ~/.bashrc` instead but it doesn't work


tomsilver (approved):
/shrug

PR #530, opened 2022-03-07
Add an optional caption to `render` and `render_state`

amburger66:
I'll use this to render videos labeled with query responses during active learning, and it could be used for other purposes too


ronuchit (approved):
ooh awesome idea!!! thanks!

tomsilver (approved):
beauty!

PR #531, opened 2022-03-07
purge misleading references to TAMP and motion planning

ronuchit:


src/approaches/random_options_approach.py
```diff
@@ -17,10 +17,9 @@ def is_learning_based(self) -> bool:
     def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
         options = sorted(self._initial_options, key=lambda o: o.name)
         cur_option = DummyOption
```
ronuchit, line 19: also snuck in this change (removing unused variable)

PR #532, opened 2022-03-07
add PyBulletBlocksEnv

tomsilver:
implements a PyBullet version of our blocks environment.


tomsilver:
the new test takes about 4 seconds. with `--longrun`, it goes up to about 10 seconds.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                    # by taking the closest one. This should be rare, but it
+                    # can happen when two blocks are stacked and the robot is
+                    # unstacking the top one.
+                    contact_distance = point[8]
+                    if contact_distance < closest_held_block_dist:
+                        closest_held_block = block_id
+                        closest_held_block_dist = contact_distance
+        return closest_held_block
+
+    ########################## Parameterized Options ##########################
```
ronuchit, line 583: i feel like some of these will be useful for other envs too, like PyBulletCover... we should really think about how to set things up so that common code is factored out (fine to do that in a different PR)

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            action = action * self._max_vel_norm / action_norm
+        action = np.r_[action, 0.0]
+        action = np.clip(action, self.action_space.low, self.action_space.high)
+        return Action(action.astype(np.float32))
+
+    @staticmethod
+    def _convert_rel_abs_to_abs(current_pose: Pose3D, rel_or_abs_pose: Pose3D,
+                                rel_or_abs: Tuple[str, str, str]) -> Pose3D:
+        abs_pos = []
+        for i in range(3):
```
ronuchit, line 721: range(len(current_pose)), and assert that `len(current_pose) == len(rel_or_abs_pose) == len(rel_or_abs)`

src/envs/pybullet_blocks.py
ronuchit, line 583: happy to see that there isn't too much code here!

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            current, target = get_current_and_target_pose(
+                state, objects, params)
+            return self._get_move_toward_waypoint_action(current, target)
+
+        def _terminal(state: State, memory: Dict, objects: Sequence[Object],
+                      params: Array) -> bool:
+            del memory  # unused
+            current, target = get_current_and_target_pose(
+                state, objects, params)
+            dist = np.sum(np.subtract(current, target)**2)
```
ronuchit, line 606: This is the squared distance, not the distance. Update name accordingly.

ronuchit: maybe slightly cleaner to do `np.sum(np.square(np.subtract(current, target)))`

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            return current_pose, target_pose
+
+        return self._move_to_pose(name, types, params_space,
+                                  _get_current_and_target_pose)
+
+    def _get_move_toward_waypoint_action(self, gripper_pose: Pose3D,
+                                         target_pose: Pose3D) -> Action:
+        action = self._move_gain * np.subtract(target_pose, gripper_pose)
+        action_norm = np.linalg.norm(action)  # type: ignore
+        if action_norm > self._max_vel_norm:
```
ronuchit, line 711: how often does this occur in practice, given how tiny our action space bounds are?

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                    # unstacking the top one.
+                    contact_distance = point[8]
+                    if contact_distance < closest_held_block_dist:
+                        closest_held_block = block_id
+                        closest_held_block_dist = contact_distance
+        return closest_held_block
+
+    ########################## Parameterized Options ##########################
+
+    def _move_to_pose(
```
ronuchit, line 585: it looks like this and other options are specific to *gripper* poses. we should clarify that by sprinkling in the word "gripper" throughout, in both the method names and the docstrings, as appropriate.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                                   policy=_policy,
+                                   initiable=lambda _1, _2, _3, _4: True,
+                                   terminal=_terminal)
+
+    def _move_relative_to_block(self, name: str,
+                                rel_or_abs_target_pose: Pose3D,
+                                rel_or_abs: Tuple[str, str,
+                                                  str], types: Sequence[Type],
+                                params_space: Box) -> ParameterizedOption:
+        """Creates a ParameterizedOption for moving to a target pose relative
```
ronuchit, line 621: for example, this should say "...for moving the gripper to a..."

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            return dist < self._grasp_tol
+
+        return ParameterizedOption(name,
+                                   types=types,
+                                   params_space=params_space,
+                                   policy=_policy,
+                                   initiable=lambda _1, _2, _3, _4: True,
+                                   terminal=_terminal)
+
+    def _move_relative_to_table(self, name: str,
```
ronuchit, line 676: move this function up to be next to move_relative_to_block

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                         params_space: Box) -> ParameterizedOption:
+        """Creates a ParameterizedOption for changing the robot grippers."""
+
+        assert types[0] == self._robot_type
+
+        def _policy(state: State, memory: Dict, objects: Sequence[Object],
+                    params: Array) -> Action:
+            del memory, objects, params  # unused
+            current_val = state.get(self._robot, "fingers")
+            gripper_action = target_val - current_val
```
ronuchit, line 656: if the gripper_action really is just this delta, instead of a discrete -1/0/1, then i'm now especially confused by the `if finger_action == 0:` line in step().

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                                     self.action_space.high[3])
+            return Action(
+                np.array([0., 0., 0., gripper_action], dtype=np.float32))
+
+        def _terminal(state: State, memory: Dict, objects: Sequence[Object],
+                      params: Array) -> bool:
+            del memory, objects, params  # unused
+            current_val = state.get(self._robot, "fingers")
+            dist = (target_val - current_val)**2
+            return dist < self._grasp_tol
```
ronuchit, line 667: it's interesting that grasp_tol has to be 0.05, while move_tol can be so tiny. any insight into why?

ronuchit (commented):
amazing work!!! 😍

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            physicsClientId=self._physics_client_id)
+        p.resetBasePositionAndOrientation(
+            self._table_id,
+            self._table_pose,
+            self._table_orientation,
+            physicsClientId=self._physics_client_id)
+
+        # Skip test coverage because GUI is too expensive to use in unit tests
+        # and cannot be used in headless mode.
+        if CFG.pybullet_draw_debug:  # pragma: no cover
```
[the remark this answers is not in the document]
ronuchit: while we're at it, maybe a video too

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+        self._block_id_to_block: Dict[int, Object] = {}
+
+        # When a block is held, a constraint is created to prevent slippage.
+        self._held_constraint_id: Optional[int] = None
+        self._held_block_id: Optional[int] = None
+
+    @property
+    def action_space(self) -> Box:
+        # dimensions: [dx, dy, dz, dfingers]
+        return Box(low=-0.05, high=0.05, shape=(4, ), dtype=np.float32)
```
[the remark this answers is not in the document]
ronuchit: i self-answered below

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                               physicsClientId=self._physics_client_id)
+            self._held_constraint_id = None
+            self._held_block_id = None
+
+        self._current_state = self._get_state()
+        return self._current_state.copy()
+
+    def _get_state(self) -> State:
+        """Create a State based on the current PyBullet state."""
+        state = self._current_state.copy()
```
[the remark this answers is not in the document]
ronuchit: i think a good idea is to construct it with an empty data dictionary and then at the end, assert that state.keys == self._current_state.keys()

tomsilver:
```
                                            NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED   LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                                                      
pybullet_blocks_nsrt_learning                  50.00 (0.00)  45.00 (7.64)   5.00 (0.00)   3.79 (1.15)  3375.92 (1933.16)  237.12 (90.40)         10
pybullet_blocks_oracle                         50.00 (0.00)  49.50 (0.67)   5.00 (0.00)   3.82 (1.80)  3706.14 (2178.78)     0.00 (0.00)         10
```
results from a local run. seed 2 is bad for `pybullet_blocks_nsrt_learning`. we believe it's due to sampler overfitting. changing `num_train_tasks` to 25 or 200 resolves it. no further action right now

src/envs/pybullet_blocks.py
ronuchit: also, this is again the squared dist, update name accordingly

tomsilver:
on supercloud (different start seed)
```
pybullet_blocks_nsrt_learning        50.00 (0.00)  48.00 (3.07)   5.00 (0.00)   2.90 (0.27)  4321.57 (2410.62)   179.84 (29.26)         10
pybullet_blocks_oracle                0.00 (0.00)  49.80 (0.40)   5.00 (0.00)   2.81 (0.24)  4233.17 (2193.30)      0.00 (0.00)         10
```

src/option_model.py
```diff
@@ -18,6 +18,10 @@ def create_option_model(name: str) -> _OptionModelBase:
         return _OracleOptionModel(env)
     if name == "behavior_oracle":
         return _BehaviorOptionModel()  # pragma: no cover
+    if name.startswith("oracle"):
+        _, env_name = name.split("_")
```
ronuchit, line 22: Let's change this to `env_name = name[name.index("_")+1:]` in anticipation of future things like oracle_cover_regrasp

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+    _finger_action_nudge_magnitude: float = 0.001
+
+    # Table parameters.
+    _table_pose: Pose3D = (1.35, 0.75, 0.0)
+    _table_orientation: Sequence[float] = [0., 0., 0., 1.]
+
+    # Block parameters.
+    _block_orientation: Sequence[float] = [0., 0., 0., 1.]
+    _block_mass = 0.5
+    _block_friction = 1.2
```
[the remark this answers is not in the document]
tomsilver: I didn't tune the friction at all, just took it from our old repo. I did increase the mass, which helped in certain situations where picking a block off another block could cause the latter block to twist.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+        (0.4, 0.05, 0.6, 1.),
+        (0.6, 0.4, 0.05, 1.),
+        (0.05, 0.04, 0.6, 1.),
+        (0.95, 0.95, 0.1, 1.),
+        (0.95, 0.05, 0.95, 1.),
+        (0.05, 0.95, 0.95, 1.),
+    ]
+    _out_of_view_xy: Sequence[float] = [10.0, 10.0]
+
+    # Camera parameters.
```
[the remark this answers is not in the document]
tomsilver: yeah, it generates videos

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+    _camera_pitch: float = -24
+    _camera_target: Pose3D = (1.65, 0.75, 0.42)
+    _debug_text_position: Pose3D = (1.65, 0.25, 0.75)
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # Override options.
+        types = [self._robot_type, self._block_type]
+        params_space = Box(0, 1, (0, ))
```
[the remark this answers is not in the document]
tomsilver: yeah that's what I had in mind. this happens to some degree already here, in PutOnTable. only a few of the children in the chain actually use the parameters.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                self._camera_yaw,
+                self._camera_pitch,
+                self._camera_target,
+                physicsClientId=self._physics_client_id)
+        else:
+            self._physics_client_id = p.connect(p.DIRECT)
+
+        p.resetSimulation(physicsClientId=self._physics_client_id)
+
+        # Load plane.
```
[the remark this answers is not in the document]
tomsilver: agreed

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            self._fetch_id,
+            self._base_pose,
+            self._base_orientation,
+            physicsClientId=self._physics_client_id)
+
+        # Extract IDs for individual robot links and joints.
+        joint_names = [
+            p.getJointInfo(
+                self._fetch_id, i,
+                physicsClientId=self._physics_client_id)[1].decode("utf-8")
```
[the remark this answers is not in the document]
tomsilver: they're bytestrings for whatever reason /shrug we had this in the old code too

src/envs/pybullet_blocks.py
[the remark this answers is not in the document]
tomsilver: (sent over slack)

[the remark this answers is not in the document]
tomsilver: I tried to do this, but it doesn't work because the drawing needs to happen after resetSimulation. but yes, this can only happen if the GUI is on. I'll add an assert

src/envs/pybullet_blocks.py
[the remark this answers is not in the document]
tomsilver: I didn't actually try anything larger (just used what we had before). I think you're right about the precision, and relatedly, we want to keep the velocity near 0 because velocity isn't modeled in the state.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            action: Optional[Action] = None,
+            caption: Optional[str] = None) -> List[Image]:  # pragma: no cover
+        # Skip test coverage because GUI is too expensive to use in unit tests
+        # and cannot be used in headless mode.
+        del caption  # unused
+
+        if not CFG.pybullet_use_gui:
+            raise Exception(
+                "Rendering only works with GUI on. See "
+                "https://github.com/bulletphysics/bullet3/issues/1157")
```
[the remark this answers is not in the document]
tomsilver: oh lol that's cool! I did not notice

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+    def step(self, action: Action) -> State:
+        ee_delta, finger_action = action.arr[:3], action.arr[3]
+
+        ee_link_state = p.getLinkState(self._fetch_id,
+                                       self._ee_id,
+                                       physicsClientId=self._physics_client_id)
+        current_position = ee_link_state[4]
+        target_position = np.add(current_position, ee_delta).tolist()
+
+        # We assume that the robot is already close enough to the target
```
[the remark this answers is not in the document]
tomsilver: yeah I think that's part of it at least (responded to the other comment too)

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+        # position that IK will succeed with one call, so validate is False.
+        # Furthermore, updating the state of the robot during simulation, which
+        # validate=True would do, is risky and discouraged by PyBullet.
+        joint_values = inverse_kinematics(
```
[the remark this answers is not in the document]
tomsilver: I would rather leave it like this because if we copy and paste this code into another repository some day (like I did in this instance), it's more obvious that we just need to change the default in the kwarg, vs having to introspect the function for CFGs.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                target_val = current_val + finger_action
+            p.setJointMotorControl2(bodyIndex=self._fetch_id,
+                                    jointIndex=finger_id,
+                                    controlMode=p.POSITION_CONTROL,
+                                    targetPosition=target_val,
+                                    physicsClientId=self._physics_client_id)
+
+        # Step the simulation here before adding or removing constraints
+        # because detect_held_block() should use the updated state.
+        for _ in range(CFG.pybullet_sim_steps_per_action):
```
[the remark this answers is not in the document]
tomsilver: I just copied it from our old code. I agree that it is somewhat important, but why do you think extremely?

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+        rf = p.getJointState(self._fetch_id,
+                             self._left_finger_id,
+                             physicsClientId=self._physics_client_id)[0]
+        state.set(self._robot, "pose_x", rx)
+        state.set(self._robot, "pose_y", ry)
+        state.set(self._robot, "pose_z", rz)
+        state.set(self._robot, "fingers", rf)
+
+        # Get block states.
+        for block_id, block in self._block_id_to_block.items():
```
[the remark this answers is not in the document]
tomsilver: yeah, also `self._held_block_id.` I'll add a comment in the docstring

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+
+        return state
+
+    def _detect_held_block(self) -> Optional[int]:
+        """Return the PyBullet object ID of the held object if one exists.
+
+        If multiple blocks are within the grasp tolerance, return the
+        one that is closest.
+        """
+        expected_finger_normals = {
```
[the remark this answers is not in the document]
tomsilver: pretty much :P but it was obvious from the values

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+            for finger_id, expected_normal in expected_finger_normals.items():
+                # Find points on the block that are within grasp_tol distance
+                # of the finger.
+                closest_points = p.getClosestPoints(
+                    bodyA=self._fetch_id,
+                    bodyB=block_id,
+                    distance=self._grasp_tol,
+                    linkIndexA=finger_id,
+                    physicsClientId=self._physics_client_id)
+                for point in closest_points:
```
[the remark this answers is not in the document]
tomsilver: yeah. also if you insist on contacts, the block gets jostled around way more

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                    # If the contact normal is substantially different from
+                    # the expected contact normal, this is probably a block
+                    # on the outside of the gripper, rather than the inside.
+                    # A perfect score here is 1.0 (normals are unit vectors).
+                    contact_normal = point[7]
+                    if expected_normal.dot(contact_normal) < 0.5:
+                        continue
+                    # Handle the case where multiple blocks pass this check
+                    # by taking the closest one. This should be rare, but it
```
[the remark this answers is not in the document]
tomsilver: the pick tests that involve picking off the top of a stack cover this

src/envs/pybullet_blocks.py
tomsilver: agreed. ignoring for now

src/envs/pybullet_blocks.py
tomsilver: I'm reserving the word "gripper" to mean "the left and right fingers". I'll sprinkle the word "end effector"

tomsilver: actually I'll just use fingers and end effector and purge gripper

src/settings.py
```diff
@@ -211,6 +209,16 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
         experiment-specific args."""
 
         return dict(
+            # Maximum number of steps to run a policy when checking if it
+            # solves a task.
+            horizon=defaultdict(
```
williamshen-nz, line 214: Maybe I'm missing something but why not do something like

```python
DEFAULT_HORIZON = 100

dict(
  horizon={
      "behavior": 1000,
      "pybullet_blocks": 1000,
  }.get(args.get("env"), DEFAULT_HORIZON)
)
```

Rather than have a default dict? Seems like the defaultdict pattern is repeated with other configs in settings as well

src/envs/__init__.py
```diff
@@ -67,6 +69,8 @@ def create_new_env(name: str, do_cache: bool = False) -> BaseEnv:
         env = BehaviorEnv()  # pragma: no cover
     elif name == "repeated_nextto":
         env = RepeatedNextToEnv()
+    elif name == "pybullet_blocks":
+        env = PyBulletBlocksEnv()
```
williamshen-nz, line 73: Just a comment for the future if we have lots and lots of environments.

Might be worth walking through all the modules in the `src/envs` directory and automatically discovering subclasses of `BaseEnv`, rather than having to manually add it here every single time a new environment is created?

src/envs/pybullet_blocks.py
tomsilver: it's zero for other options, not this one. but I changed that line anyway

src/envs/pybullet_blocks.py
tomsilver: it's because the fingers are way more fickle than the end effector (related to the nudging stuff).

src/envs/pybullet_blocks.py
tomsilver: the difference between target_pose and gripper_pose can be very large, so this occurs all the time

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+from predicators.src.envs.pybullet_utils import get_kinematic_chain, \
+    inverse_kinematics
+from predicators.src.settings import CFG
+
+
+class PyBulletBlocksEnv(BlocksEnv):
+    """PyBullet Blocks domain."""
+    # Parameters that aren't important enough to need to clog up settings.py
+
+    # Fetch robot parameters.
```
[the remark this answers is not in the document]
williamshen-nz: Maybe we could slowly build up our own version of what will essentially be a replacement for pybullet-planning in another module within predicators

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                self._change_grippers("Grasp", self.closed_fingers, types,
+                                      params_space),
+                # Move up.
+                self._move_relative_to_block(
+                    "MoveToAboveBlock", (0., 0., self.pick_z),
+                    ("rel", "rel", "abs"), types, params_space),
+            ])
+
+        types = [self._robot_type, self._block_type]
+        params_space = Box(0, 1, (0, ))
```
williamshen-nz: Repeated from above? I can see why `params_space` might need to be a different object but `types` seems identical

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,728 @@
...
+                    "MoveToAboveBlock", (0., 0., self.pick_z),
+                    ("rel", "rel", "abs"), types, params_space),
+            ])
+
+        types = [self._robot_type]
+        params_space = Box(0, 1, (2, ))  # normalized w.r.t. workspace
+        self._PutOnTable = utils.LinearChainParameterizedOption(
+            "PutOnTable",
+            [
+                # Move to above the block on which we will stack.
```
williamshen-nz, line 121: nit: comment should probably say move to above the table rather than block

tomsilver:

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+                # Move up.
+                self._move_end_effector_relative_totable(
+                    "MoveEndEffectorToAboveTable", (0., 0., self.pick_z),
+                    ("rel", "rel", "abs"), types, params_space),
+            ])
+
+        # One-time initialization of pybullet assets. Note that this happens
+        # in __init__ because many instance attributes are created.
+        # Skip test coverage because GUI is too expensive to use in unit tests
+        # and cannot be used in headless mode.
```
williamshen-nz, line 142: `__init__` is getting super long. I could imagine a lot of this pybullet stuff will be repeated for new environments as well, so might be worth breaking out at least to different instance methods for readability.

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+                self._block_orientation,
+                physicsClientId=self._physics_client_id)
+
+        # Assert that the state was properly reconstructed.
+        reconstructed_state = self._get_state()
+        if not reconstructed_state.allclose(state):
+            print("Desired state:")
+            print(state.pretty_str())
+            print("Reconstructed state:")
+            print(reconstructed_state.pretty_str())
```
williamshen-nz, line 340: Just a general comment but might be worth slowly migrating to logging from print in the long term - would make it easier to debug things as well

src/envs/pybullet_blocks.py
williamshen-nz: Looking below maybe I can start moving things into a `predicators/src/pybullet_helpers` module or something when I start working on my pybullet stuff

tests/envs/test_pybullet_blocks.py
```diff
@@ -0,0 +1,498 @@
...
+            self.step(action)
+        return self._current_state.copy()
+
+
+def _get_exposed_pybullet_env():
+    global _EXPOSED_PYBULLET_ENV  # pylint:disable=global-statement
+    if _EXPOSED_PYBULLET_ENV is None:
+        _EXPOSED_PYBULLET_ENV = _ExposedPyBulletBlocksEnv()
+    return _EXPOSED_PYBULLET_ENV
+
```
williamshen-nz, line 72: There probably is a way to do this with pytest fixtures scoped to the module. So the fixture function is only invoked once for this entire test module.

https://docs.pytest.org/en/6.2.x/fixture.html

Doesn't matter too much though

src/settings.py
ronuchit: i prefer this pattern because we want to keep the default value close to the overriden values (it would be bad to have DEFAULT_HORIZON at the top of file but then you have to scroll to some other place to figure out what the horizon actually is for your environment)

src/envs/__init__.py
ronuchit: yeah, i'm in favor of this! feel free to open a PR if you know how to do it :)

ronuchit: one problem will be the string names here; how would we handle that in your proposal?

src/envs/pybullet_blocks.py
ronuchit: there's already a predicators/src/pybullet_utils.py where a lot of this stuff can be moved to. in the meantime though, why can't we just have the _initialize_pybullet() method like we usually do, called from __init__?

src/envs/pybullet_blocks.py
ronuchit: i agree. there aren't that many print statements in our code (we try to keep them out unless they're essential), so it wouldn't take too long to do this. do you want to open an issue with some pointers regarding the best library to use @williamshen-nz ? i can work on it at some point

tests/envs/test_pybullet_blocks.py
ronuchit: it would be cool to get this to work, which would allow us to remove the ugly `global` statement too!

README.md
```diff
@@ -51,13 +51,13 @@ Please make sure to `export PYTHONHASHSEED=0` when running the code. You can add
...
 * `--behavior_randomize_init_state` can be set to True if you want to generate multiple different initial states that correspond to the BDDL init conditions of a particular task.
 * If you'd like to see a visual of the agent planning in iGibson, set the command line argument `--behavior_mode simple`. If you want to run in headless mode without any visuals, leave the default (i.e `--behavior_mode headless`).
-* Example command: `python src/main.py --env behavior --approach oracle --seed 0 --timeout 1000 --sesame_max_samples_per_step 20 --behavior_mode simple --max_num_steps_check_policy 1000 --option_model_name behavior_oracle --num_train_tasks 2 --num_test_tasks 2 --behavior_randomize_init_state True --behavior_scene_name Pomaria_1_int --behavior_task_name re-shelving_library_books`.
+* Example command: `python src/main.py --env behavior --approach oracle --seed 0 --timeout 1000 --sesame_max_samples_per_step 20 --behavior_mode simple --option_model_name behavior_oracle --num_train_tasks 2 --num_test_tasks 2 --behavior_randomize_init_state True --behavior_scene_name Pomaria_1_int --behavior_task_name re-shelving_library_books`.
```
ronuchit, line 60: update option_model_name

README.md
```diff
@@ -51,13 +51,13 @@ Please make sure to `export PYTHONHASHSEED=0` when running the code. You can add
 
 ### Running Experiments on BEHAVIOR
 * Currently, only the `oracle` approach is implemented to integrate with BEHAVIOR.
-* Note that you'll probably want to provide the command line arguments `--timeout 1000`, `--max_num_steps_check_policy 1000` to prevent early stopping.
+* Note that you'll probably want to provide the command line argument `--timeout 1000` to prevent early stopping.
 * Set `--option_model_name behavior_oracle` to use the behavior option model and speed up planning by a significant factor.
```
ronuchit, line 55: update option_model_name

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+                    ("rel", "rel", "abs"), types, params_space),
+            ])
+
+        types = self._PutOnTable.types
+        params_space = self._PutOnTable.params_space
+        self._PutOnTable: ParameterizedOption = \
+            utils.LinearChainParameterizedOption("PutOnTable",
+            [
+                # Move to above the block on which we will stack.
+                self._move_end_effector_relative_totable(
```
ronuchit, line 122: add a space: "relative_to_table" to match "relative_to_block" for the other one

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+        self._block_id_to_block: Dict[int, Object] = {}
+
+        # When a block is held, a constraint is created to prevent slippage.
+        self._held_constraint_id: Optional[int] = None
+        self._held_block_id: Optional[int] = None
+
+    @property
+    def action_space(self) -> Box:
+        # dimensions: [dx, dy, dz, dfingers]
+        return Box(low=-0.05, high=0.05, shape=(4, ), dtype=np.float32)
```
ronuchit, line 263: if velocity isn't modeled in the state, do we need to require it to be 0 in the options' terminal functions?

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+        # Get block states.
+        for block_id, block in self._block_id_to_block.items():
+            (bx, by, bz), _ = p.getBasePositionAndOrientation(
+                block_id, physicsClientId=self._physics_client_id)
+            held = (block_id == self._held_block_id)
+            # pose_x, pose_y, pose_z, held
+            state_dict[block] = np.array([bx, by, bz, held], dtype=np.float32)
+
+        state = State(state_dict)
+        assert set(state) == set(self._current_state)
```
ronuchit, line 562: add an informative error message

src/envs/pybullet_blocks.py
tomsilver: @ronuchit how would we deal with all of the class instances?

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+                    # unstacking the top one.
+                    contact_distance = point[8]
+                    if contact_distance < closest_held_block_dist:
+                        closest_held_block = block_id
+                        closest_held_block_dist = contact_distance
+        return closest_held_block
+
+    ########################## Parameterized Options ##########################
+
+    def _move_end_effector_to_pose(
```
ronuchit, line 611: For these four methods:
`_move_end_effector_to_pose`
`_move_end_effector_relative_to_block`
`_move_end_effector_relative_totable`
`_change_fingers`
let's rename them to `_create_*_option`, e.g.:
`_create_move_end_effector_to_pose_option`
`_create_change_fingers_option`

ronuchit (commented):
couple more minor things, almost there!

src/envs/pybullet_blocks.py
ronuchit: idgi, let's discuss on slack

tomsilver: TIL that linting only complains with `attribute-defined-outside-init` when the attribute is modified elsewhere in the class

src/envs/pybullet_blocks.py
tomsilver: (resolved on slack: this can't happen, because the terminal function is a function of state)

src/envs/__init__.py
williamshen-nz: We could add an abstract class method to BaseEnv and use that to compare strings

src/envs/pybullet_blocks.py
williamshen-nz: Given that there aren't too many we could just use python `logging` and write some wrappers/helpers around that. 

Otherwise, there's stuff like `loguru` which can handle setups and everything in one import: https://github.com/Delgan/loguru

src/envs/__init__.py
ronuchit: that would work, but may be just as annoying as the current solution? we can let @tomsilver tiebreak

tests/envs/test_pybullet_blocks.py
williamshen-nz: I can work on refactoring this when I work on my pybullet stuff, as I'll imagine needing to do something similar

src/envs/pybullet_blocks.py
```diff
@@ -0,0 +1,754 @@
...
+                                   params_space=params_space,
+                                   policy=_policy,
+                                   initiable=lambda _1, _2, _3, _4: True,
+                                   terminal=_terminal)
+
+    def _get_end_effector_move_toward_waypoint_action(
+            self, ee_pose: Pose3D, target_pose: Pose3D) -> Action:
+        action = self._move_gain * np.subtract(target_pose, ee_pose)
+        action_norm = np.linalg.norm(action)  # type: ignore
+        if action_norm > self._max_vel_norm:
```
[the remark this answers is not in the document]
tomsilver: so suppose that `action` is `[100, 1, 0]`. if we removed this if statement, then the return would be `[0.05, 0.05, 0]`. but instead, we would want it to be closer to `[0.049, 0.01, 0]` or something. (right?) because we want to move in the right direction, toward the target pose

src/envs/__init__.py
williamshen-nz: Something like this (simplified) should work and only require giving each env a name. `discover_environments` would discover any new environments so I would think it's slightly easier.

```python
for env in discover_environments():
    if name == env.name():
        return env
```
    
That being said all this isn't a big issue

tomsilver: I like @williamshen-nz 's proposal, it's nice to let environments have names anyway.

tomsilver: btw, we may want to use a similar pattern elsewhere, e.g., `create_approach`

ronuchit (approved):
amazing!!!!!! we should rename ourselves to the pybulletors

tomsilver:
final supercloud run before merging:
```
pybullet_blocks_nsrt_learning    50.00 (0.00)  48.30 (2.57)   5.00 (0.00)   2.99 (0.22)  4066.66 (2111.86)  181.56 (17.54)         10
pybullet_blocks_oracle            0.00 (0.00)  49.30 (0.78)   5.00 (0.00)   2.97 (0.35)  4233.17 (2193.30)     0.00 (0.00)         10
```

PR #534, opened 2022-03-07
GNN core code mostly copy-pasted from previous projects

ronuchit:
the only change is that i moved some of the normalization functions into
gnn_utils.py: `compute_normalizers` and `normalize_graph`

### DIFF
--- a/src/approaches/__init__.py
+++ b/src/approaches/__init__.py
@@ -8,7 +8,8 @@ from predicators.src.approaches.random_actions_approach import \
     RandomActionsApproach
 from predicators.src.approaches.random_options_approach import \
     RandomOptionsApproach
-from predicators.src.approaches.tamp_approach import TAMPApproach
+from predicators.src.approaches.bilevel_planning_approach import \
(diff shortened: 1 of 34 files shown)


src/gnn/gnn.py
```diff
@@ -0,0 +1,331 @@
+"""Graph neural network code. Here for learning search guidance.
```
tomsilver, line 1: outdated comment -- check for others

src/approaches/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Default imports for approaches folder."""
```
tomsilver, line 1: weird, why is this only being caught now (by the linter)?

tomsilver (approved):

src/approaches/__init__.py
ronuchit: i changed a line in `pylintrc`:
```
# Tells whether we should check for unused import in __init__ files.
- init-import=no
+ init-import=yes
```

PR #535, opened 2022-03-07
Video rendering during teacher interaction

amburger66:
- modified `TeacherInteractionMonitor` to include features of `VideoMonitor`


tomsilver:
nice! what are your thoughts about consolidating `TeacherInteractionVideoMonitor` and `TeacherInteractionMonitor` and just having the class check the `CFG.make_interaction_videos` flag internally to decide whether to render?

amburger66:
> nice! what are your thoughts about consolidating `TeacherInteractionVideoMonitor` and `TeacherInteractionMonitor` and just having the class check the `CFG.make_interaction_videos` flag internally to decide whether to render?

I had that before but then I realized that adds like 2 minutes of unit test time on my computer...

src/approaches/interactive_learning_approach.py
```diff
@@ -265,6 +265,8 @@ def _query_policy(s: State) -> Optional[GroundAtomsHoldQuery]:
                    (not strict and score >= self._best_score):
                     atoms_to_query.add(atom)
                     self._best_score = score
+            if not atoms_to_query:
```
tomsilver, line 268: interesting, why this change? will this have the effect that the episode terminates as soon as there are no more atoms to query? but wouldn't that not be what we want, if for example, some tantalizing atoms are waiting for us later on in the episode?

src/approaches/interactive_learning_approach.py
```diff
@@ -281,6 +283,8 @@ def _query_policy(s: State) -> Optional[GroundAtomsHoldQuery]:
                 score = self._score_atom_set({atom}, s)
                 if score > CFG.interactive_score_threshold:
                     atoms_to_query.add(atom)
+            if not atoms_to_query:
```
tomsilver, line 286: ditto

src/teacher.py
```diff
@@ -77,23 +77,32 @@ def _answer_Demonstration_query(
 
 
 @dataclass
-class TeacherInteractionMonitor(utils.Monitor):
-    """Wraps the interaction between agent and teacher to include generating
-    and answering queries."""
+class TeacherInteractionMonitor(utils.VideoMonitor):
```
tomsilver, line 80: nice inheritance!

src/teacher.py
```diff
@@ -77,23 +77,32 @@ def _answer_Demonstration_query(
...
+    """A monitor that renders each state and action encountered and queries and
+    responses during interaction with the teacher.
+
+    The render_fn is generally env.render.
+    """
     _request: InteractionRequest
     _teacher: Teacher
+    _render_fn: Callable[[Optional[Action], Optional[str]], List[Image]]
```
tomsilver, line 88: I would've expected that you don't need this or `_video` because the parent class already has those fields. Am I right?

tomsilver: Also, isn't the argument order here different from how it's invoked?

tomsilver (commented):
really great stuff!

src/approaches/interactive_learning_approach.py
amburger66: ohhh I forgot that's what the `None` value means, I was annoyed b/c it made my videos ugly but I'll change it back lmao

amburger66:
@tomsilver can I get some help on why this test is failing? :")

tomsilver:
@amburger66 I just pushed some test fixes! (hopefully)

src/approaches/interactive_learning_approach.py
tomsilver: is the ugliness because of the empty set? feel free to special case that in the caption generation

src/teacher.py
```diff
@@ -77,23 +77,30 @@ def _answer_Demonstration_query(
 
 
 @dataclass
-class TeacherInteractionMonitor(utils.Monitor):
-    """Wraps the interaction between agent and teacher to include generating
-    and answering queries."""
+class TeacherInteractionMonitor(utils.VideoMonitor):
+    """A monitor that renders each state and action encountered and queries and
```
tomsilver, line 81: nit: grammar

tomsilver (approved):
LGTM! but let's let @ronuchit review too

src/teacher.py
ronuchit: Let's rewrite this comment. The important thing about this Monitor isn't that it renders, it's about the querying. Maybe: "A monitor that wraps interaction with an environment to include generating and answering queries. Optionally, if CFG.make_interaction_videos is True, also renders every state and action encountered."

src/teacher.py
```diff
@@ -77,23 +77,30 @@ def _answer_Demonstration_query(
...
+            response = None
+            caption = "None"
         else:
-            self._responses.append(self._teacher.answer_query(state, query))
+            response = self._teacher.answer_query(state, query)
             self._query_cost += query.cost
+            caption = f"{response}, cost={query.cost}"
+        self._responses.append(response)
+        if CFG.make_interaction_videos:
+            self._video.extend(self._render_fn(action, caption))
```
ronuchit, line 103: @tomsilver are we 1000% sure that we want to be using the stateful render function here? We are given a state in this function, and I don't see reset being called anywhere. So I think the other one is the right one?

src/structs.py
```diff
@@ -1223,12 +1223,25 @@ class GroundAtomsHoldQuery(Query):
     def cost(self) -> float:
         return len(self.ground_atoms)
 
+    def __str__(self) -> str:
```
ronuchit, line 1226: Where is this used? It doesn't seem to be part of the caption in the generated video. I'm confused how it's getting covered.

src/teacher.py
```diff
@@ -77,23 +77,30 @@ def _answer_Demonstration_query(
 
 
 @dataclass
-class TeacherInteractionMonitor(utils.Monitor):
-    """Wraps the interaction between agent and teacher to include generating
-    and answering queries."""
+class TeacherInteractionMonitor(utils.VideoMonitor):
```
ronuchit, line 80: I'm not really a fan of this inheritance.. there's nothing inherent about a teacher that makes it have to be a VideoMonitor. This is one of those rare instances where multiple inheritance feels better. One idea would be:
```
class Monitor(abc.ABC):
  def observe(self, state: State, action: Optional[Action]) -> None:

class VideoMonitor(Monitor):
  def observe(self, state: State, action: Optional[Action]) -> None:
  def get_video(self) -> Video:

class TeacherInteractionMonitor(Monitor):
  (as before)

class TeacherInteractionMonitorWithVideo(TeacherInteractionMonitor, VideoMonitor):
  (an observe() method that calls both parents')
```
This might be messy though. Open to other thoughts as well.

src/teacher.py
tomsilver: yeah we want the stateful renderer. for example, pybullet environments will render based on their current internal state (which can include things that are not expressed in the State).

ronuchit: yup, verified the state will always equal the `env._current_state`

src/structs.py
amburger66: uhhh I can't tell either, I was using it for debugging. should I get rid of it?

ronuchit: you can keep it, but is there a reason you don't want it in your video caption?

ronuchit: oh bc it's redundant with the info in the response str?

ronuchit: wateva, fine to keep it

ronuchit: for closure's sake: it's called as a total fluke, by this method https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/test_online_learning.py#L73

ronuchit: in particular the "Disallowed Query" part prints out the query lol

src/teacher.py
```diff
@@ -192,3 +192,33 @@ def get_responses(self) -> List[Optional[Response]]:
...
+class TeacherInteractionMonitorWithVideo(TeacherInteractionMonitor,
+                                         utils.VideoMonitor):
+    """A monitor that wraps interaction with an environment to include
+    generating and answering queries.
+
+    Optionally, if CFG.make_interaction_videos is True, also renders
+    every state and action encountered. The render_fn is generally
+    env.render.
+    """
+    _request: InteractionRequest
```
ronuchit, line 207: we don't need these 4 lines right? cuz they're automatically handled by the inheritance?

src/main.py
```diff
@@ -180,15 +180,19 @@ def _generate_or_load_offline_dataset(env: BaseEnv,
...
+        cycle_num: Optional[int] = None
 ) -> Tuple[List[InteractionResult], float]:
     """Given a sequence of InteractionRequest objects, handle the requests and
     return a list of InteractionResult objects."""
     print("Generating interaction results...")
     results = []
     query_cost = 0.0
     for request in requests:
-        monitor = TeacherInteractionMonitor(request, teacher)
+        monitor = TeacherInteractionMonitorWithVideo(env.render, request,
```
tomsilver, line 194: I know we're going around and around on this, sorry 😬 but... now that we have both `TeacherInteractionMonitorWithVideo` and `TeacherInteractionMonitor`, maybe we should go back to something like the `if` statement that you had originally, and create a `TeacherInteractionMonitorWithVideo` here if `CFG.make_interaction_videos` is True, and `TeacherInteractionMonitor` otherwise? if you make this change, don't forget to remove `CFG.make_interaction_videos` within `TeacherInteractionMonitorWithVideo`.

but if this makes testing annoying, feel free to ignore me (unless @ronuchit ) chimes in strongly

src/teacher.py
```diff
@@ -192,3 +192,33 @@ def get_responses(self) -> List[Optional[Response]]:
     def get_query_cost(self) -> float:
         """Return the query cost."""
         return self._query_cost
+
+
+@dataclass
+class TeacherInteractionMonitorWithVideo(TeacherInteractionMonitor,
+                                         utils.VideoMonitor):
+    """A monitor that wraps interaction with an environment to include
```
ronuchit, line 200: This docstring is out of date, it can now read:
"""A monitor that wraps a TeacherInteractionMonitor to optionally also render every state and action encountered, if CFG.make_interaction_videos is True."""

src/settings.py
```diff
@@ -33,6 +33,7 @@ class GlobalSettings:
     pretty_print_when_loading = False
     # Used for random seeding in test environment.
     test_env_seed_offset = 10000
+    make_interaction_videos = False
```
ronuchit, line 36: while we're here... two additional tiny changes would be nice...
* take out the `make_failure_videos` line from args.py and put it here instead: `make_failure_videos = False`
* for the `make_videos` in args.py, rename it to `make_test_videos` (but keep it in args.py). you'll have to update make_videos -> make_test_videos in two places in main.py as well

src/main.py
ronuchit: sgtm, yeah, in that case `TeacherInteractionMonitorWithVideo` doesn't need to care about `CFG.make_interaction_videos`

ronuchit: but that might make mypy typing confusing.... fine to leave it too

src/settings.py
tomsilver: copying from slack: these should all live in `args.py` instead, and we should still make the second change

src/main.py
amburger66: ok yeah mypy is not happy... should i override stuff or leave it as is?

tomsilver: All good let's leave it

tomsilver (approved):
Lgtm!

ronuchit (approved):
thank you, great stuff!

PR #536, opened 2022-03-07
add a small utility in preparation for GNNPolicyApproach

ronuchit:


src/main.py
```diff
@@ -255,7 +257,8 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
             num_solved += 1
             total_suc_time += (time.time() - start)
         else:
-            print(f"Task {test_task_idx+1} / {len(test_tasks)}: Policy failed")
```
ronuchit, line 258: lol this line has never happened until GNN stuff 🤷

tomsilver (approved):

src/main.py
tomsilver: i've seen it plenty!

PR #537, opened 2022-03-07
changes to blocks env in advance of pybullet blocks env 

tomsilver:
* a number of changes to the dimensions to make them physically realistic
* changing the robot features to include pose, which is needed for the options in pybullet
* several other minor tweaks, mostly for clarity

### DIFF
--- a/src/envs/blocks.py
+++ b/src/envs/blocks.py
@@ -23,25 +23,34 @@ class BlocksEnv(BaseEnv):
     """Blocks domain."""
     # Parameters that aren't important enough to need to clog up settings.py
     table_height = 0.2
-    block_size = 0.1
-    x_lb = 1.3
-    x_ub = 1.4
-    y_lb = 0.15
-    y_ub = 20.85
+    block_size = 0.045
+    # The table x bounds are (1.1, 1.6), but the workspace is smaller.
+    # Make it narrow enough that blocks can be only horizontally arranged.
+    x_lb = 1.35 - block_size
+    x_ub = 1.35 + block_size
+    # The table y bounds are (0.3, 1.2), but the workspace is smaller.
+    y_lb = 0.4
+    y_ub = 1.1
+    pick_z = 0.75
+    robot_init_x = (x_lb + x_ub) / 2
+    robot_init_y = (y_lb + y_ub) / 2
+    robot_init_z = pick_z
     held_tol = 0.5
-    open_fingers = 0.8
-    pick_tol = 0.08
+    open_fingers = 0.04
+    closed_fingers = 0.01
+    finger_tol = 0.00001
+    pick_tol = 0.0001
+    on_tol = 0.01
+    collision_padding = 2.0
     assert pick_tol < block_size
-    pick_z = 1.5
-    num_blocks_train = [3, 4]
-    num_blocks_test = [5, 6]
 
     def __init__(self) -> None:
         super().__init__()
         # Types
         self._block_type = Type("block",
                                 ["pose_x", "pose_y", "pose_z", "held"])
-        self._robot_type = Type("robot", ["fingers"])
+        self._robot_type = Type("robot",
+                                ["pose_x", "pose_y", "pose_z", "fingers"])
         # Predicates
         self._On = Predicate("On", [self._block_type, self._block_type],
                              self._On_holds)
@@ -53,19 +62,20 @@ class BlocksEnv(BaseEnv):
                                   self._Holding_holds)
         self._Clear = Predicate("Clear", [self._block_type], self._Clear_holds)
         # Options
-        self._Pick = utils.SingletonParameterizedOption(
+        self._Pick: ParameterizedOption = utils.SingletonParameterizedOption(
             # variables: [robot, object to pick]
             # params: []
             "Pick",
             self._Pick_policy,
             types=[self._robot_type, self._block_type])
-        self._Stack = utils.SingletonParameterizedOption(
+        self._Stack: ParameterizedOption = utils.SingletonParameterizedOption(
             # variables: [robot, object on which to stack currently-held-object]
             # params: []
             "Stack",
             self._Stack_policy,
             types=[self._robot_type, self._block_type])
-        self._PutOnTable = utils.SingletonParameterizedOption(
+        self._PutOnTable: ParameterizedOption = \
+            utils.SingletonParameterizedOption(
             # variables: [robot]
             # params: [x, y] (normalized coordinates on the table surface)
             "PutOnTable",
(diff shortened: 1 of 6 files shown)


tomsilver:
@amburger66 this impacts playroom superficially, but hopefully does not actually lead to any changes in practice.

src/envs/blocks.py
```diff
@@ -23,25 +23,36 @@ class BlocksEnv(BaseEnv):
...
+    pick_z = 0.75
+    robot_init_x = (x_lb + x_ub) / 2
+    robot_init_y = (y_lb + y_ub) / 2
+    robot_init_z = pick_z
     held_tol = 0.5
-    open_fingers = 0.8
-    pick_tol = 0.08
+    open_fingers = 0.04
+    closed_fingers = 0.01
+    finger_tol = 0.00001
```
ronuchit, line 41: What's up with these tiny tolerances? How are we able to get away with them being so small? Is it because we're not learning option models anymore?

src/envs/blocks.py
```diff
@@ -53,19 +64,20 @@ def __init__(self) -> None:
                                   self._Holding_holds)
         self._Clear = Predicate("Clear", [self._block_type], self._Clear_holds)
         # Options
-        self._Pick = utils.SingletonParameterizedOption(
+        self._Pick: ParameterizedOption = utils.SingletonParameterizedOption(
```
ronuchit, line 67: how was this not an issue before?

src/envs/blocks.py
```diff
@@ -102,14 +114,14 @@ def _transition_pick(self, state: State, x: float, y: float, z: float,
         next_state.set(block, "pose_y", y)
         next_state.set(block, "pose_z", self.pick_z)
         next_state.set(block, "held", 1.0)
-        next_state.set(self._robot, "fingers", fingers)
+        next_state.set(self._robot, "fingers", self.closed_fingers)
```
ronuchit, line 117: do we still need the fingers feature?

src/envs/blocks.py
```diff
@@ -312,8 +324,14 @@ def _sample_state_from_piles(self, piles: List[List[Object]],
...
-        # [fingers]
-        data[self._robot] = np.array([1.0], dtype=np.float32)
+        # [pose_x, pose_y, pose_z, fingers]
+        # Note: the robot poses are not used in this environment, but they
+        # are used in the PyBullet subclass.
+        data[self._robot] = np.array([
+            self.robot_init_x, self.robot_init_y, self.robot_init_z,
+            self.open_fingers
+        ],
+                                     dtype=np.float32)
```
ronuchit, line 334: weirdness with indentation, can you do it a different way?

src/envs/blocks.py
```diff
@@ -312,8 +324,14 @@ def _sample_state_from_piles(self, piles: List[List[Object]],
             z = self.table_height + self.block_size * (0.5 + pile_j)
             # [pose_x, pose_y, pose_z, held]
             data[block] = np.array([x, y, z, 0.0])
-        # [fingers]
-        data[self._robot] = np.array([1.0], dtype=np.float32)
+        # [pose_x, pose_y, pose_z, fingers]
+        # Note: the robot poses are not used in this environment, but they
```
ronuchit, line 328: when you say "not used" do you mean they are always constant? or they're updated but not used for any logic within the simulator?

src/envs/blocks.py
```diff
@@ -23,25 +23,36 @@ class BlocksEnv(BaseEnv):
...
+    robot_init_y = (y_lb + y_ub) / 2
+    robot_init_z = pick_z
     held_tol = 0.5
-    open_fingers = 0.8
-    pick_tol = 0.08
+    open_fingers = 0.04
+    closed_fingers = 0.01
+    finger_tol = 0.00001
+    pick_tol = 0.0001
+    on_tol = 0.01
```
ronuchit, line 43: why does on_tol need to be higher?

src/envs/blocks.py
```diff
@@ -370,19 +389,20 @@ def _On_holds(self, state: State, objects: Sequence[Object]) -> bool:
...
-            (desired_z-self.pick_tol < z < desired_z+self.pick_tol)
+            (desired_z-self.on_tol < z < desired_z+self.on_tol)
 
-    @staticmethod
-    def _GripperOpen_holds(state: State, objects: Sequence[Object]) -> bool:
+    def _GripperOpen_holds(self, state: State,
+                           objects: Sequence[Object]) -> bool:
         robot, = objects
-        return state.get(robot, "fingers") >= BlocksEnv.open_fingers
+        return state.get(robot,
```
ronuchit, line 404: split into two lines to avoid weird indent issue

src/envs/blocks.py
```diff
@@ -433,10 +453,12 @@ def _Stack_policy(self, state: State, memory: Dict,
     def _PutOnTable_policy(self, state: State, memory: Dict,
                            objects: Sequence[Object], params: Array) -> Action:
         del state, memory, objects  # unused
-        # Un-normalize parameters to actual table coordinates
+        # De-normalize parameters to actual table coordinates.
         x_norm, y_norm = params
-        x = self.x_lb + (self.x_ub - self.x_lb) * x_norm
-        y = self.y_lb + (self.y_ub - self.y_lb) * y_norm
+        x = self.x_lb + self.block_size / 2 + (self.x_ub - self.x_lb -
```
ronuchit, line 458: why does the second part of this expression not need to use the half_size as well? we're trying to normalize to within `[self.x_lb + half_size, self.x_ub - half_size]` right?

ronuchit: oh, so `(self.x_ub - half_size) - (self.x_lb + half_size) = self.x_ub - self.x_lb - self.block_size` i see. this is really confusing; would it be better to change x_lb/x_ub/y_lb/y_ub to factor in the half sizes?

src/settings.py
```diff
@@ -52,6 +52,14 @@ class GlobalSettings:
     cover_multistep_bimodal_goal = False
     cover_multistep_goal_conditioned_sampling = False  # assumes one goal
 
+    # block parameters
```
ronuchit, line 55: blocks env parameters*

src/settings.py
```diff
@@ -52,6 +52,14 @@ class GlobalSettings:
...
+    blocks_num_blocks_train = [3, 4]
+    blocks_num_blocks_test = [5, 6]
+
+    # playroom parameters
```
ronuchit, line 59: playroom env parameters*

ronuchit: (and similar changes if elsewhere)

ronuchit (approved):

src/envs/blocks.py
tomsilver: yes that's the case for `finger_tol`. the small `pick_tol`can actually matter, but it's small because it's used to check the square of an already small number.

src/envs/blocks.py
tomsilver: it's fine without for the sake of this PR, but I'm overriding these parameterized options in pybullet env with non-singletons, so mypy complains.

src/envs/blocks.py
tomsilver: yeah we still need it for checking if the gripper is open or closed.

src/envs/blocks.py
tomsilver: they're constant within this class

src/envs/blocks.py
tomsilver: the blocks can be slightly off center from each other after a stack.

src/envs/blocks.py
ronuchit: update comment to reflect that

tomsilver:
@ronuchit this PR adds about 5 seconds to the overall unit test time. the culprit seems to be more difficult sampling during task initialization and planing because the table got smaller. I reduced the number of blocks where I could, but I want to leave the default parameters for oracle and nsrt learning tests so we issues don't go unnoticed.

PR #539, opened 2022-03-08
Revert "changes to blocks env in advance of pybullet blocks env "

tomsilver:
Reverts Learning-and-Intelligent-Systems/predicators#537

at least one bad seed found on supercloud, reverting until we investigate...


ronuchit (approved):
dunno what's going on but here have a free approval

PR #540, opened 2022-03-08
increase default neural_gaus_regressor_sample_clip to 3

tomsilver:
nsrt_learning looks fine on the four main environments on 1 seed. I'm going to test this branch with the nightly script now.

PR #541, opened 2022-03-08
changes to blocks env in advance of pybullet blocks env 

tomsilver:
this is identical to #537 (which was reverted), except that now it's branching off #540, which seems important for getting good blocks performance (to be confirmed).


ronuchit (approved):
this is still identical to what i reviewed before right? i wont look then

PR #543, opened 2022-03-08
PyBullet utils

tomsilver:


tests/test_pybullet_utils.py
```diff
@@ -0,0 +1,157 @@
...
+from predicators.src.pybullet_utils import get_kinematic_chain, \
+    inverse_kinematics
+from predicators.src import utils
+from predicators.src.settings import CFG
+
+_PYBULLET_SCENE_IS_SETUP = False
+_PYBULLET_SCENE_ATTRIBUTES: Dict[str, Any] = {}
+
+
+def _setup_pybullet_test_scene():
```
ronuchit, line 16: Are these the tests we want to mark as expensive?

src/pybullet_utils.py
```diff
@@ -0,0 +1,110 @@
+"""Generic utilities for Pybullet."""
```
ronuchit, line 1: should we move this file, and the test file, into envs?

requirements.txt
```diff
@@ -13,3 +13,4 @@ pyperplan
 pathos
 requests
 slack_bolt
+pybullet>=3.2.0
```
ronuchit, line 16: oh no! i need to upgrade!

src/pybullet_utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+from typing import List, Sequence, Tuple
+import numpy as np
+import pybullet as p
+from predicators.src.settings import CFG
+
+Pose3D = Tuple[float, float, float]
```
ronuchit, line 8: how were you able to avoid, throughout this file, mypy issues with the calls to pybullet methods?

tests/test_pybullet_utils.py
ronuchit: How's their speed?

ronuchit (approved):
nice!!!

tests/test_pybullet_utils.py
tomsilver: no, these tests are actually fast -- 0.9 seconds for the whole file.

src/pybullet_utils.py
tomsilver: btw, I'm just including `Pose3D` here because I use it for clarity in the pybullet blocks env.

PR #545, opened 2022-03-08
Harmlessness sidelining

NishanthJKumar:
Includes new approach for sidelining (via checking harmlessness preservation).

This approach works in particular situations, but isn't as powerful as once hoped. Still, it serves as a useful baseline for our eventual (hopefully wicked good) sidelining approach!


src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -224,3 +256,88 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+            expected_next_hl_state = hl_traj[idx_into_traj + 1]
+
+            for applicable_nsrt in utils.get_applicable_operators(
+                    ground_nsrts, state):
+                # NOTE: we check that the ParameterizedOptions are equal before
+                # attempting to ground because otherwise, we might
+                # get a parameter mismatch and trigger an AssertionError
+                # during grounding.
+
+                if applicable_nsrt.option == gt_option.parent:
```
ronuchit, line 319: similarly here, would be cleaner to do:
```
if applicable_nsrt.option != gt_option.parent:
    continue
if applicable_nsrt.option_objs != gt_option.objects:
    continue
# etc etc...
```

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -224,3 +256,88 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+                    if applicable_nsrt.option_objs == gt_option.objects:
+                        next_hl_state = utils.apply_operator(
+                            applicable_nsrt, set(state))
+                        # Here, we check whether all atoms that differ
+                        # between next_hl_state and state are part of
```
ronuchit, line 324: do you mean: between next_hl_state and expected_next_hl_state?

ronuchit, line 324: are you actually checking that they're all part of the operator's side predicates? it looks like you're just checking issubset, which is a weaker condition right?

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -224,3 +256,88 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+                        # the operator's side predicates. If so, this nsrt
```
ronuchit, line 325: NSRT*

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -224,3 +256,88 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+                            yield (applicable_nsrt, (frozenset(next_hl_state),
+                                                     idx_into_traj + 1), 1.0)
+
+    init_atoms_frozen = frozenset(init_atoms)
+    init_searchnode_state = (init_atoms_frozen, 0)
+    state_seq, _ = utils.run_gbfs(
+        init_searchnode_state, _check_goal, _get_successor_with_correct_option,
+        lambda searchnode_state: heuristic(searchnode_state[0]))
+
+    return _check_goal(state_seq[-1])
```
ronuchit, line 343: What would be a situation where this `_check_goal` does not return True, since you're using `_check_goal` as the termination condition of the `run_gbfs` call above? Oh, are you worried about the case that `run_gbfs` hits its max_expansions or max_evals limit? That makes sense. Did that come up in practice?

src/settings.py
```diff
@@ -146,6 +146,7 @@ class GlobalSettings:
     # NSRT learning parameters
     min_data_for_nsrt = 0
     learn_side_predicates = False
+    sidelining_approach = "naive"
```
ronuchit, line 149: I feel like we should just combine this with the `learn_side_predicates` setting. One setting is easier to keep track of than two! How about a single setting `side_predicate_learner`, with possible options: `no_learning`, `prediction_error_hill_climbing`, `preserve_skeletons_hill_climbing`?

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -97,6 +99,24 @@ def test_nsrt_learning_approach():
...
+    _test_approach(env_name="repeated_nextto",
+                   approach_name="nsrt_learning",
+                   try_solving=False,
+                   sampler_learner="random",
+                   learn_side_predicates=True,
+                   sidelining_approach="preserve_skeletons")
+
+
+def test_unknown_sidelining_approach():
+    """Test that arbitrary sidelining approach throws error."""
```
ronuchit, line 111: throws an error*

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -97,6 +99,24 @@ def test_nsrt_learning_approach():
...
+    with pytest.raises(Exception) as e:
```
ronuchit, line 112: sort of weird to catch a general Exception here -- since you're raising a ValueError in your code, do `pytest.raises(ValueError)` here

ronuchit (commented):
it's a comment extravaganza!

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -107,32 +111,57 @@ def _get_successors(
...
+            strips_ops = [pnad.op for pnad in s]
+            option_specs = [pnad.option_spec for pnad in s]
+            preserves_harmlessness = check_harmlessness(
+                predicates, train_tasks, ground_atom_dataset, strips_ops,
+                option_specs)
+            # NOTE: Arbitrary large number bigger than total number of
+            # operators
+            score = 10000
+            if preserves_harmlessness:
+                score = 2 * len(strips_ops)
```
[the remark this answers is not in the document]
NishanthJKumar: didn't really need to tune it - i think it's fine to leave in like this for now

src/nsrt_learning/nsrt_learning_main.py
NishanthJKumar: indeed - this comment is outdated!

src/nsrt_learning/nsrt_learning_main.py
NishanthJKumar: well also remember that `run_gbfs` might not be able to succeed if we've sidelined too aggressively or something (it'll just hit a bunch of dead-ends). So I think this check is necessary

src/settings.py
NishanthJKumar: great idea!

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -113,32 +116,59 @@ def _get_successors(
...
-        return score
+        def _evaluate(s: Tuple[PartialNSRTAndDatastore, ...]) -> float:
+            # Score function for search. Lower is better.
+            strips_ops = [pnad.op for pnad in s]
+            option_specs = [pnad.option_spec for pnad in s]
+            preserves_harmlessness = check_harmlessness(
+                predicates, train_tasks, trajectories, segmented_trajs,
+                strips_ops, option_specs)
+            # NOTE: Arbitrary large number bigger than total number of
+            # operators.
```
ronuchit, line 160: here, i think we should say "bigger than the total number of operators that we had at the start of search" -- is that right?

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -113,32 +116,59 @@ def _get_successors(
...
+            strips_ops = [pnad.op for pnad in s]
+            option_specs = [pnad.option_spec for pnad in s]
+            score = score_func.evaluate_with_operators(frozenset(),
+                                                       trajectories,
+                                                       segmented_trajs,
+                                                       strips_ops,
+                                                       option_specs)
+            return score
+
+    elif CFG.side_predicate_learner == "preserve_skeletons":
```
ronuchit, line 150: rename to `preserve_skeleton_hillclimbing`

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -230,3 +263,100 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+                       trajectories: List[LowLevelTrajectory],
+                       segmented_trajs: List[List[Segment]],
+                       strips_ops: List[STRIPSOperator],
+                       option_specs: List[OptionSpec]) -> bool:
+    """Function to check whether a given set of operators and predicates
+    preserves harmlessness over demonstrations on some number of training
+    tasks.
+
+    Preserving harmlessness roughly means that the set of operators and
+    predicates supports the agent's ability to plan to achieve all of
```
ronuchit, line 278: of the*

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -230,3 +263,100 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+        searchnode_state: Tuple[FrozenSet[GroundAtom], int]
+    ) -> Iterator[Tuple[_GroundNSRT, Tuple[FrozenSet[GroundAtom], int],
+                        float]]:
+        state = searchnode_state[0]
+        idx_into_traj = searchnode_state[1]
+
+        if idx_into_traj > len(ll_traj.actions) - 1:
+            return
+
+        assert ll_traj.actions[idx_into_traj].has_option()
```
ronuchit, line 328: remove this line; `get_option()` already performs this assertion

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -230,3 +263,100 @@ def _learn_pnad_samplers(pnads: List[PartialNSRTAndDatastore],
...
+            if exp_state_matches:
+                # The returned cost is uniform because we don't
+                # actually care about finding the shortest path;
+                # just one that matches!
+                yield (applicable_nsrt, (frozenset(next_hl_state),
+                                         idx_into_traj + 1), 1.0)
+
+    init_atoms_frozen = frozenset(init_atoms)
+    init_searchnode_state = (init_atoms_frozen, 0)
+    # NOTE: each state in the below gbfs is a tuple of
```
ronuchit, line 353: GBFS*

ronuchit (approved):
looks very good to me, some very very minor final comments!

PR #546, opened 2022-03-08
painfully make checks pass on src/gnn/

ronuchit:
Removed things we aren't using in src/gnn/, like recurrent stuff and
use_gpu/cuda things. Can retrieve them from this repo's commit history
or another repo if we ever need them.

No functional changes to GNNPolicyApproach (hopefully).


tomsilver (approved):
🙈

PR #550, opened 2022-03-09
catch rare case where option plan fails during data collection

tomsilver:


ronuchit (approved):
can you add a warning, like in the previous try/except?

PR #552, opened 2022-03-09
update option model API to report num actions

ronuchit:


tomsilver (commented):
This is going to be wrong for pybullet. Do we care?

ronuchit:
For pybullet it'll be an underestimate (1 instead of like 50+), right? So if planning with this, the agent will optimistically produce plans that it thinks are within the horizon, but actually exceed it. It will therefore fail in main.py with the error "Policy failed to reach goal" -- right? Should be fine.

cc @NishanthJKumar

tomsilver (approved):

PR #554, opened 2022-03-10
fix fixture

ronuchit:
I figured it would be good for me to practice using our new tools.
Verified manually that the fixture is only called once (and in
particular, due to the `scope=module` argument).


tomsilver (approved):
Thanks for taking care of this!

PR #563, opened 2022-03-10
rename `analysis/` to `scripts/` and add `run_checks.sh` for local checking

### LINKED ISSUE #556: add script for running all the github action checks locally
Should be quick, just need to copy-paste things from the README (and update README to reference it)

tomsilver:
closes #556


scripts/analyze_results_directory.py
```diff
@@ -1,5 +1,5 @@
 """Script to analyze experiments resulting from running the script
```
ronuchit, line 1: This docstring is out of date. We can just say `"""Script to analyze results of experiments, stored in the results/ directory.`

scripts/run_checks.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
ronuchit, line 1: make sure to `chmod +x scripts/run_checks.sh` so that it's executable

ronuchit (approved):
thank you so much! RIP muscle memory

williamshen-nz (approved):

PR #564, opened 2022-03-10
assortment of fixes for pybullet predicate invention

ronuchit:
* make oracle approach expose its last plan
* new flag for utils.run_policy and utils.run_policy_with_simulator that lets them terminate upon encountering an exception
when calling the policy
* demonstrations don't terminate when the goal holds, instead the
termination condition is now defined by reaching the end of the plan
* fix ExpectedNodesScoreFunction and HeuristicBasedScoreFunction to use option-level atoms sequence of
the demonstrations
* remove suspicious scores stuff from ExpectedNodesScoreFunction


src/approaches/bilevel_planning_approach.py
```diff
@@ -66,15 +66,7 @@ def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
...
-        option_policy = utils.option_plan_to_policy(plan)
-
-        def _policy(s: State) -> Action:
-            try:
-                return option_policy(s)
-            except utils.OptionPlanExhausted:
-                raise ApproachFailure("Option plan exhausted.")
-
-        return _policy
+        return utils.option_plan_to_policy(plan)
```
tomsilver, line 69: undo this change now, right?

src/predicate_search_score_functions.py
```diff
@@ -129,16 +129,17 @@ def evaluate(self, candidate_predicates: FrozenSet[Predicate]) -> float:
...
+            segment_trajectory(traj) for traj in pruned_atom_data
         ]
+        segments = [seg for traj in segmented_trajs for seg in traj]
         pnads = learn_strips_operators(segments, verbose=False)
         strips_ops = [pnad.op for pnad in pnads]
         option_specs = [pnad.option_spec for pnad in pnads]
         op_score = self.evaluate_with_operators(candidate_predicates,
-                                                pruned_atom_data, segments,
-                                                strips_ops, option_specs)
+                                                pruned_atom_data,
```
tomsilver, line 140: can we get rid of `pruned_atom_data` everywhere now? (after `segmented_trajs` is created)

tomsilver (commented):
these changes look good to me

PR #565, opened 2022-03-10
`logging`

### LINKED ISSUE #555: switch print statements to logging
using python logging library https://docs.python.org/3/howto/logging.html

tomsilver:
closes #555


src/envs/behavior_options.py
```diff
@@ -28,7 +29,7 @@
     from igibson.robots.behavior_robot import BRBody  # pylint: disable=unused-import
 
 except (ImportError, ModuleNotFoundError) as e:
-    print(e)
+    pass
```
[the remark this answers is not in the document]
tomsilver: I removed this because I want the first instance of `logging.blah` to be in `main.py`, otherwise the configuration doesn't get used

williamshen-nz:
For complex projects with lots of modules it's common to get a logger for each module rather than use the root logger with `logging.info|debug|etc.` everywhere. This allows you to identify the module log messages come from which can be helpful when debugging. 

The "Advanced Logging Tutorial" section in the python tutorials has some info: https://docs.python.org/3/howto/logging.html

It doesn't make too much of a difference, so you can decide whether it's important or not.

e.g. in `module.py`
```python

_log = logging.getLogger(__name__)

def my_method():
    _log.info("log message")

**Comment by tomsilver** [comment]
thanks for the pointer! that's good to know. I think for now, just `logging` alone should suffice. but if it gets confusing in the future we can revisit.

File: src/gnn/gnn_utils.py [hunk 2]
```diff
@@ -76,35 +77,31 @@ def train_model(model: Any, dataloaders: Dict,
                 # statistics
                 running_loss[phase] += loss.item()
 
-        if epoch % 100 == 0:
-            print("running_loss:", running_loss, flush=True)
+        if epoch % 1000 == 0:
```
ronuchit, line 80: Note that this may change the performance of the GNN approach slightly. Instead, let's leave the GNN numbers at 100. Anyway 1000 would print things out way too slowly

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -75,12 +76,12 @@ def learn_nsrts_from_data(trajectories: List[LowLevelTrajectory],
     # STEP 6: Learn samplers (sampler_learning.py) and update PNADs.
     _learn_pnad_samplers(pnads, sampler_learner)  # in-place update
 
-    # STEP 7: Print and return the NSRTs.
+    # STEP 7: log and return the NSRTs.
```
ronuchit, line 79: capitalize

src/planning.py
```diff
@@ -219,9 +220,9 @@ def _skeleton_generator(
         _, _, node = hq.heappop(queue)
         # Good debug point #1: print out the skeleton here to see what
         # the high-level search is doing. You can accomplish this via:
```
ronuchit, line 222: we should revert this part, it'll just dominate the printouts when running with debug level. let's leave it commented out

src/gnn/gnn_utils.py
```diff
@@ -27,9 +28,9 @@ def train_model(model: Any, dataloaders: Dict,
     best_seen_running_validation_loss = np.inf
 
     for epoch in range(num_epochs):
-        if epoch % 100 == 0:
-            print(f'Epoch {epoch}/{num_epochs-1}', flush=True)
-            print('-' * 10, flush=True)
+        if epoch % 1000 == 0:
```
ronuchit, line 31: change this as well (see comment below)

src/utils.py
```diff
@@ -1667,15 +1668,6 @@ def string_to_python_object(value: str) -> Any:
     return value
 
 
-def print_args(args: argparse.Namespace) -> None:
```
ronuchit, line 1670: are we wiping these printouts? (fine with me)

ronuchit (approved):
this is a huge and great change, thank you for doing it!

src/approaches/grammar_search_invention_approach.py
```diff
@@ -492,14 +493,14 @@ def __post_init__(self) -> None:
     def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
         # Predicates are identified based on their evaluation across
         # all states in the dataset.
-        seen = {}  # maps identifier to previous predicate
+        seen = {}  # type: ignore
```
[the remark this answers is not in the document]
tomsilver: we can, but it would be disgusting. do you want me to do it anyway?

src/utils.py
tomsilver: yeah

src/approaches/grammar_search_invention_approach.py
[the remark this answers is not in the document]
ronuchit: i fixed it

PR #567, opened 2022-03-10
change PrunedGrammar to be faster if segmenter is option_changes

ronuchit:


src/approaches/grammar_search_invention_approach.py
```diff
@@ -469,6 +469,26 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+            # If the segmenter is based on changes in options, then
+            # because it doesn't depend on atoms, we can be very
+            # efficient during pruning by pre-computing the segments.
+            # Then, we only need to care about the initial and final
+            # states in each segment, which we store into
+            # self._state_sequence.
+            atom_dataset = utils.create_ground_atom_dataset(
+                self.dataset.trajectories, set())
+            for traj in atom_dataset:
+                if not traj[0].is_demo:
```
tomsilver, line 486: the pruner here should probably be agnostic to demo vs not demo, so remove this line

src/approaches/grammar_search_invention_approach.py
```diff
@@ -469,6 +469,26 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+
+    def __post_init__(self) -> None:
+        if CFG.segmenter == "option_changes":
+            # If the segmenter is based on changes in options, then
+            # because it doesn't depend on atoms, we can be very
+            # efficient during pruning by pre-computing the segments.
+            # Then, we only need to care about the initial and final
+            # states in each segment, which we store into
+            # self._state_sequence.
+            atom_dataset = utils.create_ground_atom_dataset(
```
tomsilver, line 483: why are we doing this? is atom_dataset used?

tomsilver (commented):
some questions but looks good

src/approaches/grammar_search_invention_approach.py
ronuchit: isn't it used in the next line? we just need an atom_dataset to be able to call segment_trajectory

ronuchit: oh, you think we should just pass in an empty thing there, i see, okay!

tomsilver: ah okay got it

src/approaches/grammar_search_invention_approach.py
ronuchit: we can't bc segmentation would crash on any traj that isn't a demo, since it doesn't have an option (this came up during unit testing...)

tomsilver: got it, but this still doesn't feel like the right check. I think that crash is actually good/informative. can we just adjust the unit test?

ronuchit: kk

tomsilver (approved):
lgtm

PR #570, opened 2022-03-10
add simple example of using fixture for test_structs

### LINKED ISSUE #559: update tests/test_structs.py to use fixtures, as an example for the future
Changing all the tests would take too long, but we can do just test_structs.py as an example to make things better for future tests.

ronuchit:
* verified via print statement that test_state() is only called once

closes #559


ronuchit:
thanks for the tip, @williamshen-nz !

tomsilver (approved):

PR #572, opened 2022-03-11
one potential solution to envs/approaches __init__

### LINKED ISSUE #558: update envs/__init__.py and approaches/__init__.py to automatically discover modules
we'll need to make the names of envs/approaches, which are currently in the `__init__.py` files, instead be exposed via abstract methods in BaseEnv/BaseApproach

ronuchit:
This PR makes us not have to touch `__init__.py` in `approaches/` or `envs/` when adding new approaches or environments. One large change is that we can't import things directly from approaches or envs anymore, e.g.:
```
from predicators.src.envs import CoverEnv
from predicators.src.approaches import OracleApproach
```
now needs to be:
```
from predicators.src.envs.cover import CoverEnv
from predicators.src.approaches.oracle_approach import OracleApproach
```

closes #558

### DIFF
--- a/src/approaches/__init__.py
+++ b/src/approaches/__init__.py
@@ -11,7 +11,6 @@ from predicators.src import utils
 
 __all__ = ["BaseApproach", "ApproachTimeout", "ApproachFailure"]
 
-
 if not TYPE_CHECKING:
     # Load all modules so that utils.get_all_subclasses() works.
     for loader, module_name, _ in pkgutil.walk_packages(__path__):
@@ -26,8 +25,8 @@ def create_approach(name: str, initial_predicates: Set[Predicate],
     """Create an approach given its name."""
     for cls in utils.get_all_subclasses(BaseApproach):
         if cls is not BaseApproach and cls.get_name() == name:
-            approach = cls(initial_predicates, initial_options,
-                           types, action_space, train_tasks)
+            approach = cls(initial_predicates, initial_options, types,
+                           action_space, train_tasks)
             break
     else:
         raise NotImplementedError(f"Unknown approach: {name}")
--- a/src/approaches/__init__.py
+++ b/src/approaches/__init__.py
@@ -1,40 +1,22 @@
-"""Default imports for approaches folder."""
+"""Handle creation of approaches."""
 
-from typing import Set, List
+import pkgutil
+from typing import Set, List, TYPE_CHECKING
 from gym.spaces import Box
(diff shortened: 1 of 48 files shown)


ronuchit:
@tomsilver @williamshen-nz here's one potential solution, how do you feel about it

tomsilver (commented):
I'm going to let @williamshen-nz take the lead on this one, but it looks reasonable to me at a high level

ronuchit:
Verified that things work fine on supercloud with this change.

williamshen-nz:
> This PR makes us not have to touch `__init__.py` in `approaches/` or `envs/` when adding new approaches or environments. One large change is that we can't import things directly from approaches or envs anymore, e.g.:
> 
> ```
> from predicators.src.envs import CoverEnv
> from predicators.src.approaches import OracleApproach
> ```
> 
> now needs to be:
> 
> ```
> from predicators.src.envs.cover import CoverEnv
> from predicators.src.approaches.oracle_approach import OracleApproach
> ```
> 
> closes #558

Hmm that's not ideal - doesn't it sort of defeat the purpose? I've worked on a similar feature in the past will have a closer look after lab meeting

ronuchit:
Yeah, would be good to avoid that if we could... okay thanks!

src/approaches/__init__.py
```diff
@@ -1,67 +1,33 @@
...
-    "BilevelPlanningApproach",
-    "NSRTLearningApproach",
-    "InteractiveLearningApproach",
-    "GrammarSearchInventionApproach",
-    "ApproachTimeout",
-    "ApproachFailure",
-]
+__all__ = ["BaseApproach", "ApproachTimeout", "ApproachFailure"]
+
+if not TYPE_CHECKING:
```
williamshen-nz, line 14: What happens if we do it if we're type checking? is it just slower?

williamshen-nz (approved):
🚢  Looks good to me! If we want to handle directly importing from `__init__` then I can look into it - let me know what you think. I'll see if there's any easy approach

src/approaches/__init__.py
ronuchit: it crashes and burns horrendously LOL, mypy starts to search over all the modules that are imported in the whole codebase for some reason...

ronuchit:
After discussing on slack, we agreed that we are okay with changing the importing mechanism, since that helps us avoid some annoying circular import issues.

PR #573, opened 2022-03-11
isort

### LINKED ISSUE #557: use isort for formatting
can use the official github action repo: https://github.com/isort/isort-action

ronuchit:
(Still being tested)

Using isort ended up being way easier than I expected. I already had
isort installed somehow, and all I had to do was run `isort .`. I
figured the remaining work is just setting up the github checks the same
way as we do for YAPF, so I went ahead and did that (only took a few
mins).

@tomsilver @williamshen-nz 

closes #557 if successful

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[isort]
+multi_line_output = 2
--- a/.github/workflows/predicators.yml
+++ b/.github/workflows/predicators.yml
@@ -71,7 +71,14 @@ jobs:
         python-version: [3.8]
     steps:
     - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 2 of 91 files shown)


.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[isort]
+multi_line_output = 2
```
ronuchit, line 2: This comes from https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html and is the style we currently use in the repo.

ronuchit:
Lol, it turns out that this exposes a bazillion circular import issues in our repo. So everything breaks, you can't even run main.py. The good news is that it seems to be fine on top of #572, so we can merge that one first, then this one.

williamshen-nz (approved):
Might be worth rerunning the checks once #572 is merged

PR #575, opened 2022-03-11
add a long integration test for option learning

tomsilver:
this new longrun test takes about 75 seconds to run. longer than ideal, but if we make the hyperparameters any weaker, I'm not sure if we'll see the option learning signs of life that we want to test.

### DIFF
--- a/tests/nsrt_learning/test_option_learning.py
+++ b/tests/nsrt_learning/test_option_learning.py
@@ -2,7 +2,9 @@
 
 import pytest
 import numpy as np
-from predicators.src.approaches import ApproachFailure
+from predicators.src.approaches import ApproachFailure, ApproachTimeout, \
+    create_approach
+from predicators.src.datasets import create_dataset
 from predicators.src.envs import create_new_env
 from predicators.src.ground_truth_nsrts import get_gt_nsrts
 from predicators.src.datasets.demo_replay import create_demo_replay_data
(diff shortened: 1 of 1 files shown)


tests/nsrt_learning/test_option_learning.py
```diff
@@ -227,3 +231,40 @@ def test_create_option_learner():
...
+                                                   task.init,
+                                                   task.goal_holds,
+                                                   max_num_steps=CFG.horizon)
+            if task.goal_holds(traj.states[-1]):
+                num_test_successes += 1
+        except (ApproachFailure, ApproachTimeout):
+            continue
+    # This number is expected to be relatively low because the number of train
+    # tasks is pretty limiting. But if it goes lower than this, that could
+    # be a performance regression that we should investigate.
```
ronuchit, line 269: feel free to make the num train tasks higher if you want. btw, this file with longrun takes 60 seconds for me locally:
`pytest -s tests/nsrt_learning/test_option_learning.py --longrun`
and gives me a num_test_successes of exactly 3. so maybe we want more train tasks or a lower threshold?

ronuchit (approved):
unblocking approval

tests/nsrt_learning/test_option_learning.py
tomsilver: resolved over slack

PR #577, opened 2022-03-11
allow interactive approach to make multiple requests per cycle

tomsilver:


src/approaches/interactive_learning_approach.py
```diff
@@ -158,13 +159,14 @@ def _score_atom_set(self, atom_set: Set[GroundAtom],
         raise NotImplementedError("Unrecognized interactive_score_function:"
                                   f" {CFG.interactive_score_function}.")
 
-    def _select_interaction_train_task_idx(self) -> int:
+    def _select_interaction_train_task_idxs(self) -> List[int]:
         # At the moment, we only have one way to select a train task idx:
```
ronuchit, line 163: update this docstring a bit?

src/main.py
```diff
@@ -202,6 +202,8 @@ def _generate_interaction_results(
     logging.info("Generating interaction results...")
     results = []
     query_cost = 0.0
+    if CFG.make_interaction_videos:
```
ronuchit, line 205: good catch....

ronuchit (approved):

amburger66 (approved):
thanks!

PR #582, opened 2022-03-12
clarification regarding potential source of bugs

ronuchit:


NishanthJKumar (approved):
LGTM! Thanks so much for the close attention to detail and catching this now; this would've been a real doozy down the line

PR #585, opened 2022-03-14
log exceptions

### LINKED ISSUE #580: crashes no longer appear in logs on supercloud
After #571, I believe that stderr isn't showing up in the logs anymore, which makes sense since we don't redirect stderr into our log files, but is very bad for figuring out which experiments crashed if some of them didn't terminate.

tomsilver:
closes #580


ronuchit (approved):
lgtm. have you tested both on supercloud and locally, both normal usage and an error?

PR #587, opened 2022-03-14
fix learning time

ronuchit:
Note one issue is that all three of these things are cumulative now --
num_transitions, cumulative_query_cost, learning_time. But we're only
naming one of them "cumulative". Should we just make them all consistent
@tomsilver?

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -133,7 +133,7 @@ def _run_pipeline(env: BaseEnv,
         # Run evaluation once before online learning starts.
         results = _run_testing(env, approach)
         results["num_transitions"] = total_num_transitions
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -156,17 +158,21 @@ def _run_pipeline(env: BaseEnv,
             logging.info(f"Query cost incurred this cycle: {query_cost}")
             if CFG.load_approach:
                 approach.load(online_learning_cycle=i)
+                learning_time += 0.0  # ignore loading time
```
tomsilver, line 161: remove?

tomsilver (approved):

tomsilver:
sure, let's just remove the one "cumulative"

PR #589, opened 2022-03-14
rename some files in scripts/

ronuchit:


tomsilver:
I think so? but I don't care that much

scripts/analyze_results_directory.py
```diff
@@ -157,8 +157,8 @@ def _main() -> None:
     pd.set_option("expand_frame_repr", False)
     print("\n\nAGGREGATED DATA OVER SEEDS:")
     print(means)
-    means.to_csv("supercloud_analysis.csv")
-    print("\n\nWrote out table to supercloud_analysis.csv")
+    means.to_csv("results_analysis.csv")
```
tomsilver, line 160: maybe `results_summary.csv`? we're not really analyzing anything

scripts/supercloud/run_active_learning_baseline.sh
```diff
@@ -3,7 +3,7 @@
 START_SEED=456
 NUM_SEEDS=10
 CYCLES=100
-FILE="scripts/submit.py"
+FILE="scripts/supercloud/submit_supercloud_job.py"
```
tomsilver, line 6: jobs?

tomsilver (approved):

scripts/supercloud/run_active_learning_baseline.sh
ronuchit: no, it just does one job

PR #598, opened 2022-03-18
fix edge case for min_skeletons_optimized metric

tomsilver:
example: ```python src/main.py --env cover --approach oracle --seed 0 --num_test_tasks 1 --timeout 0.001```

before:
```
Test results: defaultdict(<class 'float'>, {... 'min_skeletons_optimized': 8, 'max_skeletons_optimized': 0.0, ...})
```

after:
```
Test results: defaultdict(<class 'float'>, {... 'min_skeletons_optimized': 0, 'max_skeletons_optimized': 0.0, ...})
```

### DIFF
--- a/src/approaches/bilevel_planning_approach.py
+++ b/src/approaches/bilevel_planning_approach.py
@@ -86,10 +86,8 @@ class BilevelPlanningApproach(BaseApproach):
 
     def reset_metrics(self) -> None:
         super().reset_metrics()
-        # Initialize min to self._max_skeletons_optimized (max gets initialized
-        # to 0 by default)
-        self._metrics[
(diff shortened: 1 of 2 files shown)


ronuchit (commented):
why are 0's the preferred value here? if min == max == 0, then it seems like we're saying that no skeletons needed to be optimized, when really what we mean is that planning failed (but it might have tried many, many skeletons in the process). right?

tomsilver:
@ronuchit no, this should only happen when we timeout while attempting to optimize the first skeleton. so I think 0 is the most intuitive number?

ronuchit (approved):
ok sure

PR #601, opened 2022-03-18
change approach and env seed to _set_seed and remove seed(123) in tests

### LINKED ISSUE #600: make env.seed() and approach.seed() private, and remove all external calls to them, since they're called at initialization

tomsilver:
closes #600

### DIFF
--- a/scripts/grammar_search_analysis.py
+++ b/scripts/grammar_search_analysis.py
@@ -235,7 +235,6 @@ def _run_proxy_analysis_for_predicates(
                                    env.options, env.types, env.action_space,
                                    train_tasks)
         approach.learn_from_offline_dataset(dataset)
-        approach.seed(CFG.seed)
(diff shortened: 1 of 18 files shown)


ronuchit (approved):
what about these lines in main.py:
```
    # The action space and options need to be seeded externally, because
    # env.action_space and env.options are often created during env __init__().
    env.action_space.seed(CFG.seed)
    for option in env.options:
        option.params_space.seed(CFG.seed)
```
i guess we leave those as is?

ronuchit:
it's weird that base_approach sets the action_space seed, but base_env doesn't, and instead the base_env's action space is seeded in main.py.... 🤯

tomsilver:
you're not wrong, but I think this will become a can of worms if we start to change stuff like that, so I would prefer to leave it for another day...

PR #605, opened 2022-03-20
fix an edge case with checking timeout in planner

ronuchit:
this was arising with planning using operators that were not sidelined,
in repeated_nextto. i noticed that sometimes the planner was far
exceeding the 10-second timeout. this should fix that.


ronuchit:
@NishanthJKumar this should fix the seed i mentioned to you on slack

tomsilver (approved):

PR #607, opened 2022-03-21
adds harmlessness based baseline to sidelining expers

NishanthJKumar:
even though we don't really expect this baseline to do well, I think it's good/useful to have it in this script to compare new approaches against!


ronuchit (approved):
LGTM, you might want to change the exp id to just `rnt_harmlessness` so it's a bit shorter/more readable on supercloud

PR #608, opened 2022-03-23
variation on repeated nextto with one parameterized option

tomsilver:


src/ground_truth_nsrts.py
```diff
@@ -1906,6 +1908,59 @@ def _get_repeated_nextto_gt_nsrts() -> Set[NSRT]:
...
+    robot_type, dot_type = _get_types_by_names(CFG.env, ["robot", "dot"])
+
+    NextTo, NextToNothing, Grasped = _get_predicates_by_names(
+        CFG.env, ["NextTo", "NextToNothing", "Grasped"])
+
+    MoveGrasp, = _get_options_by_names(CFG.env, ["MoveGrasp"])
+
+    nsrts = set()
+
+    # Move
```
ronuchit, line 1922: the only things that change are the option and the sampler, right? can we just call the repeatd_nextto_get_gt_nsrts() to get its two NSRTs, and then create new NSRTs using the NSRT constructor, with the option and sampler being the only differences?

tests/approaches/test_oracle_approach.py
```diff
@@ -120,6 +121,7 @@ def test_check_nsrt_parameters():
         "playroom": PlayroomEnv(),
         "cover_multistep_options": CoverMultistepOptions(),
         "repeated_nextto": RepeatedNextToEnv(),
+        "repeated_nextto_single_option": RepeatedNextToSingleOptionEnv(),
```
ronuchit, line 124: update `test_base_env.py` too

ronuchit (approved):
approving to unblock but i think it should be possible to do the gt nsrt stuff without duplicating code

tomsilver:
@ronuchit good idea thanks. can you check again?

ronuchit (approved):
lgtm again!

PR #609, opened 2022-03-23
(Hacky) Intersection-based sidelining

NishanthJKumar:
Does sidelining by simply (1) clustering all transitions by option and then (2) inducing add and delete effects as well as preconditions via intersection over states (things not in the intersection are sidelined).

Code quality is poor (rewrite necessary) but fundamental idea seems to work on rnt env! Changes build from #604


NishanthJKumar:
Superseded by #618

PR #611, opened 2022-03-24
don't try to instantiate abstract subclasses of BaseEnv or BaseApproach

tomsilver:


ronuchit:
have you printed things out to confirm this is skipping over the classes you expect it to?

tomsilver:
yeah, the only one that it currently skips is `BilevelPlanningApproach`. but I tested with an abstract environment too that is not yet merged.

PR #613, opened 2022-03-24
Automatically create symbolic-only environments from PDDL

tomsilver:
Includes the ability to generate environments from a fixed set of PDDL problem files (as in PDDLGym), and also the ability to generate environments via procedural problem generation. There is a Blocks example for both cases.

Examples of things that work (perfect test performance):
* `python src/main.py --env pddl_blocks_fixed_tasks --approach oracle --seed 0 --num_train_tasks 5 --num_test_tasks 5`
* `python src/main.py --env pddl_blocks_fixed_tasks --approach nsrt_learning --seed 0 --num_train_tasks 5 --num_test_tasks 5`
* `python src/main.py --env pddl_blocks_fixed_tasks --approach grammar_search_invention --seed 0 --num_train_tasks 5 --num_test_tasks 5`
* `python src/main.py --env pddl_blocks_procedural_tasks --approach oracle --seed 0`
* `python src/main.py --env pddl_blocks_procedural_tasks --approach nsrt_learning --seed 0`
* `python src/main.py --env pddl_blocks_procedural_tasks --approach grammar_search_invention --seed 0`

Examples of things that run and don't crash:
* `python src/main.py --env pddl_blocks_fixed_tasks --approach random_actions --seed 0 --num_train_tasks 5 --num_test_tasks 5`
* `python src/main.py --env pddl_blocks_fixed_tasks --approach random_options --seed 0 --num_train_tasks 5 --num_test_tasks 5`
* `python src/main.py --env pddl_blocks_fixed_tasks --approach grammar_search_invention --seed 0 --num_train_tasks 5 --num_test_tasks 5 --excluded_predicates all`
*  `python src/main.py --env pddl_blocks_procedural_tasks --approach random_actions --seed 0`
* `python src/main.py --env pddl_blocks_procedural_tasks --approach random_options --seed 0 `
*  `python src/main.py --env pddl_blocks_procedural_tasks --approach grammar_search_invention --seed 0 --excluded_predicates all`


src/envs/pddl_procedural_generation.py
```diff
@@ -0,0 +1,92 @@
+"""Procedurally generates PDDL problem strs."""
```
ronuchit, line 1: blank line after?

src/structs.py
```diff
@@ -1318,3 +1318,8 @@ class PathToStateResponse(Response):
                                      _GroundSTRIPSOperator)
 SamplerDatapoint = Tuple[State, Dict[Variable, Object], _Option,
                          Optional[Set[GroundAtom]]]
+# For PDDLEnv environments, given a desired number of problems and an rng,
+# returns a list of that many PDDL problem strs.
```
ronuchit, line 1322: strings

src/envs/pddl_procedural_generation.py
```diff
@@ -0,0 +1,92 @@
...
+    return partial(_generate_blocks_problems, min_num_blocks, max_num_blocks,
+                   min_num_blocks_goal, max_num_blocks_goal, new_pile_prob)
+
+
+def _generate_blocks_problems(min_num_blocks: int, max_num_blocks: int,
+                              min_num_blocks_goal: int,
+                              max_num_blocks_goal: int, new_pile_prob: float,
+                              num_problems: int,
+                              rng: np.random.Generator) -> List[str]:
+    problems = []
```
ronuchit, line 24: assert max_num_blocks_goal <= min_num_blocks?

src/envs/pddl_procedural_generation.py
```diff
@@ -0,0 +1,92 @@
...
+    # Create piles for the initial state and goal.
+    piles: List[List[str]] = []
+    goal_piles: List[List[str]] = []
+    for block_group, pile_group in ((blocks, piles), (goal_blocks,
+                                                      goal_piles)):
+        for block in block_group:
+            if not pile_group or rng.uniform() < new_pile_prob:
+                # Create a new pile.
+                pile_group.append([])
+            # Add the block to the pile.
```
ronuchit, line 53: to the most recently created pile

src/ground_truth_nsrts.py
```diff
@@ -2107,3 +2110,22 @@ def _get_predicate(base_pred_name: str,
                 f"Unexpected base option name: {base_option_name}")
 
     return nsrts
+
+
+def _get_pddl_env_gt_nsrts(name: str) -> Set[NSRT]:
+    env_base = get_or_create_env(name)
+    env = cast(_PDDLEnv, env_base)
```
ronuchit, line 2117: before casting, should we assert isinstance(env_base, _PDDLEnv)?

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    """An environment induced by PDDL.
+
+    The state space is mostly unused. The continuous vectors are dummies. What
+    is actually used is state.simulator_state, which holds the current ground
+    atoms (see PDDLEnvSimulatorState). Note that we need to use this pattern,
+    as opposed to just maintaining the ground atoms internally in the env,
+    because the predicate classifiers need access to the ground atoms.
+
+    The action space is hacked to conform to our convention that actions
+    are fixed dimensional vectors. Users of this class should not need
```
ronuchit, line 41: fixed-dimensional

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    to worry about the action space because it would never make sense to
+    learn anything using this action space. The dimensionality is 1 +
+    max operator arity. The first dimension encodes the operator. The
+    next dimensions encode the object used to ground the operator. The
```
ronuchit, line 45: but doesn't this require knowing the number of objects? (or at least an upper bound on it)

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    return cast(PDDLEnvSimulatorState, state.simulator_state)
+
+
+def _ground_atoms_to_state(ground_atoms: Set[GroundAtom],
+                           objects: Collection[Object]) -> State:
+    dummy_state_dict = {o: np.zeros(1, dtype=np.float32) for o in objects}
+    return State(dummy_state_dict, simulator_state=ground_atoms)
+
+
+def _parse_pddl_domain(
```
ronuchit, line 349: would prefer this method to be in the base class as a staticmethod since it's so long

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+            tasks.append(task)
+        return tasks
+
+    @property
+    def predicates(self) -> Set[Predicate]:
+        return self._predicates
+
+    @property
+    def goal_predicates(self) -> Set[Predicate]:
+        # For now, we assume that all predicates may be included as part of the
```
ronuchit, line 118: i feel like we should just fix this now. can't we just make this env-specific? alternatively we can just set self._train_tasks and self._test_tasks in `__init__` itself, and extract the goal predicates from the task goals. note that the comment in base_env.py "# These are generated lazily when..." doesn't matter here because we always have everything we need to be able to generate tasks in the init, for these pddl envs. in this case, _generate_train/test_tasks would be no-ops

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+                                            max_num_blocks_goal, new_pile_prob)
+
+
+###############################################################################
+#                            Utility functions                                #
+###############################################################################
+
+
+def _state_to_ground_atoms(state: State) -> Set[GroundAtom]:
+    return cast(PDDLEnvSimulatorState, state.simulator_state)
```
ronuchit, line 340: i'm confused by this cast, don't we want to be returning a Set[GroundAtom] according to the function annotation? (i know they're the same, but we should be using one or the other). i feel like this should be casting to Set[GroundAtom] instead (or maybe you don't even need to cast...?)

ronuchit, line 340: can we remove this PDDLEnvSimulatorState thing entirely? i think it's more confusing than helpful

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+
+
+def _ground_atoms_to_state(ground_atoms: Set[GroundAtom],
+                           objects: Collection[Object]) -> State:
+    dummy_state_dict = {o: np.zeros(1, dtype=np.float32) for o in objects}
```
ronuchit, line 345: why can't this be empty? is it because you want `list(state)` to give you the objects?

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+                                            min_num_blocks_goal,
+                                            max_num_blocks_goal, new_pile_prob)
+
+
+###############################################################################
+#                            Utility functions                                #
+###############################################################################
+
+
+def _state_to_ground_atoms(state: State) -> Set[GroundAtom]:
```
ronuchit, line 339: why is this out here but action_to_ground_strips_op isn't? can we just put all these methods in the base class?

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def _action_to_ground_strips_op(
+            self, action: Action,
+            ordered_objects: List[Object]) -> _GroundSTRIPSOperator:
+        action_arr = action.arr
+        assert all(float(a).is_integer() for a in action_arr)
+        op_idx = int(action_arr[0])
+        op = self._ordered_strips_operators[op_idx]
+        op_arity = len(op.parameters)
+        obj_idxs = [int(i) for i in action_arr[1:op_arity + 1]]
```
ronuchit, line 155: parenthesize (op_arity + 1)

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    def render_state(self,
+                     state: State,
+                     task: Task,
+                     action: Optional[Action] = None,
+                     caption: Optional[str] = None) -> List[Image]:
+        raise NotImplementedError("Render not implemented for PDDLEnv.")
+
+    def _pddl_problem_str_to_task(self, pddl_problem_str: str) -> Task:
+        # Let pyperplan do most of the heavy lifting.
+        # Pyperplan needs the domain to parse the problem.
```
ronuchit, line 168: let's cache the domain part of this out in the init? you can make your call to _parse_pddl_domain return the pyperplan_domain in addition to current return values

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+        return self._predicates
+
+    @property
+    def types(self) -> Set[Type]:
+        return self._types
+
+    @property
+    def options(self) -> Set[ParameterizedOption]:
+        # The parameterized options are 1:1 with the STRIPS operators.
+        return {
```
ronuchit, line 132: let's precompute this into `self._options` in init

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    def _pddl_test_problem_generator(self) -> PDDLProblemGenerator:
+        raise NotImplementedError("Override me!")
+
+    @property
+    def strips_operators(self) -> Set[STRIPSOperator]:
+        """Expose the STRIPSOperators for use by oracles."""
+        return self._strips_operators
+
+    def simulate(self, state: State, action: Action) -> State:
+        objs = list(state)
```
ronuchit, line 82: call this ordered_objs

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    def _strips_operator_to_parameterized_option(
+            self, op: STRIPSOperator) -> ParameterizedOption:
+        name = op.name
+        types = [p.type for p in op.parameters]
+        op_idx = self._ordered_strips_operators.index(op)
+        act_dims = self.action_space.shape[0]
+
+        def policy(s: State, m: Dict, o: Sequence[Object], p: Array) -> Action:
+            del m, p  # unused
+            ordered_objs = list(s)
```
ronuchit, line 214: tricky, add a comment to clarify what's going on in this and the next line

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+            obj_idxs = [ordered_objs.index(obj) for obj in o]
+            act_arr = np.zeros(act_dims, dtype=np.float32)
+            act_arr[0] = op_idx
+            act_arr[1:len(obj_idxs) + 1] = obj_idxs
```
ronuchit, line 218: parenthesize (len(obj_idxs) + 1)

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+            return ground_op.preconditions.issubset(ground_atoms)
+
+        return utils.SingletonParameterizedOption(name,
+                                                  policy,
+                                                  types,
+                                                  initiable=initiable)
+
+
+class _FixedTasksPDDLEnv(_PDDLEnv):
+    """An env where tasks are induced by static PDDL problem files."""
```
ronuchit, line 235: environment

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    @property
+    @abc.abstractmethod
+    def _pddl_problem_asset_dir(self) -> str:
+        """The name of the problem file directory in assets/pddl."""
+        raise NotImplementedError("Override me!")
+
+    @property
+    @abc.abstractmethod
+    def _train_problem_indices(self) -> List[int]:
+        """The problem file names should be task{idx}.pddl."""
```
ronuchit, line 246: Return a list of indices corresponding to training problems. For each idx in the returned list, it is expected that there will be a problem file named task{idx}.pddl.

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+        raise NotImplementedError("Override me!")
+
+    @property
+    @abc.abstractmethod
+    def _test_problem_indices(self) -> List[int]:
```
ronuchit, line 251: Return a list of indices corresponding to test problems. For each idx in the returned list, it is expected that there will be a problem file named task{idx}.pddl.

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+        return _file_problem_generator(self._pddl_problem_asset_dir,
+                                       self._test_problem_indices)
+
+
+###############################################################################
+#                              Concrete Envs                                  #
+###############################################################################
+
+
+class _BlocksPDDLEnv(_PDDLEnv):
```
ronuchit, line 273: isn't this abstract too?

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    problems = []
+    for idx in indices:
+        path = utils.get_env_asset_path(f"pddl/{dir_name}/task{idx}.pddl")
+        with open(path, encoding="utf-8") as f:
+            problem_str = f.read()
+        problems.append(problem_str)
+
+    def _problem_gen(num: int, rng: np.random.Generator) -> List[str]:
+        del rng  # unused
+        return problems[:num]
```
ronuchit, line 440: assert len(problems) >= num

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+    def get_name(cls) -> str:
+        return "pddl_blocks_fixed_tasks"
+
+    @property
+    def _pddl_problem_asset_dir(self) -> str:
+        return "blocks"
+
+    @property
+    def _train_problem_indices(self) -> List[int]:
+        return list(range(1, 6))
```
ronuchit, line 297: instead of magic numbers 6 and 11, let's grab the number of problems in the folder and use half for train, half for test?

src/envs/pddl_env.py
ronuchit, line 345: can we make this zero-dimensional?

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+        for t in pyperplan_types
+    }
+    # Convert the predicates.
+    predicate_name_to_predicate = {}
+    for pyper_pred in pyperplan_predicates.values():
+        name = pyper_pred.name
+        pred_types = [
+            pyperplan_type_to_type[t] for _, (t, ) in pyper_pred.signature
+        ]
+        # This is incredibly hacky, but necessary, because the classifier needs
```
ronuchit, line 372: sigh yeah. i can't think of a way around this either...

src/envs/pddl_env.py
```diff
@@ -0,0 +1,442 @@
...
+        predicate_name_to_predicate[name] = pred
+    # Convert the operators.
+    operators = set()
+    for pyper_op in pyperplan_operators.values():
+        name = pyper_op.name
+        parameters = [
+            Variable(n, pyperplan_type_to_type[t])
+            for n, (t, ) in pyper_op.signature
+        ]
+        name_to_param = {p.name: p for p in parameters}
```
ronuchit, line 389: rename to `param_name_to_param`

ronuchit (commented):
looking great!!

src/envs/pddl_env.py
tomsilver: it only requires knowing the max operator arity

src/envs/pddl_env.py
tomsilver: I don't care either way, but what's the reason to prefer staticmethods for longer things?

ronuchit: i was thinking all these methods should go into the base class... that way they're only accessible to subclasses

src/envs/pddl_env.py
ronuchit: oh yeah that's cuz your action space has infs on the upper bound

src/ground_truth_nsrts.py
tomsilver: ignoring because of the weirdness with isinstance

src/envs/pddl_env.py
tomsilver: discussed on slack, ignoring

src/envs/pddl_env.py
tomsilver: yeah that's why

src/envs/pddl_env.py
tomsilver: yeah. I changed "Concrete" to "Specific" in the heading. I still think it's good to group all the blocks stuff together

src/envs/pddl_env.py
tomsilver: I prefer this way because I want it to be easy to experiment with different train and test problem sets. Manipulating the files themselves is a pain.

ronuchit: what do you mean by manipulating the files? wouldn't it just be `len(os.listdir(path to blocks folder))`?

tomsilver: I mean in the course of doing research, if I want to experiment with a certain subset of all the PDDL files I have, I don't want to have to change the files in my directory manually, I would rather change the numbers here

src/envs/pddl_env.py
```diff
@@ -0,0 +1,470 @@
...
+
+    def _classifier(s: State, objs: Sequence[Object]) -> bool:
+        return GroundAtom(pred, objs) in _state_to_ground_atoms(s)
+
+    return _classifier
+
+
+def _file_problem_generator(dir_name: str,
+                            indices: Sequence[int]) -> PDDLProblemGenerator:
+    # Load all of the PDDL problem strs from files.
```
ronuchit, line 457: strings*

src/envs/pddl_procedural_generation.py
```diff
@@ -0,0 +1,100 @@
+"""Procedurally generates PDDL problem strs."""
```
ronuchit, line 1: strings*

src/envs/pddl_env.py
```diff
@@ -0,0 +1,470 @@
...
+    def get_name(cls) -> str:
+        return "pddl_blocks_fixed_tasks"
+
+    @property
+    def _pddl_problem_asset_dir(self) -> str:
+        return "blocks"
+
+    @property
+    def _train_problem_indices(self) -> List[int]:
+        return list(range(1, 6))
```
ronuchit, line 235: fine with me, but let's at least move this into settings.py? `pddl_blocks_fixed_train_indices` and `pddl_blocks_fixed_test_indices` for instance

ronuchit (approved):
lgtm after testing

tomsilver:
@ryangpeixu this pull request is with your project in mind! I will explain in person

tests/envs/test_pddl_procedural_generation.py
```diff
@@ -0,0 +1,73 @@
...
+                                       max_num_blocks_goal=2,
+                                       new_pile_prob=1.0,
+                                       force_goal_not_achieved=False)
+    rng = np.random.default_rng(123)
+    problem_strs = gen(10, rng)
+    for problem_str in problem_strs:
+        _, init_str, goal_str = _split_pddl_problem_str(problem_str)
+        assert init_str.count("ontable") == 3
+        assert goal_str.count("ontable") == 2
+        assert init_str.count("on ") == 0
```
ronuchit, line 72: why spaces here but not after ontable above? can we add spaces there too?

src/envs/pddl_env.py
```diff
@@ -0,0 +1,492 @@
...
+                          objects: Collection[Object]) -> _PDDLEnvState:
+        """Create a state from ground atoms and objects."""
+        # Keep a dummy state dict so we know what objects are in the state.
+        dummy_state_dict = {o: np.zeros(0, dtype=np.float32) for o in objects}
+        return _PDDLEnvState(dummy_state_dict, simulator_state=ground_atoms)
+
+    def allclose(self, other: State) -> bool:
+        return self.simulator_state == other.simulator_state
+
+    def copy(self) -> State:
```
ronuchit, line 51: so we're not using the base copy anymore?

ronuchit (approved):
lgtm again. since we are splitting test_pddl_env.py and test_pddl_blocks_env.py into separate files, do we also need to split pddl_env.py into a new file pddl_blocks_env.py? or not worth?

src/envs/pddl_env.py
tomsilver: that's right

tomsilver:
good point, I just consolidated the two test files for consistency with the rest of the test file structure

PR #616, opened 2022-03-25
fix lint error caused by pylint upgrade

ronuchit:
can be replicated locally with `pip install --upgrade pylint`

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -1006,8 +1006,8 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
         m["absolute_params"] = s.vec(o) + p
         return self._Pick_initiable(s, m, o, p)
 
-    def _Pick_policy(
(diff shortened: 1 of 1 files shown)


src/envs/cover.py
```diff
@@ -382,10 +382,10 @@ def get_name(cls) -> str:
     def options(self) -> Set[ParameterizedOption]:
         return {self._Pick, self._Place}
 
-    @staticmethod
-    def _Pick_policy(s: State, m: Dict, o: Sequence[Object],
+    def _Pick_policy(self, s: State, m: Dict, o: Sequence[Object],
                      p: Array) -> Action:
         del m  # unused
+        _ = self  # unused
```
ronuchit, line 388: can't use del here, pylint complains that method could be staticmethod bc self is unused

tomsilver (approved):
thx!

PR #617, opened 2022-03-25
disallow using state allclose when simulator_state is defined

tomsilver:
@NishanthJKumar @wmcclinton FYI this will break behavior. to repair, we'll need to create a subclass of `State` for behavior that overrides `allclose`. I can do this when the time comes.

### DIFF
--- a/tests/test_structs.py
+++ b/tests/test_structs.py
@@ -124,6 +124,11 @@ def test_state():
     state4 = State({obj3: np.array([1, 2])}, simulator_state="dummy")
     assert state4.simulator_state == "dummy"
     assert state4.copy().simulator_state == "dummy"
+    # Cannot use allclose with non-None simulator states.
+    with pytest.raises(NotImplementedError):
+        state4.allclose(state3)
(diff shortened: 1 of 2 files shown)


tests/test_structs.py
```diff
@@ -124,6 +124,11 @@ def test_state():
     state4 = State({obj3: np.array([1, 2])}, simulator_state="dummy")
     assert state4.simulator_state == "dummy"
     assert state4.copy().simulator_state == "dummy"
+    # Cannot use allclose with non-None simulator states.
+    with pytest.raises(NotImplementedError):
+        state4.allclose(state3)
+    with pytest.raises(NotImplementedError):
+        state3.allclose(state4)
```
ronuchit, line 131: also test allclose between a sim-state and a non-sim-state in both directions

ronuchit (approved):

tests/test_structs.py
tomsilver: resolved on slack

PR #625, opened 2022-03-28
Test predicate classifiers in CoverEnv

amburger66:


scripts/evaluate_predicate_classifiers.py
```diff
@@ -0,0 +1,140 @@
...
+        # [is_block, is_target, width, pose]
+        state.set(target, "pose", pose)
+    state.set(robot, "hand", 0.0)
+    # Test 1: no blocks overlap any targets, none are held
+    atoms = utils.abstract(state, (Holding, Covers))
+    atoms_gt = utils.abstract(state, (HoldingGT, CoversGT))
+    print(f"False positives: {atoms - atoms_gt}\n"
+          f"False negatives: {atoms_gt - atoms}")
+    # Pick up block0 and place it over target0
+    action = Action(np.array([0.15], dtype=np.float32))
```
tomsilver, line 118: remove some of these magic numbers

scripts/evaluate_predicate_classifiers.py
```diff
@@ -0,0 +1,140 @@
...
+    approach.load(online_learning_cycle=None)
+    teacher = Teacher(train_tasks)
+    for i in range(CFG.num_online_learning_cycles):
+        print(f"\n\nONLINE LEARNING CYCLE {i}\n")
+        if total_num_transitions > CFG.online_learning_max_transitions:
+            break
+        interaction_requests = approach.get_interaction_requests()
+        if not interaction_requests:
+            break  # agent doesn't want to learn anything more; terminate
+        interaction_results, _ = _generate_interaction_results(
```
tomsilver, line 69: is the reason that we need to go through this process just because we want to use `total_num_transitions` to break above? if so, can we instead just break whenever `load` fails? and then, can we simplify this function a lot? something like:

```
def _run_pipeline(env: BaseEnv,
                  approach: BilevelPlanningApproach) -> None:
    approach.load(online_learning_cycle=None)
    for i in range(CFG.num_online_learning_cycles):
        print(f"\n\nONLINE LEARNING CYCLE {i}\n")
        try:
            approach.load(online_learning_cycle=i)
        _evaluate_preds(
            approach._get_current_predicates(),  # pylint: disable=protected-access
            env)
        except FileNotFoundError:
            break
```

scripts/evaluate_predicate_classifiers.py
```diff
@@ -0,0 +1,140 @@
...
+from predicators.src.envs import BaseEnv, create_new_env
+from predicators.src.main import _generate_interaction_results, \
+    _generate_or_load_offline_dataset
+from predicators.src.settings import CFG
+from predicators.src.structs import Action, Dataset, Predicate, Task
+from predicators.src.teacher import Teacher
+
+
+def _main() -> None:
+    # Parse & validate args
```
tomsilver, line 21: see my comment below first, and then after that, I'm wondering how much of this function can potentially be removed

tomsilver (commented):
very nice!! this is going to be super useful, and it sounds like it already has been.

ronuchit:
agreed with tom, there is a principle in software engineering that says, roughly, 'if you are ever copy-pasting more than 5 lines of code, there's a better way to do what you're trying to do` -- let's try to remove the duplicated code!

scripts/evaluate_predicate_classifiers.py
amburger66: if i make this specific to InteractiveLearningApproach, most of it can go away right? we don't need to seed or anything since we're just loading every cycle

amburger66: actually it doesn't need to be specific to InteractiveLearningApproach

tomsilver (approved):
Nice! Much simpler now.

PR #629, opened 2022-03-29
rewrite backchaining algorithm from scratch (collab with Nishanth)

ronuchit:
We rewrote backchaining based on the ideas discussed in last week's meetings. It seems to work fine on repeated nextto, but will not work in rnt_painting even if we enforce the 1:1 operators:options assumption. We will discuss further on Tuesday 3/29's meeting.

Example of working command:
`python src/main.py --env repeated_nextto --approach nsrt_learning --side_predicate_learner backchaining --seed 502`
```
Learned NSRTs:
NSRT-Move:
    Parameters: [?x0:dot, ?x2:robot]
    Preconditions: []
    Add Effects: [NextTo(?x2:robot, ?x0:dot)]
    Delete Effects: [NextToNothing(?x2:robot)]
    Side Predicates: [NextTo]
    Option Spec: Move(?x2:robot, ?x0:dot)
NSRT-Grasp:
    Parameters: [?x0:dot, ?x1:robot]
    Preconditions: [NextTo(?x1:robot, ?x0:dot)]
    Add Effects: [Grasped(?x1:robot, ?x0:dot)]
    Delete Effects: [NextTo(?x1:robot, ?x0:dot)]
    Side Predicates: [NextToNothing]
    Option Spec: Grasp(?x1:robot, ?x0:dot)

Tasks solved: 50 / 50
Average time for successes: 0.02918 seconds
```


src/nsrt_learning/side_predicate_learning.py
```diff
@@ -76,12 +89,16 @@ def _recompute_datastores_from_segments(
...
-                                                     segment.init_atoms)
-                        if not atoms.issubset(segment.final_atoms):
-                            continue
+                        # Check if effects match.
+                        if semantics == "apply_operator":
+                            atoms = utils.apply_operator(
+                                ground_op, segment.init_atoms)
+                            if not atoms.issubset(segment.final_atoms):
+                                continue
+                        elif semantics == "add_effects":
```
ronuchit, line 98: We added these new semantics because we want to use them when recomputing the datastore of a PNAD after updating its add effects. I'm not convinced it's necessary or correct.

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -455,70 +377,183 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                if all(o in objects for o in atom.objects)
+            }
+            lifted_atoms = {atom.lift(obj_to_var) for atom in atoms}
+            delete_effects |= lifted_atoms
+        pnad.op = pnad.op.copy_with(delete_effects=delete_effects)
+
+    @staticmethod
+    def _finalize_pnad_side_predicates(pnad: PartialNSRTAndDatastore) -> None:
+        """Update the given PNAD to change the side predicates to ones that
+        include every unmodeled add or delete effect seen in the data."""
```
tomsilver, line 546: for discussion later, would you agree with the following? side predicates serve two purposes:
1. there are some add effects that we would prefer not to model for generalization and planning efficiency reasons. side predicates give us a mechanism to ignore these add effects (this was their original purpose).
2. there are some delete effects that we _can't_ model, given the parameters that we have, so side predicates give us a way to throw these out anyway without actually including "deictic references" to new objects. we could instead introduce new parameters, but we don't want to, again for generalization and planning efficiency reasons.

tomsilver (commented):
very nice work!

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: Exactly right! I fully agree and was planning to bring this up during our meeting today.

src/nsrt_learning/side_predicate_learning.py
ronuchit: we did this because we saw a case where it mattered, although i can't seem to replicate empirically (even buggy code can work fine on RNT so it's hard to tell). @NishanthJKumar do you remember

src/nsrt_learning/side_predicate_learning.py
ronuchit: yes, but we'll want to revise these claims after discussion later

src/structs.py
```diff
@@ -617,6 +617,21 @@ def __gt__(self, other: object) -> bool:
         assert isinstance(other, STRIPSOperator)
         return str(self) > str(other)
 
+    def copy_with(self, **kwargs: Any) -> STRIPSOperator:
```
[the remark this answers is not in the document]
ronuchit: actually since there's two changes to structs.py i don't think there's much point splitting up the PR

ronuchit:
@tomsilver ready for another look, now with tests. note the annoying circular import issue in the test file... i have run into this before, it stems from the fact that planning.py imports ApproachFailure/Timeout, which causes all sorts of circular dependencies in import statements... maybe we should just fix this in a separate PR?

tomsilver (approved):
approving to unblock, but unless there is any urgency, it may be good to wait until after our meeting today to merge, in case anything else comes up in discussion

ronuchit:
sure, we can wait. i think our course might change significantly after our meeting today, though

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -455,70 +382,193 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
                 option = segment.get_option()
                 # Find the PNAD associated with this option.
                 pnad = param_opt_to_pnad[option.parent]
-                var_to_obj = pnad.get_sub_for_member_segment(segment)
                 # Compute the ground atoms that must be added on this timestep.
+                # Note that they must be a subset of the add effects.
```
NishanthJKumar, line 386: Nit: could we mention "subset of the current pnad's add effects"

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -455,70 +382,193 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
-                # effects, then we need to make the add effects more specific.
-                if not necessary_add_effects.issubset(predicted_add_effects):
-                    obj_to_var = {o: v for v, o in var_to_obj.items()}
-                    assert len(obj_to_var) == len(var_to_obj)
+                assert necessary_add_effects.issubset(segment.add_effects)
+
+                # Find a mapping from the variables in the PNAD add effects
+                # and option to the objects in necessary_add_effects and the
+                # segment's option. If one exists, we don't need to modify this
+                # PNAD. Otherwise, we must make its add effects more specific.
```
NishanthJKumar, line 393: Could we add a note to this comment that we assume all variables in the parameters of the PNAD must appear in the option args or add effects?

NishanthJKumar (approved):
LGTM! Thanks for such quick turnaround on this 💯 . I left two small notes related to comments. Also, I like your test cases 😸

ronuchit:
@NishanthJKumar how's this https://github.com/Learning-and-Intelligent-Systems/predicators/pull/629/commits/e0edceb4c9423fa48bb21173ed51077ee4fc72fd

NishanthJKumar:
Awesome - thanks! Let's see what next steps should be after meeting today!

ronuchit:
going to merge this PR since we understand that we need further algorithmic insights

Issue #615, opened 2022-03-24
`isinstance` fails nondeterministically with envs / approaches


tomsilver:
likely because of the use of `load_module` in `envs/__init__.py` and `approaches/__init__.py`, related to this: https://stackoverflow.com/questions/15876395/isinstance-behavior-with-module-reload

tomsilver:
I can't figure out any good workaround, and I don't expect this to really cause issues down the line, but in case it does, we'll have this as a record at least.

tomsilver:
@amburger66 encountered a version of this issue independently, so I no longer feel like we can ignore it. FYI @ronuchit. We can discuss how to proceed...

PR #630, opened 2022-03-29
Use absolute imports when dynamically loading modules

williamshen-nz:
To fix: https://github.com/Learning-and-Intelligent-Systems/predicators/issues/615

By using the package loader and find_modules, that was sometimes giving us the relative import instead of absolute.


src/envs/__init__.py
```diff
@@ -10,11 +10,14 @@
 __all__ = ["BaseEnv"]
 _MOST_RECENT_ENV_INSTANCE = {}
 
+
```
ronuchit, line 13: added a blank line here? let's be consistent and do it in both of the files or neither of them

src/envs/__init__.py
```diff
@@ -10,11 +10,14 @@
...
 if not TYPE_CHECKING:
     # Load all modules so that utils.get_all_subclasses() works.
-    for loader, module_name, _ in pkgutil.walk_packages(__path__):
+    for _, module_name, _ in pkgutil.walk_packages(__path__):
         if "__init__" not in module_name:
-            loader.find_module(module_name).load_module(module_name)
+            # Important! We use an absolute import here to avoid issues
+            # with isinstance checking when using relative imports.
+            importlib.import_module(f"{__name__}.{module_name}")
```
ronuchit, line 22: added a new file pybullet_cover.py? can we delete that?

ronuchit (commented):
thank you!! very minor things

ronuchit (approved):
thank you!

PR #635, opened 2022-03-29
add option execution failure

### LINKED ISSUE #633: fix last remaining circular import issue
In tests/nsrt_learning/test_nsrt_learning_main.py, we have these lines:
```
# We need this unused import to prevent cyclic import issues when running
# this file as a standalone test (pytest -s tests/test_nsrt_learning.py).
from predicators.src import approaches  # pylint: disable=unused-import
```
(actually, that should say `pytest -s tests/nsrt_learning/test_nsrt_learning_main.py`)

I did some work in #631 to remove the circular dependencies which were causing this, but one remains: the use of `ApproachFailure` in option_learning.py. Can we fix this? This file isn't technically an approach, so ideally we don't want it to know about `ApproachFailure`s or `ApproachTimeout`s. The right thing to do is to raise an OptionPolicyFailure or something, and then catch that and raise ApproachFailure in the appropriate place in `approaches/`. I'm having trouble figuring out where it would be used, though...

After this is fixed, we should be able to remove those lines from tests/nsrt_learning/test_nsrt_learning_main.py, and `pytest -s tests/nsrt_learning/test_nsrt_learning_main.py` should work.

tomsilver:
closes #633


ronuchit:
thanks, this looks like the right idea! i verified that it fixes the standalone execution of that test file. let's get it to the finish line!

tomsilver:
@ronuchit ready for review

ronuchit (approved):
nice idea to combine with OptionPlanExhausted!!

PR #642, opened 2022-03-31
Refactors sidelining to use the new `_try_specifizing_pnad` method

NishanthJKumar:
Refactors sidelining code in preparation for non 1:1 sidelining implementation merge!

@ronuchit - I couldn't figure out how to nicely include the new test I want to. Could you fix up `test_side_predicate_learning.py` if you have some time tomorrow? I unfortunately have a bunch of other things to do, but could look into this more tomorrow evening ish.


src/nsrt_learning/side_predicate_learning.py
```diff
@@ -409,53 +409,24 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                    del new_pnad  # unused from here
+                    param_opt_to_pnad[option.parent] = pnad
                     # After all this, the unification call that failed earlier
-                    # (leading us to enter this if statement) should now work.
-                    unifies, _ = utils.find_substitution(
+                    # (leading us into the current if statement) should work.
+                    unifies, obj_to_var_ent = utils.find_substitution(
                         pnad.op.add_effects | opt_vars_atom,
                         necessary_add_effects | opt_objs_atom)
+                    obj_to_var = cast(Dict[Object, Variable], obj_to_var_ent)
```
tomsilver, line 429: i know that this pattern is used elsewhere, but don't we have an `ObjToVarSub` struct that we can use here?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -484,6 +455,61 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+        # effects are a subset of necessary_add_effects.
+        ground_op = self._get_partially_satisfying_grounding(
+            necessary_add_effects, pnad, segment)
+        # If no such grounding exists, specifizing is not possible.
+        if ground_op is None:
+            return None
+        # To figure out the effects we need to add to this PNAD,
+        # we first look at the ground effects that are missing
+        # from this arbitrary ground operator.
+        missing_effects = (necessary_add_effects - ground_op.add_effects)
```
tomsilver, line 484: remove parens?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -484,6 +455,61 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+        assert all(name.startswith("?x") for name in var_names)
+        max_var_name = max(var_names,
+                           key=lambda name: int(name[2:]),
+                           default="?x-1")
+        max_var_num = int(max_var_name[2:])
+        for effect in missing_effects:
+            for obj in effect.objects:
+                if obj not in obj_to_var:
+                    max_var_num += 1
+                    new_var = Variable("?x" + str(max_var_num), obj.type)
```
tomsilver, line 500: sorry I missed this before. so now it looks like we have dashes in between ?x and the number. should we add a dash here too then?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -484,6 +455,61 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+        # To figure out the effects we need to add to this PNAD,
+        # we first look at the ground effects that are missing
+        # from this arbitrary ground operator.
+        missing_effects = (necessary_add_effects - ground_op.add_effects)
+        obj_to_var = dict(zip(ground_op.objects, pnad.op.parameters))
+        # Before we can lift missing_effects, we need to add new
+        # entries to obj_to_var to account for the situation where
+        # missing_effects contains objects that were not in
+        # the ground operator's parameters.
+        var_names = [var.name for var in obj_to_var.values()]
```
tomsilver, line 490: you can leave an issue for me, but it would be good to factor out this stuff and make a utility that creates fresh variable names for us, and use that utility both here and in strips learning

tests/nsrt_learning/test_side_predicate_learning.py
```diff
@@ -85,26 +89,31 @@ def test_backchaining():
...
+def test_get_partially_satisfying_grounding_and_try_refining_pnad():
+    """Test the _get_partially_satisfying_grounding() and try_refining_pnad()
+    methods in the BackchainingSidePredicateLearner."""
 
     human_type = Type("human_type", ["feat"])
     Asleep = Predicate("Asleep", [human_type], lambda s, o: s[o[0]][0] > 0.5)
     Happy = Predicate("Happy", [human_type], lambda s, o: s[o[0]][0] > 0.5)
     opt = SingletonParameterizedOption("Move", lambda s, m, o, p: None)
-    human_var = human_type("?human")
+    human_var = human_type("?x0")
```
tomsilver, line 100: `?x-0`?

tomsilver (approved):

src/nsrt_learning/side_predicate_learning.py
ronuchit: no, the `?x-1` is a negative one. so that when we increment, we start at 0 as intended. i'll add a comment

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -507,29 +533,31 @@ def _get_partially_satisfying_grounding(
             return ground_op
         return None
 
-    def _update_pnad_with_params_and_add_effects(
+    def _create_new_pnad_with_params_and_add_effects(
```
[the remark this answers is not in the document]
ronuchit: yessir

src/nsrt_learning/side_predicate_learning.py
tomsilver: oh i see. okay, I definitely want to do the utility function refactor that I mentioned then!

src/nsrt_learning/side_predicate_learning.py
ronuchit: thx, it wasn't in too many places, so i fixed them all

ronuchit:
Supercloud confirms that repeated_nextto is still fine. Also, both repeated_nextto_single_option and repeated_nextto_painting give the error:
`AssertionError: 1:1 assumption violated`
which is exactly what we expect.

However, there are at least two weirdnesses that need to be investigated:
This leads to an assertion failing, which I thought should never fail: `python src/main.py --env repeated_nextto_painting --approach nsrt_learning --side_predicate_learner backchaining --seed 502 --num_train_tasks 50 --rnt_painting_num_objs_train [1] --rnt_painting_num_objs_test [1]`
This leads to harmlessness failing, which I thought should not happen in the 1-object painting case since the machine problem doesn't occur: `python src/main.py --env repeated_nextto_painting --approach nsrt_learning --side_predicate_learner backchaining --seed 502 --num_train_tasks 50 --rnt_painting_num_objs_train [1] --rnt_painting_num_objs_test [1] --painting_initial_holding_prob 0.0`

cc @NishanthJKumar

ronuchit (approved):
as far as i'm concerned, this refactor can be merged. but note the weirdnesses in my comment that need to be investigated (we can investigate after merging, but i would really like to understand what's happening there before we move to the non-1:1 PR)

PR #646, opened 2022-03-31
minor naming changes to side predicate learning file

ronuchit:
no functionality changes. we generally avoid "hl_state" (after this
change, `git grep "hl_state" | wc -l` gives 0) in favor of "atoms". also
some consistency issues and confusion with state as in search state vs
state as in atoms


NishanthJKumar (approved):
LGTM!

PR #647, opened 2022-03-31
fix a bunch of bugs in sidelining

ronuchit:
* fix incorrect usage of `ll_traj.actions` (which doesn't matter in
current envs, but would break in envs with multistep option policies)
* allow noops in call to `task_plan_grounding()` in harmlessness check,
because we want to consider operators like OpenLid in painting. note
that the planner would use such operators, only because of the
DiscoveredFailure stuff...sigh
* the most general PNADs now have preconditions computed via an
intersection, so that these operators are better than before. specifically,
this helps the OpenLid operator: it never gets specifized, but we still
want GripperOpen to appear in its precondition


tomsilver (approved):
Should we actually deprecate the discovered failure junk?

PR #648, opened 2022-03-31
add box only or shelf only setting for painting

ronuchit:
* no functional change to default code path (verified)
* checked that with box only or shelf only goals, nsrt_learning produces
1:1 operators/options that don't have any HoldingSide/HoldingTop
anywhere respectively, *as long as you also set
painting_initial_holding_prob to 0.0*


tomsilver (commented):
this logic is getting pretty difficult to follow, is there anything we can do to make it more readable?

ronuchit:
> this logic is getting pretty difficult to follow, is there anything we can do to make it more readable?

i dunno... hm... any ideas? one complication is that max_objs_in_goal could be float(inf) so we can't just make a list and mask out entries using max_objs_in_goal

ronuchit:
@tomsilver good now?

PR #651, opened 2022-04-02
Non 1 to 1 sidelining

NishanthJKumar:
ready for review! will fix tests and revert/edit the supercloud script; but would love feedback on everything else!

### DIFF
--- a/src/nsrt_learning/side_predicate_learning.py
+++ b/src/nsrt_learning/side_predicate_learning.py
@@ -427,9 +427,9 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
                             param_opt_to_nec_pnads[option.parent].append(pnad)
                         break
 
-                # If we weren't able to find a substitution (i.e, the above)
+                # If we weren't able to find a substitution (i.e, the above
                 # for loop did not break), we need to try
-                # specialize each of our PNADs.
+                # specializing each of our PNADs.
                 else:
                     for pnad in pnads_for_option:
                         new_pnad = self._try_specializing_pnad(
--- a/src/nsrt_learning/side_predicate_learning.py
+++ b/src/nsrt_learning/side_predicate_learning.py
@@ -431,7 +431,7 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
                         # and we are making them more specific, it is guaranteed
                         # that such a grounding exists, otherwise our 1:1
                         # option:operator assumption is violated.
-                        new_pnad = self._try_specifizing_pnad(
+                        new_pnad = self._try_specializing_pnad(
                             necessary_add_effects, pnad, segment)
                         if new_pnad is not None:
                             assert new_pnad.option_spec == pnad.option_spec
@@ -447,7 +447,7 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
                     # to split from the most general PNAD and make a new PNAD
                     # to cover these necessary add effects.
                     if should_split_off_new_pnad:
-                        new_pnad = self._try_specifizing_pnad(
+                        new_pnad = self._try_specializing_pnad(
                             necessary_add_effects,
                             param_opt_to_general_pnad[option.parent],
                             segment)
--- a/src/nsrt_learning/side_predicate_learning.py
+++ b/src/nsrt_learning/side_predicate_learning.py
@@ -363,12 +363,13 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
     def _sideline(self) -> List[PartialNSRTAndDatastore]:
         # Initialize the most general PNADs by merging self._initial_pnads.
         # As a result, we will have one very general PNAD per option.
-        param_opt_to_pnad = {}
+        param_opt_to_general_pnad = {}
+        param_opt_to_nec_pnad = {}
         parameterized_options = {p.option_spec[0] for p in self._initial_pnads}
         total_datastore_len = 0
         for param_opt in parameterized_options:
             pnad = self._initialize_general_pnad_for_option(param_opt)
-            param_opt_to_pnad[param_opt] = pnad
+            param_opt_to_general_pnad[param_opt] = pnad
             total_datastore_len += len(pnad.datastore)
         del self._initial_pnads  # no longer used
         # Assert that all data is in some PNAD's datastore.
@@ -394,46 +395,87 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
                 segment = seg_traj[t]
                 option = segment.get_option()
                 # Find the PNAD associated with this option.
-                pnad = param_opt_to_pnad[option.parent]
+                if param_opt_to_nec_pnad.get(option.parent) is None:
+                    pnads_for_option = [param_opt_to_general_pnad[option.parent]]
+                else:
+                    pnads_for_option = param_opt_to_nec_pnad[option.parent]
                 # Compute the ground atoms that must be added on this timestep.
                 # They must be a subset of the current PNAD's add effects.
                 necessary_add_effects = necessary_image - atoms_seq[t]
                 assert necessary_add_effects.issubset(segment.add_effects)
 
-                # Find a mapping from the variables in the PNAD add effects
-                # and option to the objects in necessary_add_effects and the
-                # segment's option. If one exists, we don't need to modify this
-                # PNAD. Otherwise, we must make its add effects more specific.
-                # Note that we are assuming all variables in the parameters
-                # of the PNAD will appear in either the option arguments or
-                # the add effects. This is in contrast to strips_learning.py,
-                # where delete effect variables also contribute to parameters.
-                opt_pred = Predicate("OPT-ARGS",
-                                     [a.type for a in pnad.option_spec[1]],
-                                     _classifier=lambda s, o: False)  # dummy
-                opt_vars_atom = frozenset(
-                    {LiftedAtom(opt_pred, pnad.option_spec[1])})
-                opt_objs_atom = frozenset(
-                    {GroundAtom(opt_pred, option.objects)})
-                unifies, obj_to_var_ent = utils.find_substitution(
-                    pnad.op.add_effects | opt_vars_atom,
-                    necessary_add_effects | opt_objs_atom)
-                obj_to_var = cast(ObjToVarSub, obj_to_var_ent)
+                substitution_exists = False
+                # We start by checking if any of the PNADs associated with the
+                # demonstrated option are able to match this transition.
+                for pnad in pnads_for_option:
+                    # Find a mapping from the variables in the PNAD add effects
+                    # and option to the objects in necessary_add_effects and the
+                    # segment's option. If one exists, we don't need to modify this
+                    # PNAD. Otherwise, we must make its add effects more specific.
+                    # Note that we are assuming all variables in the parameters
+                    # of the PNAD will appear in either the option arguments or
+                    # the add effects. This is in contrast to strips_learning.py,
+                    # where delete effect variables also contribute to parameters.
+                    opt_pred = Predicate(
+                        "OPT-ARGS", [a.type for a in pnad.option_spec[1]],
+                        _classifier=lambda s, o: False)  # dummy
+                    opt_vars_atom = frozenset(
+                        {LiftedAtom(opt_pred, pnad.option_spec[1])})
+                    opt_objs_atom = frozenset(
+                        {GroundAtom(opt_pred, option.objects)})
+                    unifies, obj_to_var_ent = utils.find_substitution(
+                        pnad.op.add_effects | opt_vars_atom,
+                        necessary_add_effects | opt_objs_atom)
+                    obj_to_var = cast(ObjToVarSub, obj_to_var_ent)
+                    if unifies:
+                        substitution_exists = True
+                        break
 
-                if not unifies:  # we need to make add effects more specific
+                # If we weren't able to find a substitution, we need to try
+                # specifize each of our PNADs.
+                if not substitution_exists:
                     del obj_to_var  # undefined; delete it for safety
-                    # We now try to make the PNAD's add effects more specific.
-                    # Since we have started with the most general operators,
-                    # and we are making them more specific, it is guaranteed
-                    # that such a grounding exists, otherwise our 1:1
-                    # option:operator assumption is violated.
-                    new_pnad = self._try_specifizing_pnad(
-                        necessary_add_effects, pnad, segment)
-                    assert new_pnad is not None, "1:1 assumption violated"
-                    assert new_pnad.option_spec == pnad.option_spec
+
+                    should_split_off_new_pnad = True
+                    for pnad in pnads_for_option:
+                        # We now try to make the PNAD's add effects more specific.
+                        # Since we have started with the most general operators,
+                        # and we are making them more specific, it is guaranteed
+                        # that such a grounding exists, otherwise our 1:1
+                        # option:operator assumption is violated.
+                        new_pnad = self._try_specifizing_pnad(
+                            necessary_add_effects, pnad, segment)
+                        if new_pnad is not None:
+                            assert new_pnad.option_spec == pnad.option_spec
+                            if param_opt_to_nec_pnad.get(option.parent) is not None:
+                                param_opt_to_nec_pnad[option.parent].remove(pnad)
+                            del pnad
+                            should_split_off_new_pnad = False
+                            break
+
+                    # If we were unable to specifize any of the PNADs, we need
+                    # to split from the most general PNAD and make a new PNAD
+                    # to cover these necessary add effects.
+                    if should_split_off_new_pnad:
+                        new_pnad = self._try_specifizing_pnad(
+                            necessary_add_effects, param_opt_to_general_pnad[option.parent],
+                            segment, check_datastore_preservation=False)
+                        assert new_pnad is not None
+                        if param_opt_to_nec_pnad.get(option.parent) is None:
+                            op_num = 0
+                        else:
+                            op_num = len(param_opt_to_nec_pnad[option.parent])
+                        new_pnad_op_name = new_pnad.op.name + str(op_num)
+                        new_pnad.op = new_pnad.op.copy_with(
+                            name=new_pnad_op_name)
+
+                    # Add the new PNAD to the dictionary mapping options to
+                    # PNADs.
                     pnad = new_pnad
-                    del new_pnad  # unused from here
-                    param_opt_to_pnad[option.parent] = pnad
+                    if param_opt_to_nec_pnad.get(option.parent) is None:
+                        param_opt_to_nec_pnad[option.parent] = [pnad]
+                    else:
+                        param_opt_to_nec_pnad[option.parent].append(pnad)
                     # After all this, the unification call that failed earlier
                     # (leading us into the current if statement) should work.
                     unifies, obj_to_var_ent = utils.find_substitution(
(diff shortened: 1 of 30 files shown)


ronuchit:
don't have time to look till next week, but just want to request @NishanthJKumar for a small QOL thing, can you fix the naming so that it's Op0, Op1, Op2, ... instead of Op, Op1, Op2, ... (that is, make sure the first one has a "0" at the end)

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -354,6 +354,10 @@ def _initialize_general_pnad_for_option(
         preconditions = induce_pnad_preconditions(pnad)
         pnad.op = pnad.op.copy_with(preconditions=preconditions)
 
+        # Finally, remove the side predicates from these operators.
+        # These will be filled in later.
+        pnad.op = pnad.op.copy_with(side_predicates=set())
```
ronuchit, line 359: are we doing this so we can remove the "add_effects" semantics stuff?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -363,12 +367,14 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
     def _sideline(self) -> List[PartialNSRTAndDatastore]:
         # Initialize the most general PNADs by merging self._initial_pnads.
         # As a result, we will have one very general PNAD per option.
-        param_opt_to_pnad = {}
+        param_opt_to_general_pnad = {}
+        param_opt_to_nec_pnad: Dict[ParameterizedOption,
```
ronuchit, line 371: rename to `param_opt_to_nec_pnads`

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -363,12 +367,14 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
...
+                                    List[PartialNSRTAndDatastore]] = {}
         parameterized_options = {p.option_spec[0] for p in self._initial_pnads}
```
ronuchit, line 373: btw, just wanted to flag that i ran `python src/main.py --env repeated_nextto_painting --approach nsrt_learning --side_predicate_learner backchaining --seed 0 --num_train_tasks 10` and something is still wrong with the naming -- it produced, for me "Place1", "Place02", and "Place22".

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +400,92 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
                 segment = seg_traj[t]
                 option = segment.get_option()
                 # Find the PNAD associated with this option.
-                pnad = param_opt_to_pnad[option.parent]
+                if param_opt_to_nec_pnad.get(option.parent) is None:
```
ronuchit, line 403: can we initialize the `param_opt_to_nec_pnad` to be an empty list for each option? so we don't have to use this `.get()` stuff anywhere, which is hard for the reader to follow what's going on

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +400,92 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
-                    # We now try to make the PNAD's add effects more specific.
-                    # Since we have started with the most general operators,
-                    # and we are making them more specific, it is guaranteed
-                    # that such a grounding exists, otherwise our 1:1
-                    # option:operator assumption is violated.
-                    new_pnad = self._try_specializing_pnad(
-                        necessary_add_effects, pnad, segment)
-                    assert new_pnad is not None, "1:1 assumption violated"
-                    assert new_pnad.option_spec == pnad.option_spec
+                substitution_exists = False
```
ronuchit, line 415: You don't need this `substitution_exists` variable; you can use the `for ... else ...` syntax instead (https://book.pythontips.com/en/latest/for_-_else.html) just change the `if not substitution_exists:` below to `else:`

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +400,92 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                # If we weren't able to find a substitution, we need to try
+                # specifize each of our PNADs.
+                if not substitution_exists:
+                    should_split_off_new_pnad = True
+                    for pnad in pnads_for_option:
+                        # We now try to make the PNAD's add effects more
+                        # specific. Since we have started with the most
+                        # general operators, and we are making them more
+                        # specific, it is guaranteed that such a
+                        # grounding exists, otherwise our 1:1
```
ronuchit, line 438: can you go through and make sure all comments are fully up-to-date? (like this one, since we're no longer doing 1:1 stuff)

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +400,92 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                    # to cover these necessary add effects.
+                    if should_split_off_new_pnad:
+                        new_pnad = self._try_specializing_pnad(
+                            necessary_add_effects,
+                            param_opt_to_general_pnad[option.parent], segment)
+                        assert new_pnad is not None
+
+                    # Add the new PNAD to the dictionary mapping options to
+                    # PNADs.
+                    assert isinstance(new_pnad, PartialNSRTAndDatastore)
```
ronuchit, line 463: do we need this assert? can we remove it?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -456,18 +495,68 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
                     for a in pnad.op.preconditions
                 }
 
+        # Now that the add effects and preconditions are correct,
+        # make a list of all final PNADs. Note
+        # that these final PNADs only come from the
+        # param_opt_to_nec_pnad dict, since we can be assured
+        # that our backchaining process ensured that the
+        # PNADs in this dict cover all of the data!
```
ronuchit, line 503: is there an easy way to assert this? (hmm, i can't think of one... because segments could be in multiple pnads' datastores... right?)

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -456,18 +495,68 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                continue
+            # If find_partial_grounding is True, we want to find a grounding
+            # that achieves some subset of the necessary add effects. Else,
+            # we want to find a grounding that is some superset of the
+            # necessary_add_effects
+            if find_partial_grounding:
+                if not ground_op.add_effects.issubset(necessary_add_effects):
+                    continue
+            else:
+                if not ground_op.add_effects.issubset(segment.final_atoms):
```
ronuchit, line 554: shouldn't this `if` clause be outside the `else:` right above? you suggested that to me haha i had it this way before

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -487,8 +576,8 @@ def _try_specializing_pnad(
         # Get an arbitrary grounding of the PNAD's operator whose
         # preconditions hold in segment.init_atoms and whose add
         # effects are a subset of necessary_add_effects.
-        ground_op = self._get_partially_satisfying_grounding(
-            necessary_add_effects, pnad, segment)
+        ground_op = self._find_unification(necessary_add_effects, pnad,
+                                           segment, True)
```
ronuchit, line 580: generally preferred to use keyword argument when passing in a constant thing, e.g. `..., find_partial_grounding=True)`

ronuchit (commented):
i will do more passes properly later, but here's some initial comments on a cursory glance!

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: yes - that was the intent

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: necessary for mypy to pass

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: no - i think this is right; we only want to run this check if we're not trying to find a partial grounding

ronuchit: gotcha -- note that this is different from before. it's okay though

src/nsrt_learning/side_predicate_learning.py
ronuchit: change to `assert new_pnad is not None`? does mypy still pass?

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: we could run through all of datapoints in our training data and check that each one is associated with at least one PNAD?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +401,80 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                    if ground_op is not None:
+                        obj_to_var = dict(
+                            zip(ground_op.objects, pnad.op.parameters))
+                        if len(param_opt_to_nec_pnads[option.parent]) == 0:
+                            pnad_op_name = pnad.op.name + "0"
+                            pnad.op = pnad.op.copy_with(name=pnad_op_name)
+                            param_opt_to_nec_pnads[option.parent].append(pnad)
+                        break
+
+                # If we weren't able to find a substitution (i.e, the above)
```
ronuchit, line 430: nits: no close paren after "above", and "specialize" -> "specializing"

src/nsrt_learning/side_predicate_learning.py
ronuchit: this would be a good check, but it doesn't really fit into the recompute_datastores API nicely...

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: huh - turns out this works (i thought i tried this before and it didn't work)

tests/nsrt_learning/test_side_predicate_learning.py
```diff
@@ -89,10 +97,38 @@ def test_backchaining():
...
+    goal3 = {Asleep([bob]), Sad([bob])}
+    act3 = Action([], Cry)
+    traj3 = LowLevelTrajectory([state_awake_and_sad, state_asleep_and_sad],
+                               [act3], True, 0)
+    traj4 = LowLevelTrajectory([state_awake_and_happy, state_asleep_and_sad],
+                               [act3], True, 0)
+    task3 = Task(state_awake_and_sad, goal3)
+    segment3 = Segment(traj3, set([Sad([bob])]), goal3, Cry)
+    segment4 = Segment(traj4, set(), goal3, Cry)
+    for pnad in initial_pnads:
```
ronuchit, line 111: instead of this for loop, do:
```
cry_pnads = [pnad for pnad in initial_pnads if pnad.op.name == "CryOp"]
assert len(cry_pnads) == 1
cry_pnad = cry_pnads[0]
```
(it's just slightly more robust this way)

tests/nsrt_learning/test_side_predicate_learning.py
```diff
@@ -89,10 +97,38 @@ def test_backchaining():
...
+    Delete Effects: []
+    Side Predicates: []
+    Option Spec: Cry()"""
+    assert str(pnads[0]) == repr(pnads[0]) == expected_str
+
 
-def test_get_partially_satisfying_grounding_and_try_refining_pnad():
-    """Test the _get_partially_satisfying_grounding() and try_refining_pnad()
-    methods in the BackchainingSidePredicateLearner."""
+def test_find_unification_and_try_refining_pnad():
```
ronuchit, line 129: i think this name is out of date(try_refining_pnad)? can you go through and make sure all the names in this test file are up to date?

tests/nsrt_learning/test_side_predicate_learning.py
```diff
@@ -89,10 +97,38 @@ def test_backchaining():
...
+    """Test the _find_unification() and try_refining_pnad() methods in the
```
ronuchit, line 130: note: for above, make sure to fix the docstrings too

ronuchit (commented):
looks great, glad you were able to figure out the testing! i just left a bunch of very minor comments

let's launch the supercloud run i suggested in one of my comments:
{oracle, no_sidepreds, backchain} x {repeated_nextto, repeated_nextto_single_option}

if that goes well (no_sidepreds should do poorly), then i think we're good to go!

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +397,79 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
-                        necessary_add_effects, pnad, segment)
-                    assert new_pnad is not None, "1:1 assumption violated"
-                    assert new_pnad.option_spec == pnad.option_spec
+                # We start by checking if any of the PNADs associated with the
+                # demonstrated option are able to match this transition.
+                for pnad in pnads_for_option:
+                    ground_op = self._find_unification(necessary_add_effects,
+                                                       pnad, segment)
+                    if ground_op is not None:
+                        obj_to_var = dict(
```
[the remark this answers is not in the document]
NishanthJKumar: yes exactly!

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +397,79 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                    # to cover these necessary add effects.
+                    else:
+                        new_pnad = self._try_specializing_pnad(
+                            necessary_add_effects,
+                            param_opt_to_general_pnad[option.parent], segment)
+                        assert new_pnad is not None
+
+                    # Add the new PNAD to the dictionary mapping options to
+                    # PNADs.
+                    if not new_pnad.op.name[-1].isdigit():
```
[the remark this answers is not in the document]
NishanthJKumar: yes - it should still work!

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,59 +397,79 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                        op_num = len(param_opt_to_nec_pnads[option.parent])
+                        new_pnad_op_name = new_pnad.op.name + str(op_num)
+                        new_pnad.op = new_pnad.op.copy_with(
+                            name=new_pnad_op_name)
                     pnad = new_pnad
-                    del new_pnad  # unused from here
```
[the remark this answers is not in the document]
NishanthJKumar: added it back in; just got taken out and forgotten during refactor.

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,53 +397,73 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                            necessary_add_effects, pnad, segment)
+                        if new_pnad is not None:
+                            assert new_pnad.option_spec == pnad.option_spec
+                            if len(param_opt_to_nec_pnads[option.parent]) > 0:
+                                param_opt_to_nec_pnads[option.parent].remove(
+                                    pnad)
+                            del pnad
+                            break
+                    # If we were unable to specialize any of the PNADs, we need
+                    # to split from the most general PNAD and make a new PNAD
```
tomsilver, line 440: can we remove occurrences of the word "split" in favor of "spawn"?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,53 +397,73 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                        new_pnad = self._try_specializing_pnad(
+                            necessary_add_effects,
+                            param_opt_to_general_pnad[option.parent], segment)
+                        assert new_pnad is not None
+
+                    # Add the new PNAD to the dictionary mapping options to
+                    # PNADs. Note that even if there are more than 10
+                    # PNADs associated with an option, the below check will
+                    # still work.
+                    if not new_pnad.op.name[-1].isdigit():
```
tomsilver, line 452: i'm confused about the interaction between this line and the place above where you add "0" to the end. when does this line get hit?

also is there any hope that we can purge all of this naming stuff and just rename the operators once at the end? i think that would be cleaner, if it doesn't break anything

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -456,18 +479,67 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
                 }
 
+        # Now that the add effects and preconditions are correct,
+        # make a list of all final PNADs. Note
+        # that these final PNADs only come from the
+        # param_opt_to_nec_pnads dict, since we can be assured
+        # that our backchaining process ensured that the
+        # PNADs in this dict cover all of the data!
+        all_pnads = []
+        for pnad_list in param_opt_to_nec_pnads.values():
```
tomsilver, line 489: i'm nervous about nondeterministic ordering here... is this dictionary always built in the same way? if we're not sure, we can sort this to ensure determinism

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -456,18 +479,67 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+    ) -> Optional[_GroundSTRIPSOperator]:
+        """Find a mapping from the variables in the PNAD add effects and option
+        to the objects in necessary_add_effects and the segment's option.
+
+        If one exists, we don't need to modify this PNAD. Otherwise, we
+        must make its add effects more specific. Note that we are
+        assuming all variables in the parameters of the PNAD will appear
+        in either the option arguments or the add effects. This is in
+        contrast to strips_learning.py, where delete effect variables
+        also contribute to parameters. If find_partial_grounding is
```
tomsilver, line 520: can we rename `find_partial_grounding` to something that makes clear this superset/subset behavior?

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -456,18 +479,67 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+        # Loop over all groundings.
+        for ground_op in utils.all_ground_operators_given_partial(
+                pnad.op, objects, isub):
+            if not ground_op.preconditions.issubset(segment.init_atoms):
+                continue
+            if find_partial_grounding:
+                if not ground_op.add_effects.issubset(necessary_add_effects):
+                    continue
+            else:
+                if not ground_op.add_effects.issubset(segment.final_atoms):
```
tomsilver, line 537: this is doing more than what is described in the docstring -- please update the docstring to describe both of these checks

tomsilver (approved):
minor stuff only. excellent and exciting work!!

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -394,53 +397,73 @@ def _sideline(self) -> List[PartialNSRTAndDatastore]:
...
+                # demonstrated option are able to match this transition.
+                for pnad in pnads_for_option:
+                    ground_op = self._find_unification(necessary_add_effects,
+                                                       pnad, segment)
+                    if ground_op is not None:
+                        obj_to_var = dict(
+                            zip(ground_op.objects, pnad.op.parameters))
+                        if len(param_opt_to_nec_pnads[option.parent]) == 0:
+                            pnad_op_name = pnad.op.name + "0"
+                            pnad.op = pnad.op.copy_with(name=pnad_op_name)
```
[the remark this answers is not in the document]
NishanthJKumar: ooh - good point; let me make a small edit to make this safer!

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: good idea on just naming things at the end! I think we should be able to do this; lemme try it!

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: Let's sort just to be safe

PR #652, opened 2022-04-02
use interaction data to learn nsrts

tomsilver:
also some refactoring and updates of the score function supercloud script; and copying in @amburger66 's plotting script

### DIFF
--- a/src/approaches/interactive_learning_approach.py
+++ b/src/approaches/interactive_learning_approach.py
@@ -305,13 +305,15 @@ class InteractiveLearningApproach(NSRTLearningApproach):
     def learn_from_interaction_results(
             self, results: Sequence[InteractionResult]) -> None:
         for result in results:
-            for state, response in zip(result.states, result.responses):
+            state_annotations: List[List[Set[GroundAtom]]] = []
+            for response in result.responses:
                 assert isinstance(response, GroundAtomsHoldResponse)
                 state_annotation: List[Set[GroundAtom]] = [set(), set()]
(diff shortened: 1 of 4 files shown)


src/approaches/interactive_learning_approach.py
```diff
@@ -305,13 +305,15 @@ def _query_policy(s: State) -> Optional[GroundAtomsHoldQuery]:
...
                 assert isinstance(response, GroundAtomsHoldResponse)
                 state_annotation: List[Set[GroundAtom]] = [set(), set()]
                 for query_atom, atom_holds in response.holds.items():
                     state_annotation[atom_holds].add(query_atom)
-                traj = LowLevelTrajectory([state], [])
-                self._dataset.append(traj, [state_annotation])
+                state_annotations.append(state_annotation)
+            traj = LowLevelTrajectory(result.states, result.actions)
+            self._dataset.append(traj, state_annotations)
         self._relearn_predicates_and_nsrts(
```
ronuchit, line 317: what's going on with these changes? was this a bug before? it looks like you've changed the indentation level at which the traj is getting created

amburger66: yeah this was kind of a bug, we want the states in each InteractionResult to form a trajectory so this transition data can be used for learning NSRTS, whereas earlier every single state was in its own "trajectory" so there were no transitions for learning nsrts

amburger66:
this lgtm, were there more changes to add?

tomsilver:
I split this code into multiple PRs

PR #653, opened 2022-04-02
toy touch point environment for option learning

tomsilver:
example with random actions:
https://user-images.githubusercontent.com/2816502/161387887-a4d97626-5993-4392-b08b-81c3b6cf7cce.mp4

example with learned nsrts:
https://user-images.githubusercontent.com/2816502/161387868-e5ad0330-f371-467a-921d-900851d056b6.mp4

things that I tested:
* random actions
* random options
* oracle
* nsrt learning

### DIFF
--- a/src/envs/touch_point.py
+++ b/src/envs/touch_point.py
@@ -94,8 +94,8 @@ class TouchPointEnv(BaseEnv):
 
     @property
     def action_space(self) -> Box:
-        # An angle in radians. Wraps around to avoid boundary issues.
-        return Box(-2 * np.pi, 2 * np.pi, (1, ))
+        # An angle in radians.
+        return Box(-np.pi, np.pi, (1, ))
 
     def render_state(self,
                      state: State,
--- a/src/envs/touch_point.py
+++ b/src/envs/touch_point.py
@@ -126,7 +126,7 @@ class TouchPointEnv(BaseEnv):
         return [img]
 
     def _get_tasks(self, num: int, rng: np.random.Generator) -> List[Task]:
-        # There is only one goal in this enviornment.
+        # There is only one goal in this environment.
         goal_atom = GroundAtom(self._Touched, [self._robot, self._target])
         goal = {goal_atom}
         # The initial positions of the robot and dot vary. The only constraint
--- a/src/envs/touch_point.py
+++ b/src/envs/touch_point.py
@@ -1,0 +1,180 @@
+"""Toy environment for testing option learning."""
+
+from typing import ClassVar, Dict, List, Optional, Sequence, Set
+
+import matplotlib.pyplot as plt
+import numpy as np
+from gym.spaces import Box
+
+from predicators.src import utils
+from predicators.src.envs import BaseEnv
+from predicators.src.settings import CFG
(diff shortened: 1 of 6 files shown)


src/envs/touch_point.py
```diff
@@ -0,0 +1,179 @@
...
+    def types(self) -> Set[Type]:
+        return {self._robot_type, self._target_type}
+
+    @property
+    def options(self) -> Set[ParameterizedOption]:
+        return {self._MoveTo}
+
+    @property
+    def action_space(self) -> Box:
+        # An angle in radians. Wraps around to avoid boundary issues.
```
ronuchit, line 97: what part of the code is handling the boundary issues of the angle?

ronuchit: the move policy seems to be just -pi to pi?

src/envs/touch_point.py
```diff
@@ -0,0 +1,179 @@
...
+        self._robot_type = Type("robot", ["x", "y"])
+        self._target_type = Type("target", ["x", "y"])
+        # Predicates
+        self._Touched = Predicate("Touched",
+                                  [self._robot_type, self._target_type],
+                                  self._Touched_holds)
+        # Options
+        self._MoveTo = ParameterizedOption(
+            "MoveTo",
+            types=[self._robot_type, self._target_type],
```
ronuchit, line 46: so plans are just length 1: MoveTo(robot, target), always?

src/settings.py
```diff
@@ -245,6 +245,9 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
                     # lower level, so tasks take more actions to complete.
                     "behavior": 1000,
                     "pybullet_blocks": 1000,
+                    # For the very simple TouchPoint environment, restrict
+                    # the horizon to be shorter.
+                    "touch_point": 15,
```
ronuchit, line 250: is this the horizon at the action level or the option level?

tests/envs/test_touch_point.py
```diff
@@ -0,0 +1,53 @@
...
+            "x": 0.5,
+            "y": 0.9,
+        }
+    })
+    assert len(task.goal) == 1
+    goal_atom = list(task.goal)[0]
+    assert goal_atom.predicate.name == "Touched"
+    assert goal_atom.objects == [robot, target]
+    assert not goal_atom.holds(state)
+    action = Action(np.array([np.pi / 2], dtype=np.float32))  # move up
```
ronuchit, line 47: if actions are -2*pi to 2*pi, are there actually two angles you could use here to move up? should we eliminate that redundancy?

ronuchit (approved):
lgtm, questions

src/envs/touch_point.py
tomsilver: yeah nvm I don't think this is necessary after all

src/envs/touch_point.py
tomsilver: yeah

src/settings.py
tomsilver: action level

PR #660, opened 2022-04-02
allow skipping until a learning cycle

### LINKED ISSUE #654: make it easy to skip ahead to later online interaction episodes when loading data/approach

tomsilver:
intended usage: when loading saved models from an online learning approach, it is often useful to skip to a later cycle and inspect things there, e.g., the data, the evaluation failures.

closes #654

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -126,14 +126,17 @@ def _run_pipeline(env: BaseEnv,
             approach.learn_from_offline_dataset(offline_dataset)
             learning_time = time.time() - learning_start
         # Run evaluation once before online learning starts.
-        results = _run_testing(env, approach)
-        results["num_transitions"] = total_num_transitions
-        results["query_cost"] = total_query_cost
-        results["learning_time"] = learning_time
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -126,14 +126,17 @@ def _run_pipeline(env: BaseEnv,
             approach.learn_from_offline_dataset(offline_dataset)
             learning_time = time.time() - learning_start
         # Run evaluation once before online learning starts.
-        results = _run_testing(env, approach)
-        results["num_transitions"] = total_num_transitions
-        results["query_cost"] = total_query_cost
-        results["learning_time"] = learning_time
-        _save_test_results(results, online_learning_cycle=None)
+        if CFG.skip_until_cycle < 0:
```
ronuchit, line 129: is there a reason we need to indent this? the testing uses its own RNGs right?

ronuchit (approved):

src/main.py
tomsilver: yeah, it's just because I want to skip the time to test when using this flag, because i'm impatient

PR #662, opened 2022-04-03
updates to ground truth nsrts for rnt painting

tomsilver:
the hope is to get them closer to what we would expect to learn with sidelining / operator learning

things I tested on seeds 0 - 4:
* `python src/main.py --env repeated_nextto_painting --approach oracle`
* `python src/main.py --env painting --approach oracle`
* `python src/main.py --env repeated_nextto_painting --approach oracle --painting_goal_receptacles box`
* `python src/main.py --env repeated_nextto_painting --approach oracle --painting_goal_receptacles shelf`

### DIFF
--- a/src/envs/painting.py
+++ b/src/envs/painting.py
@@ -279,8 +279,7 @@ class PaintingEnv(BaseEnv):
         if receptacle == "table" and \
            collider is not None and \
            collider != held_obj:
-            raise utils.EnvironmentFailure("Collision during place on table.",
-                                           {"offending_objects": {collider}})
+            return next_state
         # Execute place
         next_state.set(self._robot, "fingers", 1.0)
         next_state.set(held_obj, "pose_x", x)
--- a/src/envs/repeated_nextto_painting.py
+++ b/src/envs/repeated_nextto_painting.py
@@ -33,8 +33,6 @@ class RepeatedNextToPaintingEnv(PaintingEnv):
                                       self._NextTo_holds)
         self._NextToTable = Predicate("NextToTable", [self._robot_type],
                                       self._NextToTable_holds)
-        self._NextToNothing = Predicate("NextToNothing", [self._robot_type],
-                                        self._NextToNothing_holds)
         # Additional Options
         self._MoveToObj = utils.SingletonParameterizedOption(
             "MoveToObj",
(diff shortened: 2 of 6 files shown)


ronuchit:
Thank you for bringing this to the finish line! I tested things myself and they look good.

I also diffed the old painting operators with the new ones, and here are the changes (note < means new, > means old)
```
3c3
<     Preconditions: [Holding(?obj:obj), IsClean(?obj:obj), IsWet(?obj:obj)]
---
>     Preconditions: [Holding(?obj:obj), IsWet(?obj:obj)]
57c57
< NSRT-PlaceOnTable:
---
> NSRT-PlaceOnTable-HoldingSide:
59c59
<     Preconditions: [Holding(?obj:obj), OnTable(?obj:obj)]
---
>     Preconditions: [Holding(?obj:obj), HoldingSide(?obj:obj), OnTable(?obj:obj)]
61c61,68
<     Delete Effects: [Holding(?obj:obj), HoldingSide(?obj:obj), HoldingTop(?obj:obj)]
---
>     Delete Effects: [Holding(?obj:obj), HoldingSide(?obj:obj)]
>     Side Predicates: []
>     Option Spec: Place(?robot:robot)
> NSRT-PlaceOnTable-HoldingTop:
>     Parameters: [?obj:obj, ?robot:robot]
>     Preconditions: [Holding(?obj:obj), HoldingTop(?obj:obj), OnTable(?obj:obj)]
>     Add Effects: [GripperOpen(?robot:robot)]
>     Delete Effects: [Holding(?obj:obj), HoldingTop(?obj:obj)]
```

Do these all make sense to you @tomsilver?

Will let the environment creators do the actual approval :)

NishanthJKumar (commented):
LGTM! Thanks for these changes! I'll let @wmcclinton do the final approval

wmcclinton (approved):
Looks good! It seems like removing the z change when picking and placing simplified the environment a lot, but I don't fully understand why you need NextToTable. Also since you have a NextToTable, I think NextToNothing is useless because your either next to the table, box, or shelf.

tomsilver:
>Looks good! It seems like removing the z change when picking and placing simplified the environment a lot,
I think you're referring to the `OnTable_holds` definition? I didn't actually change anything there. but thinking about it some more, I do think we should have the same definition for both rnt_painting and painting. I also think that the rnt_painting one, which uses the z pose, is better. I don't want to change too many things in one PR before we run the nightly though. so I'm going to make an issue and then a separate PR for this change.

ronuchit:
>  I don't fully understand why you need NextToTable. 

Because that stops us from having weird plans where the robot places the currently held object on the table while it's not actually at the table. This matters in the episodes where the robot starts off holding something in the initial state, which happens with 50% probability (see `painting_initial_holding_prob` in settings.py) @wmcclinton

PR #664, opened 2022-04-04
a few small changes to rnt_painting and painting

### LINKED ISSUE #663: consolidate the definitions of `OnTable_holds` between rnt_painting and painting

tomsilver:
* consolidates the two definitions of `OnTable_holds`
* removes all occurrences of `CFG.env` in rnt painting, because if someone wanted to create an instance of painting or rnt painting while `CFG.env` is set to something else, they should be able to do that
* one intentional functional change that should have no practical effect: in get_tasks, when initially holding an object, update the y pose to be equal to the robot's y pose in painting (previously this was only done in rnt_painting). I'm doing this just because it doesn't make sense to have this happen in one env but not the other

closes #663

### DIFF
--- a/src/envs/painting.py
+++ b/src/envs/painting.py
@@ -284,7 +284,8 @@ class PaintingEnv(BaseEnv):
         next_state.set(self._robot, "fingers", 1.0)
         next_state.set(held_obj, "pose_x", x)
         next_state.set(held_obj, "pose_y", y)
-        next_state.set(held_obj, "pose_z", z)
+        if self._update_z_poses:
+            next_state.set(held_obj, "pose_z", z)
         next_state.set(held_obj, "grasp", 0.5)
         next_state.set(held_obj, "held", 0.0)
         return next_state
--- a/src/envs/painting.py
+++ b/src/envs/painting.py
@@ -279,8 +279,7 @@ class PaintingEnv(BaseEnv):
         if receptacle == "table" and \
            collider is not None and \
            collider != held_obj:
-            raise utils.EnvironmentFailure("Collision during place on table.",
-                                           {"offending_objects": {collider}})
+            return next_state
         # Execute place
         next_state.set(self._robot, "fingers", 1.0)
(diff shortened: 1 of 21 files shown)


NishanthJKumar (commented):
LGTM! Will let @wmcclinton do the final approval

src/envs/painting.py
```diff
@@ -535,9 +542,9 @@ def _get_tasks(self, num_tasks: int, num_objs_lst: List[int],
                 state.set(self._robot, "fingers", 0.0)
                 state.set(target_obj, "grasp", grasp)
                 state.set(target_obj, "held", 1.0)
-                if CFG.env == "repeated_nextto_painting":
-                    state.set(target_obj, "pose_y",
-                              state.get(self._robot, "pose_y"))
+                state.set(target_obj, "pose_y",
+                          state.get(self._robot, "pose_y"))
+                if self._update_z_poses:
                     state.set(target_obj, "pose_z",
```
ronuchit, line 548: just curious what would break if we did this in regular painting too (not just rnt painting)

ronuchit (approved):
lgtm, but can't update_z_poses always be true?

how do the supercloud results look for this PR?

wmcclinton (approved):
Looks Good! Makes sense, I approve.

src/envs/painting.py
tomsilver: you're right, I think nothing would break. I'll experiment with changing this after the current nightly runs finish (before merging this PR).

ronuchit: i think you can kill the master supercloud (your account) right? it looks fine enough

tomsilver: I attempted to make this change, but it became a rabbit hole that was leading to changes in unit tests and ground truth nsrts, not to mention I don't know how it would affect nsrt learning and predicate invention. so I'm going to abandon this.

ronuchit: kk thx for trying

PR #665, opened 2022-04-04
fix action space bound crash for option learning

tomsilver:
this fixes a crash where the learned option policies would suggest out of bounds actions

before this change, this crashes:
`python src/main.py --env touch_point --approach nsrt_learning --seed 1 --option_learner neural`

now it doesn't crash


ronuchit (approved):
nice idea!

PR #666, opened 2022-04-04
handle case where option raises error during option model

tomsilver:
before this PR, this command would fail when the learned policy returns nans:

`python src/main.py --env cover_multistep_options --approach nsrt_learning --seed 0 --option_learner neural`


ronuchit (approved):
looks reasonable. what were you seeing that led you to make this change? maybe put a command line example in the PR description?

tomsilver:
updated the PR comment

PR #670, opened 2022-04-05
implicit regressor and implicit behavior cloning for option learning

tomsilver:
Results below on seeds 0-4 with the touch point environment. Results on cover are no good yet, likely due to higher dimensional action space.

## Direct BC
```python src/main.py --approach nsrt_learning --env touch_point --option_learner direct_bc```
*Results*: 47, 45, 45, 48, 40

## Implicit BC
```python src/main.py --approach nsrt_learning --env touch_point --option_learner implicit_bc```
*Results*: 49, 50, 50, 50, 50

### Visualization of the learned energy function
good_touch_point_learned_energy

### DIFF
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -148,7 +148,11 @@ class MLPRegressor(Regressor, nn.Module):
 
 
 class ImplicitMLPRegressor(Regressor):
-    """A regressor implemented via an energy function (binary classifier).
+    """A regressor implemented via an "energy function".
+
+    Currently the energy function is treated as a binary classifier, which is
+    not consistent with how energy functions are usually treated/trained.
+    This will change soon.
 
     Negative examples are generated within the class.
 
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -180,7 +180,8 @@ class ImplicitMLPRegressor(Regressor):
         self._y_dim = Y.shape[1]
         max_itr = CFG.implicit_mlp_regressor_max_itr
         self._classifier = MLPClassifier(in_size=(self._x_dim + self._y_dim),
-                                         max_itr=max_itr, balance_data=False)
+                                         max_itr=max_itr,
+                                         balance_data=False)
         # Create the negative data.
         neg_X, neg_Y = self._create_negative_data(X, Y)
         # Set up the data for the classifier.
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -156,16 +156,17 @@ class ImplicitMLPRegressor(Regressor):
     classifying to 1, under the learned classifier. Other inference methods are
     coming soon.
     """
+
     def __init__(self) -> None:
         self._rng = np.random.default_rng(CFG.seed)
         # Set in fit().
         self._x_dim = 0
         self._y_dim = 0
-        self._input_shift = np.zeros(0, dtype=np.float32)
-        self._input_scale = np.zeros(0, dtype=np.float32)
-        self._output_shift = np.zeros(0, dtype=np.float32)
-        self._output_scale = np.zeros(0, dtype=np.float32)
-        self._classifier = MLPClassifier(0, 0)
+        self._input_shift = np.zeros(1, dtype=np.float32)
+        self._input_scale = np.zeros(1, dtype=np.float32)
+        self._output_shift = np.zeros(1, dtype=np.float32)
+        self._output_scale = np.zeros(1, dtype=np.float32)
+        self._classifier = MLPClassifier(1, 1)
 
     def fit(self, X: Array, Y: Array) -> None:
         # Normalize everything right off the bat for simplicity.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/torch_models.py [hunk 1]
```diff
@@ -146,6 +146,105 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
         logging.info(f"Loaded best model with loss: {loss:.5f}")
 
 
+class ImplicitMLPRegressor(Regressor):
+    """A regressor implemented via an energy function (binary classifier).
```
ronuchit, line 150: wait, how is a binary classifier an energy function? An energy function is a mapping from the input to an unnormalized score.

src/torch_models.py
```diff
@@ -146,6 +146,105 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+
+    Negative examples are generated within the class.
+
+    Inference is currently performed by sampling a fixed number of possible
+    inputs and returning the sample that has the highest probability of
+    classifying to 1, under the learned classifier. Other inference methods are
```
ronuchit, line 156: is this just the input that gets the highest energy?

src/torch_models.py
```diff
@@ -146,6 +146,105 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+        self._classifier = MLPClassifier(1, 1)
+
+    def fit(self, X: Array, Y: Array) -> None:
+        # Normalize everything right off the bat for simplicity.
+        X, self._input_shift, self._input_scale = self._normalize_data(X)
+        Y, self._output_shift, self._output_scale = self._normalize_data(Y)
+        # Initialize the classifier.
+        num_data, self._x_dim = X.shape
+        assert Y.shape[0] == num_data
+        logging.info(f"Training {self.__class__.__name__} on {num_data} "
```
ronuchit, line 178: let's fix the print in the other class to use `self.__class__.__name__` too

src/torch_models.py
```diff
@@ -146,6 +146,105 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+    def _create_negative_data(self, X: Array, Y: Array) -> Tuple[Array, Array]:
+        """This makes the assumption that negative data are far, far more
+        common than positive data.
+
+        Under this assumption, the negative data are simply randomly
+        sampled. There may be false negatives in general, but they
+        should be rare, under the assumption.
+        """
+        # Note that the data has already been normalized.
+        del Y  # unused for now, but may be used in the future.
```
ronuchit, line 206: nit: capitalize or remove period

src/torch_models.py
```diff
@@ -146,6 +146,105 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+        return neg_inputs, neg_outputs
+
+    def predict(self, arr: Array) -> Array:
+        # This sampling-based inference method is okay in 1 dimension, but
+        # won't work well with higher dimensions.
+        assert arr.shape == (self._x_dim, )
+        # Normalize.
+        x = (arr - self._input_shift) / self._input_scale
+        num_samples = CFG.implicit_mlp_regressor_num_samples_per_inference
+        sample_ys = self._rng.uniform(size=(num_samples, self._y_dim))
```
ronuchit, line 226: how does it make sense to do this uniformly? what if certain object states are nonsensical or infeasible to reach?

src/torch_models.py
```diff
@@ -146,6 +146,105 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
...
+        sampled. There may be false negatives in general, but they
+        should be rare, under the assumption.
+        """
+        # Note that the data has already been normalized.
+        del Y  # unused for now, but may be used in the future.
+        num_samples = CFG.implicit_mlp_regressor_num_negative_data_per_input
+        neg_input_lst = []
+        neg_output_lst = []
+        for pos_input in X:
+            samples = self._rng.uniform(size=(num_samples, self._y_dim))
```
ronuchit, line 211: shouldn't this be sampling from the action space instead?

src/torch_models.py
tomsilver: agreed this is wrong. discussed offline, going to merge this anyway and then try to do it correctly after (issue #672)

src/torch_models.py
tomsilver: yeah

src/torch_models.py
tomsilver: well, they're negative data, so we would hope to not strive toward the nonsensical or infeasible things. but you're right, it feels like there is something better that can be done here, using the distribution seen in the data. since the implicit BC paper just does what we're doing here, I'm going to leave it for now.

src/torch_models.py
```diff
@@ -401,6 +502,10 @@ def normalize(self, x: Array) -> Array:
         """Apply shift and scale to the input."""
         return (x - self._input_shift) / self._input_scale
 
+    def predict_proba(self, X: Array) -> Array:
```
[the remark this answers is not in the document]
tomsilver: ditto

src/torch_models.py
ronuchit: fix the comment though

src/torch_models.py
```diff
@@ -146,6 +147,107 @@ def _fit(self, inputs: Array, outputs: Array) -> None:
         logging.info(f"Loaded best model with loss: {loss:.5f}")
 
 
+class ImplicitMLPRegressor(Regressor):
+    """A regressor implemented via an energy function (binary classifier).
```
ronuchit, line 151: let's fix this comment to apologize, maybe saying like, currently we use a binary classifier, but this is not ideal

tomsilver:
I verified that the visualization improves with more data. This is with 500 demos:

better_touch_point_learned_energy

ronuchit:
Thank you for testing that!

PR #674, opened 2022-04-06
large refactor of torch models file

tomsilver:
cleaning up the torch models file. there should be NO functional changes. highlights:

* removed a lot of redundant code, especially the training loops
* added a few new base classes
* removed all occurrences of CFG to promote modularity

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -170,7 +170,6 @@ class GlobalSettings:
     side_predicate_learner = "no_learning"
 
     # torch model parameters
-    normalization_scale_clip = 1
     learning_rate = 1e-3
     mlp_regressor_max_itr = 10000
     mlp_regressor_hid_sizes = [32, 32]
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -26,8 +26,8 @@ class Regressor(abc.ABC):
     All regressors normalize the input and output data.
     """
 
-    def __init__(self) -> None:
-        self._seed = CFG.seed
+    def __init__(self, seed: int) -> None:
+        self._seed = seed
         self._rng = np.random.default_rng(self._seed)
         # Set in fit().
         self._x_dim = 0
@@ -42,13 +42,13 @@ class Regressor(abc.ABC):
 
         X and Y are both two-dimensional.
         """
-        X, self._input_shift, self._input_scale = _normalize_data(X)
-        Y, self._output_shift, self._output_scale = _normalize_data(Y)
         num_data, self._x_dim = X.shape
         _, self._y_dim = Y.shape
         assert Y.shape[0] == num_data
         logging.info(f"Training {self.__class__.__name__} on {num_data} "
                      "datapoints")
+        X, self._input_shift, self._input_scale = _normalize_data(X)
+        Y, self._output_shift, self._output_scale = _normalize_data(Y)
         self._fit(X, Y)
 
     def predict(self, x: Array) -> Array:
@@ -80,13 +80,14 @@ class Regressor(abc.ABC):
 class PyTorchRegressor(Regressor, nn.Module):
     """ABC for PyTorch regression models."""
 
-    def __init__(self) -> None:
-        Regressor.__init__(self)
+    def __init__(self, seed: int, max_train_iters: int, clip_gradients: bool,
+                 clip_value: float, learning_rate: float) -> None:
+        Regressor.__init__(self, seed)
         nn.Module.__init__(self)  # type: ignore
-        self._max_train_iters = CFG.mlp_regressor_max_itr
-        self._clip_gradients = CFG.mlp_regressor_clip_gradients
-        self._clip_value = CFG.mlp_regressor_gradient_clip_value
-        self._learning_rate = CFG.learning_rate
+        self._max_train_iters = max_train_iters
+        self._clip_gradients = clip_gradients
+        self._clip_value = clip_value
+        self._learning_rate = learning_rate
 
     @abc.abstractmethod
     def forward(self, tensor_X: Tensor) -> Tensor:
@@ -137,29 +138,49 @@ class PyTorchRegressor(Regressor, nn.Module):
         return y
 
 
-class Classifier(abc.ABC):
+class BinaryClassifier(abc.ABC):
     """ABC for binary classifier classes.
 
     All binary classifiers normalize the input data.
     """
 
-    def __init__(self) -> None:
-        self._rng = np.random.default_rng(CFG.seed)
+    def __init__(self, seed: int, balance_data: bool) -> None:
+        self._seed = seed
+        self._rng = np.random.default_rng(seed)
+        self._balance_data = balance_data
         # Set in fit().
         self._x_dim = 0
         self._input_shift = np.zeros(1, dtype=np.float32)
         self._input_scale = np.zeros(1, dtype=np.float32)
+        self._do_single_class_prediction = False
+        self._predicted_single_class = False
 
     def fit(self, X: Array, y: Array) -> None:
         """Train the classifier on the given data.
 
         X is two-dimensional, y is one-dimensional.
         """
-        X, self._input_shift, self._input_scale = _normalize_data(X)
         num_data, self._x_dim = X.shape
         assert y.shape == (num_data, )
         logging.info(f"Training {self.__class__.__name__} on {num_data} "
                      "datapoints")
+        # If there is only one class in the data, then there's no point in
+        # learning, since any predictions other than that one class could
+        # only be strange generalization issues.
+        if np.all(y == 0):
+            self._do_single_class_prediction = True
+            self._predicted_single_class = False
+            return
+        if np.all(y == 1):
+            self._do_single_class_prediction = True
+            self._predicted_single_class = True
+            return
+        # Balance the classes.
+        if self._balance_data and len(y) // 2 > sum(y):
+            old_len = len(y)
+            X, y = _balance_binary_classification_data(X, y, self._rng)
+            logging.info(f"Reduced dataset size from {old_len} to {len(y)}")
+        X, self._input_shift, self._input_scale = _normalize_data(X)
         self._fit(X, y)
 
     def classify(self, x: Array) -> bool:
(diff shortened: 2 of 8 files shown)


src/settings.py
```diff
@@ -170,7 +170,6 @@ class GlobalSettings:
     side_predicate_learner = "no_learning"
 
     # torch model parameters
-    normalization_scale_clip = 1
```
ronuchit, line 173: isn't removing this a functional change?

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+        Regressor.__init__(self, seed)
+        nn.Module.__init__(self)  # type: ignore
+        self._max_train_iters = max_train_iters
+        self._clip_gradients = clip_gradients
+        self._clip_value = clip_value
+        self._learning_rate = learning_rate
+
+    @abc.abstractmethod
+    def forward(self, tensor_X: Tensor) -> Tensor:
+        """Pytorch forward method."""
```
ronuchit, line 97: PyTorch?

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+    def _predict(self, x: Array) -> Array:
+        tensor_x = torch.from_numpy(np.array(x, dtype=np.float32))
+        tensor_X = tensor_x.unsqueeze(dim=0)
+        tensor_Y = self(tensor_X)
+        tensor_y = tensor_Y.squeeze(dim=0)
+        y = tensor_y.detach().numpy()  # type: ignore
+        return y
+
+
+class BinaryClassifier(abc.ABC):
```
ronuchit, line 143: since there's non-torch stuff now, do we need to rename this file again? x.x (and the corresponding test file)

ronuchit: i never liked the name models.py though, too generic

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
-    @abc.abstractmethod
-    def predict(self, arr: Array) -> Array:
+    def predict(self, x: Array) -> Array:
         """Return a prediction for the given datapoint.
 
-        arr is single-dimensional.
+        x is single-dimensional.
         """
-        raise NotImplementedError("Override me")
+        assert x.shape == (self._x_dim, )
```
ronuchit, line 61: even before this, assert fit was called by checking that `self._x_dim > 0 and self._y_dim > 0`? (similar for the classifier below)

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
 
-    @abc.abstractmethod
     def fit(self, X: Array, Y: Array) -> None:
-        """Train regressor on the given data.
+        """Train the regressor on the given data.
 
         X and Y are both two-dimensional.
         """
-        raise NotImplementedError("Override me")
+        num_data, self._x_dim = X.shape
```
ronuchit, line 47: why would x_dim and y_dim ever be different?

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+
+        X is two-dimensional, y is one-dimensional.
+        """
+        num_data, self._x_dim = X.shape
+        assert y.shape == (num_data, )
+        logging.info(f"Training {self.__class__.__name__} on {num_data} "
+                     "datapoints")
+        # If there is only one class in the data, then there's no point in
+        # learning, since any predictions other than that one class could
+        # only be strange generalization issues.
```
ronuchit, line 171: remove "strange"

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+        if self._do_single_class_prediction:
+            return self._predicted_single_class
+        # Normalize.
+        x = (x - self._input_shift) / self._input_scale
+        # Make prediction.
+        return self._classify(x)
+
+    @abc.abstractmethod
+    def _fit(self, X: Array, y: Array) -> None:
+        """Train the regressor on normalized data."""
```
ronuchit, line 203: classifier*?

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+        raise NotImplementedError("Override me!")
+
+    @abc.abstractmethod
+    def _classify(self, x: Array) -> bool:
+        """Return a predicted class for the normalized input."""
+        raise NotImplementedError("Override me!")
 
-    def forward(self, inputs: Array) -> Tensor:
+
+class PyTorchClassifier(BinaryClassifier, nn.Module):
```
ronuchit, line 212: call this PyTorchBinaryClassifier

src/torch_models.py
```diff
@@ -602,3 +606,85 @@ def classifier(self, state: State, objects: Sequence[Object]) -> bool:
...
+                                    loss_fn: Callable[[Tensor, Tensor],
+                                                      Tensor],
+                                    optimizer: optim.Optimizer,
+                                    tensor_X: Tensor,
+                                    tensor_Y: Tensor,
+                                    max_iters: int,
+                                    print_every: int = 1000,
+                                    clip_gradients: bool = False,
+                                    clip_value: float = 5,
+                                    n_iter_no_change: int = 10000000) -> None:
```
ronuchit, line 651: use `1e7`

src/torch_models.py
```diff
@@ -602,3 +606,85 @@ def classifier(self, state: State, objects: Sequence[Object]) -> bool:
...
+        if loss.item() < best_loss:
+            best_loss = loss.item()
+            best_itr = itr
+            # Save this best model.
+            torch.save(model.state_dict(), model_name)
+        if itr % print_every == 0:
+            logging.info(f"Loss: {loss:.5f}, iter: {itr}/{max_iters}")
+        optimizer.zero_grad()
+        loss.backward()  # type: ignore
+        if clip_gradients:
```
ronuchit, line 674: is this a new feature?

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+        self._initialize_net()
+        # Create the loss function.
+        loss_fn = self._create_loss_fn()
+        # Create the optimizer.
+        optimizer = self._create_optimizer()
+        # Convert data to tensors.
+        tensor_X = torch.from_numpy(np.array(X, dtype=np.float32))
+        tensor_y = torch.from_numpy(np.array(y, dtype=np.float32))
+        # Run training.
+        _train_predictive_pytorch_model(
```
ronuchit, line 254: why not just call this `_train_pytorch_model`?

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+        y = tensor_y.detach().numpy()  # type: ignore
+        proba = y.item()
+        assert 0 <= proba <= 1
+        return proba
+
+    def _classify(self, x: Array) -> bool:
+        return self._forward_single_input_np(x) > 0.5
+
+    def predict_proba(self, denorm_x: Array) -> float:
+        """Get the predicted probability that the input classifies to 1.
```
ronuchit, line 279: move this fn up, so that public methods come before private ones in general

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
+
+        The input is unnormalized.
```
ronuchit, line 281: The input is NOT normalized.

src/torch_models.py
```diff
@@ -1,153 +1,323 @@
...
-
-class ImplicitMLPRegressor(Regressor):
+    def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
+        return nn.MSELoss()
+
+
+class ImplicitMLPRegressor(PyTorchRegressor):
     """A regressor implemented via an "energy function".
 
     Currently the energy function is treated as a binary classifier, which is
```
ronuchit, line 323: treated -> implemented

src/torch_models.py
```diff
@@ -295,305 +502,102 @@ def predict_sample(self, x: Array, rng: np.random.Generator) -> Array:
...
-        self.load_state_dict(torch.load(model_name))  # type: ignore
-        os.remove(model_name)
-        self.eval()  # switch to eval mode
-        yhat = self(X)
-        loss = loss_fn(yhat, y)
-        logging.info(f"Loaded best model with loss: {loss:.5f}")
-
-
-class MLPClassifierEnsemble(Classifier):
+class MLPClassifierEnsemble(BinaryClassifier):
```
ronuchit, line 559: is it on purpose that this inherits from the base class, and not pytorchclassifier?

src/torch_models.py
```diff
@@ -295,305 +502,102 @@ def predict_sample(self, x: Array, rng: np.random.Generator) -> Array:
...
 
     def classify(self, x: Array) -> bool:
+        # Each member maintains its own normalizers.
         avg = np.mean(self.predict_member_probas(x))
         classification = avg > 0.5
         assert classification in [True, False]
         return classification
 
+    def _fit(self, X: Array, y: Array) -> None:
+        raise NotImplementedError("Not used.")
```
ronuchit, line 586: maybe change this to `raise Exception` since it never makes sense to implement it?

src/torch_models.py
```diff
@@ -295,305 +502,102 @@ def predict_sample(self, x: Array, rng: np.random.Generator) -> Array:
...
+
+    def _classify(self, x: Array) -> bool:
+        raise NotImplementedError("Not used.")
```
ronuchit, line 589: maybe change this to raise Exception since it never makes sense to implement it?

ronuchit (approved):
great refactor!

src/settings.py
tomsilver: well in the place where it's used, it's a kwarg that defaults to 1. but you're right that technically this changes some (unused) functionality... gonna leave it though

src/torch_models.py
tomsilver: that's a float though

src/torch_models.py
tomsilver: no, we already had it for regressor training

src/torch_models.py
tomsilver: yeah

src/torch_models.py
ronuchit: nvm, got confused

src/torch_models.py
ronuchit: gotcha

PR #676, opened 2022-04-06
more correct implementation of implicit regressor 

### LINKED ISSUE #672: experiment with InfoNCE loss instead of BCE loss for implicit regression
as mentioned here https://arxiv.org/pdf/2109.00137.pdf

tomsilver:
closes #672 

on the first 5 seeds, `python src/main.py --approach nsrt_learning --env touch_point --option_learner implicit_bc` now gets 50/50 in all cases (previously, one seed would get 49).

visualization with 50 demos (note: unlike before, the colors are now energies, not probabilities):
touch_point_learned_energy

visualization with 500 demos:
touch_point_learned_energy

a few observations:
* the difference between 50 and 500 is no longer as obvious. but the results with 50 look good.
* the NN apparently hasn't figured out that -np.pi == np.pi in this action space, as evidenced by the discontinuity in the color map on the left. I'm not sure what to think about that... maybe it's related to the fact that we're only training on expert data with no noise in it.

### DIFF
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -436,8 +436,8 @@ class ImplicitMLPRegressor(PyTorchRegressor):
                                 self._x_dim + self._y_dim)
             # Create labels for multiclass loss. Note that the true inputs
             # are first, so the target labels are all zeros (see docstring).
-            indices = torch.zeros([num_samples], dtype=torch.int64)
-            labels = F.one_hot(indices, num_classes=num_negatives + 1).float()
+            idxs = torch.zeros([num_samples], dtype=torch.int64)
+            labels = F.one_hot(idxs, num_classes=(num_negatives + 1)).float()
             assert labels.shape == (num_samples, num_negatives + 1)
             # Note that XY is flattened and labels is not. XY is flattened
             # because we need to feed each entry through the network during


src/torch_models.py
```diff
@@ -364,26 +363,43 @@ def _initialize_net(self) -> None:
...
-            # Convert data to tensors.
-            tensor_X = torch.from_numpy(
-                np.array(concat_inputs, dtype=np.float32))
-            yield (tensor_X, tensor_Y)
+            # Resample negative examples on each iteration.
+            neg_Y = torch.rand(size=(num_samples, num_negatives, self._y_dim),
+                               dtype=tensor_Y.dtype)
+            # Create a multiclass classification-style target vector.
+            combined_Y = torch.cat([tensor_Y, neg_Y], axis=1)  # type: ignore
+            combined_Y = combined_Y.reshape([-1, tensor_Y.shape[-1]])
```
ronuchit, line 393: i don't understand what this line is doing. can you go through a concrete example of the shapes?
so tensor_X is (num_samples, state_dim)
and tensor_Y is (num_samples, action_dim)
we are creating num_neg_samples negative examples of actions, so combined_Y should be (num_samples + num_neg_samples, action_dim)
then what's the reshape doing? and why is the `neg_Y = ...` 3-dimensional? i'm clearly misunderstanding something...

src/torch_models.py
```diff
@@ -364,26 +363,43 @@ def _initialize_net(self) -> None:
         self._linears.append(nn.Linear(self._hid_sizes[-1], 1))
 
     def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
-        return nn.BCEWithLogitsLoss()
+
+        def _loss_fn(Y_hat: Tensor, Y: Tensor) -> Tensor:
+            pred = Y_hat.reshape(Y.shape)  # (num data, num classes)
```
ronuchit, line 368: why is Y_hat not already the shape of Y?

ronuchit (approved):
just generally very confused on the shapes of things but i trust that it's correct based on the results

src/torch_models.py
```diff
@@ -364,26 +363,43 @@ def _initialize_net(self) -> None:
...
+            log_probs = F.log_softmax(pred / self._temperature, dim=-1)
+            loss = F.kl_div(log_probs, Y, reduction='batchmean')
```
ronuchit, line 370: why this setting of reduction?

src/torch_models.py
```diff
@@ -364,26 +363,43 @@ def _initialize_net(self) -> None:
...
+            combined_Y = torch.cat([tensor_Y, neg_Y], axis=1)  # type: ignore
+            combined_Y = combined_Y.reshape([-1, tensor_Y.shape[-1]])
+            # Concatenate to create the final input to the network.
+            XY = torch.cat([extended_X, combined_Y], axis=1)  # type: ignore
+            assert XY.shape == ((num_negatives + 1) * num_samples,
+                                self._x_dim + self._y_dim)
+            # Create labels for multiclass loss. Note that the true inputs
+            # are first, so the target labels are all zeros.
+            indices = torch.zeros([num_samples], dtype=torch.int64)
+            labels = F.one_hot(indices, num_classes=num_negatives + 1).float()
```
ronuchit, line 401: waitttt, why aren't there just 2 classes? positive and negative?

src/torch_models.py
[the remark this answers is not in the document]
tomsilver: it's recommended here: https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html#torch.nn.KLDivLoss adding a comment

src/torch_models.py
tomsilver: it's recommended here: https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html#torch.nn.KLDivLoss adding a comment

src/torch_models.py
tomsilver: adding a docstring comment to hopefully clarify

src/torch_models.py
```diff
@@ -364,26 +363,43 @@ def _initialize_net(self) -> None:
...
-        targets = np.array([1] * num_pos_inputs + [0] * num_neg_inputs,
-                           dtype=np.float32)
-        tensor_Y = torch.from_numpy(np.array(targets, dtype=np.float32))
+        num_samples = X.shape[0]
+        num_negatives = self._num_negatives_per_input
+        # Cast to torch first.
+        tensor_X = torch.from_numpy(np.array(X, dtype=np.float32))
+        tensor_Y = torch.from_numpy(np.array(Y, dtype=np.float32))
+        # Expand tensor_Y in preparation for concat in the loop below.
+        tensor_Y = tensor_Y[:, None, :]
```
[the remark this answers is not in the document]
tomsilver: this is pytorch!

tomsilver:
@ronuchit I added some comments and asserts to try to address your questions. can you take another look at let me know which parts I should try to further clarify? (sorry, this is hard... I implemented it this way mainly by looking at references)

src/torch_models.py
```diff
@@ -319,13 +319,32 @@ def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
...
+        L = E[-log(p(y | x, {y'}))]
+
+        p(y | x, {y'})) = exp(-f(x, y)) / [
+            (exp(-f(x, y)) + sum_{y'} exp(-f(x, y')))
+        ]
+
+    where (x, y) is an example "positive" input/output from (X, Y), f is
+    the model that we are learning in this class, and {y'} is a set of
+    "negative" output examples for input x. The size of that set is
+    self._num_negatives_per_input (num_negatives).
```
ronuchit, line 340: what does the `(num_negatives)` mean here -- i thought this sentence is sufficiently clear without it

src/torch_models.py
```diff
@@ -319,13 +319,32 @@ def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
...
+    loss that looks like this:
+
+        L = E[-log(p(y | x, {y'}))]
+
+        p(y | x, {y'})) = exp(-f(x, y)) / [
+            (exp(-f(x, y)) + sum_{y'} exp(-f(x, y')))
+        ]
+
+    where (x, y) is an example "positive" input/output from (X, Y), f is
+    the model that we are learning in this class, and {y'} is a set of
```
ronuchit, line 338: model -> energy function

src/torch_models.py
```diff
@@ -319,13 +319,32 @@ def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
...
+    "negative" output examples for input x. The size of that set is
+    self._num_negatives_per_input (num_negatives).
+
+    One way to interpret the expression is that the numerator exp(-f(x, y))
```
ronuchit, line 342: smart! did you come up with this, or is this normally how people implement it?

src/torch_models.py
```diff
@@ -319,13 +319,32 @@ def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
...
+    represents an unnormalized probability that the same (x, y) belongs to
+    a certain ground truth "class". Each of the exp(-f(x, y') in the
```
ronuchit, line 344: missing a `)`

src/torch_models.py
```diff
@@ -319,13 +319,32 @@ def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
...
+            (exp(-f(x, y)) + sum_{y'} exp(-f(x, y')))
+        ]
+
+    where (x, y) is an example "positive" input/output from (X, Y), f is
+    the model that we are learning in this class, and {y'} is a set of
+    "negative" output examples for input x. The size of that set is
+    self._num_negatives_per_input (num_negatives).
+
+    One way to interpret the expression is that the numerator exp(-f(x, y))
+    represents an unnormalized probability that the same (x, y) belongs to
```
ronuchit, line 343: what do you mean by `the same (x, y)`? maybe just `this (x, y)`?

src/torch_models.py
```diff
@@ -364,26 +384,60 @@ def _initialize_net(self) -> None:
...
     def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
-        return nn.BCEWithLogitsLoss()
+
+        # See the class docstring for context.
+        def _loss_fn(Y_hat: Tensor, Y: Tensor) -> Tensor:
+            pred = Y_hat.reshape(Y.shape)  # (num data, num classes)
+            log_probs = F.log_softmax(pred / self._temperature, dim=-1)
+            # Note: batchmean is recommended in the PyTorch documentation
+            # and will become the default in a future version.
+            loss = F.kl_div(log_probs, Y, reduction='batchmean')
```
ronuchit, line 394: i thought the loss would be a multiclass classification loss like you were saying above. is this equivalent?

src/torch_models.py
```diff
@@ -364,26 +384,60 @@ def _initialize_net(self) -> None:
...
+                                self._x_dim + self._y_dim)
+            # Create labels for multiclass loss. Note that the true inputs
+            # are first, so the target labels are all zeros.
+            indices = torch.zeros([num_samples], dtype=torch.int64)
+            labels = F.one_hot(indices, num_classes=num_negatives + 1).float()
+            assert labels.shape == (num_samples, num_negatives + 1)
+            # Note that XY is flattened and labels is not. XY is flattened
+            # because we need to feed each entry through the network during
+            # training. Labels is unflattened because we will want to use
+            # KLDivLoss.
```
ronuchit, line 439: i don't see KLDivLoss anywhere? do you mean F.kl_div?

ronuchit (approved):
lots of questions, but unblocking

src/torch_models.py
tomsilver: I did not come up with it! I'm heavily relying on a private reference implementation.

src/torch_models.py
tomsilver: yeah, I think they should be equivalent: https://stats.stackexchange.com/questions/357963/what-is-the-difference-cross-entropy-and-kl-divergence

(I'm opting for KL divergence just in case, because of the reference implementation that I'm following.)

to your second question, I'll add more comments in the code

PR #677, opened 2022-04-06
update pylint to ignore newly added cache-max-size-none warning

tomsilver:
pylint updated again, causing lint failures in #676 that are unrelated to that PR


ronuchit (approved):
sighhh

PR #678, opened 2022-04-07
derivate free optimization for implicit regressor 

tomsilver:
`python src/main.py --approach nsrt_learning --env touch_point --option_learner implicit_bc` continues to get 50/50 on the first five seeds with this alternative inference method

### DIFF
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -8,12 +8,13 @@ import logging
 import os
 import tempfile
 from dataclasses import dataclass
-from typing import Callable, Iterator, List, Sequence, Tuple
+from typing import Callable, Iterator, List, Optional, Sequence, Tuple
 
 import numpy as np
 import torch
 import torch.nn.functional as F
 from torch import Tensor, nn, optim
+from torch.distributions.categorical import Categorical
 
 from predicators.src.structs import Array, Object, State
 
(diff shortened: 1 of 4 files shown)


tomsilver:
there appears to be nondeterminism in the predict function for the new inference method (`derivate_free`). I'll investigate later

src/torch_models.py
```diff
@@ -476,6 +504,43 @@ def _predict(self, x: Array) -> Array:
         sample_idx = torch.argmax(scores)
         return sample_ys[sample_idx]
 
+    def _predict_derivative_free(self, x: Array) -> Array:
+        # Reference: https://arxiv.org/pdf/2109.00137.pdf (Algorithm 1).
+        # This method reportedly works well in up to 5 dimensions.
+        # Since we are using torch for random sampling, and since we want
```
ronuchit, line 510: this is shady; why do we need to reseed? we already seed at the start. is it because training could use up an unknown amount of seeding? if so, i'd rather we just call torch.manual_seed() at the end of train_pytorch_model, right around where we call `model.eval()`

src/torch_models.py
```diff
@@ -352,12 +353,31 @@ class ImplicitMLPRegressor(PyTorchRegressor):
...
-                 num_negative_data_per_input: int, temperature: float) -> None:
+                 num_negative_data_per_input: int,
+                 temperature: float,
+                 inference_method: str,
+                 derivate_free_num_iters: Optional[int] = None,
+                 derivate_free_sigma_init: Optional[float] = None,
+                 derivate_free_shrink_scale: Optional[float] = None) -> None:
         super().__init__(seed, max_train_iters, clip_gradients, clip_value,
                          learning_rate)
+        assert inference_method in ("sample_once", "derivate_free")
```
ronuchit, line 372: can remove this assert too, you'll hit the exception below

src/torch_models.py
```diff
@@ -476,6 +504,43 @@ def _predict(self, x: Array) -> Array:
         sample_idx = torch.argmax(scores)
         return sample_ys[sample_idx]
 
+    def _predict_derivative_free(self, x: Array) -> Array:
+        # Reference: https://arxiv.org/pdf/2109.00137.pdf (Algorithm 1).
+        # This method reportedly works well in up to 5 dimensions.
```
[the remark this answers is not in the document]
tomsilver: wouldn't that be nice...

src/torch_models.py
```diff
@@ -476,6 +504,43 @@ def _predict(self, x: Array) -> Array:
...
+        sigma = self._derivate_free_sigma_init
+        K = self._derivate_free_shrink_scale
+        assert num_samples is not None and num_samples > 0
+        assert num_iters is not None and num_iters > 0
+        assert sigma is not None and sigma > 0
+        assert K is not None and 0 < K < 1
+        tensor_x = torch.from_numpy(np.array(x, dtype=np.float32))
+        repeated_x = tensor_x.repeat(num_samples, 1)
+        # Initialize candidate outputs.
+        Y = torch.rand(size=(num_samples, self._y_dim), dtype=tensor_x.dtype)
```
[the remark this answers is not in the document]
tomsilver: right

src/torch_models.py
tomsilver: discussed on slack. it's not just load, but also option model vs execution

src/torch_models.py
tomsilver: i'm confused, did you comment on the same line twice somehow?

src/torch_models.py
```diff
@@ -352,12 +353,26 @@ class ImplicitMLPRegressor(PyTorchRegressor):
     coming soon.
```
ronuchit, line 353: update this paragraph about inference, or remove it

src/torch_models.py
```diff
@@ -482,6 +505,43 @@ def _predict(self, x: Array) -> Array:
         sample_idx = torch.argmax(scores)
         return sample_ys[sample_idx]
 
+    def _predict_derivative_free(self, x: Array) -> Array:
+        # Reference: https://arxiv.org/pdf/2109.00137.pdf (Algorithm 1).
+        # This method reportedly works well in up to 5 dimensions.
+        # Since we are using torch for random sampling, and since we want
+        # to ensure deterministic predictions, we need to reseed torch.
```
ronuchit, line 512: clarify that we can't seed in the init because we want consistency across multiple calls to this predictor for the same input

ronuchit (approved):
some finishing touches

src/torch_models.py
ronuchit: (i think this is what's happening in the option model stuff)

PR #679, opened 2022-04-07
grid method for implicit regressor inference

tomsilver:
`python src/main.py --approach nsrt_learning --env touch_point --option_learner implicit_bc --implicit_mlp_regressor_inference_method grid` 50/50 on the first 5 seeds

### DIFF
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -522,7 +522,7 @@ class ImplicitMLPRegressor(PyTorchRegressor):
         assert self._grid_num_ticks_per_dim > 0
         dy = 1.0 / self._grid_num_ticks_per_dim
         ticks = [np.arange(0.0, 1.0, dy)] * self._y_dim
-        grid = np.meshgrid(*ticks)   # type: ignore
+        grid = np.meshgrid(*ticks)  # type: ignore
         candidate_ys = np.transpose(grid).reshape((-1, self._y_dim))
         num_samples = candidate_ys.shape[0]
         assert num_samples == self._grid_num_ticks_per_dim**self._y_dim
--- a/src/torch_models.py
+++ b/src/torch_models.py
@@ -518,10 +518,12 @@ class ImplicitMLPRegressor(PyTorchRegressor):
         return sample_ys[sample_idx]
 
     def _predict_grid(self, x: Array) -> Array:
+        assert self._grid_num_ticks_per_dim is not None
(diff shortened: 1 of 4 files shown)


src/torch_models.py
```diff
@@ -351,6 +351,11 @@ class ImplicitMLPRegressor(PyTorchRegressor):
     inputs and returning the sample that has the highest probability of
```
ronuchit, line 351: returns*

src/torch_models.py
```diff
@@ -508,6 +517,25 @@ def _predict_sample_once(self, x: Array) -> Array:
         sample_idx = torch.argmax(scores)
         return sample_ys[sample_idx]
 
+    def _predict_grid(self, x: Array) -> Array:
+        assert self._grid_num_ticks_per_dim is not None
+        assert self._grid_num_ticks_per_dim > 0
+        dy = 1.0 / self._grid_num_ticks_per_dim
+        ticks = [np.arange(0.0, 1.0, dy)] * self._y_dim
+        grid = np.meshgrid(*ticks)  # type: ignore
+        candidate_ys = np.transpose(grid).reshape((-1, self._y_dim))
```
ronuchit, line 526: isn't the grid symmetric? what does this transpose do?

src/torch_models.py
```diff
@@ -351,6 +351,11 @@ class ImplicitMLPRegressor(PyTorchRegressor):
     inputs and returning the sample that has the highest probability of
     classifying to 1, under the learned classifier.
 
+    Inference with the "grid" method is similar to "sample_once", except that
```
ronuchit, line 354: move this para down after derivative_free one, for ordering consistency

src/torch_models.py
```diff
@@ -488,6 +495,8 @@ def _predict(self, x: Array) -> Array:
         assert x.shape == (self._x_dim, )
         if self._inference_method == "sample_once":
             return self._predict_sample_once(x)
+        if self._inference_method == "grid":
```
ronuchit, line 498: move this down after derivative_free one, for ordering consistency

ronuchit (approved):

src/torch_models.py
tomsilver: without the transpose, the shape would be (num_samples, y_dim) instead of (y_dim, num_samples)

PR #680, opened 2022-04-07
Change to how datastore is calculated under add effect semantics

NishanthJKumar:
Introduces a change to how we assign transitions to a particular PNAD's datastore under the `add_effects` semantics. This change does not affect results on RNT or RNT_single_option (Supercloud run over 100 seeds presented below), but is necessary for new keep-effect code to work on RNT_painting. Also, this new check just seems more correct.

[results_summary.csv](https://github.com/Learning-and-Intelligent-Systems/predicators/files/8445825/results_summary.csv)

### DIFF
--- a/src/nsrt_learning/side_predicate_learning.py
+++ b/src/nsrt_learning/side_predicate_learning.py
@@ -627,11 +627,7 @@ class BackchainingSidePredicateLearner(GeneralToSpecificSidePredicateLearner):
         self._recompute_datastores_from_segments([new_pnad],
                                                  semantics="add_effects")
         # Determine the preconditions.
-        try:
-            preconditions = induce_pnad_preconditions(new_pnad)
-        except AssertionError:
-            import ipdb
-            ipdb.set_trace()
+        preconditions = induce_pnad_preconditions(new_pnad)
         # Update the preconditions of the new PNAD's operator.
         new_pnad.op = new_pnad.op.copy_with(preconditions=preconditions)
         return new_pnad
--- a/src/nsrt_learning/side_predicate_learning.py
+++ b/src/nsrt_learning/side_predicate_learning.py
@@ -66,7 +66,8 @@ class SidePredicateLearner(abc.ABC):
         that (1) the add effects of the segment intersected with the
         ground op's add effects must not be empty [because otherwise, a
         rational demonstrator would not have called this operator] and
-        (2) all the ground op's add effects must hold in the final atoms.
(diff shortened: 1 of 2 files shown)


src/nsrt_learning/side_predicate_learning.py
```diff
@@ -96,9 +96,27 @@ def _recompute_datastores_from_segments(
...
+                            # op's add effects must hold in the final atoms
+                            # [since these add effects are really just 'set'
+                            # effects].
+
                             if not ground_op.add_effects.issubset(
-                                    segment.add_effects):
+                                    segment.final_atoms):
+                                continue
+
+                            if not len(segment.add_effects
```
ronuchit, line 116: change to `if len(segment.add_effects & ground_op.add_effects) == 0`?

ronuchit (approved):
looks good. did you find the new check (the intersection thing) to be necessary for some example?

src/nsrt_learning/side_predicate_learning.py
NishanthJKumar: sure - either way works; I was just keeping with the `not` trend used in the previous statements lol

ronuchit: you can do `if not (segment.add_effects & ground_op.add_effects)`

src/nsrt_learning/side_predicate_learning.py
```diff
@@ -62,9 +62,11 @@ def _recompute_datastores_from_segments(
...
         satisfied and apply_operator() results in an abstract next state
         that is a subset of the segment's final_atoms. If semantics is
-        "add_effects", then rather than using apply_operator(), we
-        simply check that the add effects are a subset of the segment's
-        add effects.
+        "add_effects", then rather than using apply_operator(), we check
+        that (1) the add effects of the segment intersected with the
+        ground op's add effects must not be empty [because otherwise, a
+        rational demonstrator would not have called this operator] and
+        (2) all the ground op's add effects must hold in the final atoms
```
ronuchit, line 69: nit: add period

NishanthJKumar:
@ronuchit yes - without this change, in `rnt_painting`, our procedure learns two different (overlapping in effects) `MoveToShelf` operators, which causes the keep effect code to induce like 8 total operators with keep effects. With this change, we should only have 1 `MoveToShelf` op before inducing keep effects, which should make debugging the keep effects stuff for this case much easier!

PR #681, opened 2022-04-08
cover multistep options with holding goals

tomsilver:
aka yet another cover environment

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -1255,3 +1255,32 @@ class CoverMultistepOptionsFixedTasks(CoverMultistepOptions):
         del rng
         zero_rng = np.random.default_rng(0)
         return super()._create_initial_state(blocks, targets, zero_rng)
+
+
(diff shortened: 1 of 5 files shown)


src/envs/cover.py
```diff
@@ -1255,3 +1255,32 @@ def _create_initial_state(self, blocks: List[Object],
         del rng
         zero_rng = np.random.default_rng(0)
         return super()._create_initial_state(blocks, targets, zero_rng)
+
+
+class CoverMultistepOptionsHolding(CoverMultistepOptions):
+    """A variation of CoverMultistepOptions where the goals only involve
+    Holding.
+
+    This environment is useful for debugging option learning.
```
ronuchit, line 1264: this is just because we only have to get picking right, not placing?

ronuchit (approved):

src/envs/cover.py
tomsilver: Yeah right

PR #695, opened 2022-04-09
default option model terminate early when stuck

tomsilver:


ronuchit (approved):
lgtm, i remember we did something similar to this before in some other part of the code... can't remember what...

tomsilver:
yeah I remember that too, I couldn't find it. we must have removed it at some point

PR #696, opened 2022-04-09
four rooms environment

### LINKED ISSUE #693: four-room TAMP environment

tomsilver:
closes #693

the idea behind this environment is:
1. it's like the classic four rooms environment for option learning ([ref](https://www.sciencedirect.com/science/article/pii/S0004370299000521))
2. unlike the original, though, the agent is a rectangle with variable width, and depending on that width, it may need to rotate before fitting through a hallway. this is to motivate the idea that we should learn to sample different continuous groundings of the option (where the rotation is the continuous parameter)
3. also unlike the original, there is a "generalize" version of the environment that involves more rooms, see below. this is to motivate learning relational skills that are tied to lifted operators that we can use for planning.

initial tests suggest that everything up to `nsrt_learning` with fixed options works okay. after this PR, what I will be working toward is hopefully a demo that shows (1) our option learning and (2) a baseline that learns non-continuously-parameterized options. the hope is that (2) should fail because of the variable width / rotation stuff discussed above. I'm not 100% confident that this is going to work out yet. it will also be easy to show that a baseline that does not learn operator-tied options will fail to generalize.

### Examples
`python src/main.py --env four_rooms --approach oracle --seed 0 --num_test_tasks 2 --make_test_videos --video_fps 10`
* https://user-images.githubusercontent.com/2816502/162589049-9339c344-066d-4f78-b0fb-c2b1c24dfc1a.mp4
* https://user-images.githubusercontent.com/2816502/162589052-d6ff7034-fff8-49b0-86ac-f8f01c87646d.mp4

`python src/main.py --env four_rooms_generalize --approach oracle --seed 0 --num_test_tasks 2 --make_test_videos --video_fps 10`
* https://user-images.githubusercontent.com/2816502/162589081-2cf25c9b-97db-4c40-b207-2d2c6741c5d2.mp4
* https://user-images.githubusercontent.com/2816502/162589083-8c92a9e7-f2b7-4ad5-ab14-2c2030cd53af.mp4


tomsilver:
TODO: I saw a case where the agent penetrated a wall. need to add test to reproduce and then fix

https://user-images.githubusercontent.com/2816502/162631243-d3e232b3-e640-43f4-a0a7-16335a9f4d30.mp4

tomsilver:
I remain unsure about whether we're going to want this PR, or some variation on it. I'm going to close it for now and reopen it if and when that becomes clear.

PR #700, opened 2022-04-10
add horizon parameter to sesame plan

### LINKED ISSUE #699: planning with default option model can return a plan that exceeds the task horizon

tomsilver:
closes #699

### DIFF
--- a/src/planning.py
+++ b/src/planning.py
@@ -45,6 +45,7 @@ def sesame_plan(
     seed: int,
     task_planning_heuristic: str,
     max_skeletons_optimized: int,
+    max_horizon: int,
     check_dr_reachable: bool = True,
     allow_noops: bool = False,
 ) -> Tuple[List[_Option], Metrics]:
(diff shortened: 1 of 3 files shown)


src/planning.py
```diff
@@ -298,34 +303,39 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
-                    state, option)
+                next_state, num_actions = \
+                    option_model.get_next_state_and_num_actions(state, option)
             except EnvironmentFailure as e:
                 can_continue_on = False
                 # Remember only the most recent failure.
                 discovered_failures[cur_idx - 1] = _DiscoveredFailure(e, nsrt)
             else:  # an EnvironmentFailure was not raised
                 discovered_failures[cur_idx - 1] = None
+                num_actions_per_option[cur_idx - 1] = num_actions
```
ronuchit, line 314: looks correct to me, but have you verified by printing out this array in an env with nontrivial options, like pybullet_blocks or cover_multistep?

ronuchit (approved):
thanks, we had something like this before but it got lost in all the refactoring. definitely a bug... our main approach wasn't using the horizon at all!

Issue #685, opened 2022-04-08
try different classifiers


amburger66:
- make scikit-learn interface within `torch_models.py`
- instantiate classifier type based on a setting

PR #701, opened 2022-04-10
more refactoring and updates to machine learning models

tomsilver:
* renames `torch_models` to `ml_models`
* add another layer of abstraction in the base classes for normalization
* add sklearn models
* update requirements.txt and mypy config for this new dependency
* should help with #685, but to close that issue, we will need to make changes in interactive learning approach too

### DIFF
--- a/src/ml_models.py
+++ b/src/ml_models.py
@@ -213,8 +213,8 @@ class _ScikitLearnBinaryClassifier(BinaryClassifier):
     def _initialize_model(self, **kwargs: Any) -> BaseEstimator:
         raise NotImplementedError("Override me!")
 
-    def fit(self, X: Array, Y: Array) -> None:
-        return self._model.fit(X, Y)
+    def fit(self, X: Array, y: Array) -> None:
+        return self._model.fit(X, y)
 
     def classify(self, x: Array) -> bool:
         class_prediction = self._model.predict([x])[0]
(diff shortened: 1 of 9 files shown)


src/ml_models.py
```diff
@@ -726,17 +814,21 @@ def classify(self, x: Array) -> bool:
         assert classification in [True, False]
         return classification
 
-    def _fit(self, X: Array, y: Array) -> None:
```
ronuchit, line 729: why is it okay to remove these?

src/ml_models.py
```diff
@@ -8,11 +8,14 @@
...
 import tempfile
 from dataclasses import dataclass
-from typing import Callable, Iterator, List, Optional, Sequence, Tuple
+from typing import Any, Callable, Iterator, List, Optional, Sequence, Tuple
 
 import numpy as np
 import torch
 import torch.nn.functional as F
+from sklearn.base import BaseEstimator
+from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier
```
ronuchit, line 17: better naming for these might be `SKLearnKNeighborsClassifier` and `SKLearnKNeighborsRegressor`

src/ml_models.py
tomsilver: because not all classes need to normalize, so they might only want to implement `fit` and `predict`

amburger66 (approved):

PR #702, opened 2022-04-10
fix that snaps objects onto the table surface in rnt-painting

NishanthJKumar:
necessary to get learned samplers to work in this environment!

### DIFF
--- a/src/envs/painting.py
+++ b/src/envs/painting.py
@@ -53,6 +53,7 @@ class PaintingEnv(BaseEnv):
     side_grasp_thresh: ClassVar[float] = 0.5 - 1e-2
     robot_x: ClassVar[float] = table_x - 0.5
     nextto_thresh: ClassVar[float] = 1.0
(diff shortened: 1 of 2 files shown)


src/envs/painting.py
```diff
@@ -285,7 +285,17 @@ def _transition_place(self, state: State, action: Action) -> State:
         next_state.set(held_obj, "pose_x", x)
         next_state.set(held_obj, "pose_y", y)
         if self._update_z_poses:
-            next_state.set(held_obj, "pose_z", z)
+            if receptacle == "table" and np.allclose(
+                    z, self.table_height + self.obj_height / 2, rtol=5e-02):
```
tomsilver, line 289: can we make this constant a class variable?

tomsilver (commented):
thanks for finding and fixing this! could you also add a unit test that fails before this change and succeeds after it?

tomsilver (approved):
lgtm after the the unit tests pass!

ronuchit (approved):
great catch!

PR #708, opened 2022-04-11
move some methods to utils.py, to prepare STRIPS learning refactor

ronuchit:


src/nsrt_learning/option_learning.py
```diff
@@ -7,13 +7,14 @@
 from typing import Dict, List, Sequence, Set, Tuple
 
 import numpy as np
+from gym.spaces import Box
```
NishanthJKumar, line 10: was this intentional? Why is this the only change to this file?

tests/test_predicate_search_score_functions.py
```diff
@@ -4,6 +4,7 @@
 
 import numpy as np
 import pytest
+from gym.spaces import Box
```
NishanthJKumar, line 7: again, don't we want to import Box from `structs.py`?

NishanthJKumar (approved):
LGTM - aside from the changes that import `Box` from `gym` instead of `structs`. Feel free to merge if that was intentional, but do fix if not!

ronuchit:
it's intentional; Box is not something we define in structs.py, it's from openai gym. we had incorrectly imported it from structs in some places

PR #711, opened 2022-04-12
Datastore len check fix

NishanthJKumar:
Adds the datastore length check back to `_try_specializing_pnad` method for backchaining. Also adds a test that distinguishes this change.

### DIFF
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -134,7 +134,7 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                         new_pnad = self._try_specializing_pnad(
                             necessary_add_effects,
                             param_opt_to_general_pnad[option.parent], segment,
-                            False)
+                            check_datastore_change=False)
                         assert new_pnad is not None
 
                     pnad = new_pnad
@@ -246,9 +246,10 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
 
         Returns the new constructed PNAD, without modifying the
         original. If the PNAD does not have a grounding that can even
-        partially satisfy the necessary add effects, or if the newly
-        created PNAD covers a different set of datapoints than the
-        original, returns None.
+        partially satisfy the necessary add effects, then returns None. 
+        If check_datastore_change is set to True, then additionally
+        checks whether the newly created PNAD covers a different set
+        of datapoints than the original, and returns None in this case.
         """
 
         # Get an arbitrary grounding of the PNAD's operator whose
@@ -291,7 +292,7 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
             # If the new PNAD has a datastore size that's not the same
             # as that of the original PNAD, then we've potentially lost some
             # data by specializing, which might do harm!
-            if len(new_pnad.datastore) != len(pnad.datastore):
+            if len(new_pnad.datastore) < len(pnad.datastore):
                 return None
 
         return new_pnad
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -1,6 +1,7 @@
 """Algorithms for STRIPS learning that start from the most general operators,
 then specialize them based on the data."""
 
+from builtins import AssertionError
 from typing import Dict, List, Optional, Set
 
 from predicators.src import utils
@@ -102,6 +103,8 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                 necessary_add_effects = necessary_image - atoms_seq[t]
                 assert necessary_add_effects.issubset(segment.add_effects)
 
+                # import ipdb; ipdb.set_trace()
+
                 # We start by checking if any of the PNADs associated with the
                 # demonstrated option are able to match this transition.
                 for pnad in pnads_for_option:
@@ -133,7 +136,8 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                     else:
                         new_pnad = self._try_specializing_pnad(
                             necessary_add_effects,
-                            param_opt_to_general_pnad[option.parent], segment)
+                            param_opt_to_general_pnad[option.parent], segment,
+                            False)
                         assert new_pnad is not None
 
                     pnad = new_pnad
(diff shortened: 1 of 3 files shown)


src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -283,6 +287,13 @@ def _try_specializing_pnad(
         new_pnad = self._create_new_pnad_with_params_and_add_effects(
             pnad, updated_params, updated_add_effects)
 
+        if check_datastore_change:
+            # If the new PNAD has a datastore size that's not the same
+            # as that of the original PNAD, then we've potentially lost some
+            # data by specializing, which might do harm!
+            if len(new_pnad.datastore) != len(pnad.datastore):
```
tomsilver, line 294: for my own understanding, could we assert here that that the lengths are related via < or > if they are not equal?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -133,7 +133,8 @@ def _learn(self) -> List[PartialNSRTAndDatastore]:
                     else:
                         new_pnad = self._try_specializing_pnad(
                             necessary_add_effects,
-                            param_opt_to_general_pnad[option.parent], segment)
+                            param_opt_to_general_pnad[option.parent], segment,
+                            False)
```
ronuchit, line 137: use keyword argument here: `check_datastore_change=False`

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -237,14 +238,17 @@ def _try_specializing_pnad(
...
+        check_datastore_change: bool = True
     ) -> Optional[PartialNSRTAndDatastore]:
         """Given a PNAD and some necessary add effects that the PNAD must
         achieve, try to make the PNAD's add effects more specific
         ("specialize") so that they cover these necessary add effects.
 
         Returns the new constructed PNAD, without modifying the
         original. If the PNAD does not have a grounding that can even
-        partially satisfy the necessary add effects, returns None.
+        partially satisfy the necessary add effects, or if the newly
```
ronuchit, line 249: Fix this docstring -- this second condition only matters if `check_datastore_change` is True.

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
ronuchit: +1, i think we can `assert len(new_pnad.datastore) < len(pnad.datastore)` in here, since data can never be added to the datastore?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -283,6 +287,13 @@ def _try_specializing_pnad(
         new_pnad = self._create_new_pnad_with_params_and_add_effects(
             pnad, updated_params, updated_add_effects)
 
+        if check_datastore_change:
+            # If the new PNAD has a datastore size that's not the same
+            # as that of the original PNAD, then we've potentially lost some
+            # data by specializing, which might do harm!
+            if len(new_pnad.datastore) != len(pnad.datastore):
+                return None
+
```
ronuchit, line 296: have you verified this PR across several seeds of rnt and rnt single option?

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -27,7 +31,7 @@ def find_unification(necessary_add_effects,
 
 
 def test_backchaining_strips_learner():
-    """Test the BackchainingSidePredicateLearner."""
+    """Test the BackchainingSTRIPSLearner."""
```
ronuchit, line 34: good catch. the word "BackchainingSidePredicateLearner" appears in one other place in this file, can you fix that too?

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -111,6 +117,134 @@ def test_backchaining_strips_learner():
     assert str(pnads[0]) == repr(pnads[0]) == expected_str
 
 
+def test_backchaining_strips_learner_order_dependence():
+    """Test that the BackchainingSTRIPSLearner is invariant to order of
+    traversal through trajectories."""
+    # Setup up the types and predicates.
```
ronuchit, line 123: Set up*

(also, this appears in behavior_options.py, so if you want, you can update "Setup" -> "Set up" there too)

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -111,6 +117,134 @@ def test_backchaining_strips_learner():
...
+    learner = MockBackchainingSTRIPSLearner(
+        [traj3, traj2, traj1], [task1, task2, task3],
+        {RobotAt, LightOn, NotLightOn, LightColorBlue, LightColorRed},
+        [[segment3], [segment2], [segment1]])
+    reverse_order_pnads = learner.learn()
+
+    # First, check that the two sets of PNADs have the same number of PNADs.
+    assert len(natural_order_pnads) == len(reverse_order_pnads) == 2
+
+    correct_pnads = [
```
ronuchit, line 219: can this be a set, or does order matter?

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -86,17 +90,19 @@ def test_backchaining_strips_learner():
 
     # Test sidelining where an existing operator needs to
     # be specialized.
-    goal3 = {Asleep([bob]), Sad([bob])}
```
ronuchit, line 89: just curious what's the motivation behind the changes to this existing test?

ronuchit (commented):
i love seeing 100 lines of tests for a 10 line fix! :D just some minor comments

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
NishanthJKumar: that test turned out to be wrong in that it was relying on the bug (i.e, not checking datastore length change) to pass. I modified the input so that it would produce the same output given our fix.

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
NishanthJKumar: good call - just verified results on both envs across 5 seeds! Things work as expected.

ronuchit (approved):

PR #714, opened 2022-04-13
implement _PyBulletState subclass of State

tomsilver:
so that we can use joint states within the forthcoming new controllers (which will perform planning in joint space), without changing the object-centric state exposed to the rest of the system

### DIFF
--- a/src/envs/pybullet_blocks.py
+++ b/src/envs/pybullet_blocks.py
@@ -17,7 +17,7 @@ from predicators.src.structs import Action, Array, Image, Object, \
     ParameterizedOption, Pose3D, State, Task, Type
 
 
-class _PybulletState(State):
+class _PyBulletState(State):
     """A PyBullet state that stores the robot joint states in addition to the
     features that are exposed in the object-centric state."""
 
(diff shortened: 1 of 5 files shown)


src/envs/pybullet_blocks.py
```diff
@@ -16,6 +17,25 @@
     ParameterizedOption, Pose3D, State, Task, Type
 
 
+class _PyBulletState(State):
+    """A PyBullet state that stores the robot joint states in addition to the
```
ronuchit, line 21: should this be in pybullet_utils.py?

src/envs/pybullet_blocks.py
```diff
@@ -452,7 +477,8 @@ def _get_state(self) -> State:
         state_dict = {}
 
         # Get robot state.
-        state_dict[self._robot] = self._pybullet_robot.get_state()
+        state_dict[self._robot] = self._pybullet_robot.get_object_state()
```
ronuchit, line 480: can we have a better name for this method? "object" seems to mean "things other than robots"

ronuchit: get_state() and get_joints()?

tests/envs/test_pybullet_blocks.py
```diff
@@ -40,10 +41,22 @@ def PutOnTable(self):
         return self._PutOnTable
 
     def set_state(self, state):
```
ronuchit, line 43: is this method not implemented in the actual env? why do we not need it there?

ronuchit (approved):

tests/envs/test_pybullet_blocks.py
tomsilver: it's not, we don't necessarily wanna assume that we can set the state in the actual env because it's a proxy for the real world

PR #715, opened 2022-04-14
Pybullet robot tweaks

tomsilver:
* move the action space into the robot
* expose the joint finger indices, since they're special for manipulation


ronuchit (approved):

ronuchit:
not sure i understand why we need this temporary implementation, but ah well

tomsilver:
i'm just trying to do this in tiny chunks because i'm scared

PR #716, opened 2022-04-14
(wip) trying to change finger action space in pybullet blocks

tomsilver:
* change the finger actions to be absolute joint positions instead of deltas (this is one step closer toward having the action space be absolute joint value)
* remove the finger nudging crap that happens inside the robot set_motors, because with the new action space, that won’t make sense anymore. also it’s just hacky

this is failing very subtly one of the longrun tests (but only if you run them all, not if you run just that test :( ) and also it’s leading to a decrease in oracle performance from ~49 to ~46


tomsilver:
superseded by #719

PR #717, opened 2022-04-14
RL framework

ashayathalye:
Outline of framework to support reinforcement learning for option learning. A new approach, ReinforcementLearningApproach, collects trajectories using the most current policy. In an updated PR, we will send these trajectories to an "RLOptionLearner". 

Want feedback on how to do the pruning: should I add that into the current `learn_nsrts_from_data()`, to minimize duplication of code in `learn_pruned_nsrts_from_data()`?


src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task]) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self._nsrts: Set[NSRT] = set()
+        self.online_learning_cycle = 0
+        self.initial_dataset: List[LowLevelTrajectory] = []
+        self.online_dataset: List[LowLevelTrajectory] = []
+        # initialize option learner?
```
tomsilver, line 34: add a `TODO` here so we make sure not to merge before resolving. (the linter will catch the TODO if it's a comment)

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,109 @@
...
+class ReinforcementLearningApproach(NSRTLearningApproach):
+    """A bilevel planning approach that learns NSRTs."""
+
+    def __init__(self, initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task]) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self._nsrts: Set[NSRT] = set()
+        self.online_learning_cycle = 0
```
tomsilver, line 31: can we make these three private?

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,109 @@
...
+        save_path = utils.get_approach_save_path_str()
+        with open(f"{save_path}_{online_learning_cycle}.NSRTs", "wb") as f:
+            pkl.dump(self._nsrts, f)
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "reinforcement_learning"
+
+    def learn_from_offline_dataset(self, dataset: Dataset) -> None:
+        # The only thing we need to do here is learn NSRTs,
```
tomsilver, line 56: remove this copied comment, and rewrite to something like:
```
self.initial_dataset = dataset.trajectories
super().learn_from_offline_dataset(dataset)
```
with the key thing being calling the parent class method instead of copying it.

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,109 @@
...
+    """A bilevel planning approach that learns NSRTs."""
+
+    def __init__(self, initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task]) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self._nsrts: Set[NSRT] = set()
+        self.online_learning_cycle = 0
+        self.initial_dataset: List[LowLevelTrajectory] = []
```
tomsilver, line 32: if these are trajectories, and not `Dataset` instances, let's rename them to something like `initial_trajectories`, etc.

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,109 @@
...
+    def _make_termination_fn(cls, goal: Set[GroundAtom]) \
+            -> Callable[[State], bool]:
+
+        def _termination_fn(s: State) -> bool:
+            return all(goal_atom.holds(s) for goal_atom in goal)
+
+        return _termination_fn
+
+    def get_interaction_requests(self) -> List[InteractionRequest]:
+        requests = []
```
tomsilver, line 72: add some comments explaining what's happening here

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,109 @@
...
+                partial_refinements = e.info.get("partial_refinements")
+                assert partial_refinements is not None
+                _, plan = max(partial_refinements, key=lambda x: len(x[1]))
+                _act_policy = utils.option_plan_to_policy(plan)
+            request = InteractionRequest(
+                train_task_idx=i,
+                act_policy=_act_policy,
+                query_policy=lambda s: None,
+                termination_function=ReinforcementLearningApproach.
+                _make_termination_fn(task.goal)  # pylint: disable=line-too-long
```
tomsilver, line 87: you can do `self._make_termination_fn` and then get rid of the pylint disable

tomsilver (commented):
good stuff! some initial comments, I'll take another look through after they're incorporated. 

regarding the larger question about the pruning part: for the sake of this PR, can we remove all of the pruning stuff, and just accept that we will learn more operators than the ones induced by the demonstrations? short term, this will lead to much simpler code and make merging this PR easier. longer term, it will be good to have a longer discussion about what we actually want. for example, I was discussing this with @ronuchit and he felt strongly that we should not prune. so either way, worth a discussion.

src/approaches/reinforcement_learning_approach.py
ashayathalye: checker complains that 'self' is unused which is why I did this. But maybe I can add some ignore statement so we can do that.

tomsilver: to clarify, the checker complains if you leave the method definition alone, and just change this line?

src/approaches/reinforcement_learning_approach.py
ashayathalye: Via double underscore / name mangling? The other approaches I see only use single underscore.

tomsilver: single underscore!

src/main.py
```diff
@@ -219,7 +220,8 @@ def _generate_interaction_results(
             request.train_task_idx,
             request.termination_function,
             max_num_steps=CFG.max_num_steps_interaction_request,
-            monitor=monitor)
+            monitor=monitor,
+            exceptions_to_break_on={OptionExecutionFailure})
```
tomsilver, line 224: why is this needed?

tomsilver (commented):
looking good! @ronuchit could you also take a look?

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,92 @@
...
+    InteractionRequest, InteractionResult, LowLevelTrajectory, \
+    ParameterizedOption, Predicate, State, Task, Type
+
+
+class ReinforcementLearningApproach(NSRTLearningApproach):
+    """A bilevel planning approach that learns NSRTs from an offline dataset,
+    and continues learning options through reinforcement learning."""
+
+    def __init__(self, initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption], types: Set[Type],
```
ronuchit, line 24: so if we're learning options in this approach, then what are these initial options going to be?

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,92 @@
...
+        # policy.
+        requests = []
+        for i in range(len(self._train_tasks)):
+            task = self._train_tasks[i]
+            try:
+                _act_policy = self.solve(task, CFG.timeout)
+            except (ApproachTimeout, ApproachFailure) as e:
+                partial_refinements = e.info.get("partial_refinements")
+                assert partial_refinements is not None
+                _, plan = max(partial_refinements, key=lambda x: len(x[1]))
```
ronuchit, line 65: very cool usage of the partial_refinements info!

src/main.py
ronuchit: +1

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,92 @@
...
+            states = result.states
+            actions = result.actions
+            traj = LowLevelTrajectory(states,
+                                      actions,
+                                      _is_demo=False,
+                                      _train_task_idx=i)
+            self._online_trajectories.append(traj)
+
+        # Replace this with an _RLOptionLearner.
+        self._learn_nsrts(self._initial_trajectories + \
```
ronuchit, line 89: oh i see, so we're not actually learning options right now, this is like a placeholder?

src/main.py
ashayathalye: Due to the following case in `get_interaction_requests()` in reinforcement_learning_approach.py:
```except (ApproachTimeout, ApproachFailure) as e:
                partial_refinements = e.info.get("partial_refinements")
                assert partial_refinements is not None
                _, plan = max(partial_refinements, key=lambda x: len(x[1]))
                _act_policy = utils.option_plan_to_policy(plan)
```
When the act policy is created from a list of options that is "incomplete"/ too short and does not reach the goal, there is a `OptionExecutionFailure("Option plan exhausted!")` that is thrown. 

Maybe we should fill this argument only when this particular approach is called?

tomsilver: yes thanks! I believe we actually already merged this change for a similar reason in a different recent PR, so when you rebase off master, you should be all set here.

ashayathalye: Do to the following case in `get_interaction_requests()` in reinforcement_learning_approach.py: 
```
    partial_refinements = e.info.get("partial_refinements")
                assert partial_refinements is not None
                _, plan = max(partial_refinements, key=lambda x: len(x[1]))
                _act_policy = utils.option_plan_to_policy(plan)
```
Because the list of options that makes the policy is incomplete / too short, `OptionExecutionFailure("Option plan exhausted!")` is thrown in `option_plan_to_policy()`.

Is there a better way to populate this argument only when this particular approach is called?

src/approaches/reinforcement_learning_approach.py
ashayathalye: These aren't used by this approach or NSRTLearningApproach, etc. The initial options seem to be `env.options` which are passed in during `create_approach()` called in `main.py`. 

This is included in the arguments to init here just because super().__init()___() is called repeatedly, and BaseApproach uses it to set `self._initial_options`. 

But maybe it's not great for an approach that is learning options (NSRTLearningApproach included) to get access to the env's options?

src/approaches/reinforcement_learning_approach.py
ashayathalye: So, we got more data after interacting with the environment during our online learning cycle. Now we get to choose how we want to update our options based on the new data we got. We want `reinforcement_learning_approach.py` to implement various RL based ways to update these options. 

So yeah– this line is currently a placeholder, where we update the options in a different manner, that is, doing behavioral cloning all over again, on the total dataset we have so far.

src/approaches/reinforcement_learning_approach.py
ronuchit: note: this was fixed recently (#800)

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,91 @@
...
+    ApproachTimeout
+from predicators.src.approaches.nsrt_learning_approach import \
+    NSRTLearningApproach
+from predicators.src.settings import CFG
+from predicators.src.structs import NSRT, Dataset, GroundAtom, \
+    InteractionRequest, InteractionResult, LowLevelTrajectory, \
+    ParameterizedOption, Predicate, State, Task, Type
+
+
+class ReinforcementLearningApproach(NSRTLearningApproach):
```
ronuchit, line 19: let's rename this to NSRTReinforcementLearningApproach since it's NSRT-specific (and that better reflects the name we've given it, which is `nsrt_rl`)

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,91 @@
...
+    @classmethod
+    def get_name(cls) -> str:
+        return "nsrt_rl"
+
+    def learn_from_offline_dataset(self, dataset: Dataset) -> None:
+        self._initial_trajectories = dataset.trajectories
+        super().learn_from_offline_dataset(dataset)
+
+    @classmethod
+    def _make_termination_fn(cls, goal: Set[GroundAtom]) \
```
ronuchit, line 42: we can delete this function (see next comment)

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,91 @@
...
+            except (ApproachTimeout, ApproachFailure) as e:
+                partial_refinements = e.info.get("partial_refinements")
+                assert partial_refinements is not None
+                _, plan = max(partial_refinements, key=lambda x: len(x[1]))
+                _act_policy = utils.option_plan_to_policy(plan)
+            request = InteractionRequest(
+                train_task_idx=i,
+                act_policy=_act_policy,
+                query_policy=lambda s: None,
+                termination_function=self._make_termination_fn(task.goal))
```
ronuchit, line 70: just do `termination_function=task.goal_holds`

src/approaches/reinforcement_learning_approach.py
```diff
@@ -0,0 +1,91 @@
...
+            -> Callable[[State], bool]:
+
+        def _termination_fn(s: State) -> bool:
+            return all(goal_atom.holds(s) for goal_atom in goal)
+
+        return _termination_fn
+
+    def get_interaction_requests(self) -> List[InteractionRequest]:
+        # For each training task, try to solve the task to get a policy. If the
+        # task can't be solved, construct a policy from the sequence of _Options
```
ronuchit, line 52: for added clarity: _Options -> _Option objects

tests/approaches/test_reinforcement_approach.py
```diff
@@ -0,0 +1,51 @@
...
+from predicators.src.approaches.reinforcement_learning_approach import \
+    ReinforcementLearningApproach
+from predicators.src.datasets import create_dataset
+from predicators.src.envs.cover import CoverMultistepOptions
+from predicators.src.main import _generate_interaction_results
+from predicators.src.settings import CFG
+from predicators.src.teacher import Teacher
+
+
+def test_reinforcement_learning_approach():
```
ronuchit, line 16: rename this file, methods in here, comments etc

tests/approaches/test_reinforcement_approach.py
```diff
@@ -0,0 +1,51 @@
...
+    interaction_results, _ = _generate_interaction_results(
+        env, teacher, interaction_requests)
+    approach.learn_from_interaction_results(interaction_results)
+    approach.load(online_learning_cycle=1)
+    with pytest.raises(FileNotFoundError):
+        approach.load(online_learning_cycle=2)
+    for task in env.get_test_tasks():
+        try:
+            approach.solve(task, timeout=CFG.timeout)
+        except (ApproachTimeout, ApproachFailure):  # pragma: no cover
```
ronuchit, line 48: do we really need this except? what happens if we remove it?

src/approaches/reinforcement_learning_approach.py
ronuchit: (rename the file too)

tests/approaches/test_reinforcement_approach.py
ashayathalye: Given that try needs an except block following it, the other two choices are:
1. just `except:`, which generates the pylint error `No exception type(s) specified (bare-except)`
2. just `except Exception:`, which generates the pylint error `Catching too general exception Exception (broad-except)`

How it is done currently without changing it is how it is done in `test_interactive_approach.py`.

tomsilver: I think @ronuchit was asking if the code would still work if we just did this instead:
```
for task in env.get_test_tasks():
    approach.solve(task, timeout=CFG.timeout)
```

tomsilver: and if not, then we should be able to remove the # pragma: no cover

PR #718, opened 2022-04-14
factor out common pybullet code into a new base class

ronuchit:


src/envs/pybullet_robots.py
```diff
@@ -307,3 +306,105 @@ def create_single_arm_pybullet_robot(
                                   finger_action_tol, max_vel_norm,
                                   physics_client_id)
     raise NotImplementedError(f"Unrecognized robot name: {robot_name}.")
+
```
ronuchit, line 309: all this is just copied over from pybullet_utils (now deleted), but PyBulletState was moved into pybullet_env.py

tests/envs/test_pybullet_robots.py
```diff
@@ -4,11 +4,153 @@
...
 
+from predicators.src import utils
 from predicators.src.envs.pybullet_robots import FetchPyBulletRobot, \
-    create_single_arm_pybullet_robot
+    create_single_arm_pybullet_robot, get_kinematic_chain, \
+    inverse_kinematics
 from predicators.src.settings import CFG
 
 
+@pytest.fixture(scope="module", name="scene_attributes")
```
ronuchit, line 14: copy pasted from test_pybullet_utils.py (now removed)

src/envs/pybullet_env.py
```diff
@@ -0,0 +1,423 @@
...
+                    # by taking the closest one. This should be rare, but it
+                    # can happen when two objects are stacked and the robot is
+                    # unstacking the top one.
+                    contact_distance = point[8]
+                    if contact_distance < closest_held_obj_dist:
+                        closest_held_obj = obj_id
+                        closest_held_obj_dist = contact_distance
+        return closest_held_obj
+
+    def _create_block(self, block_num: int, size: float) -> int:
```
tomsilver, line 300: i'm trying to reason about the purview of this base class. on one hand, we're changing `block` to `obj` in some variable names above. on the other hand, we are including this method in this class, so the class is specific to blocks, in some sense... what's our guiding philosophy?

tomsilver (approved):
thanks a lot for doing this!

src/envs/pybullet_blocks.py
```diff
@@ -219,69 +159,55 @@ def _initialize_pybullet(self) -> None:
...
+    def _create_pybullet_robot(self) -> _SingleArmPyBulletRobot:
+        ee_home = (self.robot_init_x, self.robot_init_y, self.robot_init_z)
+        return create_single_arm_pybullet_robot(CFG.pybullet_robot, ee_home,
+                                                self.open_fingers,
+                                                self.closed_fingers,
+                                                self._finger_action_tol,
+                                                self._max_vel_norm,
+                                                self._physics_client_id)
+
+    def _extract_robot_state(self, state: State) -> Array:
```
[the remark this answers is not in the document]
ronuchit: resolved on slack, bc there could be other features of the robot in general, or the order could change someday

src/envs/pybullet_env.py
ronuchit: pulled this method out as a top-level method in pybullet_envs.py

PR #719, opened 2022-04-14
move finger nudging into options 

tomsilver:


src/envs/pybullet_env.py
```diff
@@ -296,15 +297,28 @@ def _detect_held_object(self) -> Optional[int]:
...
-                                              Tuple[Pose3D, Pose3D]]
+                                              Tuple[Pose3D, Pose3D]],
+        finger_status: str,
     ) -> ParameterizedOption:
         """A generic utility that creates a ParameterizedOption for moving the
         end effector to a target pose, given a function that takes in the
         current state, objects, and parameters, and returns the current pose
         and target pose of the end effector."""
 
+        # Fingers drift if left alone. When the fingers are not explicitly
```
ronuchit, line 313: put this info in the docstring instead, since it's an argument

ronuchit (approved):
is this important for performance?

tomsilver:
i wouldn't expect this PR to make a performance difference. if you're asking whether nudging in general makes a performance difference, yeah it definitely matters

PR #720, opened 2022-04-14
make finger values state-dependent in pybullet options

ronuchit:


tomsilver (approved):
it's not exactly clear to me how you're going to use this for cover, but if it's clear to you then great

PR #723, opened 2022-04-14
pybullet cover

### LINKED ISSUE #704: pybullet version of cover
visualizing hand regions is hard, so we can leave that out

ronuchit:
Supercloud results look good:
```
f97c8c95504d93f8a43f8ecf44a6caa6f7408918


AGGREGATED DATA OVER SEEDS:
                                  NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED     LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                                              
blocks_invent_allexclude             50.00 (0.00)  49.95 (0.22)   6.10 (0.44)   0.33 (0.10)  3563.46 (1482.24)   989.83 (182.14)         20
blocks_invent_noexclude              50.00 (0.00)  49.95 (0.22)   5.30 (0.78)   0.26 (0.07)  3604.61 (1492.54)   228.86 (145.50)         20
blocks_nsrt_learning                 50.00 (0.00)  50.00 (0.00)   5.00 (0.00)   0.24 (0.07)  3563.90 (1476.55)      10.40 (0.97)         20
blocks_oracle                         0.00 (0.00)  50.00 (0.00)   5.00 (0.00)   0.24 (0.07)  3492.69 (1467.75)       0.00 (0.00)         20
cover_invent_allexclude              50.00 (0.00)  49.05 (1.20)   5.80 (0.60)   0.01 (0.02)        4.85 (0.23)    139.12 (17.98)         20
cover_invent_noexclude               50.00 (0.00)  49.05 (1.20)   7.05 (0.38)   0.01 (0.02)        4.85 (0.23)     89.81 (13.55)         20
cover_nsrt_learning                  50.00 (0.00)  49.00 (1.18)   5.00 (0.00)   0.05 (0.02)        6.69 (0.36)      20.59 (1.91)         20
cover_oracle                          0.00 (0.00)  50.00 (0.00)   5.00 (0.00)   0.03 (0.01)        6.71 (0.35)       0.00 (0.00)         20
pybullet_blocks_invent_allexclude    50.00 (0.00)  49.85 (0.36)   6.30 (0.56)   0.43 (0.11)  3703.91 (1558.61)  2368.53 (554.00)         20
pybullet_blocks_invent_noexclude     50.00 (0.00)  49.85 (0.36)   5.20 (0.68)   0.36 (0.07)  3601.93 (1494.34)   629.96 (298.19)         20
pybullet_blocks_nsrt_learning        50.00 (0.00)  49.85 (0.36)   5.00 (0.00)   0.36 (0.08)  3563.90 (1476.55)      18.91 (1.61)         20
pybullet_blocks_oracle                0.00 (0.00)  49.75 (0.54)   5.00 (0.00)   0.36 (0.07)  3492.69 (1467.75)       0.00 (0.00)         20
pybullet_cover_invent_allexclude     50.00 (0.00)  47.40 (1.77)   5.80 (0.60)   0.05 (0.02)        4.86 (0.23)    238.00 (29.74)         20
pybullet_cover_invent_noexclude      50.00 (0.00)  47.40 (1.77)   7.05 (0.38)   0.05 (0.02)        4.86 (0.23)    156.36 (19.74)         20
pybullet_cover_nsrt_learning         50.00 (0.00)  47.25 (1.79)   5.00 (0.00)   0.10 (0.04)        6.70 (0.38)      22.40 (2.09)         20
pybullet_cover_oracle                 0.00 (0.00)  47.05 (1.24)   5.00 (0.00)   0.08 (0.01)        6.71 (0.35)       0.00 (0.00)         20
```

As we expected, there are some instabilities in pybullet_cover, particularly when the planner prefers to do a placement that's on the edge of being correct. The way to get around this would be to plan with a _conservative option model_, but that's more than I'm willing to do, especially for this PR.

closes #704

### DIFF
--- a/scripts/supercloud/run_nightly_experiments.sh
+++ b/scripts/supercloud/run_nightly_experiments.sh
@@ -11,11 +11,11 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     python $FILE --experiment_id cover_invent_noexclude --env cover --approach grammar_search_invention --seed $SEED --num_train_tasks 50
     python $FILE --experiment_id cover_invent_allexclude --env cover --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 50
 
-    # blocks
-    python $FILE --experiment_id blocks_oracle --env blocks --approach oracle --seed $SEED --num_train_tasks 0
-    python $FILE --experiment_id blocks_nsrt_learning --env blocks --approach nsrt_learning --seed $SEED --num_train_tasks 50
-    python $FILE --experiment_id blocks_invent_noexclude --env blocks --approach grammar_search_invention --seed $SEED --num_train_tasks 50
-    python $FILE --experiment_id blocks_invent_allexclude --env blocks --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 50
+    # pybullet_blocks
+    python $FILE --experiment_id pybullet_blocks_oracle --env pybullet_blocks --approach oracle --seed $SEED --num_train_tasks 0
+    python $FILE --experiment_id pybullet_blocks_nsrt_learning --env pybullet_blocks --approach nsrt_learning --seed $SEED --num_train_tasks 50
+    python $FILE --experiment_id pybullet_blocks_invent_noexclude --env pybullet_blocks --approach grammar_search_invention --seed $SEED --num_train_tasks 50
+    python $FILE --experiment_id pybullet_blocks_invent_allexclude --env pybullet_blocks --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 50
 
     # painting
     python $FILE --experiment_id painting_oracle --env painting --approach oracle --seed $SEED --num_train_tasks 0
@@ -30,15 +30,13 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     python $FILE --experiment_id tools_invent_noexclude --env tools --approach grammar_search_invention --seed $SEED --num_train_tasks 200
     python $FILE --experiment_id tools_invent_allexclude --env tools --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 200
 
+    # playroom
+    python $FILE --experiment_id playroom_oracle --env playroom --approach oracle --seed $SEED --num_train_tasks 0
+    python $FILE --experiment_id playroom_nsrt_learning --env playroom --approach nsrt_learning --seed $SEED --num_train_tasks 50
+
     # pybullet_cover
     python $FILE --experiment_id pybullet_cover_oracle --env pybullet_cover --approach oracle --seed $SEED --num_train_tasks 0
     python $FILE --experiment_id pybullet_cover_nsrt_learning --env pybullet_cover --approach nsrt_learning --seed $SEED --num_train_tasks 50
     python $FILE --experiment_id pybullet_cover_invent_noexclude --env pybullet_cover --approach grammar_search_invention --seed $SEED --num_train_tasks 50
     python $FILE --experiment_id pybullet_cover_invent_allexclude --env pybullet_cover --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 50
-
-    # pybullet_blocks
-    python $FILE --experiment_id pybullet_blocks_oracle --env pybullet_blocks --approach oracle --seed $SEED --num_train_tasks 0
-    python $FILE --experiment_id pybullet_blocks_nsrt_learning --env pybullet_blocks --approach nsrt_learning --seed $SEED --num_train_tasks 50
-    python $FILE --experiment_id pybullet_blocks_invent_noexclude --env pybullet_blocks --approach grammar_search_invention --seed $SEED --num_train_tasks 50
-    python $FILE --experiment_id pybullet_blocks_invent_allexclude --env pybullet_blocks --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 50
 done
--- a/scripts/supercloud/run_nightly_experiments.sh
+++ b/scripts/supercloud/run_nightly_experiments.sh
@@ -30,9 +30,11 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     python $FILE --experiment_id tools_invent_noexclude --env tools --approach grammar_search_invention --seed $SEED --num_train_tasks 200
     python $FILE --experiment_id tools_invent_allexclude --env tools --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 200
 
-    # playroom
-    python $FILE --experiment_id playroom_oracle --env playroom --approach oracle --seed $SEED --num_train_tasks 0
-    python $FILE --experiment_id playroom_nsrt_learning --env playroom --approach nsrt_learning --seed $SEED --num_train_tasks 50
+    # pybullet_cover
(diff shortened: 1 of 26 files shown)


scripts/supercloud/run_nightly_experiments.sh
```diff
@@ -30,9 +30,11 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     python $FILE --experiment_id tools_invent_noexclude --env tools --approach grammar_search_invention --seed $SEED --num_train_tasks 200
     python $FILE --experiment_id tools_invent_allexclude --env tools --approach grammar_search_invention --excluded_predicates all --seed $SEED --num_train_tasks 200
 
-    # playroom
```
tomsilver, line 33: instead of removing playroom, can we remove (non pybullet) blocks?

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+from predicators.src.structs import Action, Array, Object, \
+    ParameterizedOption, Pose3D, State, Type
+
+
+class PyBulletCoverEnv(PyBulletEnv, CoverEnv):
+    """PyBullet Cover domain."""
+    # Parameters that aren't important enough to need to clog up settings.py
+
+    # Option parameters.
+    _open_fingers: ClassVar[float] = 0.04
```
tomsilver, line 25: we don't need to do this now, but i'm guessing we will wanna do some refactoring that moves this into the robots, since these refer to joint values that may be different for other robots

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+    _target_height: ClassVar[float] = 0.0001
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # Override PickPlace option
+        types = self._PickPlace.types
+        params_space = self._PickPlace.params_space
+        toggle_fingers_func = lambda s, _1, _2: (
+            (self._open_fingers, self._closed_fingers)
```
tomsilver, line 56: this feels a bit wrong to me because the first argument here is not going to be the real "current" finger state, which is what is expected by the option creator. if you ever see behavior where the robot hits the object as it's trying to grasp it, leading to undesired rotations etc, then this could be to blame

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+        p.resetBasePositionAndOrientation(
+            self._table_id,
+            self._table_pose,
+            self._table_orientation,
+            physicsClientId=self._physics_client_id)
+        p.loadURDF(utils.get_env_asset_path("urdf/table.urdf"),
+                   useFixedBase=True,
+                   physicsClientId=self._physics_client_id2)
+        p.resetBasePositionAndOrientation(
+            self._table_id,
```
tomsilver, line 101: don't we need to use the Id from the second client here? what guarantees that they are the same? (realizing this may be an issue in the PR we already merged too)

src/envs/pybullet_cover.py
tomsilver: okay I think I get why this is happening though -- it's because the finger joint state isn't actually in the State, right?

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def _reset_state(self, state: State) -> None:
+        """Run super(), then handle cover-specific resetting."""
+        super()._reset_state(state)
+
+        # Reset blocks based on the state.
+        block_objs = state.get_objects(self._block_type)
+        self._block_id_to_block = {}
+        for i, block_obj in enumerate(block_objs):
+            block_id = self._block_ids[i]
```
tomsilver, line 160: can you add some assertions that the widths of the block objects here match the widths of the pybullet blocks (up to normalization)? i think they will match right now, but it is a little tenuous

tomsilver: (and same for the targets)

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+            self._block_id_to_block[block_id] = block_obj
+            bx = self._workspace_x
+            # De-normalize block y to actual coordinates.
+            y_norm = state.get(block_obj, "pose")
+            by = self._y_lb + (self._y_ub - self._y_lb) * y_norm
+            if state.get(block_obj, "grasp") != -1:
+                # If an object starts out held, it has a different z.
+                bz = self._workspace_z - 0.01
```
tomsilver, line 168: make 0.01 here a class variable

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+            else:
+                bz = self._table_height + self._obj_len_hgt * 0.5
+            p.resetBasePositionAndOrientation(
+                block_id, [bx, by, bz], [0.0, 0.0, 0.0, 1.0],
```
tomsilver, line 172: this default block orientation is used in enough places that we should make it a class variable

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+        # In the cover environment, we need to first check the hand region
+        # constraint before we can call PyBullet.
+        # Use self._pybullet_robot2 to run forward kinematics, since that
+        # method shouldn't be run on the client that is doing simulation.
+        _, ry, rz = self._pybullet_robot2.forward_kinematics(action.arr)
+        hand = (ry - self._y_lb) / (self._y_ub - self._y_lb)
+        hand_regions = self._get_hand_regions(self._current_state)
+        # If we're going down to grasp, we need to be in a hand region.
+        # Otherwise, we don't care if we're between hand regions.
+        z_thresh = (self._pickplace_z + self._workspace_z) / 2
```
tomsilver, line 218: explain the z thresh part in a comment

src/envs/pybullet_cover.py
```diff
@@ -0,0 +1,313 @@
...
+        rx, ry, rz, _ = self._pybullet_robot.get_state()
+        hand = (ry - self._y_lb) / (self._y_ub - self._y_lb)
+        state_dict[self._robot] = np.array([hand, rx, rz], dtype=np.float32)
+        joint_state = self._pybullet_robot.get_joints()
+
+        # Get block states.
+        for block_id, block in self._block_id_to_block.items():
+            width_unnorm = p.getVisualShapeData(
+                block_id, physicsClientId=self._physics_client_id)[0][3][1]
+            width = width_unnorm / self._max_obj_width * max_width
```
tomsilver, line 240: why can't we just use `CFG.cover_block_widths` directly to get the block widths?

tests/envs/test_pybullet_cover.py
```diff
@@ -0,0 +1,156 @@
...
+    action = env.action_space.sample()
+    with pytest.raises(NotImplementedError):
+        env.simulate(state, action)
+    task = env.get_train_tasks()[0]
+    with pytest.raises(NotImplementedError):
+        env.render_state(state, task, action)
+
+
+def test_pybullet_cover_step(env):
+    """Tests cases for taking actions in PyBulletCoverEnv."""
```
tomsilver, line 104: can you add a test for picking and placing at the boundaries of the workspace?

tomsilver (commented):
great work!

src/envs/pybullet_cover.py
ronuchit: right yeah. i agree with your statement, but i never saw any behavior like that. i'm inclined to leave this as is since it seems to work fine, but i'll be on the lookout if we see this behavior in the future. i'll put a comment.

also, i'll add a video to this PR

ronuchit:
Example video:
https://user-images.githubusercontent.com/3521150/164042563-62fdcfc3-7fb4-4fc4-bbb8-12db31470d49.mp4

PR #725, opened 2022-04-15
change pybullet env action space to joint positions, move IK into options

### LINKED ISSUE #705: change pybullet blocks action space to joint positions
this is in service of #692

tomsilver:
closes #705

### DIFF
--- a/src/envs/pybullet_robots.py
+++ b/src/envs/pybullet_robots.py
@@ -133,9 +133,8 @@ class _SingleArmPyBulletRobot(abc.ABC):
         raise NotImplementedError("Override me!")
 
     @abc.abstractmethod
-    def run_inverse_kinematics(self,
-                               end_effector_pose: Pose3D,
-                               validate: bool = True) -> List[float]:
+    def run_inverse_kinematics(self, end_effector_pose: Pose3D,
+                               validate: bool) -> List[float]:
         """Run inverse kinematics."""
         raise NotImplementedError("Override me!")
 
@@ -179,7 +178,7 @@ class FetchPyBulletRobot(_SingleArmPyBulletRobot):
         self._arm_joints.append(self._right_finger_id)
 
         self._initial_joint_values = self.run_inverse_kinematics(
-            self._ee_home_pose)
+            self._ee_home_pose, validate=True)
         # The initial joint values for the fingers should be open. IK may
         # return anything for them.
         self._initial_joint_values[-2] = self._open_fingers
@@ -285,9 +284,8 @@ class FetchPyBulletRobot(_SingleArmPyBulletRobot):
                                     targetPosition=joint_val,
                                     physicsClientId=self._physics_client_id)
 
-    def run_inverse_kinematics(self,
-                               end_effector_pose: Pose3D,
-                               validate: bool = True) -> List[float]:
+    def run_inverse_kinematics(self, end_effector_pose: Pose3D,
+                               validate: bool) -> List[float]:
         return inverse_kinematics(self._fetch_id,
                                   self._ee_id,
                                   end_effector_pose,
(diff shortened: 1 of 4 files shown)


tests/envs/test_pybullet_blocks.py
```diff
@@ -148,22 +151,6 @@ def test_pybullet_blocks_picking(env):
         state = env.step(action)
     # The block should now be held.
     assert state.get(block, "held") == 1.0
-    # Test the case where the right finger is on the left side of the block,
```
tomsilver, line 151: was too lazy to deal with this because hardcoding actions in joint space is now annoying, but if you yell at me i will figure it out

tomsilver:
supercloud results:
```
EXPERIMENT_ID                                                                                                                      
pybullet_blocks_nsrt_learning    50.00 (0.00)  49.70 (0.46)   5.00 (0.00)   0.39 (0.15)  3696.50 (1766.09)  17.98 (1.64)         20
pybullet_blocks_oracle            0.00 (0.00)  49.65 (0.57)   5.00 (0.00)   0.36 (0.09)  3731.87 (1821.66)   0.00 (0.00)         20

```

src/envs/pybullet_robots.py
```diff
@@ -129,6 +132,13 @@ def set_motors(self, action_arr: Array) -> None:
         """Update the motors to execute the given action in PyBullet."""
         raise NotImplementedError("Override me!")
 
+    @abc.abstractmethod
+    def run_inverse_kinematics(self,
+                               end_effector_pose: Pose3D,
+                               validate: bool = True) -> List[float]:
```
ronuchit, line 138: do we really want validate = True as a default arg? aren't we using it as False? do we ever use it as True? maybe make it a non-option arg so people have to think about it?

tests/envs/test_pybullet_blocks.py
ronuchit: no problem

src/envs/pybullet_robots.py
```diff
@@ -250,42 +275,26 @@ def get_joints(self) -> Sequence[float]:
...
         # Set arm joint motors.
-        for joint_idx, joint_val in zip(self._arm_joints, joint_values):
+        for joint_idx, joint_val in zip(self._arm_joints, action_arr):
             p.setJointMotorControl2(bodyIndex=self._fetch_id,
                                     jointIndex=joint_idx,
                                     controlMode=p.POSITION_CONTROL,
                                     targetPosition=joint_val,
                                     physicsClientId=self._physics_client_id)
 
-        # Set finger joint motors.
```
ronuchit, line 282: why is this removed?

src/envs/pybullet_robots.py
```diff
@@ -250,42 +275,26 @@ def get_joints(self) -> Sequence[float]:
...
-                                       self._ee_id,
-                                       physicsClientId=self._physics_client_id)
-        current_position = ee_link_state[4]
-        target_position = np.add(current_position, ee_delta).tolist()
-
-        # We assume that the robot is already close enough to the target
-        # position that IK will succeed with one call, so validate is False.
-        # Furthermore, updating the state of the robot during simulation, which
-        # validate=True would do, is risky and discouraged by PyBullet.
-        joint_values = inverse_kinematics(
```
ronuchit, line 265: so we're now doing IK in option policy, but setJointMotorControl2 in step(), right?

ronuchit (approved):

src/envs/pybullet_robots.py
tomsilver: we don't need it because the fingers are in the arm joints, so they're set above

src/envs/pybullet_robots.py
tomsilver: right

PR #733, opened 2022-04-18
execution failure metric tweak & plotting different error types for interactive learning

### LINKED ISSUE #729: log number of different kinds of errors
main thing is we want to know how many option execution failures there are vs. planning failures. but breaking down the different kinds of planning failures could also be useful.

tomsilver:
closes #729

also removed commented out / unused code from analysis scripts. to see what columns are available, we can refer to `analyze_results_directory.py`, we don't need them listed in the other files.

I tested the update to the interactive learning plot by running the "silent kid" command over 3 seeds and 1 learning cycle, and then ran the script. the plots from that are below. (the data are not meaningful, this just shows that the script runs and the axes are what we want.)

coverenv_excluding_covers,holding_num_transitions_perc_plan_fail
coverenv_excluding_covers,holding_num_transitions_perc_exec_fail
coverenv_excluding_covers,holding_num_transitions_query_cost
coverenv_excluding_covers,holding_num_transitions_perc_solved

### DIFF
--- a/scripts/create_interactive_predicate_learning_plots.py
+++ b/scripts/create_interactive_predicate_learning_plots.py
@@ -33,8 +33,7 @@ COLUMN_NAMES_AND_KEYS = [
 ]
 
 DERIVED_KEYS = [
-    ("perc_solved",
-     lambda r: 100 * r["num_solved"] / r["num_test_tasks"]),
+    ("perc_solved", lambda r: 100 * r["num_solved"] / r["num_test_tasks"]),
     ("perc_exec_fail",
      lambda r: 100 * r["num_execution_failures"] / r["num_test_tasks"]),
     ("perc_plan_fail",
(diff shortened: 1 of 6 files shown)


scripts/create_interactive_predicate_learning_plots.py
```diff
@@ -28,10 +28,15 @@
...
     ("NUM_TRANSITIONS", "num_transitions"),
     ("QUERY_COST", "query_cost"),
+    ("NUM_EXECUTION_FAILURES", "num_execution_failures"),
+    ("NUM_PLANNING_FAILURES", "num_planning_failures"),
 ]
 
 DERIVED_KEYS = [("perc_solved",
-                 lambda r: 100 * r["num_solved"] / r["num_test_tasks"])]
+                 lambda r: 100 * r["num_solved"] / r["num_test_tasks"]),
+                ("num_planning_failures",
```
ronuchit, line 37: is it guaranteed that all non-solved tasks will be one of these two types of failures? are we ignoring things like code crashes?

ronuchit: why not just have another entry in the results for num_planning_failures?

ronuchit (approved):

amburger66 (approved):
thanks!

PR #734, opened 2022-04-18
2D classification plot is now video

amburger66:
example:
https://user-images.githubusercontent.com/26547725/163830796-6ac58cc8-43f6-4c09-928c-35fb95c18d64.mp4


tomsilver (approved):
better than netflix!

PR #738, opened 2022-04-18
move two generic option creators into pybullet robot

tomsilver:
this is so that the panda can run IK once and motion plan, if it wants to, but the fetch can still run IK at every time step, which appears to be more stable and effective.

### DIFF
--- a/src/envs/pybullet_env.py
+++ b/src/envs/pybullet_env.py
@@ -25,7 +25,7 @@ class PyBulletEnv(BaseEnv):
     # General robot parameters.
     _max_vel_norm: ClassVar[float] = 0.05
     _grasp_tol: ClassVar[float] = 0.05
-    _finger_action_tol: ClassVar[float] = 0.0001
+    _finger_action_tol: ClassVar[float] = 1e-4
 
     # Object parameters.
     _obj_mass: ClassVar[float] = 0.5
--- a/src/envs/pybullet_env.py
+++ b/src/envs/pybullet_env.py
@@ -4,8 +4,7 @@ Contains useful common code.
 """
 
 import abc
-from typing import Callable, ClassVar, Dict, List, Optional, Sequence, Tuple, \
-    cast
+from typing import ClassVar, List, Optional, Sequence, Tuple, cast
 
 import numpy as np
 import pybullet as p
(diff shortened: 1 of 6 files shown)


src/envs/pybullet_robots.py
```diff
@@ -294,15 +320,124 @@ def run_inverse_kinematics(self, end_effector_pose: Pose3D,
...
 
+    def create_move_end_effector_to_pose_option(
+        self,
+        name: str,
+        types: Sequence[Type],
+        params_space: Box,
+        get_current_and_target_pose_and_finger_status: Callable[
+            [State, Sequence[Object], Array], Tuple[Pose3D, Pose3D, str]],
+    ) -> ParameterizedOption:
+        """Fingers drift if left alone.
```
ronuchit, line 331: this docstring doesn't make sense here, and we shouldn't have one anyway since the parent method has one. move this to where we say "Handle fingers."?

src/envs/pybullet_env.py
```diff
@@ -26,9 +25,7 @@ class PyBulletEnv(BaseEnv):
     # General robot parameters.
     _max_vel_norm: ClassVar[float] = 0.05
     _grasp_tol: ClassVar[float] = 0.05
-    _move_to_pose_tol: ClassVar[float] = 0.0001
```
ronuchit, line 29: have you checked that these are the only paramters that need to be moved?

src/envs/pybullet_robots.py
```diff
@@ -146,6 +170,8 @@ class FetchPyBulletRobot(_SingleArmPyBulletRobot):
     _base_pose: ClassVar[Pose3D] = (0.75, 0.7441, 0.0)
     _base_orientation: ClassVar[Sequence[float]] = [0., 0., 0., 1.]
     _ee_orientation: ClassVar[Sequence[float]] = [1., 0., -1., 0.]
+    _finger_action_nudge_magnitude: ClassVar[float] = 0.001
+    _move_to_pose_tol: ClassVar[float] = 0.0001
```
ronuchit, line 174: rename to 1e-4 for both this line and the previous one

ronuchit (approved):

src/envs/pybullet_env.py
tomsilver: yep

PR #739, opened 2022-04-18
Greedy explore

### LINKED ISSUE #689: greedy exploration

amburger66:
closes #689

### DIFF
--- a/scripts/supercloud/run_action_policy_experiments.sh
+++ b/scripts/supercloud/run_action_policy_experiments.sh
@@ -14,8 +14,10 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     COMMON_ARGS="--env cover --approach interactive_learning --seed $SEED --excluded_predicates Covers,Holding --interactive_query_policy threshold --interactive_score_function entropy --interactive_score_threshold 0.05 --num_online_learning_cycles $CYCLES --online_learning_max_transitions $MAX_TRANSITIONS --interactive_num_requests_per_cycle $REQUESTS --max_num_steps_interaction_request $MAX_STEPS --min_data_for_nsrt $MIN_DATA --sampler_disable_classifier True --mlp_classifier_balance_data False"
 
     # glib
-    python $FILE $COMMON_ARGS --experiment_id glib --interactive_action_strategy glib
+    # python $FILE $COMMON_ARGS --experiment_id glib --interactive_action_strategy glib
     # greedy lookahead
-    python $FILE $COMMON_ARGS --experiment_id greedy_lookahead --interactive_action_strategy greedy_lookahead
+    # python $FILE $COMMON_ARGS --experiment_id greedy_lookahead --interactive_action_strategy greedy_lookahead
+    # glib + 10k max iters
+    python $FILE $COMMON_ARGS --experiment_id glib_10k --interactive_action_strategy glib --predicate_mlp_classifier_max_itr 10000
 
 done
--- a/scripts/supercloud/run_action_policy_experiments.sh
+++ b/scripts/supercloud/run_action_policy_experiments.sh
@@ -16,6 +16,6 @@ for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
     # glib
     python $FILE $COMMON_ARGS --experiment_id glib --interactive_action_strategy glib
     # greedy lookahead
-    python $FILE $COMMON_ARGS --experiment_id glib --interactive_action_strategy greedy_lookahead
+    python $FILE $COMMON_ARGS --experiment_id greedy_lookahead --interactive_action_strategy greedy_lookahead
 
 done
--- a/scripts/supercloud/run_action_policy_experiments.sh
+++ b/scripts/supercloud/run_action_policy_experiments.sh
@@ -11,7 +11,11 @@ FILE="scripts/supercloud/submit_supercloud_job.py"
 
 for SEED in $(seq $START_SEED $((NUM_SEEDS+START_SEED-1))); do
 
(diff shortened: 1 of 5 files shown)


tomsilver:
I feel like I'm too close to this since we pair programmed it, so when you're ready for a review, let's ask ronuchit

scripts/supercloud/run_action_policy_experiments.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
ronuchit, line 1: could you add the results of this experiment to this PR? (can be done after merging too.) it helps keep things organized & lets us easily look back to performance of old things that were tried

src/approaches/interactive_learning_approach.py
```diff
@@ -255,6 +258,75 @@ def _create_glib_interaction_strategy(
...
+            trajectory_length = 0
+            total_score = 0.0
+            while trajectory_length < CFG.interactive_max_trajectory_length:
+                # Sample an NSRT that has preconditions satisfied in the
+                # current state.
+                ground_nsrt = self._sample_applicable_ground_nsrt(state)
+                if ground_nsrt is None:  # No applicable NSRTs
+                    break
+                assert all(a.holds for a in ground_nsrt.preconditions)
+                # Sample an option. Note that goal is assumed not used.
```
ronuchit, line 282: we can verify this via `assert not CFG.sampler_learning_use_goals`

src/settings.py
```diff
@@ -206,12 +206,16 @@ class GlobalSettings:
 
     # interactive learning parameters
     interactive_num_ensemble_members = 10
-    interactive_action_strategy = "glib"
-    interactive_query_policy = "strict_best_seen"
-    interactive_score_function = "frequency"
-    interactive_score_threshold = 0.5
+    interactive_action_strategy = "greedy_lookahead"
```
ronuchit, line 209: is there a reason you want to change these defaults? (doesn't matter much, just wondering)

src/approaches/interactive_learning_approach.py
```diff
@@ -255,6 +258,75 @@ def _create_glib_interaction_strategy(
...
+        best_options = []
+        for _ in range(CFG.interactive_max_num_trajectories):
+            state = init.copy()
+            options = []
+            trajectory_length = 0
+            total_score = 0.0
+            while trajectory_length < CFG.interactive_max_trajectory_length:
+                # Sample an NSRT that has preconditions satisfied in the
+                # current state.
+                ground_nsrt = self._sample_applicable_ground_nsrt(state)
```
ronuchit, line 278: inside this function, you're re-computing all the ground nsrts, but that only has to be done once at the start of this function (using the objects in `init`). so, we should just make _sample_applicable_ground_nsrt take in all the ground  nsrts.

src/approaches/interactive_learning_approach.py
```diff
@@ -255,6 +258,75 @@ def _create_glib_interaction_strategy(
...
+                options.append(option)
+                trajectory_length += num_actions
+                # Update the total score.
+                atoms = utils.abstract(state, self._predicates_to_learn)
+                total_score += self._score_atom_set(atoms, state)
+            else:  # Sampled a complete trajectory
+                if total_score > best_score:
+                    best_score = total_score
+                    best_options = options
+
```
ronuchit, line 306: should we `assert not np.isinf(best_score)`?

src/approaches/interactive_learning_approach.py
```diff
@@ -255,6 +258,75 @@ def _create_glib_interaction_strategy(
...
+                # and we don't want to use this option after all. To prevent
+                # possible infinite loops, just break immediately in this case.
+                if num_actions == 0:
+                    break
+                options.append(option)
+                trajectory_length += num_actions
+                # Update the total score.
+                atoms = utils.abstract(state, self._predicates_to_learn)
+                total_score += self._score_atom_set(atoms, state)
+            else:  # Sampled a complete trajectory
```
ronuchit, line 302: is there a reason you want to force all the trajectories to be exactly of length `interactive_max_trajectory_length`? what's the problem if they're shorter than that?

ronuchit (commented):
looks good!

src/settings.py
amburger66: the combination of frequency + strict_best_seen is just not very good

src/approaches/interactive_learning_approach.py
```diff
@@ -255,6 +258,78 @@ def _create_glib_interaction_strategy(
...
+                if num_actions == 0:
+                    break
+                options.append(option)
+                trajectory_length += num_actions
+                # Update the total score.
+                atoms = utils.abstract(state, self._predicates_to_learn)
+                total_score += self._score_atom_set(atoms, state)
+            if total_score > best_score:
+                best_score = total_score
+                assert not np.isinf(best_score)
```
ronuchit, line 310: oh i actually meant that we should do this assert after the loop, right before (and at the same indentation level as) `act_policy = utils.option_plan_to_policy(best_options)`. the idea is to make sure that we saw at least one trajectory, so best_score isn't just equal to its initialization (`-np.inf`)

src/main.py
```diff
@@ -219,6 +219,7 @@ def _generate_interaction_results(
             request.train_task_idx,
             request.termination_function,
             max_num_steps=CFG.max_num_steps_interaction_request,
+            exceptions_to_break_on={utils.OptionExecutionFailure},
```
ronuchit, line 222: are there any others using online stuff that this might affect? @tomsilver

src/approaches/interactive_learning_approach.py
amburger66: i was thinking about this too... i guess it's not a problem and if it's the best trajectory then we should use it!

src/main.py
tomsilver: yeah Ashay actually added this line too in his open PR, for the same reason as we're adding it here

ronuchit (approved):
lgtm! remember to add results when they're ready!

amburger66:
coverenv_glib_vs_greedy_lookahead_num_transitions_query_cost
coverenv_glib_vs_greedy_lookahead_num_transitions_perc_solved
coverenv_glib_vs_greedy_lookahead_num_transitions_perc_plan_fail
coverenv_glib_vs_greedy_lookahead_num_transitions_perc_exec_fail

amburger66:
https://user-images.githubusercontent.com/26547725/164309727-e7c39b5a-a8d6-4bbd-a266-38141cd6cf09.mp4

PR #741, opened 2022-04-18
changes to pybullet to prep for pybullet_cover

ronuchit:
* add second pybullet connection for stateless things like collision checking and forward kinematics
* make `move_to_pose_tol` and `ee_orientation` env-specific
* move grasp constraint creation to its own method (no code changes)

### DIFF
--- a/src/envs/pybullet_blocks.py
+++ b/src/envs/pybullet_blocks.py
@@ -30,7 +30,8 @@ class PyBulletBlocksEnv(PyBulletEnv, BlocksEnv):
     _table_orientation: ClassVar[Sequence[float]] = [0., 0., 0., 1.]
 
     # Robot parameters.
-    _ee_orn = p.getQuaternionFromEuler([0.0, np.pi / 2, -np.pi])
+    _ee_orn: ClassVar[Sequence[float]] = p.getQuaternionFromEuler(
+        [0.0, np.pi / 2, -np.pi])
     _move_to_pose_tol: ClassVar[float] = 1e-4
 
     def __init__(self) -> None:
--- a/src/envs/pybullet_blocks.py
+++ b/src/envs/pybullet_blocks.py
@@ -24,6 +24,7 @@ class PyBulletBlocksEnv(PyBulletEnv, BlocksEnv):
 
     # Option parameters.
     _offset_z: ClassVar[float] = 0.01
+    _move_to_pose_tol: ClassVar[float] = 1e-4
 
     # Table parameters.
     _table_pose: ClassVar[Pose3D] = (1.35, 0.75, 0.0)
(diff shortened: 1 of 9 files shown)


src/envs/pybullet_env.py
```diff
@@ -251,8 +244,10 @@ def _detect_held_object(self) -> Optional[int]:
         one that is closest.
         """
         expected_finger_normals = {
-            self._pybullet_robot.left_finger_id: np.array([0., 1., 0.]),
-            self._pybullet_robot.right_finger_id: np.array([0., -1., 0.]),
+            self._pybullet_robot.left_finger_id: np.array([np.pi / 2., 1.,
```
ronuchit, line 247: i don't understand what this change does, but somehow it works for both cover and blocks (which use two different ee_orientations). the previous thing didn't work for cover.

tomsilver: sounds a little fluky. can we try to understand what's going on here before merging?

tests/envs/test_pybullet_robots.py
```diff
@@ -174,7 +177,7 @@ def test_fetch_pybullet_robot():
     assert np.allclose(robot_state, recovered_state, atol=1e-3)
     assert np.allclose(robot.get_joints(),
                        robot.initial_joint_values,
-                       atol=1e-3)
+                       atol=1e-2)
```
tomsilver, line 180: what happened here?

src/envs/pybullet_blocks.py
```diff
@@ -29,6 +29,10 @@ class PyBulletBlocksEnv(PyBulletEnv, BlocksEnv):
     _table_pose: ClassVar[Pose3D] = (1.35, 0.75, 0.0)
     _table_orientation: ClassVar[Sequence[float]] = [0., 0., 0., 1.]
 
+    # Robot parameters.
+    _ee_orn = p.getQuaternionFromEuler([0.0, np.pi / 2, -np.pi])
```
tomsilver, line 33: do these need to be implemented by all subclasses of pybullet env? if so, can we make them abstract properties in the pybullet env?

ronuchit: we don't need to, it's correct as is. because `create_pybullet_robot` is abstract already, and that's where the subclasses use this `ee_orn` thing. it's not used in the pybullet base env.

src/envs/pybullet_env.py
ronuchit: sure will look into it tomorrow

tests/envs/test_pybullet_robots.py
ronuchit: we changed the reset_state() code to first reset to `self._initial_joint_values` and then call IK + reset joint states again, this time to (rx, ry, rz). so there will be larger-than-before errors in this allclose. but i still prefer to keep it this way

tomsilver: Kk, fine with me as long as supercloud checks out

src/envs/pybullet_env.py
ronuchit: nvm looked into it now, tl;dr is you're right it's a bad mistake, i'll make this env-specific.

ronuchit: resolved in latest commit

tomsilver (approved):
Lgtm

PR #742, opened 2022-04-18
utilities for visualizing and collision checking simple 2D geoms

tomsilver:
I didn't expect to go down this rabbit hole when I woke up today, but here we are...

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -287,6 +287,10 @@ class _Geom2DBody(abc.ABC):
         """Checks if a point is contained in the body."""
         raise NotImplementedError("Override me!")
 
+    def intersects(self, other: _Geom2DBody) -> bool:
+        """Checks if this body intersects with another one."""
+        return geom2d_bodies_intersect(self, other)
+
 
 @dataclass(frozen=True)
 class LineSegment(_Geom2DBody):
@@ -522,6 +526,17 @@ def line_segment_intersects_circle(seg: LineSegment,
     return circ.contains_point(dx, dy)
 
 
+def line_segment_intersects_rectangle(seg: LineSegment, rect: Rectangle) -> bool:
+    """Checks if a line segment intersects a rectangle."""
+    # Case 1: one of the end points of the segment are in the rectangle.
+    if rect.contains_point(seg.x1, seg.y1):
+        return True
+    if rect.contains_point(seg.x2, seg.y2):
+        return True
+    # Case 2: the segment intersects with one of the rectangle sides.
+    return any(line_segments_intersect(s, seg) for s in rect.line_segments)
+
+
 def rectangle_intersects_circle(rect: Rectangle, circ: Circle) -> bool:
     """Checks if a rectangle intersects a circle."""
     # Case 1: the circle's center is in the rectangle.
@@ -544,6 +559,10 @@ def geom2d_bodies_intersect(body1: Union[Rectangle, Circle],
         return line_segments_intersect(body1, body2)
     if isinstance(body1, LineSegment) and isinstance(body2, Circle):
         return line_segment_intersects_circle(body1, body2)
+    if isinstance(body1, LineSegment) and isinstance(body2, Rectangle):
+        return line_segment_intersects_rectangle(body1, body2)
+    if isinstance(body1, Rectangle) and isinstance(body2, LineSegment):
+        return line_segment_intersects_rectangle(body2, body1)
     if isinstance(body1, Circle) and isinstance(body2, LineSegment):
         return line_segment_intersects_circle(body2, body1)
     if isinstance(body1, Rectangle) and isinstance(body2, Rectangle):
--- a/src/utils.py
+++ b/src/utils.py
@@ -371,11 +371,47 @@ class Rectangle:
         ax.add_patch(patch)
 
 
+def line_segments_intersect(seg1: LineSegment, seg2: LineSegment) -> bool:
+    """Checks if two line segments intersect.
+
+    This method, which works by checking relative orientation, allows
+    for collinearity, and only checks if each segment straddles the line
+    containing the other.
+    """
+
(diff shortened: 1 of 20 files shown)


tests/test_utils.py
```diff
@@ -221,58 +222,251 @@ def test_create_state_from_dict():
...
-    l1, r1 = (1, 7), (3, 1)
-    l2, r2 = (1, 8), (6, 1)
-    assert utils.overlap(l1, r1, l2, r2)
+def test_geom2d_bodies_intersect():
+    """Tests for geom2d_bodies_intersect()."""
 
-    l1, r1 = (1, 7), (5, 1)
-    l2, r2 = (2, 4), (4, 2)
-    assert utils.overlap(l1, r1, l2, r2)
+    class _MockGeom2DBody(utils._Geom2DBody):  # pylint: disable=protected-access
```
ronuchit, line 451: you can get rid of this pylint disable by subclassing from one of the public ones instead (Circle, Rectangle)

tests/test_utils.py
```diff
@@ -221,58 +222,251 @@ def test_create_state_from_dict():
...
-    l1, r1 = (1, 4), (3, 1)
-    l2, r2 = (5, 8), (7, 6)
-    assert not utils.overlap(l1, r1, l2, r2)
+    body = _MockGeom2DBody()
 
-    l1, r1 = (1, 4), (6, 1)
-    l2, r2 = (2, 7), (5, 5)
-    assert not utils.overlap(l1, r1, l2, r2)
+    _, ax = plt.subplots(1, 1)
+    with pytest.raises(NotImplementedError):
```
ronuchit, line 462: why do you need to test these? it's just going to run the methods here in the test file, so it's not actually covering anything in src right?

src/utils.py
```diff
@@ -219,53 +221,296 @@ def create_state_from_dict(data: Dict[Object, Dict[str, float]],
...
         return False
-    return True
+    # So d is on the segment. Check if it's in the circle.
+    return circ.contains_point(dx, dy)
+
+
+def line_segment_intersects_rectangle(seg: LineSegment,
+                                      rect: Rectangle) -> bool:
+    """Checks if a line segment intersects a rectangle."""
+    # Case 1: one of the end points of the segment are in the rectangle.
```
ronuchit, line 472: is*

src/utils.py
```diff
@@ -219,53 +221,296 @@ def create_state_from_dict(data: Dict[Object, Dict[str, float]],
     return State(state_dict, simulator_state)
 
 
-def intersects(p1: Tuple[float, float], p2: Tuple[float, float],
-               p3: Tuple[float, float], p4: Tuple[float, float]) -> bool:
-    """Checks if line segment p1p2 and p3p4 intersect.
+class _Geom2DBody(abc.ABC):
+    """A 2D body that contains some points."""
```
ronuchit, line 225: is 2D going to be enough for us, thinking ahead?

src/utils.py
```diff
@@ -219,53 +221,296 @@ def create_state_from_dict(data: Dict[Object, Dict[str, float]],
     return State(state_dict, simulator_state)
 
 
-def intersects(p1: Tuple[float, float], p2: Tuple[float, float],
-               p3: Tuple[float, float], p4: Tuple[float, float]) -> bool:
-    """Checks if line segment p1p2 and p3p4 intersect.
+class _Geom2DBody(abc.ABC):
```
ronuchit, line 224: this all looks reasonable. but really, why are we not just using a library for all this? we can have our own wrappers around it to conform to this API so you don't have to change the tests, but that will be a lot easier to extend in the future, no?

ronuchit: for instance sympy provides basically everything you're doing here:
https://docs.sympy.org/latest/modules/geometry/index.html

like in the Polygon class (https://docs.sympy.org/latest/modules/geometry/polygons.html#sympy.geometry.polygon.Polygon), which you can use to implement rectangles, there is a method `intersection` and another method `encloses_point`

tomsilver: I definitely should've looked into this first, oops. Looking into it now, here's what I'm seeing:
* sympy is extremely slow. replacing this code with sympy code, `pytest -s tests/envs/test_cover.py  -k test_cover_multistep_options` goes from ~3 seconds to ~87 seconds. there are also some stack overflow whispers confirming that other people have found it slow, and it's not intended for this kind of usage.
* [shapely](https://shapely.readthedocs.io/en/latest/manual.html#geometric-objects) has a similar API, but it makes me nervous because it seems to discretize everything (and for example, there's no way to make a perfect circle). this is because it descends from GIS software where they mostly care about maps.
* [scikit-geometry](https://scikit-geometry.github.io/scikit-geometry/introduction.html#Basic-Geometric-Types) looks promising, but it's too new and small right now, and there's apparently [no way to pip install it](https://stackoverflow.com/questions/71591926/how-to-install-the-package-scikit-geometry).
* I don't see any other good candidates listed [here](https://stackoverflow.com/questions/1076778/good-geometry-library-in-python) or from googling around.

src/utils.py
tomsilver: I think so? for 3D things, I imagine we'll use PyBullet

tests/test_utils.py
tomsilver: I can't do that because then the `isinstance` checks will get caught

tests/test_utils.py
tomsilver: coverage complains if I don't do this. I'm sure there's a better way, and I feel like we've run into this before, but I can't figure out how to do it...

tests/test_utils.py
tomsilver: okay I think I found a workaround (pushed)

tests/test_utils.py
tomsilver: okay I think I found a workaround (pushed)

ronuchit (approved):

PR #744, opened 2022-04-19
fix importing of PyBulletState

ronuchit:


src/envs/pybullet_env.py
```diff
@@ -15,7 +15,6 @@
 from predicators.src.envs.pybullet_robots import _SingleArmPyBulletRobot
 from predicators.src.settings import CFG
 from predicators.src.structs import Action, Array, Image, Pose3D, State, Task
-from predicators.src.utils import PyBulletState
```
ronuchit, line 18: double importing utils was the original sin!

tomsilver (approved):
😬

PR #754, opened 2022-04-20
special case null samplers during planning

tomsilver:
this doesn't seem to make a difference in our existing environments (but we should supercloud it). i verified that it makes a huge difference in the stick point environment that's under development though.


ronuchit (approved):
yes please 😋

PR #755, opened 2022-04-20
GNNMetacontrollerApproach

ronuchit:


src/approaches/gnn_metacontroller_approach.py
```diff
@@ -0,0 +1,200 @@
...
+        NSRT."""
+        for _ in range(CFG.gnn_metacontroller_max_samples):
+            # Invoke the ground NSRT's sampler to produce an option.
+            opt = ground_nsrt.sample_option(state, goal, self._rng)
+            if not opt.initiable(state):
+                # The option is not initiable. Continue on to the next sample.
+                continue
+            next_state, _ = self._option_model.get_next_state_and_num_actions(
+                state, opt)
+            next_atoms = utils.abstract(next_state, self._initial_predicates)
```
tomsilver, line 176: can you use the same pattern as we have in planning.py here instead?

tomsilver (approved):
really awesome!

PR #761, opened 2022-04-21
fix `has_option` for loading

tomsilver:
if we save data that has no options (dummy options) and then load a dataset, `has_option` incorrectly reports True, because the `DummyOption` instance is different now, because it's a different Python process.

I looked for other similar cases in the code and didn't see any. there are a few other instances of `is DummyOption`, but those look correct, because they could only refer to the dummy option for the current process.

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -950,7 +950,7 @@ class Action:
 
     def has_option(self) -> bool:
         """Whether this action has a non-default option attached."""
-        return self._option.name != DummyOption.name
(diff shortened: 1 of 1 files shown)


ronuchit (commented):
should we do `self._option.parent != DummyOption.parent`? i feel like that's more clear, and `__eq__` on parameterized options is defined to use the name already

ronuchit (approved):

PR #764, opened 2022-04-21
oracle segmenter

### LINKED ISSUE #760: create oracle segmenter
should use the oracle operators to perform segmentation. the goal is that this command should run and work as expected:
```
python src/main.py --env stick_point --approach nsrt_learning --option_learner direct_bc --num_train_tasks 500 --min_data_for_nsrt 50 --seed 0 --strips_learner oracle --segmenter oracle
```

if options are known, this should just call the options segmentation. but if they're not, then we should use the operators.

tomsilver:
closes #760

### DIFF
--- a/src/nsrt_learning/segmentation.py
+++ b/src/nsrt_learning/segmentation.py
@@ -72,23 +72,24 @@ def _segment_with_oracle(trajectory: GroundAtomTrajectory) -> List[Segment]:
         for ground_nsrt in utils.all_ground_nsrts(nsrt, objects)
     }
     atoms = all_atoms[0]
-    expected_next_atoms = [
+    all_expected_next_atoms = [
         utils.apply_operator(n, atoms)
         for n in utils.get_applicable_operators(ground_nsrts, atoms)
     ]
 
     def _switch_fn(t: int) -> bool:
-        nonlocal expected_next_atoms  # update at each switch point
-        atoms = all_atoms[t + 1]
+        nonlocal all_expected_next_atoms  # update at each switch point
+        next_atoms = all_atoms[t + 1]
         # Check if any of the current NSRT effects hold.
-        for next_atoms in expected_next_atoms:
+        for expected_next_atoms in all_expected_next_atoms:
             # Check if we have reached the expected next atoms.
-            if next_atoms != atoms:
+            if expected_next_atoms != next_atoms:
                 continue
             # Time to segment. Update the expected next atoms.
-            expected_next_atoms = [
-                utils.apply_operator(n, atoms)
-                for n in utils.get_applicable_operators(ground_nsrts, atoms)
+            applicable_nsrts = utils.get_applicable_operators(
+                ground_nsrts, next_atoms)
+            all_expected_next_atoms = [
+                utils.apply_operator(n, next_atoms) for n in applicable_nsrts
             ]
             return True
         # Not yet time to segment.
--- a/src/nsrt_learning/segmentation.py
+++ b/src/nsrt_learning/segmentation.py
@@ -60,8 +60,6 @@ def _segment_with_oracle(trajectory: GroundAtomTrajectory) -> List[Segment]:
     which oracle ground NSRTs are applicable. When any of them have their
     effects achieved, that marks the switch point between segments.
     """
-    import ipdb; ipdb.set_trace()
-
     traj, all_atoms = trajectory
     if not traj.actions or traj.actions[0].has_option():
         return _segment_with_option_changes(trajectory)
--- a/src/nsrt_learning/segmentation.py
+++ b/src/nsrt_learning/segmentation.py
@@ -78,11 +78,9 @@ def _segment_with_oracle(trajectory: GroundAtomTrajectory) -> List[Segment]:
         atoms = all_atoms[t + 1]
         # Check if any of the current NSRT effects hold.
         for ground_nsrt in current_nsrts:
-            # Check add effects.
-            if not ground_nsrt.add_effects.issubset(atoms):
-                continue
-            # Check delete effects.
-            if ground_nsrt.delete_effects & atoms:
+            # Check effects.
+            if not ground_nsrt.add_effects.issubset(atoms) or (
+                    ground_nsrt.delete_effects & atoms):
                 continue
             # Time to segment. Update the current NSRTs.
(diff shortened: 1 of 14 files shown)


src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,47 @@ def _switch_fn(t: int) -> bool:
...
+    """Segment a trajectory using oracle NSRTs.
+
+    If options are known, just uses _segment_with_option_changes().
+
+    Otherwise, starting at the beginning of the trajectory, keeps track of
+    which oracle ground NSRTs are applicable. When any of them have their
+    effects achieved, that marks the switch point between segments.
+    """
+    traj, all_atoms = trajectory
+    if not traj.actions or traj.actions[0].has_option():
```
ronuchit, line 64: the first condition here is weird -- is it covering an edge case?

src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,47 @@ def _switch_fn(t: int) -> bool:
...
+        return _segment_with_option_changes(trajectory)
+    env = get_or_create_env(CFG.env)
+    gt_nsrts = get_gt_nsrts(env.predicates, env.options)
+    ground_nsrts = {
+        ground_nsrt
+        for nsrt in gt_nsrts
+        for ground_nsrt in utils.all_ground_nsrts(nsrt, list(traj.states[0]))
```
ronuchit, line 71: save `objects = list(traj.states[0])` before doing this double-loop

src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,47 @@ def _switch_fn(t: int) -> bool:
...
+    }
+    atoms = all_atoms[0]
+    current_nsrts = list(utils.get_applicable_operators(ground_nsrts, atoms))
+
+    def _switch_fn(t: int) -> bool:
+        nonlocal current_nsrts  # update at each switch point
+        atoms = all_atoms[t + 1]
+        # Check if any of the current NSRT effects hold.
+        for ground_nsrt in current_nsrts:
+            # Check effects.
```
ronuchit, line 81: is it worth having an `prev_atoms = all_atoms[t]; assert ground_nsrt.preconditions.issubset(prev_atoms)` here?

src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,47 @@ def _switch_fn(t: int) -> bool:
...
+            if not ground_nsrt.add_effects.issubset(atoms) or (
```
ronuchit, line 82: highly confusing...can we just use utils.apply_operator somehow?

at the very least, split this into two if statements, one after the other.

src/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -94,6 +94,14 @@ def _check_harmlessness(self,
                 seg_traj, ll_traj.states[0], atoms_seq, traj_goal, strips_ops,
                 option_specs)
             if not demo_preserved:
+                # Useful debug point.
```
ronuchit, line 97: this is what we have `logging.debug` for :)

src/nsrt_learning/segmentation.py
tomsilver: nvm not needed

src/nsrt_learning/segmentation.py
tomsilver: i could imagine that assertion not always holding, and I don't think we need it to?

tomsilver:

src/nsrt_learning/segmentation.py
ronuchit: how is that possible? didn't you create the current_nsrts list to exactly be the states where the preconditions hold?

oh, i think i see. in the process of "executing" this operator, you may go through some states where the preconditions don't hold (but the effects haven't been reached yet). because the GroundAtomTrajectory is at the level of the low-level actions. is that what you mean?

src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,52 @@ def _switch_fn(t: int) -> bool:
...
+    objects = list(traj.states[0])
+    ground_nsrts = {
+        ground_nsrt
+        for nsrt in gt_nsrts
+        for ground_nsrt in utils.all_ground_nsrts(nsrt, objects)
+    }
+    atoms = all_atoms[0]
+    expected_next_atoms = [
+        utils.apply_operator(n, atoms)
+        for n in utils.get_applicable_operators(ground_nsrts, atoms)
```
ronuchit, line 77: can we rename: `expected_next_atoms` -> `all_expected_next_atoms`?

src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,52 @@ def _switch_fn(t: int) -> bool:
...
+    ]
+
+    def _switch_fn(t: int) -> bool:
+        nonlocal expected_next_atoms  # update at each switch point
+        atoms = all_atoms[t + 1]
+        # Check if any of the current NSRT effects hold.
+        for next_atoms in expected_next_atoms:
```
ronuchit, line 84: `for expected_next_atoms in all_expected_next_atoms`?

src/nsrt_learning/segmentation.py
```diff
@@ -46,6 +51,52 @@ def _switch_fn(t: int) -> bool:
...
+    }
+    atoms = all_atoms[0]
+    expected_next_atoms = [
+        utils.apply_operator(n, atoms)
+        for n in utils.get_applicable_operators(ground_nsrts, atoms)
+    ]
+
+    def _switch_fn(t: int) -> bool:
+        nonlocal expected_next_atoms  # update at each switch point
+        atoms = all_atoms[t + 1]
```
ronuchit, line 82: rename to `next_atoms`?

src/nsrt_learning/segmentation.py
ronuchit: or `possible_expected_next_atoms`

ronuchit (approved):
just some renaming suggestions

src/nsrt_learning/segmentation.py
tomsilver: yeah exactly

PR #765, opened 2022-04-22
raise terminal state in stick point

tomsilver:


src/planning.py
```diff
@@ -313,8 +313,16 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
                     option_model.get_next_state_and_num_actions(state, option)
             except EnvironmentFailure as e:
                 can_continue_on = False
-                # Remember only the most recent failure.
-                discovered_failures[cur_idx - 1] = _DiscoveredFailure(e, nsrt)
+                # If there is an environment failure with no offending objects,
```
ronuchit, line 316: is the idea that instead of raising the discovered failure back up to the skeleton search, we want to keep going with backtracking search?

if so, then another option here is to use `sesame_propagate_failures = "after_exhaust"` for this env. because the change you've made is a bit hacky in the sense that if you think of the robot as an object (which it is in many of our envs), then it's valid to claim that the robot is an offending object (since robot.pose is the offending variable), and this check doesn't help

tomsilver: but if we use `sesame_propagate_failures = "after_exhaust"`, wouldn't the planner change the NSRTs after the skeleton is exhausted? I don't want that to happen.

is there a better way to basically introduce "terminal states" into the environment?

tomsilver:

ronuchit (approved):

PR #768, opened 2022-04-22
catch environment failures too in demo data generation

tomsilver:


src/datasets/demo_only.py
```diff
@@ -47,7 +47,8 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task]) -> Dataset:
             # that the goal holds at the end.
             assert task.goal_holds(traj.states[-1]), \
                 "Oracle failed on training task"
-        except (ApproachTimeout, ApproachFailure, AssertionError) as e:
+        except (ApproachTimeout, ApproachFailure, utils.EnvironmentFailure,
+                AssertionError) as e:
```
ronuchit, line 51: are we worried about catching AssertionError here, which is overly specific? it could hide some real issues potentially. it's intended to just catch the `assert task.goal_holds(traj.states[-1]), "Oracle failed on training task"` line right? so maybe we could just make that a separate check afterward, like `if not task.goal_holds(traj.states[-1]): logging.warning("WARNING: Approach failed to solve because oracle failed on training task"); continue`

ronuchit (approved):

PR #770, opened 2022-04-22
env-specific `max_num_steps_option_rollout`

tomsilver:
this seems to help stick point with learned options


ronuchit (approved):

ronuchit:
do we really need this change in light of the degenerate parameter spaces idea? i feel like that's a better way to achieve a similar effect

tomsilver:
I think they're orthogonal. the learned options, even if they're not paramterized, will be much faster to simulate in the option model with this change.

PR #778, opened 2022-04-23
add flag to disable angles in stick point env

tomsilver:
example:
https://user-images.githubusercontent.com/2816502/164981509-be8036f0-2043-422b-8559-d7bee4f711ff.mp4

### DIFF
--- a/src/envs/stick_point.py
+++ b/src/envs/stick_point.py
@@ -237,10 +237,15 @@ class StickPointEnv(BaseEnv):
         assert isinstance(circ, utils.Circle)
         circ.plot(ax, facecolor="red", edgecolor="black")
         # Show the direction that the robot is facing.
-        theta = state.get(robot, "theta")
-        l = 1.5 * self.robot_radius  # arrow length
-        w = 0.1 * self.robot_radius  # arrow width
-        ax.arrow(circ.x, circ.y, l * np.cos(theta), l * np.sin(theta), width=w)
+        if not CFG.stick_point_disable_angles:
+            theta = state.get(robot, "theta")
+            l = 1.5 * self.robot_radius  # arrow length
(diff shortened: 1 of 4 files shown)


tests/approaches/test_oracle_approach.py
```diff
@@ -862,6 +862,27 @@ def test_oracle_approach_stick_point():
         "num_test_tasks": 2,
         "stick_point_num_points_train": [1],
         "stick_point_num_points_test": [2],
+        "stick_point_disable_angles": False,
```
ronuchit, line 865: looks like a lot of duplicated code in this function... let's use the pytest parametrize pattern here? see `tests/approaches/test_gnn_metacontroller_approach.py` for an example of how to use it with multiple arguments

src/envs/stick_point.py
```diff
@@ -111,7 +111,10 @@ def simulate(self, state: State, action: Action) -> State:
         # Actions are normalized to [-1, 1]. Denormalize them here.
         dx = norm_dx * self.max_speed
         dy = norm_dy * self.max_speed
-        dtheta = norm_dtheta * self.max_angular_speed
+        if CFG.stick_point_disable_angles:
+            dtheta = 0.0
+        else:
+            dtheta = norm_dtheta * self.max_angular_speed
         # Update the robot state.
         rx = state.get(self._robot, "x")
```
ronuchit, line 119: in the example video you showed, the arrow looks a little silly / out of place. can we just not show the arrow when `CFG.stick_point_disable_angles`?

ronuchit (approved):
can we also

tests/approaches/test_oracle_approach.py
tomsilver: you're right, but this entire file is a disaster in this respect. i'm going to make an issue to refactor the whole file.

PR #780, opened 2022-04-24
Variance score function for interactive learning

amburger66:
and changes entropy calculation to use log base 2 instead of 10


tomsilver (approved):
Lgtm! With the change of log base, we will probably need to fiddle again with the threshold for the entropy score functions right?

amburger66:
> Lgtm! With the change of log base, we will probably need to fiddle again with the threshold for the entropy score functions right?

yup currently experimenting with that

PR #781, opened 2022-04-24
fix gnn shooting edge case

tomsilver:


tests/approaches/test_gnn_policy_approach.py
```diff
@@ -200,3 +200,5 @@ def _solved_classifier(s, o):
                                         test_task.goal_holds,
                                         max_num_steps=CFG.horizon)
     assert "Option plan exhausted" in str(e)
+    # TODO: test that shooting does not infinitely hang in the case where the
```
ronuchit, line 203: `git grep "MockOptionModel"` might help, and then just the existence of that test would be sufficient (otherwise unit tests would take infinite time)

tomsilver: thanks, yeah I saw that, but can you think of an easy way to get the GNN approach to use the mock option model?

ronuchit: ...it should be clear from the above no? you can see lines like `approach._option_model = _MockOptionModel1(_simulator)  # pylint: disable=protected-access`

tomsilver: thx sry was being dum

ronuchit (approved):
thanks for catching these edge cases! i wonder if this will change our paper results (esp. the "Break early if we have timed out" change)

tomsilver:
if anything, it would make the GNN a bit worse

PR #782, opened 2022-04-24
add `min_perc_data_for_nsrt` flag

tomsilver:
when running experiments that vary the number of training data, it's easier to specify this new flag instead of `min_data_for_nsrt`. but I think we should still leave the old flag since it's useful for testing and is currently being used in a number of @amburger66 's scripts.

### DIFF
--- a/tests/nsrt_learning/test_nsrt_learning_main.py
+++ b/tests/nsrt_learning/test_nsrt_learning_main.py
@@ -14,6 +14,7 @@ def test_nsrt_learning_specific_nsrts():
     """Tests with a specific desired set of NSRTs."""
     utils.reset_config({
         "min_data_for_nsrt": 0,
+        "min_perc_data_for_nsrt": 0,
         "sampler_mlp_classifier_max_itr": 1000,
         "neural_gaus_regressor_max_itr": 1000
     })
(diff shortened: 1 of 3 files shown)


tests/nsrt_learning/test_nsrt_learning_main.py
```diff
@@ -201,6 +201,13 @@ def test_nsrt_learning_specific_nsrts():
                                   action_space,
                                   sampler_learner="random")
     assert len(nsrts) == 0
+    # Test minimum percent of examples parameter
+    utils.update_config({"min_perc_data_for_nsrt": 50})
```
ronuchit, line 205: don't you also have to reset the `min_data_for_nsrt` parameter that was set in the above call to `update_config` since it's not `reset_config`

ronuchit: and maybe explicitly set `min_perc_data_for_nsrt` to 0 up there?

ronuchit (approved):

amburger66 (approved):

tests/nsrt_learning/test_nsrt_learning_main.py
tomsilver: thanks good catch

PR #786, opened 2022-04-25
Interactive learning baseline: Random query policy

amburger66:


tomsilver (approved):
perf!

ronuchit (approved):
lgtm. for your "final" experiments, you'll want to set `interactive_random_query_prob` to the same proportion as your main method uses. that way the only difference between random queries and your main method is the queries themselves, not the frequency of querying. and so any difference in performance will be clearly due to the queries themselves.

PR #791, opened 2022-04-25
try once more to fix flakey tests

### LINKED ISSUE #790: MASTER IS FAILING
due to flakiness in GNN policy tests

tomsilver:
closes #790 (hopefully)


ronuchit (approved):
why does this help?

tomsilver:
the flakiness seems to be due to the timeout being hit before one full pass through the loop happens. with this change, there will always be at least one pass.

PR #797, opened 2022-04-26
refactor GNN base approach to prep for GNN action policy

ronuchit:


src/approaches/gnn_metacontroller_approach.py
```diff
@@ -42,24 +41,40 @@ def __init__(self, initial_predicates: Set[Predicate],
...
-        """In this approach, we learned NSRTs, so we just return the segmented
-        trajectories that NSRT learning returned to us."""
-        return self._segmented_trajs
-
-    def _extract_target_from_data(self, segment: Segment,
-                                  segment_traj: List[Segment],
-                                  ll_traj: LowLevelTrajectory) -> _GroundNSRT:
+    def _generate_data_from_dataset(
+        self, dataset: Dataset
+    ) -> List[Tuple[State, Set[GroundAtom], Set[GroundAtom], _GroundNSRT]]:
```
tomsilver, line 46: I'm surprised that you don't need to use _Output here instead?

tomsilver: I guess mypy understands inheritance here and doesn't care

tomsilver (approved):
Beaut!

src/approaches/gnn_metacontroller_approach.py
ronuchit: it's important to not use output here -- that type is too generic, since we need GroundNSRT-specific fields in other places in this file, so it's important that mypy knows the targets are GroundNSRTs. this all works because the abstract GNNApproach class has the signature:
`class GNNApproach(BaseApproach, Generic[_Output]):` (note that last part)

PR #800, opened 2022-04-26
dont pass in env options to approach when learning options

### LINKED ISSUE #796: don't pass `env.options` into `create_approach` in main.py when doing option_learning
the approaches can currently cheat by looking at `self._initial_options` when option learning is on. we should make this impossible, like we did with predicate invention.

tomsilver:
closes #796

### DIFF
--- a/src/ground_truth_nsrts.py
+++ b/src/ground_truth_nsrts.py
@@ -2172,6 +2172,7 @@ def _get_behavior_gt_nsrts() -> Set[NSRT]:  # pragma: no cover
     op_name_count_pick = itertools.count()
     op_name_count_place = itertools.count()
 
+    assert CFG.option_learner == "no_learning"
     for option in env.options:
         split_name = option.name.split("-")
         base_option_name = split_name[0]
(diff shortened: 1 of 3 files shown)


src/ground_truth_nsrts.py
```diff
@@ -2172,6 +2172,7 @@ def _get_predicate(base_pred_name: str,
     op_name_count_pick = itertools.count()
     op_name_count_place = itertools.count()
 
+    assert CFG.option_learner == "no_learning"
```
ronuchit, line 2175: don't we need a similar assert in `get_options_by_names`? but this would break data collection. maybe we shouldn't have these asserts at all actually?

tests/test_main.py
```diff
@@ -113,6 +113,14 @@ def test_main():
         "--load_experiment_id", "foobar", "--experiment_id", "baz"
     ]
     main()
+    # Run actual main approach, but with option learning.
```
ronuchit, line 116: wdym by "actual main approach"?

src/ground_truth_nsrts.py
tomsilver: good point, i'll remove

tests/test_main.py
tomsilver: I was just copying the language from above in the same file. I can say "NSRT learning" instead

ronuchit (approved):

PR #805, opened 2022-04-26
keep effects for backchaining

ronuchit:
result of pair programming with @NishanthJKumar 

* update score function for `recompute_datastores_from_segments()` to consider keep effects
* algorithm for maintaining possible keep effects during backchaining and finalizing them afterward
* add a test case that shows we need to also keep around the old operator when adding PNADs with keep effects. this is why `_get_pnads_with_keep_effects` returns new PNADs

supercloud results:
```
cf31c0a69fdded87f8e1502d15ac17f650089c72


AGGREGATED DATA OVER SEEDS:
                               NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED    LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID
rnt_backchain                     50.00 (0.00)  50.00 (0.00)   3.00 (0.00)   0.04 (0.00)       88.50 (2.35)     12.51 (0.98)         10
rnt_no_sidepreds                  50.00 (0.00)   4.89 (4.70)   3.00 (0.00)   3.46 (3.51)  1504.36 (1394.92)    81.42 (86.52)          9
rnt_oracle                         0.00 (0.00)  50.00 (0.00)   3.00 (0.00)   0.02 (0.00)       88.50 (2.35)      0.00 (0.00)         10
rnt_painting_backchain            50.00 (0.00)  49.40 (0.80)  18.00 (0.00)   0.74 (0.06)   1822.34 (420.94)     95.58 (7.88)         10
rnt_painting_oracle                0.00 (0.00)  49.40 (0.92)  18.00 (0.00)   1.34 (0.18)  6870.27 (2368.25)      0.00 (0.00)         10
rnt_pred_error                    50.00 (0.00)  45.60 (4.41)   3.00 (0.00)   0.48 (0.33)       91.18 (2.39)    50.95 (74.97)         10
rnt_single_option_backchain       50.00 (0.00)  50.00 (0.00)   3.00 (0.00)   0.03 (0.01)       88.50 (2.35)     23.50 (1.42)         10
rnt_single_option_no_sidepreds    50.00 (0.00)   5.00 (4.59)   3.00 (0.00)   3.48 (3.67)  1457.09 (1336.64)  146.68 (178.69)          9
rnt_single_option_oracle           0.00 (0.00)  50.00 (0.00)   3.00 (0.00)   0.02 (0.00)       88.50 (2.35)      0.00 (0.00)         10
rnt_single_option_pred_error      50.00 (0.00)  43.40 (2.65)   3.00 (0.00)   0.25 (0.19)      67.75 (27.03)   84.23 (137.16)         10
```

### DIFF
--- a/src/ground_truth_nsrts.py
+++ b/src/ground_truth_nsrts.py
@@ -1986,10 +1986,10 @@ def _get_stick_point_gt_nsrts() -> Set[NSRT]:
     """Create ground truth NSRTs for StickPointEnv."""
     robot_type, point_type, stick_type = _get_types_by_names(
         CFG.env, ["robot", "point", "stick"])
-    Touched, InContactRobotPoint, InContactRobotStick, InContactStickPoint, \
-        Grasped, HandEmpty, NoPointInContact = _get_predicates_by_names(
-            CFG.env, ["Touched", "InContactRobotPoint", "InContactRobotStick",
-            "InContactStickPoint", "Grasped", "HandEmpty", "NoPointInContact"])
+    Touched, RobotAbovePoint, StickAbovePoint, \
+        Grasped, HandEmpty, AboveNoPoint = _get_predicates_by_names(
+            CFG.env, ["Touched", "RobotAbovePoint",
+            "StickAbovePoint", "Grasped", "HandEmpty", "AboveNoPoint"])
     RobotTouchPoint, PickStick, StickTouchPoint = _get_options_by_names(
         CFG.env, ["RobotTouchPoint", "PickStick", "StickTouchPoint"])
 
@@ -2003,13 +2003,13 @@ def _get_stick_point_gt_nsrts() -> Set[NSRT]:
     option = RobotTouchPoint
     preconditions = {
         LiftedAtom(HandEmpty, [robot]),
-        LiftedAtom(NoPointInContact, []),
+        LiftedAtom(AboveNoPoint, []),
     }
     add_effects = {
         LiftedAtom(Touched, [point]),
-        LiftedAtom(InContactRobotPoint, [robot, point])
+        LiftedAtom(RobotAbovePoint, [robot, point])
     }
-    delete_effects = {LiftedAtom(NoPointInContact, [])}
+    delete_effects = {LiftedAtom(AboveNoPoint, [])}
     side_predicates: Set[Predicate] = set()
     robot_touch_point_nsrt = NSRT("RobotTouchPointFromNothing", parameters,
                                   preconditions, add_effects, delete_effects,
@@ -2026,13 +2026,13 @@ def _get_stick_point_gt_nsrts() -> Set[NSRT]:
     option = RobotTouchPoint
     preconditions = {
         LiftedAtom(HandEmpty, [robot]),
-        LiftedAtom(InContactRobotPoint, [robot, from_point]),
+        LiftedAtom(RobotAbovePoint, [robot, from_point]),
     }
     add_effects = {
         LiftedAtom(Touched, [point]),
-        LiftedAtom(InContactRobotPoint, [robot, point])
+        LiftedAtom(RobotAbovePoint, [robot, point])
     }
-    delete_effects = {LiftedAtom(InContactRobotPoint, [robot, from_point])}
+    delete_effects = {LiftedAtom(RobotAbovePoint, [robot, from_point])}
     side_predicates = set()
     robot_touch_point_nsrt = NSRT("RobotTouchPointFromPoint", parameters,
                                   preconditions, add_effects, delete_effects,
@@ -2048,11 +2048,10 @@ def _get_stick_point_gt_nsrts() -> Set[NSRT]:
     option = PickStick
     preconditions = {
         LiftedAtom(HandEmpty, [robot]),
-        LiftedAtom(NoPointInContact, []),
+        LiftedAtom(AboveNoPoint, []),
     }
     add_effects = {
         LiftedAtom(Grasped, [robot, stick]),
-        LiftedAtom(InContactRobotStick, [robot, stick])
     }
     delete_effects = {LiftedAtom(HandEmpty, [robot])}
     side_predicates = set()
@@ -2080,16 +2079,15 @@ def _get_stick_point_gt_nsrts() -> Set[NSRT]:
     option = PickStick
     preconditions = {
         LiftedAtom(HandEmpty, [robot]),
-        LiftedAtom(InContactRobotPoint, [robot, point])
+        LiftedAtom(RobotAbovePoint, [robot, point])
     }
     add_effects = {
         LiftedAtom(Grasped, [robot, stick]),
-        LiftedAtom(InContactRobotStick, [robot, stick]),
-        LiftedAtom(NoPointInContact, [])
+        LiftedAtom(AboveNoPoint, [])
     }
     delete_effects = {
         LiftedAtom(HandEmpty, [robot]),
-        LiftedAtom(InContactRobotPoint, [robot, point]),
+        LiftedAtom(RobotAbovePoint, [robot, point]),
     }
     side_predicates = set()
     pick_stick_nsrt = NSRT("PickStickFromPoint", parameters, preconditions,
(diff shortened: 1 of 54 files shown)


tomsilver:
Yayyyy

src/ground_truth_nsrts.py
```diff
@@ -1002,13 +1007,15 @@ def moveto_sampler(state: State, goal: Set[GroundAtom],
         option = MoveToObj
         preconditions = {
             LiftedAtom(GripperOpen, [robot]),
+            LiftedAtom(OnTable, [targetobj]),
```
ronuchit, line 1010: adding this precond shouldn't affect anything, but it helps with clarity -- we only want to allow MoveToObj on objects that are on the table.

src/ground_truth_nsrts.py
```diff
@@ -1029,14 +1036,15 @@ def moveto_sampler(state: State, goal: Set[GroundAtom],
...
+            LiftedAtom(NextTo, [robot, obj]),
+            LiftedAtom(NotOnTable, [obj])
         }
         delete_effects = {
             LiftedAtom(NextToTable, [robot]),
             LiftedAtom(OnTable, [obj])
         }
         # Moving could have us end up NextTo other objects.
-        side_predicates = {NextTo, NextToBox, NextToShelf}
+        side_predicates = {NextTo}
```
ronuchit, line 1047: i feel like we should commit this change to master -- MoveToBox shouldn't have NextToBox or NextToShelf as side predicates right?

src/ground_truth_nsrts.py
```diff
@@ -1055,14 +1063,15 @@ def moveto_sampler(state: State, goal: Set[GroundAtom],
...
```
ronuchit, line 1074: i feel like we should commit this change to master -- MoveToShelf shouldn't have NextToBox or NextToShelf as side predicates right?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -78,6 +78,9 @@ def _learn(self) -> List[PartialNSRTAndDatastore]:
 
         # Finish learning by adding in the delete effects and side predicates.
         all_pnads = self._finish_learning(param_opt_to_nec_pnads)
+
+        # Recompute datastores one final time.
+        self._recompute_datastores_from_segments(all_pnads)
```
ronuchit, line 83: i don't think this is necessary but i got scared since now i'm messing with add effects and preconds in finish_learning()

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -377,7 +391,24 @@ def _finalize_pnad_side_predicates(pnad: PartialNSRTAndDatastore) -> None:
                 side_predicates.add(atom.predicate)
             for atom in next_atoms - segment.final_atoms:
                 side_predicates.add(atom.predicate)
-        pnad.op = pnad.op.copy_with(side_predicates=side_predicates)
```
ronuchit, line 380: if this ever makes it into master, we should rename this function to something like `_finalize_pnad_side_predicates_and_keep_effects`

src/ground_truth_nsrts.py
tomsilver: ditto

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -185,6 +188,17 @@ def _backchain_one_pass(
                     obj_to_var = dict(
                         zip(ground_op.objects, pnad.op.parameters))
 
+                # For any atom in the necessary_image that wasn't in the
+                # necessary_add_effects, add it as a possible keep effect.
+                for atom in necessary_image - necessary_add_effects:
+                    for obj in atom.objects:
+                        if obj not in obj_to_var:
+                            var, = utils.create_new_variables(
```
tomsilver, line 196: this is interesting. so with this, would it be possible for keep effects to expand the set of objects in the scope for an NSRT? if so, I'm surprised that you needed to add NotOnTable, because I thought the motivation for that predicate was to deal with scope issues

tests/approaches/test_oracle_approach.py
```diff
@@ -615,9 +615,6 @@ def test_oracle_approach_repeated_nextto_painting():
         "env": "repeated_nextto_painting",
         "num_test_tasks": 1,
         "painting_num_objs_test": [1],
-        # NOTE: We found hff to make planning for this problem significantly
-        # faster, and thus use it here so that the tests run quickly!
-        "sesame_task_planning_heuristic": "hff"
```
tomsilver, line 620: is this no longer true?

ronuchit: correct, due to the notontable change

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
ronuchit: discussed offline -- with NotOnTable, we are assuming keep effects cannot introduce new variables. this simplifies the logic too. cc @NishanthJKumar

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
ronuchit: update: this is necessary with some changes that are coming up

src/nsrt_learning/strips_learning/clustering_learner.py
```diff
@@ -313,6 +313,8 @@ def _postprocessing_learn_side_predicates(
         pnads = list(path[-1])
         # Because the PNADs have been modified, recompute the datastores.
         self._recompute_datastores_from_segments(pnads)
+        # Filter out PNADs that have an empty datastore.
+        pnads = [pnad for pnad in pnads if pnad.datastore]
```
ronuchit, line 317: fixes some test that was breaking due to the change in score function... i think it's correct anyway

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -379,6 +408,29 @@ def _finalize_pnad_side_predicates(pnad: PartialNSRTAndDatastore) -> None:
...
+        }
+        if not keep_effects:
+            return set()
+        preconditions = pnad.op.preconditions | keep_effects
+        add_effects = pnad.op.add_effects | keep_effects
+        new_pnad_op = pnad.op.copy_with(name=f"{pnad.op.name}-KEEP",
+                                        preconditions=preconditions,
+                                        add_effects=add_effects)
+        new_pnad = PartialNSRTAndDatastore(new_pnad_op, [], pnad.option_spec)
+        return {new_pnad}
```
ronuchit, line 432: right now we only return one new pnad with ALL keep effects included. soon we will change this to generate combinatorially many (@wmcclinton is working on a test case for this)

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -331,3 +332,147 @@ def test_find_unification_and_try_specializing_pnad():
...
     learner.recompute_datastores_from_segments([new_pnad])
     assert len(new_pnad.datastore) == 1
+
+
+def test_keep_effect_data_partitioning():
+    """Test that the BackchainingSTRIPSLearner is able to correctly induce
+    operators with keep effects in a case where a naive procedure that always
+    induces potential keep effects would fail."""
+
+    utils.reset_config({"segmenter": "atom_changes"})
```
ronuchit, line 342: high-level idea:
two demos: [TurnOn, Configure, Run] and [Configure, TurnOn, Run]
TurnOn always just turns on your machine, Configure makes it configured but may turn on/off arbitrary other machines. you are allowed to Configure when the machine is either off or on.
our algorithm will say that MachineOn is a keep effect of Configure, but then this will be harmful to the second demo, where the machine was off when you configured it. so you need to keep around the original (non-KEEP) operator for Configure. see ground truth operators at end of function

tomsilver: add this as a comment?

tomsilver: to clarify,
>you are allowed to Configure when the machine is either off or on.

but only if `MachineConfigurableWhileOff` holds?

tomsilver (approved):
awesome work and exciting results!!

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
ronuchit: yup that's right, that precond we just added to test that the original operator (the Configure without keep effects) needs its preconditions re-induced, since they can get more specific now that you've thrown the keep effect data into the keep effect operator.

PR #806, opened 2022-04-27
add stick holder to stick point env

tomsilver:
```python src/main.py --approach oracle --env stick_point --seed 0 --make_test_videos --video_fps 10```
* https://user-images.githubusercontent.com/2816502/165422693-579694f4-63d4-4e6b-820c-035c263bd26a.mp4
* https://user-images.githubusercontent.com/2816502/165422703-54bc2c83-3831-4e90-83a9-b6a2a53453b6.mp4
* https://user-images.githubusercontent.com/2816502/165422722-c9d1fb84-3f37-431c-a456-4c587dab7b56.mp4

```python src/main.py --approach oracle --env stick_point --seed 0 --make_test_videos --video_fps 10 --stick_point_disable_angles False```
* https://user-images.githubusercontent.com/2816502/165422558-8d66b37c-17b7-4dfd-a009-90415190e95b.mp4
* https://user-images.githubusercontent.com/2816502/165422586-865f458c-7d4e-4250-a1a2-7ce666ebd68c.mp4
* https://user-images.githubusercontent.com/2816502/165422594-bd53bdc6-abef-4f81-a87f-650254ea7759.mp4

```python src/main.py --approach oracle --env stick_point --seed 0 --make_test_videos --video_fps 10 --stick_point_holder_scale 0.5```
* https://user-images.githubusercontent.com/2816502/165422808-f3d90e5b-fe18-4e8b-b0d5-fc37e29209b4.mp4
* https://user-images.githubusercontent.com/2816502/165422839-191eb8a9-b82e-4eeb-b98d-d33b3eaa8fc3.mp4
* (third was a planning failure)

### DIFF
--- a/src/envs/stick_point.py
+++ b/src/envs/stick_point.py
@@ -35,8 +35,12 @@ class StickPointEnv(BaseEnv):
     # Note that the stick_width is the longer dimension.
     stick_width: ClassVar[float] = 3.0
     stick_height: ClassVar[float] = 0.05
+    # Note that the holder width is set in the class because it uses CFG.
+    holder_height: ClassVar[float] = 2.5 * stick_height
     stick_tip_width: ClassVar[float] = 0.05
     init_padding: ClassVar[float] = 0.5  # used to space objects in init states
+    stick_init_lb: ClassVar[float] = 0.6
+    stick_init_ub: ClassVar[float] = 1.6  # start low in the reachable zone
     pick_grasp_tol: ClassVar[float] = 1e-3
 
     def __init__(self) -> None:
@@ -50,6 +54,8 @@ class StickPointEnv(BaseEnv):
         # The (x, y) is the bottom left-hand corner of the stick, and theta
         # is CCW angle in radians, consistent with utils.Rectangle.
         self._stick_type = Type("stick", ["x", "y", "theta", "held"])
+        # Holds the stick up so that it can be grasped by the robot.
+        self._holder_type = Type("holder", ["x", "y", "theta"])
         # Predicates
         self._Touched = Predicate("Touched", [self._point_type],
                                   self._Touched_holds)
@@ -100,6 +106,10 @@ class StickPointEnv(BaseEnv):
         # Static objects (always exist no matter the settings).
         self._robot = Object("robby", self._robot_type)
         self._stick = Object("stick", self._stick_type)
+        self._holder = Object("holder", self._holder_type)
+
+        assert 0 < CFG.stick_point_holder_scale < 1
+        self._holder_width = self.stick_width * CFG.stick_point_holder_scale
 
     @classmethod
     def get_name(cls) -> str:
@@ -122,6 +132,7 @@ class StickPointEnv(BaseEnv):
         new_rx = rx + dx
         new_ry = ry + dy
         new_rtheta = rtheta + dtheta
+
         # The robot cannot leave the reachable zone. If it tries to, raise
         # an EnvironmentFailure, which represents a terminal state.
         rad = self.robot_radius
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tomsilver** [comment]
@ashayathalye I thought some more about changing the reachable zone like we discussed, but I decided to go this way instead, because I felt like this is more realistic. (I'm imagining a robot standing in front of a table, and between different tasks, the reachable zone would be roughly the same, but the holder for the stick could move where ever).

File: src/settings.py [hunk 1]
```diff
@@ -126,6 +126,7 @@ class GlobalSettings:
     stick_point_num_points_train = [1, 2]
     stick_point_num_points_test = [3, 4]
     stick_point_disable_angles = True
+    stick_point_holder_scale = 0.1
```
ronuchit, line 129: is this a fraction of the stick's total width? it doesn't seem like it from the videos, since for the holder_scale=0.5 videos, it's clearly not covering half the stick

src/envs/stick_point.py
```diff
@@ -153,6 +164,17 @@ def simulate(self, state: State, action: Action) -> State:
             # Check if the stick is now held for the first time.
             if state.get(self._stick, "held") <= 0.5 and \
                 stick_rect.intersects(robot_circ):
+                # Check for a collision with the stick holder. The reason that
+                # we only check for a collision here, as opposed to every time
+                # step, is that we imagine the robot moving down in the z
+                # direction to pick up the stick, at which point it may
+                # collide with the stick holder. Otherwise, the robot would
```
ronuchit, line 171: Otherwise -> On other timesteps

src/envs/stick_point.py
```diff
@@ -153,6 +164,17 @@ def simulate(self, state: State, action: Action) -> State:
...
+                # be high enough above the stick holder to avoid collisions.
+                holder_rect = self._object_to_geom(self._holder, state)
+                if robot_circ.intersects(holder_rect):
+                    # Immediately fail in case of collision.
+                    raise utils.EnvironmentFailure("Collided with holder.")
```
ronuchit, line 176: we're still ignoring these env failures at planning time right? i think we should include the holder as an offending object here just for correctness's sake. it won't matter anyway

ronuchit, line 176: if possible, would be good to add a test in test_stick_point.py that explicitly tries to hit this line, like your reachable zone test

src/envs/stick_point.py
```diff
@@ -320,6 +348,47 @@ def _get_tasks(self, num: int, num_point_lst: List[int],
...
+            # Create the holder for the stick, sampling the position so that it
+            # is somewhere along the long dimension of the stick. To make sure
+            # that the problem is solvable, check that if the stick were
+            # grasped at the lowest reachable position, it would still be
+            # able to touch the highest point.
+            max_point_y = max(state_dict[p]["y"] for p in points)
+            necessary_reach = max_point_y - self.rz_y_ub
+            while True:
+                # Allow the stick to start in the middle of the holder.
+                x_offset = rng.uniform(-self._holder_width,
```
ronuchit, line 360: why is it the case that in the videos, the holder can extend off the end of the stick? i would have thought it should be completely contained within the stick.

src/settings.py
ronuchit: copying from another comment bc it addresses this question but is still confusing me -- why is it the case that in the videos, the holder can extend off the end of the stick? i would have thought it should be completely contained within the stick.

src/envs/stick_point.py
```diff
@@ -320,6 +348,47 @@ def _get_tasks(self, num: int, num_point_lst: List[int],
...
+                if x_offset > 2 * self.robot_radius:
+                    break
+                # Case 2: we can grasp the stick above the holder, but we can
+                # still reach the highest point.
+                min_rel_grasp = x_offset + self._holder_width
+                grasp_to_top = self.stick_width - min_rel_grasp
+                if grasp_to_top > necessary_reach:
+                    break
+            # First orient the rectangle at 0 and then rotate it.
+            # Displace the y because the holder is bigger.
```
ronuchit, line 376: i didn't understand this "Displace..." sentence; elaborate?

ronuchit:
remmeber to link to issue

src/settings.py
tomsilver: it is a fraction of the stick's total width. in the holder_scale=0.5 video, the size of the holder is half the size of the stick.

I made it possible for the holder to extend off the stick because I wanted it to be common that the stick cannot just be grasped from the bottom. if I made it such that the holder was always somewhere contained in the stick's length, and I randomized that relative location, then it would be too often possible to grasp the stick from the end.

ronuchit: ah okay that makes a lot of sense

tests/envs/test_stick_point.py
```diff
@@ -228,3 +230,34 @@ def test_stick_point():
...
+    y = (env.rz_y_ub + env.rz_y_lb) / 2
+    state.set(robot, "x", x)
+    state.set(robot, "y", y)
+    state.set(stick, "x", x)
+    state.set(stick, "y", y)
+    state.set(holder, "x", x - (env.holder_height - env.stick_height) / 2)
+    state.set(holder, "y", y)
+    # Press to pick up the stick.
+    action = Action(np.array([0., 0., 0., 1.], dtype=np.float32))
+    try:
```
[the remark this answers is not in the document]
tomsilver: (note that this syntax is safe because coverage would complain if the except wasn't hit)

PR #810, opened 2022-04-27
fix metacontroller when NSRTs are filtered out due to min_data

ronuchit:


src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -73,16 +82,20 @@ def learn_nsrts_from_data(
     _learn_pnad_samplers(pnads, sampler_learner)  # in-place update
 
     # STEP 6: Make, log, and return the NSRTs.
-    nsrts = [pnad.make_nsrt() for pnad in pnads]
+    nsrts = []
+    seg_to_nsrt = {}
+    for pnad in pnads:
+        nsrt = pnad.make_nsrt()
+        nsrts.append(nsrt)
+        for (seg, _) in pnad.datastore:
```
tomsilver, line 90: one day we may also want to include the substitutes in the returns here? but no need for now

tomsilver (approved):
thanks a lot!

PR #811, opened 2022-04-27
add NotOnTable to painting

ronuchit:


src/envs/repeated_nextto_painting.py
```diff
@@ -107,7 +109,11 @@ def _transition_move(self, state: State, action: Action) -> State:
     @property
     def predicates(self) -> Set[Predicate]:
         return super().predicates | {
-            self._NextTo, self._NextToBox, self._NextToShelf, self._NextToTable
+            self._NextTo,
+            self._NextToBox,
+            self._NextToShelf,
+            self._NextToTable,
+            self._NotOnTable,
```
NishanthJKumar, line 116: `NotOnTable` doesn't feel specific to `rnt_painting`, I think we should just make it be part of regular painting as well.

src/envs/repeated_nextto_painting.py
```diff
@@ -172,3 +178,7 @@ def _NextToTable_holds(self, state: State,
                            objects: Sequence[Object]) -> bool:
         robot, = objects
         return self.table_lb < state.get(robot, "pose_y") < self.table_ub
+
+    def _NotOnTable_holds(self, state: State,
+                          objects: Sequence[Object]) -> bool:
+        return not self._OnTable_holds(state, objects)
```
NishanthJKumar, line 184: If you agree with above comment, we should move this into `painting.py`

NishanthJKumar (commented):
Mostly LGTM - I just think that we should make `NotOnTable` part of the regular `painting` env itself; lmk if you agree!

NishanthJKumar (approved):
LGTM!

PR #813, opened 2022-04-27
creates new hard test case for keep effect induction

NishanthJKumar:


tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -326,3 +325,169 @@ def test_find_unification_and_try_specializing_pnad():
...
+        }, run, goal)
+    # Now, run the learner on the two demos and make sure check_harmlessness
+    # is set to True.
+    learner = _MockBackchainingSTRIPSLearner(
+        [traj1, traj2], [task1, task2],
+        set([MachineOn, NotMachineOn, MachineConfigured,
+             MachineRun]), [[segment1_1, segment1_2, segment1_3],
+                            [segment2_1, segment2_2, segment2_3]], True)
+    output_pnads = learner.learn()
+    # There should be exactly 3 output PNADs: 2 for Configuring, and 1 for
```
ronuchit, line 491: your comment says 3 but you have `== 4`

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -326,3 +325,169 @@ def test_find_unification_and_try_specializing_pnad():
...
+    # each of TurningOn and Running.
+    assert len(output_pnads) == 4
```
ronuchit, line 493: can we further verify the str() of the operators is what we would expect, like we do in the other tests? because we want to make sure e.g. that the preconditions are specific enough for both Configuring PNADs

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -326,3 +325,169 @@ def test_find_unification_and_try_specializing_pnad():
...
+            NotMachineOn([m2]),
+            NotMachineOn([m3])
+        }, run, goal)
+    # Now, run the learner on the two demos and make sure check_harmlessness
+    # is set to True.
+    learner = _MockBackchainingSTRIPSLearner(
+        [traj1, traj2], [task1, task2],
+        set([MachineOn, NotMachineOn, MachineConfigured,
+             MachineRun]), [[segment1_1, segment1_2, segment1_3],
+                            [segment2_1, segment2_2, segment2_3]], True)
```
ronuchit, line 489: explicitly write: `verify_harmlessness=True`

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -326,3 +325,169 @@ def test_find_unification_and_try_specializing_pnad():
...
+            NotMachineOn([m2]),
+            NotMachineOn([m3])
+        }, {
+            MachineOn([m1]),
+            MachineConfigured([m1]),
+            MachineRun([m1]),
+            NotMachineOn([m2]),
+            NotMachineOn([m3])
+        }, run, goal)
+    # Now, run the learner on the two demos and make sure check_harmlessness
```
ronuchit, line 483: do you mean `verify_harmlessness`?

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -326,3 +325,169 @@ def test_find_unification_and_try_specializing_pnad():
...
+    # Now create the trajectories, goals and tasks.
+    traj1 = LowLevelTrajectory([
+        all_off_not_configed, m1_off_configed_m2_on, m1_on_configed_m2_on,
+        m1_on_configed_run_m2_on
+    ], [configure_act, turn_on_act, run_act], True, 0)
+    traj2 = LowLevelTrajectory(
+        [m3_on, m1_on_m3_on, m1_on_configed, m1_on_configed_run],
+        [turn_on_act, configure_act, run_act], True, 1)
+    goal = {
+        MachineOn([m1]),
```
ronuchit, line 412: why do we need MachineOn and MachineConfigured in the goal? would MachineRun be sufficient? preferable to keep the goal as compact as possible

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -326,3 +325,169 @@ def test_find_unification_and_try_specializing_pnad():
...
+        MachineConfigured([m1]),
+        MachineRun([m1]),
+    }
+    task1 = Task(all_off_not_configed, goal)
+    task2 = Task(m3_on, goal)
+
+    # Define the 2 demos (each with 3 segments) to backchain over.
```
ronuchit, line 419: instead of defining these segments manually, can we call `segment_trajectory` in segmentation.py on each traj? it should give the same result hopefully...

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
NishanthJKumar: sounds good - i'll try to write the expected results up and include them

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
NishanthJKumar: we don't - good catch

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
NishanthJKumar: damn - i didn't know this method existed lol - will try!

NishanthJKumar:
closed since integration into #805

PR #816, opened 2022-04-27
make verify_harmlessness NOT have a default

### LINKED ISSUE #814: clean up `verify_harmlessness`
it's a mess right now -- it's a default arg that defaults to False, but we actually want tests to be using it...

ronuchit:
* only place it is True within `src/` is `nsrt_learning_main.py`
* but we generally want it to be True for unit tests as well, so that errors don't get swept under the rug

closes #814


NishanthJKumar (approved):
LGTM!

PR #817, opened 2022-04-27
stick point env predicate changes

tomsilver:
* rename all `InContact` to `Above`
* remove redundant predicate between robot and stick (redundant with `Grasped`)

### DIFF
--- a/src/envs/stick_point.py
+++ b/src/envs/stick_point.py
@@ -59,22 +59,19 @@ class StickPointEnv(BaseEnv):
         # Predicates
         self._Touched = Predicate("Touched", [self._point_type],
                                   self._Touched_holds)
-        self._InContactStickPoint = Predicate(
-            "InContactStickPoint", [self._stick_type, self._point_type],
-            self._InContact_holds)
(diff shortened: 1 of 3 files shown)


src/envs/stick_point.py
```diff
@@ -59,22 +59,19 @@ def __init__(self) -> None:
...
+                                          [self._robot_type, self._point_type],
+                                          self._Above_holds)
         self._Grasped = Predicate("Grasped",
                                   [self._robot_type, self._stick_type],
                                   self._Grasped_holds)
         self._HandEmpty = Predicate("HandEmpty", [self._robot_type],
                                     self._HandEmpty_holds)
-        self._NoPointInContact = Predicate("NoPointInContact", [],
-                                           self._NoPointInContact_holds)
+        self._AboveNoPoint = Predicate("AboveNoPoint", [],
```
ronuchit, line 73: what is this referring to? the stick or the robot?

ronuchit (approved):

src/envs/stick_point.py
tomsilver: either one

PR #820, opened 2022-04-28
disable harmlessness check for oracle method in option script

tomsilver:
without this, the harmlessness check crashes in the rare cases when the stick presses two points at once. I don't fully understand why this seems to happen only for oracle strips learning, but that seems to be the case.


ronuchit (approved):
worth understanding this more. why is the stick able to press two points at once? arent the points pretty far apart?

tomsilver:
yeah I'll dig in tomorrow. and yeah they are far apart radius wise, but can still line up vertically in rare cases.

ronuchit:
fixed in #823

PR #823, opened 2022-04-28
fix stick_button to not allow pressing with both robot and stick simultaneously

ronuchit:
command that failed before but works now:
`python src/main.py --env stick_button --segmenter contacts --num_train_tasks 500 --seed 0 --approach nsrt_learning --strips_learner oracle`


tomsilver (approved):
lgtm thanks, I'll be curious to see whether this makes `min_perc_data_for_nsrt` unneeded too

ronuchit:
can you check that @tomsilver? what exactly changed to make that not be needed anymore? i thought we needed that bc you could move above two things at once

tomsilver:
you're right, thanks.

PR #827, opened 2022-04-28
increase timeout for stick button nightly

tomsilver:


ronuchit (approved):
did you check that min perc is still necessary?

PR #830, opened 2022-04-28
refactor test_oracle_approach.py and remove CoverMultistepOptionsFixedTasks, CoverMultistepOptionsHolding

### LINKED ISSUE #784: refactor `tests/approaches/test_oracle_approach.py` to remove redundant code
use parameterized tests

ronuchit:
uses `pytest.mark.parametrize` to significantly shorten test_oracle_approach.py

removes two old unneeded cover environments

closes #784


tests/approaches/test_oracle_approach.py
```diff
@@ -26,306 +25,58 @@
...
-        # Test that a repeated random action fails.
-        assert not policy_solves_task(lambda s: random_action, task,
-                                      env.simulate)
-
-
-def test_oracle_approach_cover_multistep_options():
-    """Tests for OracleApproach class with CoverMultistepOptions."""
-    utils.reset_config({
-        "env": "cover_multistep_options",
+ENV_NAME_AND_CLS = [
```
tomsilver, line 31: can't we get all these names from the class themselves? (with `get_name`)

tests/approaches/test_oracle_approach.py
```diff
@@ -336,25 +87,86 @@ def test_oracle_approach_cover_multistep_options():
...
+@pytest.mark.parametrize("env_name,env_cls", ENV_NAME_AND_CLS)
+def test_oracle_approach(env_name, env_cls):
+    """Tests for OracleApproach class with all environments."""
+    for extra_args in EXTRA_ARGS_ORACLE_APPROACH[env_name]:
+        common_cfg_dict = {
+            "env": env_name,
+            "num_train_tasks": 2,
+            "num_test_tasks": 2,
+        }
+        cfg_dict = dict(common_cfg_dict, **extra_args)
```
tomsilver, line 149: this is fine but if you prefer, could also do:
```
cfg_dict = {
    "env": env_name,
    "num_train_tasks": 2,
    "num_test_tasks": 2,
    **extra_args
}
```

tests/envs/test_base_env.py
```diff
@@ -4,34 +4,13 @@
...
-            "cover_multistep_options_fixed_tasks",
-            "cover_multistep_options_holding",
-            "pybullet_cover",
-            "pybullet_blocks",
-            "pddl_blocks_fixed_tasks",
-            "pddl_blocks_procedural_tasks",
-            "touch_point",
-            "stick_button",
-    ]:
+    for name, _ in ENV_NAME_AND_CLS:
```
tomsilver, line 13: is there a reason to not make this one parameterized as well?

tomsilver (approved):
thanks!! so much better ow

tests/approaches/test_oracle_approach.py
ronuchit: i'd rather not because i want people who make a new env to be aware that it should be inserted into this test file. because you might want to change some parameters with the `EXTRA_ARGS_ORACLE_APPROACH` to make the oracle not be super slow

ronuchit: oh i see what you mean... eh whatever

tests/envs/test_base_env.py
ronuchit: too lazy to pull the last few lines (`with pytest.raises(NotImplementedError): create_new_env("Not a real env")`) into a new function

PR #832, opened 2022-04-29
stick point -> stick button

ronuchit:


tomsilver (approved):
Thanks, sorry thought I got 'em all

PR #836, opened 2022-04-29
move plot scripts to new subfolder

### LINKED ISSUE #834: move plotting scripts into subfolder
def too annoying to make plotting code modular, but at least we can do this...

ronuchit:
closes #834


scripts/plotting/create_numdemos_plots.py
```diff
@@ -1,4 +1,4 @@
-"""Create plots for online learning."""
+"""Create plots for learning from varying numbers of demonstrations."""
```
tomsilver, line 1: can you add an underscore between num and demos in the filename?

tomsilver (approved):
thank you!

PR #837, opened 2022-04-29
catch environment failures in run policy

### LINKED ISSUE #835: handle case where an environment failure is raised during `run_policy` in `_generate_interaction_results`
an annoyance is that we can't just add `utils.EnvironmentFailure` to `exceptions_to_break_on` because those exceptions are only caught when they get raised in `policy(state)`, not when they're in `env.step(act)`

tomsilver:
closes #835

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -839,7 +839,11 @@ def run_policy(
     Terminates when any of these conditions hold:
     (1) the termination_function returns True
     (2) max_num_steps is reached
-    (3) policy() raises any exception of type in exceptions_to_break_on
+    (3) policy() or step() raise an exception of type in exceptions_to_break_on
+
+    Note that in the case where the exception is raised in step, we exclude the
+    last action from the returned trajectory to maintain the invariant that
+    the trajectory states are always one more than the actions.
     """
     state = env.reset(train_or_test, task_idx)
     states = [state]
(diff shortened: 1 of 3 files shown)


src/utils.py
```diff
@@ -839,7 +839,11 @@ def run_policy(
     Terminates when any of these conditions hold:
     (1) the termination_function returns True
     (2) max_num_steps is reached
-    (3) policy() raises any exception of type in exceptions_to_break_on
+    (3) policy() or step() raise an exception of type in exceptions_to_break_on
+
+    Note that in the case where the exception is raised in step, we exclude the
+    last action from the returned trajectory to maintain the invariant that
+    the trajectory states are always one more than the actions.
```
ronuchit, line 846: are of length one greater than the actions*

src/utils.py
```diff
@@ -895,7 +899,11 @@ def run_policy_with_simulator(
...
```
ronuchit, line 906: copy-paste change above

src/utils.py
```diff
@@ -839,7 +839,11 @@ def run_policy(
     Terminates when any of these conditions hold:
     (1) the termination_function returns True
     (2) max_num_steps is reached
-    (3) policy() raises any exception of type in exceptions_to_break_on
+    (3) policy() or step() raise an exception of type in exceptions_to_break_on
```
ronuchit, line 842: raises*

src/utils.py
```diff
@@ -895,7 +899,11 @@ def run_policy_with_simulator(
...
```
ronuchit, line 902: raises*

ronuchit (approved):

src/utils.py
tomsilver: You want this change even though the line will break?

ronuchit: no

PR #842, opened 2022-05-01
a few minor pybullet improvements

### LINKED ISSUE #841: pybullet: minor QOL improvements
* make inverse kinematics public and mirror forward kinematics
* make a type alias for joint states
* be consistent about terminology for joint states

tomsilver:
closes #841


ronuchit (approved):
nice qol change. but are you sure we don't want this to be called JointsState or JointStates instead? it's not the state of a single joint

ronuchit (changes requested):
oh, just realized you forgot to change `action_arr` in `set_motors()` -- maybe you should grep for `action_arr` more generally? btw, this will need to change anyway in light of #849

src/envs/pybullet_cover.py
```diff
@@ -228,7 +228,7 @@ def step(self, action: Action) -> State:
         # constraint before we can call PyBullet.
         # Use self._pybullet_robot2 to run forward kinematics, since that
         # method shouldn't be run on the client that is doing simulation.
-        _, ry, rz = self._pybullet_robot2.forward_kinematics(action.arr)
+        _, ry, rz = self._pybullet_robot2.forward_kinematics(list(action.arr))
```
ronuchit, line 231: should we use `tolist()` instead? from https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html:
"For a 1D array, a.tolist() is almost the same as list(a), except that tolist changes numpy scalars to Python scalars"

src/envs/pybullet_env.py
```diff
@@ -235,7 +235,7 @@ def render(
 
     def step(self, action: Action) -> State:
         # Send the action to the robot.
-        self._pybullet_robot.set_motors(action.arr)
+        self._pybullet_robot.set_motors(list(action.arr))
```
ronuchit, line 238: same q about tolist

src/envs/pybullet_robots.py
```diff
@@ -138,29 +138,36 @@ def get_state(self) -> Array:
...
+        """Update the motors to move toward the given joints state."""
         raise NotImplementedError("Override me!")
 
     @abc.abstractmethod
-    def forward_kinematics(self, action_arr: Array) -> Pose3D:
-        """Compute the end effector pose that would result from executing the
-        given action in PyBullet. The action_arr is an array of desired arm
-        joint values.
+    def forward_kinematics(self, joints_state: JointsState) -> Pose3D:
+        """Compute the end effector pose that if the robot arm joints state
```
ronuchit, line 152: that would result if*

src/envs/pybullet_robots.py
```diff
@@ -138,29 +138,36 @@ def get_state(self) -> Array:
...
 
         WARNING: This method will make use of resetJointState(), and so it
         should NOT be used during simulation.
         """
         raise NotImplementedError("Override me!")
 
+    @abc.abstractmethod
+    def inverse_kinematics(self, end_effector_pose: Pose3D,
+                           validate: bool) -> JointsState:
+        """Compute a joints state from an end effector pose.
```
ronuchit, line 163: from a target end effector pose*

src/envs/pybullet_robots.py
```diff
@@ -138,29 +138,36 @@ def get_state(self) -> Array:
...
+
+        If validate is True, guarantee that the returned joints state
```
ronuchit, line 165: add a word of caution about validate=True overriding physics and so it shouldn't be used with stepsimulation

src/envs/pybullet_robots.py
```diff
@@ -580,12 +586,11 @@ def inverse_kinematics(
     # Reset the joint states to their initial values to avoid modifying the
     # PyBullet internal state.
     if validate:
-        for joint, joint_state in zip(free_joints, initial_joint_states):
-            position, velocity, _, _ = joint_state
+        for joint, (pos, vel, _, _) in zip(free_joints, initial_joints_states):
```
ronuchit, line 589: just a readability change i guess

ronuchit (approved):

PR #843, opened 2022-05-02
BiRRT

tomsilver:


src/utils.py
```diff
@@ -1394,6 +1394,122 @@ def run_hill_climbing(
...
+                    path1, path2 = path2, path1
+                assert path1[0] == root1
+                path = path1[:-1] + path2[::-1]
+                return [node.data for node in path]
+        return None
+
+    def _smooth_path(self, path: List[_S]) -> List[_S]:
+        # This is a tricky case to cover and may not ever happen in practice,
+        # but leaving it here just in case.
+        if len(path) <= 2:  # pragma: no cover
```
ronuchit, line 1478: i don't get how it can happen in practice... if the path length is 0 or 1, wouldn't `try_direct_path` have worked?

should we just make this an assert?

src/utils.py
```diff
@@ -1394,6 +1394,122 @@ def run_hill_climbing(
     return states, actions, heuristics
 
 
+class BiRRT(Generic[_S]):
```
ronuchit, line 1397: thank you for bringing this in and annotating it!

src/envs/pybullet_robots.py
```diff
@@ -363,12 +382,7 @@ def set_motors(self, action_arr: Array) -> None:
                                     physicsClientId=self._physics_client_id)
 
     def forward_kinematics(self, joint_state: JointState) -> Pose3D:
-        assert len(joint_state) == len(self._arm_joints)
```
ronuchit, line 366: check for any other places where we could call set_joints too?

src/envs/pybullet_robots.py
```diff
@@ -606,3 +620,53 @@ def inverse_kinematics(
         joint_vals.append(joint_val)
 
     return joint_vals
+
+
+def run_motion_planning(
+        robot: _SingleArmPyBulletRobot, initial_state: JointState,
+        target_state: JointState, collision_bodies: Collection[int], seed: int,
+        physics_client_id: int) -> Optional[Sequence[JointState]]:
+    """Run BiRRT to find a collision-free sequence of joint states.
```
ronuchit, line 629: the idea is to use this with pybullet_robot2?

tests/envs/test_pybullet_robots.py
```diff
@@ -213,3 +214,109 @@ def test_create_single_arm_pybullet_robot():
                                          ee_orn, move_to_pose_tol,
                                          max_vel_norm, grasp_tol,
                                          physics_client_id)
+
+
+def test_run_motion_planning():
```
ronuchit, line 219: awesome comprehensive testing thank you!

tests/envs/test_pybullet_robots.py
```diff
@@ -213,3 +214,109 @@ def test_create_single_arm_pybullet_robot():
...
+    joint_target = robot.inverse_kinematics(ee_target, validate=True)
+    path = run_motion_planning(robot,
+                               joint_initial,
+                               joint_target,
+                               collision_bodies=set(),
+                               seed=seed,
+                               physics_client_id=physics_client_id)
+    assert np.allclose(path[0], joint_initial)
+    assert np.allclose(path[-1], joint_target)
+    # Should fail because the target collides with the table.
```
ronuchit, line 256: what about if the start is in collision?

tests/envs/test_pybullet_robots.py
```diff
@@ -213,3 +214,109 @@ def test_create_single_arm_pybullet_robot():
                                          ee_orn, move_to_pose_tol,
                                          max_vel_norm, grasp_tol,
                                          physics_client_id)
+
+
+def test_run_motion_planning():
+    """Tests for run_motion_planning()."""
```
ronuchit, line 220: how's the speed of this test?

src/envs/pybullet_robots.py
tomsilver: yeah that's the idea. actually wiring this together is going to be a bit annoying, because we're going to have to have a copy of all of the collision bodies too. if you have any thoughts about the best way to do this lmk...

ronuchit: yeah i feel like it's cleanest to just have a second carbon copy of the env that is updated forcibly on each timestep? bc we need to be robust to robot2 doing stupid things during planning and therefore messing up the object states

tests/envs/test_pybullet_robots.py
tomsilver: 0.55 seconds

src/envs/pybullet_robots.py
```diff
@@ -142,6 +143,16 @@ def get_joints(self) -> JointsState:
         """Get the joints state from the current PyBullet state."""
         raise NotImplementedError("Override me!")
 
+    @abc.abstractmethod
+    def set_joints(self, joint_state: Sequence[float]) -> None:
```
ronuchit, line 147: should this be JointState?

src/envs/pybullet_robots.py
```diff
@@ -351,6 +353,15 @@ def get_joints(self) -> JointsState:
             joints_state.append(joint_val)
         return joints_state
 
+    def set_joints(self, joint_state: Sequence[float]) -> None:
```
ronuchit, line 356: should this be JointState?

ronuchit (approved):
couple more comments

src/envs/pybullet_robots.py
```diff
@@ -351,6 +353,15 @@ def get_joints(self) -> JointsState:
...
+        assert len(joint_state) == len(self._arm_joints)
+        for joint_id, joint_val in zip(self._arm_joints, joint_state):
+            p.resetJointState(self._fetch_id,
+                              joint_id,
+                              targetValue=joint_val,
+                              targetVelocity=0,
+                              physicsClientId=self._physics_client_id)
+
     def set_motors(self, joints_state: JointsState) -> None:
```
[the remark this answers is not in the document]
tomsilver: they're included in the joints state

PR #845, opened 2022-05-02
fix for cover multistep option action space limits

tomsilver:
before this change, if one were to change `cover_multistep_action_limits` from the default infs to small numbers, then oracle fails because the actions returned by the oracle policy would violate the gripper dimension of the action space.


ronuchit (commented):
i don't understand. why can dgrip be anything? surely there must be some bounds on it, e.g. `|closed_gripper - open_gripper|`?

tomsilver:
@ronuchit good point, I changed that

ronuchit (approved):

PR #851, opened 2022-05-02
Interactive learning supercloud script

amburger66:
contains main approach and all baselines


scripts/supercloud/run_random_query_experiments.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
tomsilver, line 1: so this is the script that you used to determine the 0.03 threshold, but we're not gonna use this script regularly going forward right? (this is fine and good to include still, just checking)

src/approaches/interactive_learning_approach.py
```diff
@@ -391,7 +391,7 @@ def _query_policy(s: State) -> Optional[GroundAtomsHoldQuery]:
             atoms_to_query = set()
             for atom in ground_atoms:
                 score = self._score_atom_set({atom}, s)
-                if score > CFG.interactive_score_threshold:
+                if score > float(CFG.interactive_score_threshold):
```
tomsilver, line 394: why is this needed?

ronuchit, line 394: do we have a bug in our type checking for the CFG?

amburger66: before i didn't know how to handle floats in bash but yeah this shouldn't be necessary

amburger66: nah this should be unnecessary, will take it out

scripts/supercloud/run_random_query_experiments.sh
amburger66: correct. also i verified that we can calculate the appropriate query probability based off of the environment and how many queries the main approach makes!

tomsilver (approved):
good stuff!

PR #854, opened 2022-05-02
remove the non-learned-equivalent options from cover multistep options

tomsilver:
hooray for deleting code. we haven't been using the options I'm deleting for a while.


ronuchit (approved):
great! make sure to `git grep "LearnedEquiv"`

PR #855, opened 2022-05-02
changes to `cover_multistep_options` to make option learning work

### LINKED ISSUE #846: cover multistep options: change collision handling
to make this environment less sensitive to tiny values, change the way that collisions are handled.

* if a collision between the floor and block or hand occurs, don't noop; instead, have the block or hand change to be exactly level with the floor.
* if a collision between the hand an the top of the block occurs (as is often the case with picking), similarly have the hand change be exactly level with the top of the block, instead of nooping.

### LINKED ISSUE #847: cover multistep options: make picks and places snap
to make the test distribution more closely align with the train distribution, snap the hand/block to a constant vertical value whenever a pick or place happens.

### LINKED ISSUE #857: get option learning working with cover_multistep_options

tomsilver:
closes #846
closes #847
closes #857

summary:
* prevents the robot from ever going below the top of the blocks, or above a max y position (`self.initial_robot_y`)
* if the robot is close enough (as defined by `self.snap_tol`) to picking or placing a block, snaps automatically into that pick or place
* removes robot-block collision checking and slightly simplifies some block-block collision checking
* simplifies the oracle option policies by trying to pick or drop as soon as vertically aligned (the environment dictates that the pick/place will only occur when the robot gets close enough the block/floor)
* makes the hand regions slightly smaller to make nonparameterized approaches worse

### DIFF
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -603,10 +603,10 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
                 hw, hh = state.get(held_block, "width"), \
                          state.get(held_block, "height")
 
-        # Prevent the robot from going below the floor.
-        y_floor = 0
-        if y + dy < y_floor:
-            dy = y_floor - y
+        # Prevent the robot from going below the top of the blocks.
+        y_min_robot = self.block_height
+        if y + dy < y_min_robot:
+            dy = y_min_robot - y
         # If the robot is close to the initial robot y position and moving up,
         # snap it into place vertically.
         if abs(y + dy - self.initial_robot_y) < self.snap_tol and dy > 0:
@@ -618,6 +618,7 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
         if held_block is not None and dy < 0:
             held_block_bottom = hy - hh
             # Always snap if below the floor.
+            y_floor = 0
             if held_block_bottom + dy < y_floor or \
                 abs(held_block_bottom + dy - y_floor) < self.snap_tol:
                 dy = y_floor - held_block_bottom
--- a/src/envs/cover.py
+++ b/src/envs/cover.py
@@ -498,17 +498,14 @@ class CoverMultistepOptions(CoverEnvTypedOptions):
     when the robot is sufficiently close to the block in the y-direction.
     Placing is allowed anywhere. Collisions are handled in simulate().
     """
-    grasp_height_tol: ClassVar[float] = 1e-2
     grasp_thresh: ClassVar[float] = 0.0
     initial_block_y: ClassVar[float] = 0.1
     block_height: ClassVar[float] = 0.1
     target_height: ClassVar[float] = 0.1  # Only for rendering purposes.
-    placing_height: ClassVar[
-        float] = 0.1  # A block's base must be below this to be placed.
     initial_robot_y: ClassVar[float] = 0.4
-    collision_threshold: ClassVar[float] = 1e-5
     grip_lb: ClassVar[float] = -1.0
     grip_ub: ClassVar[float] = 1.0
+    snap_tol: ClassVar[float] = 1e-2
 
     def __init__(self) -> None:
         super().__init__()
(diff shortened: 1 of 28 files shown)


ronuchit:
so what's the plan with this PR now? @tomsilver not necessary?

tomsilver:
nah i'm sure that this will be necessary in some form. my plan is to try to add back a few of the things i removed, run again, and eventually merge

ronuchit:
can you include a summary of the changes in the PR description, in preparation for reviewing?

tomsilver:
@ronuchit done

src/envs/cover.py
```diff
@@ -625,48 +622,60 @@ def simulate(self, state: State, action: Action) -> State:
                         y2) in zip(held_rect.vertices, next_held_rect.vertices)
                 ]
 
-        # Ensure neither the gripper nor the possible held block go below the
-        # y-axis.
-        if y + dy < 0 - self.collision_threshold:
-            return state.copy()
+        # Prevent the robot from going below the top of the blocks.
```
ronuchit, line 625: with all these snapping hacks, are we giving up hope that option learning could work in pybullet?

src/envs/cover.py
```diff
@@ -1019,26 +1014,21 @@ def _Pick_policy(
...
         lb, ub = CFG.cover_multistep_action_limits
-        # If we're already above the object and prepared to pick,
-        # then execute the pick (turn up the magnet).
-        if at_desired_x and at_desired_y:
-            return Action(np.array([0., 0., 1.0], dtype=np.float32))
-        # If we're above the object but not yet close enough, move down.
+        # If we're above the object, move down and turn on the gripper.
         if at_desired_x:
             delta_y = np.clip(desired_y - y, lb, ub)
-            return Action(np.array([0., delta_y, -1.0], dtype=np.float32))
```
ronuchit, line 1032: why are all these -1/1's being flipped in the last dimension of the action? did you reverse the meaning of the dgrip dimension?

tests/approaches/test_oracle_approach.py
```diff
@@ -137,12 +139,17 @@ def _policy_solves_task(policy, task, simulator):
 def test_oracle_approach(env_name, env_cls):
     """Tests for OracleApproach class with all environments."""
     for extra_args in EXTRA_ARGS_ORACLE_APPROACH[env_name]:
-        utils.reset_config({
+        args = {
```
ronuchit, line 142: would it be cleaner to do this by doing:
```
args_dict = {
            "env": env_name,
            "num_train_tasks": 2,
            "num_test_tasks": 2,
}
args_dict.update(extra_args)  # note: extra_args can override the default values in args_dict
utils.reset_config(args_dict)
```
? is this less clear than what you have?

tests/test_teacher.py
```diff
@@ -128,8 +128,8 @@ def test_PathToStateQuery():
     assert isinstance(response, PathToStateResponse)
     assert response.query is query
     assert isinstance(response.teacher_traj, LowLevelTrajectory)
```
ronuchit, line 130: how did the traj lengths change?

ronuchit (approved):

src/envs/cover.py
tomsilver: hope is diminished, but some aspects of pybullet are easier than what we had here before, e.g., there are natural limits on the robot position (kinematics and table collisions)

tests/test_teacher.py
tomsilver: due to snapping, one fewer step is required to pick/place

src/envs/cover.py
tomsilver: no reversal, I just changed it because I figured it out would be easier to imitate the pick policy if this dimension is constant for the entire duration of picking, and i realized it doesn't otherwise change the result because you won't actually pick until you're close enough

tomsilver: (before, the expert would wait until it was in position to turn the gripping on)

tests/approaches/test_oracle_approach.py
tomsilver: I think mine is slightly more clear but I don't feel strongly

tomsilver: I'm just gonna leave it

PR #860, opened 2022-05-03
exclude static features from learned option param spaces

### LINKED ISSUE #856: option learning: remove static features from option parameter space

tomsilver:
closes #856 

I'm going to test this on supercloud with stick button. I would expect that the results for the parameterized methods stay the same or improve. if they don't improve, I can think of two possibilities: there's a bug in this PR, or the `_create_absolute_option_param` function is significantly slow, leading to more timeouts.

### DIFF
--- a/src/nsrt_learning/option_learning.py
+++ b/src/nsrt_learning/option_learning.py
@@ -244,9 +244,11 @@ class _LearnedNeuralParameterizedOption(ParameterizedOption):
       below), all concatenated into a 1D vector, and outputs an action.
 
     The continuous parameters of the option are the main thing to note: they
-    correspond to a desired change in state for a subset of the option objects.
-    The objects that are changing correspond to changing_parameters; all other
-    objects are assumed to have no change in their state.
+    correspond to a desired change in state for a subset of the option objects
+    and a subset of those object features. The objects that are changing are
+    changing_var_order, and also the keys of changing_var_to_feat. The feature
+    indices that are changing are the values in changing_var_to_feat. All other
+    objects and features are assumed to have no change in their state.
 
     Note that the option terminal, which corresponds to the operator effects,
     already describes how we want the objects to change. But these effects only
@@ -279,7 +281,8 @@ class _LearnedNeuralParameterizedOption(ParameterizedOption):
                  name: str,
                  operator: STRIPSOperator,
                  regressor: Regressor,
-                 changing_var_to_feat: Dict[Variable, Set[int]],
+                 changing_var_to_feat: Dict[Variable, List[int]],
+                 changing_var_order: List[Variable],
                  action_space: Box,
                  is_parameterized: bool = True) -> None:
         assert set(changing_var_to_feat).issubset(set(operator.parameters))
@@ -296,9 +299,7 @@ class _LearnedNeuralParameterizedOption(ParameterizedOption):
         self._operator = operator
         self._regressor = regressor
         self._changing_var_to_feat = changing_var_to_feat
-        self._changing_vars = [
-            v for v in operator.parameters if v in changing_var_to_feat
-        ]
+        self._changing_var_order = changing_var_order
         self._action_space = action_space
         self._is_parameterized = is_parameterized
         super().__init__(name,
@@ -317,7 +318,7 @@ class _LearnedNeuralParameterizedOption(ParameterizedOption):
             memory["params"] = params  # store for sanity checking in policy
             var_to_obj = dict(zip(self._operator.parameters, objects))
             state_params = _create_absolute_option_param(
-                state, self._changing_var_to_feat, self._changing_vars,
+                state, self._changing_var_to_feat, self._changing_var_order,
                 var_to_obj)
             memory["absolute_params"] = state_params + params
         # Check if initiable based on preconditions.
@@ -332,7 +333,7 @@ class _LearnedNeuralParameterizedOption(ParameterizedOption):
             assert np.allclose(params, memory["params"])
             var_to_obj = dict(zip(self._operator.parameters, objects))
             state_params = _create_absolute_option_param(
-                state, self._changing_var_to_feat, self._changing_vars,
+                state, self._changing_var_to_feat, self._changing_var_order,
                 var_to_obj)
             relative_goal_vec = memory["absolute_params"] - state_params
         else:
@@ -420,11 +421,8 @@ class _BehaviorCloningOptionLearner(_OptionLearnerBase):
             changing_var_to_feat = self._get_changing_features(datastore)
             # Just to avoid confusion, we will insist that the order of the
             # changing parameters is consistent with the order of the original.
-            changing_parameters = sorted(changing_var_to_feat,
-                                         key=op.parameters.index)
-            assert changing_parameters == [
-                v for v in op.parameters if v in changing_parameters
-            ]
+            changing_var_order = sorted(changing_var_to_feat,
+                                        key=op.parameters.index)
             for segment, var_to_obj in datastore:
                 all_objects_in_operator = [
                     var_to_obj[v] for v in op.parameters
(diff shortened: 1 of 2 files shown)


src/nsrt_learning/option_learning.py
```diff
@@ -546,3 +566,18 @@ def _create_regressor(self) -> Regressor:
...
+                                  changing_var_to_feat: Dict[Variable,
+                                                             Set[int]],
+                                  var_order: Sequence[Variable],
+                                  var_to_obj: VarToObjSub) -> Array:
+
+    vec = []
+    for v in var_order:
+        obj = var_to_obj[v]
+        obj_vec = state[obj]
+        for idx in sorted(changing_var_to_feat[v]):
```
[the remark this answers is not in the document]
ronuchit: (i actually would prefer this)

ronuchit (commented):
looks reasonable. how are the results?

src/nsrt_learning/option_learning.py
```diff
@@ -298,7 +295,10 @@ def __init__(self,
             params_space = Box(0, 1, (0, ), dtype=np.float32)
         self._operator = operator
         self._regressor = regressor
-        self._changing_parameter_idxs = changing_parameter_idxs
+        self._changing_var_to_feat = changing_var_to_feat
+        self._changing_vars = [
+            v for v in operator.parameters if v in changing_var_to_feat
```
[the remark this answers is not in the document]
ronuchit: after reading the rest of the PR, coming back here... i think we can do
`changing_parameters = sorted(changing_var_to_feat, key=op.parameters.index)` like you do elsewhere?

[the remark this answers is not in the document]
ronuchit: oh but also, why do we need to compute changing_parameters elsewhere? we can always just use this self._changing_vars right?

ronuchit:
update docstring of the LearnedNeuralParameterizedOption class, esp the part that says
The objects that are changing correspond to changing_parameters; all other
    objects are assumed to have no change in their state.

src/nsrt_learning/option_learning.py
```diff
@@ -546,3 +566,18 @@ def _create_regressor(self) -> Regressor:
...
             grid_num_ticks_per_dim=num_ticks)
+
+
+def _create_absolute_option_param(state: State,
+                                  changing_var_to_feat: Dict[Variable,
+                                                             Set[int]],
+                                  var_order: Sequence[Variable],
+                                  var_to_obj: VarToObjSub) -> Array:
+
+    vec = []
```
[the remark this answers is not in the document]
tomsilver: no, I didn't notice any slowdown yet

src/nsrt_learning/option_learning.py
```diff
@@ -412,31 +415,32 @@ def learn_option_specs(self, strips_ops: List[STRIPSOperator],
...
-                    absolute_params = final_state.vec(changing_objects)
-                    option_param = absolute_params - initial_state.vec(
-                        changing_objects)
+                    init_param = _create_absolute_option_param(
+                        init_state, changing_var_to_feat, changing_parameters,
+                        var_to_obj)
+                    final_param = _create_absolute_option_param(
+                        final_state, changing_var_to_feat, changing_parameters,
+                        var_to_obj)
+                    option_param = final_param - init_param
```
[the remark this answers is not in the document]
tomsilver: yeah

src/nsrt_learning/option_learning.py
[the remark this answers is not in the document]
tomsilver: I tried to address this, lmk what you think

tomsilver:
>looks reasonable. how are the results?

I'll update the PR when they come in, it will be several hours

tomsilver:
oops did not mean to close

src/nsrt_learning/option_learning.py
```diff
@@ -244,9 +244,11 @@ class _LearnedNeuralParameterizedOption(ParameterizedOption):
       below), all concatenated into a 1D vector, and outputs an action.
 
     The continuous parameters of the option are the main thing to note: they
-    correspond to a desired change in state for a subset of the option objects.
-    The objects that are changing correspond to changing_parameters; all other
-    objects are assumed to have no change in their state.
+    correspond to a desired change in state for a subset of the option objects
+    and a subset of those object features. The objects that are changing are
+    changing_var_order, and also the keys of changing_var_to_feat. The feature
```
ronuchit, line 249: and also the keys -> which are the same as the keys

src/nsrt_learning/option_learning.py
```diff
@@ -315,10 +316,11 @@ def _precondition_based_initiable(self, state: State, memory: Dict,
             # The memory here is used to store the absolute params, based on
             # the relative params and the object states.
             memory["params"] = params  # store for sanity checking in policy
```
ronuchit, line 318: is this sanity check just verifying that the params here are always the same as when passed into the policy()? this is always guaranteed for any option in ANY situation, right?

src/nsrt_learning/option_learning.py
```diff
@@ -329,9 +331,11 @@ def _regressor_based_policy(self, state: State, memory: Dict,
         if self._is_parameterized:
             # Compute the updated relative goal.
             assert np.allclose(params, memory["params"])
-            changing_objs = [objects[i] for i in self._changing_parameter_idxs]
-            relative_goal_vec = memory["absolute_params"] - state.vec(
-                changing_objs)
+            var_to_obj = dict(zip(self._operator.parameters, objects))
+            state_params = _create_absolute_option_param(
+                state, self._changing_var_to_feat, self._changing_var_order,
```
ronuchit, line 336: the state here can and will be different than the above one in `initiable()` right? (it'll match for the first timestep, but then there will be many more after that)

src/nsrt_learning/option_learning.py
```diff
@@ -412,31 +416,29 @@ def learn_option_specs(self, strips_ops: List[STRIPSOperator],
...
-            ]
-
+            changing_var_order = sorted(changing_var_to_feat,
+                                        key=op.parameters.index)
             for segment, var_to_obj in datastore:
                 all_objects_in_operator = [
                     var_to_obj[v] for v in op.parameters
                 ]
 
                 # First, determine the absolute goal vector for this segment.
```
ronuchit, line 431: is this comment up to date? i don't understand what here is the "absolute goal vector". i feel like this chunk is actually computing the option parameters.

ronuchit: oh, is final_params the absolute goal vector? can we just clarify this comment significantly, to state that we're accomplishing two things here: (1) computing option_param so they can later be stored into the segments, (2) computing final_param which is the absolute goal vector so we can later compute relative goal vectors when we iterate through the segment's states

tests/nsrt_learning/test_option_learning.py
```diff
@@ -182,15 +182,18 @@ def test_learned_neural_parameterized_option():
                                    pick_nsrt.delete_effects,
                                    pick_nsrt.side_predicates)
     # In this example, both of the parameters (block and robot) are changing.
-    changing_parameters = pick_operator.parameters
+    # For simplicity, assume that the first and third features for the
+    # block and the robot are changing.
+    changing_var_to_feat = {p: {0, 2} for p in pick_operator.parameters}
```
ronuchit, line 187: type error: should be a list here: [0, 2]

src/nsrt_learning/option_learning.py
```diff
@@ -412,31 +416,29 @@ def learn_option_specs(self, strips_ops: List[STRIPSOperator],
             # We do this for learning efficiency; including all objects would
             # likely work too, but may require more data for the model to
             # realize that those objects' parameters can be ignored.
-            changing_parameter_set = self._get_changing_parameters(datastore)
+            # Furthermore, we only want to include the features of the objects
+            # that exhibit some change in the data.
+            changing_var_to_feat = self._get_changing_features(datastore)
             # Just to avoid confusion, we will insist that the order of the
             # changing parameters is consistent with the order of the original.
```
ronuchit, line 423: original -> operator parameters

ronuchit (approved):
nice work!

src/nsrt_learning/option_learning.py
tomsilver: yeah

src/nsrt_learning/option_learning.py
tomsilver: right

tomsilver:
this made a very large difference for cover.

## Before
cover_num_train_tasks_perc_solved (1)

## After
cover_num_train_tasks_perc_solved

the results for stick button look comparable to before.

PR #861, opened 2022-05-03
add allow_interaction_in_demo_tasks flag

tomsilver:


src/main.py
```diff
@@ -215,6 +215,10 @@ def _generate_interaction_results(
     if CFG.make_interaction_videos:
         video = []
     for request in requests:
+        if request.train_task_idx >= CFG.max_initial_demos and \
```
ronuchit, line 218: shouldn't this be `if request.train_task_idx < CFG.max_initial_demos`?

tests/test_online_learning.py
```diff
@@ -135,3 +135,21 @@ def test_interaction():
...
+        "cover_initial_holding_prob": 0.0,
+        "approach": "unittest",
+        "timeout": 1,
+        "num_train_tasks": 2,
+        "num_test_tasks": 1,
+        "make_interaction_videos": True,
+        "max_num_steps_interaction_request": 3,
+    })
+    with pytest.raises(RuntimeError):
+        _run_pipeline(env, approach, train_tasks, dataset)
```
ronuchit, line 155: assert something about str(e)

ronuchit (changes requested):

src/main.py
tomsilver: thanks, brainfart...

ronuchit (approved):

PR #864, opened 2022-05-04
add flag to make videos of demos

tomsilver:


ronuchit (approved):
cool, what's this being used for?

tomsilver:
pretend play at the moment, but i think it will be nice to have in general

PR #865, opened 2022-05-04
clip parameters inside sample_option() instead of sampler learning

ronuchit:
note: the only places that call sample_option() are planner.py and gnn_metacontroller_approach.py, both of which use sampler_learning.py, so i think i've cleaned out all the redundant calls to clip().

### DIFF
--- a/src/nsrt_learning/sampler_learning.py
+++ b/src/nsrt_learning/sampler_learning.py
@@ -286,11 +286,11 @@ class _LearnedSampler:
         sub = dict(zip(self._variables, objects))
         for var in self._variables:
             x_lst.extend(state[sub[var]])
-        # For sampler learning, we currently make the extremely limiting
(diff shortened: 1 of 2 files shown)


src/nsrt_learning/sampler_learning.py
```diff
@@ -312,12 +309,6 @@ def sampler(self, state: State, goal: Set[GroundAtom],
                self._classifier.classify(np.r_[x, params]):
                 break
             num_rejections += 1
-        else:
-            # Edge case: we exceeded the number of sampling tries
-            # and we might be left with a params that is not in
-            # bounds. If so, fall back to sampling from the space.
-            if not self._param_option.params_space.contains(params):
-                params = self._param_option.params_space.sample()
```
ronuchit, line 320: note: we're changing the behavior from random sample to clip here. shouldn't matter hopefully

PR #866, opened 2022-05-04
add setting to disable expected atoms check

### LINKED ISSUE #862: implement planning baseline that does not use expected atoms check

ronuchit:
findings on repeated_nextto:
- oracle & our backchaining approach work the same whether it's enabled or disabled, no difference in planning time
- cluster_and_intersect sucks whether it's enabled or disabled, because the skeleton search itself is so difficult with so many high-arity operators. so planning times out in skeleton search itself
- one interesting thing i tried: **if you change the oracle operators to not contain side predicates** (just do side_predicates = set() for both Move and Grasp), **then planning works** (albeit slowly; gets ~40/50) **when the expected atoms check is disabled, but fails unilaterally when it's enabled**. this was interesting to me because i always thought that the expected atoms check is strictly beneficial from a success rate perspective (ignoring timeout), but this is not true. the reason planning fails when expected_atoms is enabled is, simply, that something like NextToNothing would be in the expected_atoms after a Move, but it's not in the actual state.

closes #862

### DIFF
--- a/src/planning.py
+++ b/src/planning.py
@@ -322,7 +322,7 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
                 # Check if we have exceeded the horizon.
                 if np.sum(num_actions_per_option[:cur_idx]) > max_horizon:
                     can_continue_on = False
-                else:
+                elif CFG.sesame_check_expected_atoms:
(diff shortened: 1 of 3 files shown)


ronuchit:
i'll post on slack

PR #871, opened 2022-05-05
fix bug in random options approach

tomsilver:


ronuchit (commented):
i don't understand the bug... if we're in this `else` then that guarantees the `for` loop could not have changed `cur_option`, and so that means we're guaranteed to re-enter the `if cur_option is DummyOption or cur_option.terminal(state):` on the next iteration right

tomsilver:
the issue happens when:
1. `cur_option.terminal(state)` is true
2. we can't sample another option, so `cur_option` remains the same, and we take a random action
3. on the next time step, `cur_option` is still what it was before. but that's not right, because as soon as an option terminates, we should never use it after that. note that `cur_option.terminal(state)` may now return True, because the state is different after a random action has been taken.

ronuchit:
oh, do you mean "cur_option.terminal(state)" may now return False?

tomsilver:
yeah sorry edited

ronuchit (approved):
got it now... no idea how you caught it but power to you

PR #877, opened 2022-05-06
fix video monitor render order bug 

### LINKED ISSUE #876: `VideoMonitor` state rendering is off by one
when this function is called: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/utils.py#L2013

the render function is using `env._current_state`, but that current state has already been advanced here:
https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/utils.py#L859

this bug was introduced by me here:
https://github.com/Learning-and-Intelligent-Systems/predicators/commit/35643231e695bcdb80461c6c9810b80bcea034ee

tomsilver:
closes #876 

I tried a lot of less gross things first, but none worked... open to other solutions though

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -856,10 +856,14 @@ def run_policy(
             # into one, but that led to subtle bugs in the order of calls to
             # monitor.observe(). This version is uglier but less bug-prone.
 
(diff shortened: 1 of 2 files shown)


ronuchit:
superseded by #878

PR #880, opened 2022-05-06
RL approach with Dummy RLOptionLearner. 

ashayathalye:
Loops through online learning trajectories and creates online experience data for each option. The online experience data consists of the lists of states, actions, rewards, and relative_parameters experienced. This data is passed to an RLOptionLearner, which is instantiated for each option (because the RLOptionLearner will need to maintain state unique to the learning process for its option). The RLOptionLearner returns an updated parameterized_option, which replaces the old one in the corresponding NSRT. 

Reviewers: the main thing to look at is the overall logic before I spend much time integrating non-dummy RLOptionLearners, in case big things need changing, and whether certain computations (especially for the option's relative parameters) can be achieved more concisely. I will continue to update some comments and add tests today, but I think an initial review can be done now.

### DIFF
--- a/src/approaches/nsrt_rl_approach.py
+++ b/src/approaches/nsrt_rl_approach.py
@@ -28,7 +28,9 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
         self._nsrts: Set[NSRT] = set()
         self._online_learning_cycle = 0
         self._initial_trajectories: List[LowLevelTrajectory] = []
-        self._online_trajectories: List[LowLevelTrajectory] = []
+        self._train_task_to_online_traj: Dict[int, List[LowLevelTrajectory]] = {}
+        self._train_task_to_option_plan: Dict[int, List[_Option]] = {}
+        self._reward_epsilon = CFG.reward_epsilon
 
     @classmethod
     def get_name(cls) -> str:
@@ -49,11 +51,15 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
             task = self._train_tasks[i]
             try:
                 _act_policy = self.solve(task, CFG.timeout)
+                # Store the list of _Option objects corresponding to this policy.
+                self._train_task_to_option_plan[i] = self._last_plan
             except (ApproachTimeout, ApproachFailure) as e:
                 partial_refinements = e.info.get("partial_refinements")
                 assert partial_refinements is not None
                 _, plan = max(partial_refinements, key=lambda x: len(x[1]))
                 _act_policy = utils.option_plan_to_policy(plan)
+                # Store the list of _Option objects corresponding to this policy.
+                self._train_task_to_option_plan[i] = plan
             request = InteractionRequest(train_task_idx=i,
                                          act_policy=_act_policy,
                                          query_policy=lambda s: None,
@@ -72,10 +78,40 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
                                       actions,
                                       _is_demo=False,
                                       _train_task_idx=i)
-            self._online_trajectories.append(traj)
+            self._train_task_to_online_traj[i] = traj
 
-        # Replace this with an _RLOptionLearner.
-        self._learn_nsrts(self._initial_trajectories + \
-            self._online_trajectories,
-            online_learning_cycle=self._online_learning_cycle
-        )
+        # for each task:
+        #    for each _Option involved in the trajectory:
+        #       1) get the trajectory that it took
+        #       2) get the sample subgoal, and determine if it was reached
+        for i in range(len(self._train_tasks)):
+            plan = self._train_task_to_option_plan[i]
+            traj = self._train_task_to_online_traj[i]
+
+            # map _Option to trajectory and reward it had
+            option_to_traj = {}
+            option_to_reward = {}
+            curr_option_idx = 0
+            curr_option = plan[curr_option_idx]
+            curr_traj = [] # List of (state, action) tuples
+
+            for s, a in zip(traj.states, traj.actions):
+                if curr_option.terminal(s):
+                    curr_option_idx += 1
+                    if curr_option_idx < len(plan):
+                        curr_option = plan[curr_option_idx]
+                    else:
+                        # If we run out of options in the plan, there should be
+                        # an _OptionPlanExhausted exception, and so there is
+                        # nothing more in the trajectory that we have not yet
+                        # assigned to an _Option already.
+                        # TODO: maybe add an assert to confirm the above ^
+                        pass
+                    option_to_traj[curr_option_idx] = list(curr_traj)
+                    curr_traj = [(s, a)]
+                else:
+                    curr_traj.append((s, a))
+
+            import pdb; pdb.set_trace()
+
+            # TODO: review how the params are used by the _Option's policy for nsrt_learning
--- a/src/approaches/nsrt_rl_approach.py
+++ b/src/approaches/nsrt_rl_approach.py
@@ -6,6 +6,7 @@ from typing import List, Sequence, Set
 from gym.spaces import Box
 
 from predicators.src import utils
+from predicators.src.nsrt_learning.option_learning import create_option_learner
 from predicators.src.approaches.base_approach import ApproachFailure, \
     ApproachTimeout
 from predicators.src.approaches.nsrt_learning_approach import \
@@ -34,6 +35,10 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
         self._reward_epsilon = CFG.reward_epsilon
         self._pos_reward = CFG.pos_reward
         self._neg_reward = CFG.neg_reward
+        # We need to create a separate RL option learner for each option because
+        # each one will maintain its own unique state associated with the
+        # learning process.
+        self._option_learners = {n.name: create_option_learner() for n in self._nsrts}
 
     @classmethod
     def get_name(cls) -> str:
@@ -179,9 +184,13 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
         # update, and all the data associated with it from the online learning
         # cycle that just happened.
         option_to_parent_and_nsrt = {}
-        parameterized_options = [(nsrt.option, nsrt.name) for nsrt in self._nsrts]
-        for option_name in option_to_data.keys():
+        parameterized_options = [(nsrt.option, nsrt) for nsrt in self._nsrts]
 
-
-        # TODO: call RL option learner's update method, passing in (s, a, s', r)
-        # TODO: replace the corresponding parameterized option
+        for option_name, experience in option_to_data.items():
+            corresponding_nsrt = [nsrt for nsrt in self._nsrts if nsrt.option.name == option_name][0]
+            corresponding_parent_option = corresponding_nsrt.option
+            updated_option = self._option_learners[corresponding_nsrt.name].update(
+                corresponding_parent_option,
+                experience
+            )
+            corresponding_nsrt.option = updated_option
--- a/src/approaches/nsrt_rl_approach.py
+++ b/src/approaches/nsrt_rl_approach.py
@@ -71,7 +71,7 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
         return requests
 
     @classmethod
-    def infer_subgoal(cls, object: Object, states: List[State], features: List[str]) -> List[float]:
+    def infer_delta(cls, object: Object, states: List[State], features: List[str]) -> List[float]:
         return [states[-1].get(object, feat) - states[0].get(object, feat) for feat in features]
 
     def learn_from_interaction_results(
@@ -88,7 +88,8 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
             self._train_task_to_online_traj[i] = traj
 
 
-        option_to_data = {i: [] for i in range(len(plan))} # idx -> list (s, a, s', r)
+        option_to_data = {} # option_name -> online experience data
+
         # for each task:
         #    for each _Option involved in the trajectory:
         #       compute (s, a, s', r) tuples
@@ -96,42 +97,50 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
             plan = self._train_task_to_option_plan[i]
             traj = self._train_task_to_online_traj[i]
 
-            # map _Option to trajectory and reward it had
-            option_to_traj = {}
-            option_to_reward = {}
             curr_option_idx = 0
             curr_option = plan[curr_option_idx]
-            curr_states = []
+            if curr_option.name not in option_to_data:
+                option_to_data[curr_option.name] = []
+            curr_states = [traj.states[0]]
             curr_actions = []
             curr_rewards = []
+            curr_relative_params = []
             actions = (a for a in traj.actions)
+            block = [b for b in curr_option.objects if b.type.name=='block'][0]
+            robot = [r for r in curr_option.objects if r.type.name=='robot'][0]
 
-            for i, s in enumerate(traj.states):
+            # Generate transition data, (s, a, s', r, relative_param). The
+            # reward R(s, a, s') = neg_reward if s' is not within epsilon of the
+            # subgoal, and pos_reward otherwise.
+            for j, s in enumerate(traj.states[1:]):
+                curr_states.append(s)
+                curr_actions.append(next(actions))
+                curr_relative_params.append(curr_option.memory["absolute_params"] - s.vec(curr_option.objects))
+
                 if curr_option.terminal(s):
-                    curr_states.append(s)
+                    # TODO: make this not hardcoded to be environment specific,
+                    # or implement specifically per environment and throw an
+                    # error for environments where this is not implemented.
 
-                    # Figure out reward.
-                    # TODO: inferring reward requires environment specific code?
-                    block = [b for b in curr_option.objects if b.type.name=='block'][0]
-                    robot = [r for r in curr_option.objects if r.type.name=='robot'][0]
-                    if curr_option.params[-1] > 0: # if holding becomes true
-                        dblock = self.infer_subgoal(block, curr_states, ['grasp'])
-                        drobot = self.infer_subgoal(robot, curr_states, ['x', 'y', 'grip', 'holding'])
+                    # Check if we reached our subgoal within a tolerance by
+                    # checking the difference between our proposed subgoal
+                    # (which is expressed relatively) and the relative changes
+                    # that actually happened.
+                    if curr_option.params[-1] > 0:  # if holding becomes true
+                        dblock = self.infer_delta(block, curr_states, ['grasp'])
+                        drobot = self.infer_delta(robot, curr_states, ['x', 'y', 'grip', 'holding'])
                     else:
-                        dblock = self.infer_subgoal(block, curr_states, ['x', 'grasp'])
-                        drobot = self.infer_subgoal(robot, curr_states, ['x', 'grip', 'holding'])
-                    subgoal = np.array(dblock + drobot)
-                    print("option params: ", curr_option.params)
-                    print("subgoal: ", subgoal)
-                    if np.allclose(curr_option.params, subgoal, atol=self._reward_epsilon):
+                        dblock = self.infer_delta(block, curr_states, ['x', 'grasp'])
+                        drobot = self.infer_delta(robot, curr_states, ['x', 'grip', 'holding'])
+                    actual_delta = np.array(dblock + drobot)
+                    if np.allclose(curr_option.params, actual_delta, atol=self._reward_epsilon):
                         reward = self._pos_reward
                     else:
                         reward = self._neg_reward
                     curr_rewards.append(reward)
 
-                    # Store trajectory and reward
-                    option_to_reward[curr_option_idx] = list(curr_rewards)
-                    option_to_traj[curr_option_idx] = (list(curr_states), list(curr_actions))
+                    # Store transition data.
+                    option_to_data[curr_option.name].append(list(curr_states, curr_actions, curr_rewards, curr_relative_params))
 
                     # Advance to next option.
                     curr_option_idx += 1
@@ -148,19 +157,31 @@ class NSRTReinforcementLearningApproach(NSRTLearningApproach):
                     # Initialize trajectory for next option.
                     curr_states = [s]
                     curr_actions = []
-                    curr_rewards = [-1]
+                    curr_rewards = []
+                    curr_relative_params = []
 
-                else:
-                    curr_states.append(s)
-                    a = next(actions)
-                    curr_actions.append(a)
-                    # If this is the last state, then we haven't gotten the reward
-                    if i == len(traj.states) - 1:
-                        option_to_traj[curr_option_idx] = (list(curr_states), list(curr_actions))
+                else:  # case where terminal state not reached.
                     curr_rewards.append(self._neg_reward)
+                    # Handle the case where we are at the last state in the
+                    # trajectory, but it is not a terminal state of the current
+                    # option. This occurs when the plan we are executing is a
+                    # partial refinement.
+                    if j+1 == len(traj.states) - 1:
+                        # Store transition data.
+                        option_to_data[curr_option.name].append(list(curr_states, curr_actions, curr_rewards, curr_relative_params))
 
-            # TODO: make a list of (s, a, s', r) for each option
-            
-            # TODO: associate each _Option we see with an nsrt's parameterized option
-            # TODO: call RL option learner's update method, passing in (s, a, s', r)
-            # TODO: replace the corresponding parameterized option
+        import pdb; pdb.set_trace()
+
+        # Associate each unique option we see in the data (that we
+        # identify by the option's name) with an nsrt's parameterized option. We
+        # need to do this because the RLOptionLearner updates the parameterized
+        # options. The RLOptionLearner will receive a parameterized option to
+        # update, and all the data associated with it from the online learning
+        # cycle that just happened.
+        option_to_parent_and_nsrt = {}
+        parameterized_options = [(nsrt.option, nsrt.name) for nsrt in self._nsrts]
+        for option_name in option_to_data.keys():
+
+
+        # TODO: call RL option learner's update method, passing in (s, a, s', r)
+        # TODO: replace the corresponding parameterized option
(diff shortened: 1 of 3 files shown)


src/approaches/nsrt_rl_approach.py
```diff
@@ -37,6 +42,11 @@ def get_name(cls) -> str:
     def learn_from_offline_dataset(self, dataset: Dataset) -> None:
         self._initial_trajectories = dataset.trajectories
         super().learn_from_offline_dataset(dataset)
+        # We need to create a separate RL option learner for each option because
+        # each one will maintain its own unique state associated with the
+        # learning process.
+        CFG.option_learner = CFG.rl_option_learner
```
tomsilver, line 48: why do we need a separate flag for rl option learner vs option learner? can we just use the existing flag?

also, I think we are implicitly assuming that `CFG` is static within the course of a single run (we should probably enforce this). I'm not sure what will go wrong if we change things like this, but something might.

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,18 +59,26 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
             task = self._train_tasks[i]
             try:
                 _act_policy = self.solve(task, CFG.timeout)
+                # Store the list of _Option objects corresponding to this policy.
+                self._train_task_to_option_plan[i] = self._last_plan
```
tomsilver, line 63: nice use of `self._last_plan`

src/approaches/nsrt_rl_approach.py
```diff
@@ -72,10 +90,128 @@ def learn_from_interaction_results(
...
+            block = [b for b in curr_option.objects if b.type.name=='block'][0]
+            robot = [r for r in curr_option.objects if r.type.name=='robot'][0]
+
+            # Generate transition data, (s, a, s', r, relative_param). The
+            # reward R(s, a, s') = neg_reward if s' is not within epsilon of the
+            # subgoal, and pos_reward otherwise.
+            for j, s in enumerate(traj.states[1:]):
+                curr_states.append(s)
+                curr_actions.append(next(actions))
+                # TODO: is there a better way to do this based on
```
tomsilver, line 122: yeah I see the issue. here are a few possibilities:
1. we could store the relevant information in the option's memory. that would involve a change in `_LearnedNeuralParameterizedOption._precondition_based_initiable`.
2. we could create a subclass of `_Option` that exposes the relevant information with a method.
3. we could refactor the code so that when the relevant information is computed the first time, it's passed back to this Approach, which then saves it.

I think (3) is preferred if we can figure out a way to do that cleanly, but I can't immediately envision such a way. absent that, (1) is probably preferred.

src/approaches/nsrt_rl_approach.py
```diff
@@ -72,10 +90,128 @@ def learn_from_interaction_results(
...
+                        s.get(block, 'grasp'),
+                        s.get(robot, 'x'),
+                        s.get(robot, 'grip'),
+                        s.get(robot, 'holding')
+                    ])
+                curr_relative_params.append(curr_option.memory["absolute_params"] - changing_obj_feats)
+
+                if curr_option.terminal(s):
+                    # TODO: make this not hardcoded to be environment specific,
+                    # or implement specifically per environment and throw an
```
tomsilver, line 145: definitely we will want to do all of this in a non environment specific way

src/approaches/nsrt_rl_approach.py
```diff
@@ -72,10 +90,128 @@ def learn_from_interaction_results(
...
+                    # (which is expressed relatively) and the relative changes
+                    # that actually happened.
+                    if curr_option.params[-1] > 0:  # if holding becomes true
+                        dblock = self.infer_delta(block, curr_states, ['grasp'])
+                        drobot = self.infer_delta(robot, curr_states, ['x', 'y', 'grip', 'holding'])
+                    else:
+                        dblock = self.infer_delta(block, curr_states, ['x', 'grasp'])
+                        drobot = self.infer_delta(robot, curr_states, ['x', 'grip', 'holding'])
+                    actual_delta = np.array(dblock + drobot)
+                    if np.allclose(curr_option.params, actual_delta, atol=self._reward_epsilon):
```
tomsilver, line 159: another thing we will probably want to do is give a large negative reward if any objects that are not supposed to change during this option execution do actually change. we can leave a TODO in the code or just remember

src/approaches/nsrt_rl_approach.py
```diff
@@ -72,10 +90,128 @@ def learn_from_interaction_results(
...
+                    if curr_option_idx < len(plan):
+                        curr_option = plan[curr_option_idx]
+                        if curr_option.name not in option_to_data:
+                            option_to_data[curr_option.name] = []
+                    else:
+                        # If we run out of options in the plan, there should be
+                        # an _OptionPlanExhausted exception, and so there is
+                        # nothing more in the trajectory that we have not yet
+                        # assigned to an _Option already.
+                        # TODO: maybe add an assert to confirm the above ^
```
tomsilver, line 179: +1 to the assert here

src/approaches/nsrt_rl_approach.py
```diff
@@ -72,10 +90,128 @@ def learn_from_interaction_results(
...
+                    curr_actions = []
+                    curr_rewards = []
+                    curr_relative_params = []
+
+                else:  # case where terminal state not reached.
+                    curr_rewards.append(self._neg_reward)
+                    # Handle the case where we are at the last state in the
+                    # trajectory, but it is not a terminal state of the current
+                    # option. This occurs when the plan we are executing is a
+                    # partial refinement.
```
tomsilver, line 193: I could also imagine this happening if we just run out of horizon in the environment. maybe we should just ditch this data because we can't know if the option would have succeeded with just a few more steps?

src/approaches/nsrt_rl_approach.py
```diff
@@ -72,10 +90,128 @@ def learn_from_interaction_results(
...
+                        option_to_data[curr_option.name].append([curr_states, curr_actions, curr_rewards, curr_relative_params])
 
-        # Replace this with an _RLOptionLearner.
-        self._learn_nsrts(self._initial_trajectories + \
-            self._online_trajectories,
-            online_learning_cycle=self._online_learning_cycle
-        )
+        # Call the RL option learner on each option.
+        for option_name, experience in option_to_data.items():
+            corresponding_nsrt = [nsrt for nsrt in self._nsrts if nsrt.option.name == option_name][0]
```
tomsilver, line 200: assert that there's exactly 1 of these

src/approaches/nsrt_rl_approach.py
tomsilver: oh looking further, I'm guessing this is actually a temporary placeholder

src/approaches/nsrt_rl_approach.py
tomsilver: maybe we can achieve (3) by using the new option learner class that you're introducing?

src/nsrt_learning/option_learning.py
```diff
@@ -578,6 +580,33 @@ def _create_regressor(self) -> Regressor:
             grid_num_ticks_per_dim=num_ticks)
 
 
+class _RLOptionLearnerBase(abc.ABC):
```
tomsilver, line 583: this all looks good. we'll have to think through how we want to reuse the rest of the code in this file in the forthcoming subclasses.

tomsilver (commented):
thanks for the clear code and very helpful comments in the PR! you raise a number of important and difficult questions. let's discuss them at our meeting tomorrow.

src/approaches/nsrt_rl_approach.py
ashayathalye: I think the cleanest way, without making any modifications, would use the `self._changing_var_to_feat` attribute of the `_LearnedNeuralParamterizedOption`. But for some reason, that's the one attribute I couldn't access– it just didn't exist when I listed the attributes (via `o.dir()`, where o is the `_Option`. What confused me was that I did see `self._changing_var_order`. We can discuss during our meeting today.

src/approaches/nsrt_rl_approach.py
ashayathalye: Solution to earlier comment about line 122 solves it here as well.

src/approaches/nsrt_rl_approach.py
ashayathalye: Added a TODO as a comment. This will also be implemented easily after we figure out the thing related to `self._changing_var_to_feat`.

src/approaches/nsrt_rl_approach.py
ashayathalye: Added.

src/approaches/nsrt_rl_approach.py
ashayathalye: To be clear, the limited horizon is coming from `for _ in range(max_num_steps):` in `utils.run_policy()`. 

This is an interesting point: 
It is true that perhaps this option would've succeeded in more steps. By penalizing it negatively in such a case, maybe it will learn that what it was doing was "wrong". The implicit assumption here is that `max_num_steps` encodes our belief for the sufficient number of steps to solve a task. As long as we take `max_num_steps` to be sufficiently high (higher than our oracle options take), then it's not a problem? 

If `max_num_steps` was set high enough, ditching this data is bad, because the option will never get a signal that what it was doing here was bad. We may see the option continuing to perform this behavior, and we'll continue to not assign it a positive or negative reward (since we ditch it).

ashayathalye: Note from discussion: 
Maybe dense rewards addresses this issue.

ashayathalye:
* Test is a bit slow due to having 10 training tasks (which I put so that there's at least one that ends up as a partial refinement, so that we get 100% coverage of the code). Maybe can find a seed or certain parameters so that this happens with fewer training tasks? 
* I put a few `# pylint: disable=protected-access` when accessing the some attributes of `_LearnedNeuralParameterizedOption.parent.[something]` => can either make these not private, or can make getters for these attributes.

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+        var_to_unchanging_feat_ind = {}
+        for var, changing_indices in changing_var_to_feat.items():
+            dim = var_to_obj[var].type.dim
+            unchanging_indices = [
+                i for i in range(dim) if i not in changing_indices
+            ]
+            var_to_unchanging_feat_ind[var] = unchanging_indices
+        return var_to_unchanging_feat_ind
+
     def learn_from_interaction_results(
```
tomsilver, line 95: this has become a very long method, can we split it up into some helper methods for easier readibility?

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+                subgoal_state_changing_feat = curr_option.memory[
+                    "absolute_params"]
+                relative_param = subgoal_state_changing_feat - \
+                    curr_state_changing_feat
+                curr_relative_params.append(relative_param)
+
+                # In the remaining code in this loop we compute the reward.
+                reward = 0
+
+                # Add large negative reward if any object's features that are
```
tomsilver, line 162: I think that we should be able to do this large negative reward computation stuff in a more concise way. for now, can we actually just take this part out of this PR, and we'll soon make another PR that is exclusively focused on this large negative reward question? my feeling is that there is already a lot of stuff going on in this PR, without this bit.

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+            plan = self._train_task_to_option_plan[i]
+            traj = self._train_task_to_online_traj[i]
+
+            curr_option_idx = 0
+            curr_option = plan[curr_option_idx]
+            parent = cast(_LearnedNeuralParameterizedOption,
+                          curr_option.parent)
+            var_to_obj = dict(
+                zip(parent._operator.parameters, curr_option.objects))  # pylint: disable=protected-access
+            var_to_unchanging_feat_ind = \
```
tomsilver, line 132: I'm hoping we can get rid of this line based on my other comments, but just FYI, in this situation, you can do `self._get_unchanging_features` instead of `NSRTReinforcementLearningApproach._get_unchanging_features`

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+                    curr_states[-1].vec(other_objects),
+                    atol=1e-5)
+                if not option_objects_unchanged or not other_objects_unchanged:
+                    reward += self._neg_reward * 10
+
+                # Add pos_reward if we got within epsilon of the option's
+                # subgoal, otherwise we add neg_reward.
+                # Ignore the optimization that checks for repeated states, which
+                # messes with the terminal state check.
+                _ = curr_option.memory.pop("last_state", None)
```
tomsilver, line 199: modifying the memory of the option could lead to unforeseen bugs in the future. can we instead add a public method in `_LearnedNeuralParameterizedOption` that checks the abstract effects, but doesn't check the last state?

tomsilver: actually, why do we even need this check? shouldn't we be rewarding the option on every time step, not just when it terminates?

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+                        curr_rewards.append(reward + self._neg_reward)
+
+                    # Store transition data.
+                    option_to_data[curr_option.name].append(
+                        (curr_states, curr_actions, curr_rewards,
+                         curr_relative_params))
+
+                    # Advance to next option.
+                    curr_option_idx += 1
+                    if curr_option_idx < len(plan):
```
tomsilver, line 216: I'm hoping that in addition to simplifying these lines with the refactoring suggested above, we can do additional refactoring to remove any duplicated logic (this all looks similar to code above)

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+                            parent._changing_var_to_feat, var_to_obj)  # pylint: disable=protected-access
+                        if curr_option.name not in option_to_data:
+                            option_to_data[curr_option.name] = []
+                    else:
+                        # If we run out of options in the plan, there should be
+                        # an _OptionPlanExhausted exception, and so there is
+                        # nothing more in the trajectory that we have not yet
+                        # assigned to an _Option already. That is, the current
+                        # state in this loop is the last state.
+                        assert s == traj.states[-1]
```
tomsilver, line 235: state equality is not defined. if you want to assert that this is the same python object, use `is` instead of `==`. if you just want to check that the states are semantically equivalent, use `state.allclose(other)`

tomsilver (commented):
good stuff! comments below. after a few more rounds, I can help push it over the finish line.

in addition to my comments, I am wondering if we can split this into multiple PRs. for example, any changes that we make to the `_LearnedNeuralParameterizedOption` class can be a standalone PR.

overall, even though the number of lines of code is not enormous, there are a lot of important and complicated decision decisions in this PR, which is why I would like to split it.

src/approaches/nsrt_rl_approach.py
ashayathalye: @ second question: I thought that the positive reward is only awarded for the transition into the terminal state (and then only if sufficiently close to the target subgoal). Instead, I guess I could remove the terminal state check and just check at each timestep if the current state is within epsilon of the subgoal. Then, if the effects hold only when within epsilon/2 of the subgoal, the agent may receive a positive reward multiple times in this trajectory. But that's fine– we can set epsilon however we want to to tune this. 

@ first question: any time option.terminal(s) is called, the last_state is set to s. Because we loop through the entire trajectory, we always end with last_state being set to what it originally was set to. So while we modify it, it gets modified back to the original thing properly. But I get that maybe modifying it at all might lead to bugs. The separate public method seems like a good idea, I'll do that.

src/approaches/nsrt_rl_approach.py
```diff
@@ -49,33 +65,215 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+            curr_option = plan[curr_option_idx]
+            parent = cast(_LearnedNeuralParameterizedOption,
+                          curr_option.parent)
+            var_to_obj = dict(
+                zip(parent._operator.parameters, curr_option.objects))  # pylint: disable=protected-access
+            var_to_unchanging_feat_ind = \
+                NSRTReinforcementLearningApproach._get_unchanging_features(
+                parent._changing_var_to_feat, var_to_obj)  # pylint: disable=protected-access
+            if curr_option.name not in option_to_data:
+                option_to_data[curr_option.name] = []
```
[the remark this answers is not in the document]
ashayathalye: The RLOptionLearner (which is instantiated for each option) will need the data of all executions for a particular option (replay buffer). 

So for each online learning cycle, we can collect the new experience for each option, and then pass it into the RLOptionLearner for that option. The RLOptionLearner will store this in ITS OWN replay buffer. So we just pass it the new stuff, and it keeps track what is new from this current online learning cycle and what is old from a previous online learning cycle. 

I think this is fine? Let's chat over slack– maybe I'm super confused / misunderstanding something.

src/approaches/nsrt_rl_approach.py
ashayathalye: Sure, will remove it from this PR.

src/approaches/nsrt_rl_approach.py
```diff
@@ -66,16 +84,121 @@ def learn_from_interaction_results(
...
+            plan = self._train_task_to_option_plan[i]
+            traj = self._train_task_to_online_traj[i]
+
+            curr_option_idx = 0
+            curr_option = plan[curr_option_idx]
+            parent_option = cast(_LearnedNeuralParameterizedOption,
+                                 curr_option.parent)
+            if parent_option not in option_to_data:
+                option_to_data[parent_option] = []
+            actions = ((i, a) for i, a in enumerate(traj.actions))
```
tomsilver, line 118: careful about reusing `i` when we're already in a loop with that variable

src/approaches/nsrt_rl_approach.py
```diff
@@ -66,16 +84,121 @@ def learn_from_interaction_results(
...
+        # contacenation of the state vector for the option's objects and the
+        # relative parameter that the sampler provided. This input vector is
+        # necessary during learning to update the option's regressor.
+        option_to_data: Dict[ParameterizedOption,
+                             List[List[Tuple[State, Array, Action, int,
+                                             State]]]] = {}
+
+        # For each training task, compute the experience data for each _Option
+        # we see used in that training task.
+        for i in range(len(self._train_tasks)):
```
tomsilver, line 108: let's make a helper function that takes in an `InteractionResult` and produces `option_to_data` for that result; then, join the outputs together in this loop.

src/approaches/nsrt_rl_approach.py
tomsilver: also it's currently unclear why this generator is getting created until one reads on to the loop below. it would be more clear if the actions were just indexed inside of that loop when they're needed.

src/approaches/nsrt_rl_approach.py
```diff
@@ -66,16 +84,121 @@ def learn_from_interaction_results(
...
+
+                # Add pos_reward if we got within epsilon of the option's
+                # subgoal, otherwise we add neg_reward. To do this, we can check
+                # the relative parameter for the next state.
+                next_rel_param = parent_option.get_rel_option_param_from_state(
+                    next_state,
+                    curr_option.memory,
+                    curr_option.objects,
+                )
+                reward: int = 0
```
tomsilver, line 144: is this type annotation needed? if so, I'm curious why...

src/approaches/nsrt_rl_approach.py
tomsilver: as part of that refactor, it may as be nice to introduce a new type for the experience tuple (see the bottom of structs.py)

src/approaches/nsrt_rl_approach.py
```diff
@@ -66,16 +84,121 @@ def learn_from_interaction_results(
...
+                # option did not reach its terminal state but had a sufficient
+                # number of steps to reach it so that it is reasonable to assign
+                # it a reward.
+                terminate = parent_option.effect_based_terminal(
+                    next_state, curr_option.objects)
+                had_sufficient_steps = next_state.allclose(
+                    traj.states[-1]) and (
+                        CFG.max_num_steps_interaction_request -
+                        (idx + 1) > CFG.valid_reward_steps_threshold)
+                if terminate:
```
tomsilver, line 164: as part of the suggested refactor into a helper function, it may also be good to have another smaller helper function that is responsible for processing the trajectory for a single option, until that option is done.

tests/approaches/test_nsrt_rl_approach.py
```diff
@@ -1,51 +1,47 @@
...
 from predicators.src.envs.cover import CoverMultistepOptions
 from predicators.src.main import _generate_interaction_results
+from predicators.src.settings import CFG
 from predicators.src.teacher import Teacher
 
 
 def test_nsrt_reinforcement_learning_approach():
     """Test for NSRTReinforcementLearningApproach class, entire pipeline."""
-    config = {
+
```
tomsilver, line 16: I can help iron out test coverage when it's the only thing left. I saw your note about test speed; we'll want to make sure these changes don't add more than 1-2 seconds overall.

tomsilver (commented):
@ashayathalye good stuff! left some comments and suggested refactors. ping me when it's ready for another look. thanks!

src/approaches/nsrt_rl_approach.py
```diff
@@ -66,16 +84,121 @@ def learn_from_interaction_results(
...
         # We get one result per training task.
         for i, result in enumerate(results):
-            states = result.states
-            actions = result.actions
-            traj = LowLevelTrajectory(states,
-                                      actions,
+            traj = LowLevelTrajectory(result.states,
+                                      result.actions,
                                       _is_demo=False,
                                       _train_task_idx=i)
```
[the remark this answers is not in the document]
ashayathalye: Well, we determine the order of the results because we define the method `get_interaction_requests()` for this class, which returns a list of requests. Then in `main.py`, the requests are handled sequentially. 

@ (1), we don't technically need the train task indices. We just need to associate the online trajectory with the plan (list of _Options) that was used to generate it. 
@ (2), InteractionResult already includes the train task index. 

I'll update the code to address these points.

src/approaches/nsrt_rl_approach.py
```diff
@@ -66,16 +84,121 @@ def learn_from_interaction_results(
...
+                        if parent_option not in option_to_data:
+                            option_to_data[parent_option] = []
+                    else:
+                        # If we run out of options in the plan, there should be
+                        # an _OptionPlanExhausted exception, and so there is
+                        # nothing more in the trajectory that we have not yet
+                        # assigned to an _Option already. That is, the current
+                        # state in this loop is the last state.
+                        assert next_state.allclose(traj.states[-1])
+                elif had_sufficient_steps:
```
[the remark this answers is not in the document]
ashayathalye: But not the same stuff happens in case <terminate> and case <had_sufficient_steps>? The contents of the if body for each is different?

src/nsrt_learning/option_learning.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import abc
+import copy
```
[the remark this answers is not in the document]
ashayathalye: Well since nsrt_rl_approach.py uses things from option_learner.py that were added here, won't that mess things up if we move it to a different PR? 

I added a test for the method added to _LearnedNeuralParameterizedOption. The refactor done to the effect_based_terminal function is already tested properly in the unit tests.

src/approaches/nsrt_rl_approach.py
```diff
@@ -61,126 +60,134 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+        parent_option = cast(_LearnedNeuralParameterizedOption,
+                             curr_option.parent)
+        if parent_option not in option_to_data:
+            option_to_data[parent_option] = []
+        experience = []
+
+        # Loop through the trajectory and compute the experience tuples for
+        # each _Option in the plan.
+        j = 0
+        for state, next_state in zip(traj.states[:-1], traj.states[1:]):
```
tomsilver, line 105: you can do instead:

```for state, action, next_state in zip(traj.states[:-1], traj.actions, traj.states[1:]):```

and then get rid of `j`

File: src/approaches/nsrt_rl_approach.py [hunk 26]
```diff
@@ -61,126 +60,134 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
+                traj.states[-1]) and (CFG.max_num_steps_interaction_request - j
+                                      > CFG.valid_reward_steps_threshold)
+            if terminate:
+                option_to_data[parent_option].append(experience)
+                curr_option_idx += 1
+                if curr_option_idx < len(plan):
+                    curr_option = plan[curr_option_idx]
+                    parent_option = cast(_LearnedNeuralParameterizedOption,
+                                         curr_option.parent)
+                    experience = []  # Reset for next option in the plan.
```
tomsilver, line 151: do you need to do this in the cases below as well?

src/approaches/nsrt_rl_approach.py
```diff
@@ -61,126 +60,134 @@ def get_interaction_requests(self) -> List[InteractionRequest]:
...
-                elif had_sufficient_steps:
-                    assert curr_option == plan[-1]
-                    option_to_data[parent_option].append(experience)
+        # For each InteractionResult, compute the experience data for each
+        # _Option we see used in that interaction.
+        for i, result in enumerate(results):
+            option_to_data_from_result = self._get_experience_from_result(
+                i, result)
+            for option, experience in option_to_data_from_result.items():
+                option_to_data.setdefault(option, []).extend(experience)
```
tomsilver, line 190: this is a little hard to read. maybe use a defaultdict, or just break it up into cases.

tomsilver, line 190: why are we only using data from the latest interaction results, rather than accumulating all data over all cycles?

tomsilver (commented):
@ashayathalye this is looking a lot better, very nice work! just a few things, and then I think we'll just have tests and checks to worry about

src/approaches/nsrt_rl_approach.py
ashayathalye: j is used in line 142.

src/approaches/nsrt_rl_approach.py
ashayathalye: Design choice: I thought I'd have the RLOptionLearner hold the history. So you send the RLOptionLearner the latest interaction results at each cycle, and it saves them.

ronuchit (approved):

PR #889, opened 2022-05-11
implement degenerate distribution regressor

tomsilver:
I'm not sure if we will end up using this, but I'm using it to debug sampler learning in the doors environment.


tomsilver:
since this should not affect any existing code, I'm going to merge

tomsilver:
since this should not affect any existing code, I'm going to merge

PR #892, opened 2022-05-12
implement oracle opendoor sampler for doors

tomsilver:


src/envs/doors.py
```diff
@@ -77,10 +77,12 @@ def __init__(self) -> None:
             terminal=self._MoveToDoor_terminal)
         self._OpenDoor = ParameterizedOption(
             "OpenDoor",
-            types=[self._robot_type, self._door_type],
-            # No parameters, since the right rotation is a deterministic
-            # function of the door state.
-            params_space=Box(0, 1, (0, )),
+            types=[self._door_type, self._robot_type],
```
tomsilver, line 80: had to swap this order because the oracle sampler learner expects the arguments in alphabetic order 🙄 one day maybe we can make it more robust...

PR #895, opened 2022-05-14
New screws pickup env

NishanthJKumar:
A simple environment for which we need to sideline something other than `NextTo` in order to do well. This is inspired by a claw-game type situation where the gripper is so big that it picks up multiple objects at once.

Video of successful execution:

https://user-images.githubusercontent.com/12446934/168487539-29acbc92-d3f5-4679-947a-dae602f11cf4.mp4

### DIFF
--- a/src/envs/screws.py
+++ b/src/envs/screws.py
@@ -22,7 +22,7 @@ class ScrewsEnv(BaseEnv):
     irrelevant screws that might be picked up when picking a desired
     screw.
     """
-    _gripper_width: ClassVar[float] = 0.5
+    gripper_width: ClassVar[float] = 0.5
     _receptacle_width: ClassVar[float] = 0.6
     _receptacle_height: ClassVar[float] = 0.2
     _screw_width: ClassVar[float] = 0.05
@@ -64,32 +64,32 @@ class ScrewsEnv(BaseEnv):
             # variables: [robot, screw to pick up]
             # params: []
             "MoveToScrew",
-            self._MoveToScrew_policy,
+            self.MoveToScrew_policy,
             types=[self._gripper_type, self._screw_type])
         self._MoveToReceptacle: ParameterizedOption = \
             utils.SingletonParameterizedOption(
             # variables: [robot, receptacle]
             # params: []
             "MoveToReceptacle",
-            self._MoveToReceptacle_policy,
+            self.MoveToReceptacle_policy,
             types=[self._gripper_type, self._receptacle_type])
         self._MagnetizeGripper: ParameterizedOption = \
             utils.SingletonParameterizedOption(
             # variables: [robot]
             # params: []
             "MagnetizeGripper",
-            self._MagnetizeGripper_policy,
+            self.MagnetizeGripper_policy,
             types=[self._gripper_type])
         self._DemagnetizeGripper: ParameterizedOption = \
             utils.SingletonParameterizedOption(
             # variables: [robot]
             # params: []
             "DemagnetizeGripper",
-            self._DemagnetizeGripper_policy,
+            self.DemagnetizeGripper_policy,
             types=[self._gripper_type])
 
         # Static objects (always exist no matter the settings).
-        self._robot = Object("robby", self._gripper_type)
+        self.robot = Object("robby", self._gripper_type)
         self._receptacle = Object("receptacle", self._receptacle_type)
 
     def _generate_train_tasks(self) -> List[Task]:
@@ -129,9 +129,9 @@ class ScrewsEnv(BaseEnv):
                  color="green")
 
         # Draw the gripper
-        gripper_x = state.get(self._robot, "pose_x")
-        gripper_y = state.get(self._robot, "pose_y")
-        gripper_width = state.get(self._robot, "width")
+        gripper_x = state.get(self.robot, "pose_x")
+        gripper_y = state.get(self.robot, "pose_y")
+        gripper_width = state.get(self.robot, "width")
         # NOTE: this 1.3 factor is just to make the gripper wide enough so that
         # even when it grabs a block on its very edge, it doesn't look strange
         # in the visualization.
@@ -289,9 +289,9 @@ class ScrewsEnv(BaseEnv):
         ])
         # Create gripper object such that it is at the middle of
         # the screen.
-        data[self._robot] = np.array([(self.rz_x_lb + self.rz_x_ub) / 2.0,
-                                      (self.rz_y_lb + self.rz_y_ub) / 2.0,
-                                      self._gripper_width])
+        data[self.robot] = np.array([(self.rz_x_lb + self.rz_x_ub) / 2.0,
+                                     (self.rz_y_lb + self.rz_y_ub) / 2.0,
+                                     self.gripper_width])
         return State(data), goal_atoms
 
     def _surface_xy_is_valid(self, x: float, y: float,
@@ -309,9 +309,9 @@ class ScrewsEnv(BaseEnv):
 
     def _transition_move(self, state: State, x: float, y: float) -> State:
         next_state = state.copy()
-        rx = state.get(self._robot, "pose_x")
-        ry = state.get(self._robot, "pose_y")
-        rwidth = state.get(self._robot, "width")
+        rx = state.get(self.robot, "pose_x")
+        ry = state.get(self.robot, "pose_y")
+        rwidth = state.get(self.robot, "width")
         new_rx = rx + x
         new_rx_min = new_rx - rwidth / 2.0
         new_rx_max = new_rx + rwidth / 2.0
@@ -327,8 +327,8 @@ class ScrewsEnv(BaseEnv):
             new_ry = self.rz_y_lb
         if new_ry > self.rz_y_ub:
             new_ry = self.rz_y_ub
-        next_state.set(self._robot, "pose_x", new_rx)
-        next_state.set(self._robot, "pose_y", new_ry)
+        next_state.set(self.robot, "pose_x", new_rx)
+        next_state.set(self.robot, "pose_y", new_ry)
 
         # Recompute dx and dy in case it's changed.
         dx = new_rx - rx
@@ -348,14 +348,14 @@ class ScrewsEnv(BaseEnv):
 
     def _transition_magnetize(self, state: State) -> State:
         next_state = state.copy()
-        ry = state.get(self._robot, "pose_y")
+        ry = state.get(self.robot, "pose_y")
         all_screws = state.get_objects(self._screw_type)
         for screw in all_screws:
             if not self._HoldingScrew_holds(state, [screw]):
                 # Check if the gripper is in the right position to
                 # pick up this screw, and if so, move the screw to
                 # the gripper.
-                if self._ScrewPickupable_holds(state, [self._robot, screw]):
+                if self._ScrewPickupable_holds(state, [self.robot, screw]):
                     screw_height = state.get(screw, "height")
                     next_state.set(screw, "pose_y", ry - (screw_height / 2.0))
                     next_state.set(screw, "held", 1.0)
@@ -372,8 +372,8 @@ class ScrewsEnv(BaseEnv):
                 # screws onto the floor. This neglects the case where the
                 # gripper is half over the receptacle and half not.
                 screw_height = state.get(screw, "height")
-                if self._AboveReceptacle_holds(
-                        state, [self._robot, self._receptacle]):
+                if self._AboveReceptacle_holds(state,
+                                               [self.robot, self._receptacle]):
                     receptacle_y = state.get(self._receptacle, "pose_y")
                     next_state.set(screw, "pose_y",
                                    receptacle_y + (screw_height / 2.0))
@@ -406,7 +406,8 @@ class ScrewsEnv(BaseEnv):
             screw_maxx > gripper_minx and \
             screw_minx < gripper_maxx
 
-    def _AboveReceptacle_holds(self, state: State,
+    @staticmethod
+    def _AboveReceptacle_holds(state: State,
                                objects: Sequence[Object]) -> bool:
         gripper, receptacle = objects
         gripper_y = state.get(gripper, "pose_y")
@@ -448,9 +449,10 @@ class ScrewsEnv(BaseEnv):
         return screw_held <= 0.5 and screw_minx > receptacle_minx \
             and screw_maxx < receptacle_maxx
 
-    def _MoveToScrew_policy(self, state: State, memory: Dict,
-                            objects: Sequence[Object],
-                            params: Array) -> Action:
+    def MoveToScrew_policy(self, state: State, memory: Dict,
+                           objects: Sequence[Object], params: Array) -> Action:
+        """Policy to return an action that moves to a position above a
+        particular screw such that the screw can be grasped."""
         del memory, params  # unused
         _, screw = objects
         screw_x = state.get(screw, "pose_x")
@@ -461,16 +463,18 @@ class ScrewsEnv(BaseEnv):
         target_y = screw_y + (screw_height /
                               2.0) + (self._magnetic_field_dist / 2.0)
 
-        current_x = state.get(self._robot, "pose_x")
-        current_y = state.get(self._robot, "pose_y")
+        current_x = state.get(self.robot, "pose_x")
+        current_y = state.get(self.robot, "pose_y")
 
         return Action(
             np.array([target_x - current_x, target_y - current_y, 0.0],
                      dtype=np.float32))
 
-    def _MoveToReceptacle_policy(self, state: State, memory: Dict,
-                                 objects: Sequence[Object],
-                                 params: Array) -> Action:
+    def MoveToReceptacle_policy(self, state: State, memory: Dict,
+                                objects: Sequence[Object],
+                                params: Array) -> Action:
+        """Policy to return an action that moves to a position above the
+        receptacle."""
         del memory, params  # unused
         _, receptacle = objects
         receptacle_x = state.get(receptacle, "pose_x")
@@ -479,23 +483,27 @@ class ScrewsEnv(BaseEnv):
         target_x = receptacle_x
         target_y = receptacle_y + (self._magnetic_field_dist)
 
-        current_x = state.get(self._robot, "pose_x")
-        current_y = state.get(self._robot, "pose_y")
+        current_x = state.get(self.robot, "pose_x")
+        current_y = state.get(self.robot, "pose_y")
 
         return Action(
             np.array([target_x - current_x, target_y - current_y, 1.0],
                      dtype=np.float32))
 
     @staticmethod
-    def _MagnetizeGripper_policy(state: State, memory: Dict,
-                                 objects: Sequence[Object],
-                                 params: Array) -> Action:
+    def MagnetizeGripper_policy(state: State, memory: Dict,
+                                objects: Sequence[Object],
+                                params: Array) -> Action:
+        """Policy to return an action that magnetizes the gripper at its
+        current position."""
         del state, memory, objects, params  # unused
         return Action(np.array([0.0, 0.0, 1.0], dtype=np.float32))
 
     @staticmethod
-    def _DemagnetizeGripper_policy(state: State, memory: Dict,
-                                   objects: Sequence[Object],
-                                   params: Array) -> Action:
+    def DemagnetizeGripper_policy(state: State, memory: Dict,
+                                  objects: Sequence[Object],
+                                  params: Array) -> Action:
+        """Policy to return an action that demagnetizes the gripper at its
+        current position."""
         del state, memory, objects, params  # unused
         return Action(np.array([0.0, 0.0, 0.0], dtype=np.float32))
--- a/src/envs/screws.py
+++ b/src/envs/screws.py
@@ -106,8 +106,10 @@ class ScrewsEnv(BaseEnv):
 
     @property
     def predicates(self) -> Set[Predicate]:
-        return {self._ScrewPickupable, self._AboveReceptacle,
-                self._HoldingScrew, self._ScrewInReceptacle}
+        return {
+            self._ScrewPickupable, self._AboveReceptacle, self._HoldingScrew,
+            self._ScrewInReceptacle
+        }
 
     @property
     def goal_predicates(self) -> Set[Predicate]:
@@ -119,8 +121,10 @@ class ScrewsEnv(BaseEnv):
 
     @property
     def options(self) -> Set[ParameterizedOption]:
-        return {self._MoveToScrew, self._MoveToReceptacle,
-                self._MagnetizeGripper, self._DemagnetizeGripper}
+        return {
+            self._MoveToScrew, self._MoveToReceptacle, self._MagnetizeGripper,
+            self._DemagnetizeGripper
+        }
 
     @property
     def action_space(self) -> Box:
--- a/src/envs/screws.py
+++ b/src/envs/screws.py
@@ -1,0 +1,236 @@
+"""Toy gem purification environment.
+
+This environment involves picking particular ores from clutter and
+running processes on these to make particular gems. It was created to
+test our approach's ability to learn compact operator models with side-
+effects.
+"""
+
+from typing import ClassVar, Dict, List, Optional, Sequence, Set
+
+import matplotlib.pyplot as plt
+import numpy as np
+from gym.spaces import Box
+
+from predicators.src import utils
+from predicators.src.envs import BaseEnv
+from predicators.src.settings import CFG
+from predicators.src.structs import Action, Array, GroundAtom, Image, Object, \
+    ParameterizedOption, Predicate, State, Task, Type
+
+
+class ScrewsEnv(BaseEnv):
+    """Toy screw picking env.
+
+    In this environment, the agent controls a magnetic gripper that can
+    be used to pick up screws of different colors/sizes. The challennge
+    is for the agent to learn actions that don't precisely model all the
+    irrelevant screws that might be picked up when picking a desired
+    screw.
+    """
+    _gripper_width: ClassVar[float] = 0.5
+    _receptacle_width: ClassVar[float] = 0.6
+    _receptacle_height: ClassVar[float] = 0.2
+    _screw_width: ClassVar[float] = 0.05
+    _screw_height: ClassVar[float] = 0.05
+    _magnetic_field_dist: ClassVar[float] = 0.1
+    # Reachable zone boundaries.
+    rz_x_lb = -5.0
+    rz_x_ub = 5.0
+    rz_y_lb = 0.0
+    rz_y_ub = 5.0
+
+    def __init__(self) -> None:
+        super.__init__()
+        # Types
+        self._screw_type = Type(
+            "screw", ["pose_x", "pose_y", "color", "width", "height", "held"])
+        self._gripper_type = Type("gripper", ["pose_x", "pose_y", "width"])
+        self._receptacle_type = Type("receptacle",
+                                     ["pose_x", "pose_y", "width"])
+        # Predicates
+        self._ScrewPickupable = Predicate("AboveScrew",
+                                     [self._gripper_type, self._screw_type],
+                                     self._ScrewPickupable_holds)
+        self._AboveReceptacle = Predicate(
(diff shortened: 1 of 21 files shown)


tomsilver:
love the rendering! can we fix this though?

image

src/envs/screws.py
```diff
@@ -0,0 +1,501 @@
...
+from predicators.src.settings import CFG
+from predicators.src.structs import Action, Array, GroundAtom, Image, Object, \
+    ParameterizedOption, Predicate, State, Task, Type
+
+
+class ScrewsEnv(BaseEnv):
+    """Toy screw picking env.
+
+    In this environment, the agent controls a magnetic gripper that can
+    be used to pick up screws of different colors/sizes. The challennge
```
tomsilver, line 20: typo: challennge

src/envs/screws.py
```diff
@@ -0,0 +1,501 @@
...
+    screw.
+    """
+    _gripper_width: ClassVar[float] = 0.5
+    _receptacle_width: ClassVar[float] = 0.6
+    _receptacle_height: ClassVar[float] = 0.2
+    _screw_width: ClassVar[float] = 0.05
+    _screw_height: ClassVar[float] = 0.05
+    _magnetic_field_dist: ClassVar[float] = 0.5
+    _num_screw_clusters: ClassVar[int] = 3
+    num_screws_train: ClassVar[List[int]] = CFG.screws_num_screws_train
```
tomsilver, line 32: tests sometimes get screwy if we do this. suggest instead setting these fields that use `CFG` in `__init__()`

tomsilver: (no pun intended)

src/envs/screws.py
```diff
@@ -0,0 +1,501 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+        # Types
+        self._screw_type = Type(
+            "screw", ["pose_x", "pose_y", "width", "height", "held"])
+        self._gripper_type = Type("gripper", ["pose_x", "pose_y", "width"])
+        self._receptacle_type = Type("receptacle",
+                                     ["pose_x", "pose_y", "width"])
+        # Predicates
+        self._ScrewPickupable = Predicate(
```
tomsilver, line 49: is this the same thing as the screw being on the ground? if so, rename to `ScrewOnGround`?

src/envs/screws.py
```diff
@@ -0,0 +1,501 @@
...
+            "AboveReceptacle", [self._gripper_type, self._receptacle_type],
+            self._AboveReceptacle_holds)
+        self._HoldingScrew = Predicate("HoldingScrew", [self._screw_type],
+                                       self._HoldingScrew_holds)
+        self._ScrewInReceptacle = Predicate(
+            "ScrewInReceptacle", [self._screw_type, self._receptacle_type],
+            self._ScrewInReceptacle_holds)
+
+        # Options
+        self._MoveToScrew: ParameterizedOption = \
```
tomsilver, line 62: are these `: ParameterizedOption` annotations needed? (if not, could avoid the new line break too)

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            utils.SingletonParameterizedOption(
+            # variables: [robot, screw to pick up]
+            # params: []
+            "MoveToScrew",
+            self.MoveToScrew_policy,
```
tomsilver, line 67: can all these policies be made private methods?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        self._DemagnetizeGripper: ParameterizedOption = \
+            utils.SingletonParameterizedOption(
+            # variables: [robot]
+            # params: []
+            "DemagnetizeGripper",
+            self.DemagnetizeGripper_policy,
+            types=[self._gripper_type])
+
+        # Static objects (always exist no matter the settings).
+        self.robot = Object("robby", self._gripper_type)
```
tomsilver, line 92: make private?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            receptacle_x - receptacle_width / 2.0,
+            receptacle_x + receptacle_width / 2.0
+        ], [receptacle_y, receptacle_y],
+                 color="green")
+
+        # Draw the gripper
+        gripper_x = state.get(self.robot, "pose_x")
+        gripper_y = state.get(self.robot, "pose_y")
+        gripper_width = state.get(self.robot, "width")
+        # NOTE: this 1.3 factor is just to make the gripper wide enough so that
```
tomsilver, line 135: the 1.3 thing here is liable to break if we later change the dimensions of stuff. can we accomplish the same thing by changing the conditions for picking so that the entirety of the screw is below the gripper (taking into account the widths of both)?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            [gripper_y - gripper_width / 4.0, gripper_y],
+            color="black")
+
+        # Find which object is the goal screw.
+        assert len(task.goal) == 1
+        goal_pred = list(task.goal)[0]
+        goal_screw = goal_pred.objects[0]
+        assert goal_screw.type.name == 'screw'
+
+        # Draw screws
```
tomsilver, line 160: nit: add period

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            if screw == goal_screw:
+                color = 'gold'
+            else:
+                color = 'grey'
+            rect = plt.Rectangle((screw_x, screw_y),
+                                 screw_width,
+                                 screw_height,
+                                 edgecolor=color,
+                                 facecolor=color,
+                                 label=screw.name)
```
tomsilver, line 175: is this `label` argument used? it doesn't look like it

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            ax.add_patch(rect)
+
+        img = utils.fig2data(fig, dpi=150)
```
tomsilver, line 178: remove `=150` since that's the default

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        return {
+            self._MoveToScrew, self._MoveToReceptacle, self._MagnetizeGripper,
+            self._DemagnetizeGripper
+        }
+
+    @property
+    def action_space(self) -> Box:
+        # dimensions: [dx, dy, magnetized vs. unmagnetized].
+        lowers = np.array([
+            -(self.rz_x_ub - self.rz_x_lb), -(self.rz_y_ub - self.rz_y_lb), 0.0
```
tomsilver, line 208: suggest making variables like `rz_width = self.rz_x_ub - self.rz_x_lb` and reusing in `lowers` and `uppers`

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        ],
+                          dtype=np.float32)
+        uppers = np.array([(self.rz_x_ub - self.rz_x_lb),
+                           (self.rz_y_ub - self.rz_y_lb), 1.0],
+                          dtype=np.float32)
+        return Box(lowers, uppers)
+
+    def simulate(self, state: State, action: Action) -> State:
+        assert self.action_space.contains(action.arr)
+        # NOTE: currently, the agent can only pick things up or
```
tomsilver, line 218: why is this necessary? can't we always just process the move first, then process the magnetization?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+
+        if magnetization > 0.5:
+            return self._transition_magnetize(state)
+        return self._transition_demagnetize(state)
+
+    def _get_tasks(self, num_tasks: int, possible_num_screws: List[int],
+                   rng: np.random.Generator) -> List[Task]:
+        tasks = []
+
+        # Sample center locations for each of the screw clusters.
```
tomsilver, line 233: I don't care that much about this, but I'm wondering if we could get away without the notion of screw clusters, and just randomly sample positions of all the screws, period?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+                rng.uniform(self.rz_x_lb,
+                            self.rz_x_ub - self._receptacle_width))
+
+        for _ in range(num_tasks):
+            screw_name_to_pos: Dict[str, Tuple[float, float]] = {}
+            num_screws = rng.choice(possible_num_screws)
+            for si in range(num_screws):
+                existing_xys = set(screw_name_to_pos.values())
+                while True:
+                    # sample a random iniitial position for the screw
```
tomsilver, line 246: capitalize here and elsewhere. also typo: iniitial

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+                    # such that it is on the ground, not in the
```
tomsilver, line 247: doesn't "on the ground" imply "not in the receptacle"?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        for _ in range(self._num_screw_clusters):
+            screw_cluster_xcenters.append(
+                rng.uniform(self.rz_x_lb,
+                            self.rz_x_ub - self._receptacle_width))
+
+        for _ in range(num_tasks):
+            screw_name_to_pos: Dict[str, Tuple[float, float]] = {}
+            num_screws = rng.choice(possible_num_screws)
+            for si in range(num_screws):
+                existing_xys = set(screw_name_to_pos.values())
```
tomsilver, line 244: maybe just keep track of the existing xs in this loop, and worry about the ys and the names after all of the xs have been determined?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+                    y_pos = self.rz_y_lb + self._screw_height / 2.0
+                    screw_cluster_xcenter = rng.choice(screw_cluster_xcenters)
+                    x_pos = rng.normal(screw_cluster_xcenter)
+                    # if this position doesn't make this screw in collision
+                    # with another screw, then put the screw here.
+                    if self._surface_xy_is_valid(x_pos, y_pos, existing_xys):
+                        screw_name_to_pos[f"screw_{si}"] = (x_pos, y_pos)
+                        break
+            init_state, goal_atoms = \
+                self._get_init_state_and_goal_atoms_from_positions(
```
tomsilver, line 258: might as well just get the task itself?

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            [(gripper_x + gripper_width / 2.0) + (self._screw_width * 1.3),
+             (gripper_x + gripper_width / 2.0) + (self._screw_width * 1.3)],
+            [gripper_y - gripper_width / 4.0, gripper_y],
+            color="black")
+
+        # Find which object is the goal screw.
+        assert len(task.goal) == 1
+        goal_pred = list(task.goal)[0]
+        goal_screw = goal_pred.objects[0]
+        assert goal_screw.type.name == 'screw'
```
tomsilver, line 158: can probably just remove this assertion, but if you want, you can do `goal_screw.is_instance(self._screw_type)` instead

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            init_state, goal_atoms = \
+                self._get_init_state_and_goal_atoms_from_positions(
+                screw_name_to_pos, rng)
+            tasks.append(Task(init_state, goal_atoms))
+        return tasks
+
+    def _get_init_state_and_goal_atoms_from_positions(
+            self, screw_name_to_pos: Dict[str, Tuple[float, float]],
+            rng: np.random.Generator) -> Tuple[State, Set[GroundAtom]]:
+        data: Dict[Object, Array] = {}
```
tomsilver, line 266: I'm recently a fan of using `utils.create_state_from_dict` instead for this, for improved readibility. see stick button or doors for examples.

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+    rz_x_lb = -5.0
+    rz_x_ub = 5.0
+    rz_y_lb = 0.0
+    rz_y_ub = 5.0
+
+    def __init__(self) -> None:
+        super().__init__()
+        # Types
+        self._screw_type = Type(
+            "screw", ["pose_x", "pose_y", "width", "height", "held"])
```
tomsilver, line 44: if all screws are the same height and width, we probably don't need to include those in their state features

tomsilver: ditto for width in the next two types

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        return State(data), goal_atoms
+
+    def _surface_xy_is_valid(self, x: float, y: float,
+                             existing_xys: Set[Tuple[float, float]]) -> bool:
+        # First, check that there isn't another screw at this location.
+        for (existing_x, existing_y) in existing_xys:
+            if (abs(existing_x - x) < self._screw_width
+                    and abs(existing_y - y) < self._screw_height):
+                return False
+        # Next, check that the location is valid.
```
tomsilver, line 304: shouldn't this part be guaranteed by the bounds with which you're sampling? if not, change those bounds

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        y_is_valid = y == self.rz_y_lb + self._screw_height / 2.0
+        x_is_valid = (self.rz_x_lb + self._screw_width / 2.0) <= x <= (
+            self.rz_x_ub - self._screw_width / 2.0)
+        return x_is_valid and y_is_valid
+
+    def _transition_move(self, state: State, x: float, y: float) -> State:
```
tomsilver, line 310: these inputs should be `dx` and `dy`

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        new_rx = rx + x
+        new_rx_min = new_rx - rwidth / 2.0
+        new_rx_max = new_rx + rwidth / 2.0
+        new_ry = ry + y
+
+        # If the new position is outside the boundaries, move till
+        # the robot collides with the boundary.
+        if new_rx_min < self.rz_x_lb:
+            new_rx = self.rz_x_lb + (rwidth / 2.0)
+        if new_rx_max > self.rz_x_ub:
```
tomsilver, line 324: elif

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+            new_rx = self.rz_x_ub - (rwidth / 2.0)
+        if new_ry < self.rz_y_lb:
+            new_ry = self.rz_y_lb
+        if new_ry > self.rz_y_ub:
```
tomsilver, line 328: elif

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        self._receptacle_type = Type("receptacle",
+                                     ["pose_x", "pose_y", "width"])
+        # Predicates
+        self._ScrewPickupable = Predicate(
+            "ScrewPickupable", [self._gripper_type, self._screw_type],
+            self._ScrewPickupable_holds)
+        self._AboveReceptacle = Predicate(
+            "AboveReceptacle", [self._gripper_type, self._receptacle_type],
+            self._AboveReceptacle_holds)
+        self._HoldingScrew = Predicate("HoldingScrew", [self._screw_type],
```
tomsilver, line 55: i know there's only one robot, but if the other predicates have the robot as an argument, it makes sense for this one to also

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        ry = state.get(self.robot, "pose_y")
+        all_screws = state.get_objects(self._screw_type)
+        for screw in all_screws:
+            if not self._HoldingScrew_holds(state, [screw]):
+                # Check if the gripper is in the right position to
+                # pick up this screw, and if so, move the screw to
+                # the gripper.
+                if self._ScrewPickupable_holds(state, [self.robot, screw]):
+                    screw_height = state.get(screw, "height")
+                    next_state.set(screw, "pose_y", ry - (screw_height / 2.0))
```
tomsilver, line 360: probably related to my rendering gripe, if `ry` is the center of the gripper, this calculation doesn't look quite right

src/envs/screws.py
```diff
@@ -0,0 +1,509 @@
...
+        return next_state
+
+    def _transition_demagnetize(self, state: State) -> State:
+        next_state = state.copy()
+        all_screws = state.get_objects(self._screw_type)
+        for screw in all_screws:
+            if self._HoldingScrew_holds(state, [screw]):
+                # NOTE: This currently will drop all screws into the receptacle
+                # iff AboveReceptacle is true. Otherwise, it will drop all
+                # screws onto the floor. This neglects the case where the
```
tomsilver, line 372: hmm yeah not ideal. can we change the meaning of AboveReceptacle so that the gripper must be fully over it?

tests/envs/test_screws.py
```diff
@@ -0,0 +1,64 @@
+"""Test cases for the cover environment."""
```
tomsilver, line 1: screws environment

src/planning.py
```diff
@@ -291,6 +291,7 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
     discovered_failures: List[Optional[_DiscoveredFailure]] = [
         None for _ in skeleton
     ]
+
```
tomsilver, line 294: revert

tomsilver (commented):
awesome work!! i'm really happy that we can learn operators for an environment like this :)

src/envs/screws.py
NishanthJKumar: no - this means that the gripper is hovering over the screw and close enough to pick it up. I originally called it `GripperAboveScrew`, but the problem is that it's not just that, but also that the gripper is close enough to magnetize. Maybe I'll call it `GripperCanPickScrew`?

src/envs/screws.py
NishanthJKumar: so the problem is that I call the `MoveToScrew_policy` in a test. Any thoughts on a good way to work around this?

src/envs/screws.py
NishanthJKumar: also need in a test...

src/envs/screws.py
NishanthJKumar: i did that initially, and it led to things generally being grouped further apart. I like the notion of clusters simply because it makes it more likely that we're going to see a lot of different number of screws being held when collecting training data.

src/envs/screws.py
NishanthJKumar: good point; this note is deprecated...

src/envs/screws.py
NishanthJKumar: i think it already does this, so this comment isn't actually accurate

src/envs/screws.py
NishanthJKumar: wait why? I feel like this is exactly right (`ry` is the center of the gripper, and `"pose_y"` is the center of the block, so we want to set `"pose_y"` to `ry - half_of_screw_height`)

src/envs/screws.py
NishanthJKumar: since we're using a normal distribution to sample, we can't enforce bounds

src/envs/screws.py
NishanthJKumar: i like the current structure; i think doing the y's separately looks clunky, and this way, we could easily extend to put things in different y positions as well in the future if necessary!

src/envs/screws.py
tomsilver: got it, yeah `GripperCanPickScrew` sounds good

src/envs/screws.py
tomsilver: what we usually do is get the option instead from `env.options` and use the policy through the option, like here: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/envs/test_doors_env.py#L188

but if you need the policy specifically, you can create a subclass of the main environment that exposes things, like here: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/envs/test_pybullet_blocks.py#L14

src/envs/screws.py
tomsilver: you can get the robot from a state, like here: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/envs/test_doors_env.py#L55

src/envs/screws.py
tomsilver: okay sounds good!

src/envs/screws.py
tomsilver: ah right. FWIW, you could use a truncated normal, but this is fine too.

src/envs/screws.py
tomsilver: don't you need to take into account the height of the gripper?

NishanthJKumar: The gripper is a flat line and so doesn't have a height - I think during rendering it's given a height because of the line thickness

tomsilver: ah, got it. in that case, you can just change the rendering. probably easier to make it a rectangle

tomsilver:
I notice when running oracle that the first skeleton often doesn't succeed, and it needs 2 skeletons to work; is that expected? @NishanthJKumar

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+        lowers = np.array([-self._rz_width, -self._rz_height, 0.0],
+                          dtype=np.float32)
+        uppers = np.array([self._rz_width, self._rz_height, 1.0],
+                          dtype=np.float32)
+        return Box(lowers, uppers)
+
+    def simulate(self, state: State, action: Action) -> State:
+        assert self.action_space.contains(action.arr)
+        dx, dy, magnetization = action.arr
+        if dx != 0.0 or dy != 0.0:
```
tomsilver, line 203: would anything be lost if we _always_ transition move, and then also transition (de)magnetize? if we ever try to learn options in this environment, the != 0.0 thing here is going to be an issue

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+        # Sample center locations for each of the screw clusters.
+        screw_cluster_xcenters = []
+        for _ in range(self._num_screw_clusters):
+            screw_cluster_xcenters.append(
+                rng.uniform(self.rz_x_lb,
+                            self.rz_x_ub - self._receptacle_width))
+
+        for _ in range(num_tasks):
+            screw_name_to_pos: Dict[str, Tuple[float, float]] = {}
+            num_screws = rng.choice(possible_num_screws)
```
tomsilver, line 223: should this be `rng.integers(1, possible_num_screws+1)`? (a) i don't think we'd want to allow for no screws and (b) we should be able to sample `possible_num_screws`, based on the name

NishanthJKumar:
Really - that's actually quite unexpected; lemme look into this.

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+            return self._transition_magnetize(state)
+        return self._transition_demagnetize(state)
+
+    def _get_tasks(self, num_tasks: int, possible_num_screws: List[int],
+                   rng: np.random.Generator) -> List[Task]:
+        tasks = []
+
+        # Sample center locations for each of the screw clusters.
+        screw_cluster_xcenters = []
+        for _ in range(self._num_screw_clusters):
```
tomsilver, line 216: why are these clusters sampled once for all tasks? wouldn't it make more sense to resample for each task?

NishanthJKumar:
confirmed i can replicate this behavior; investigating what the first skeleton is/why it's failing rn.

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+
+        # Create gripper object such that it is at the middle of
+        # the screen.
+        state_dict[self._robot] = {
+            "pose_x": (self.rz_x_lb + self.rz_x_ub) / 2.0,
+            "pose_y": (self.rz_y_lb + self.rz_y_ub) / 2.0
+        }
+
+        init_state = utils.create_state_from_dict(state_dict)
+        task = Task(init_state, goal_atoms)
```
tomsilver, line 281: add an assertion that `len(goal_atoms) == 1`

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+        return tasks
+
+    def _get_single_task(self, screw_name_to_pos: Dict[str, Tuple[float,
+                                                                  float]],
+                         rng: np.random.Generator) -> Task:
+        state_dict = {}
+        goal_atoms: Set[GroundAtom] = set()
+        # Select a random screw that needs to be in the receptacle as the goal
+        # GroundAtom.
+        goal_screw_name = rng.choice(list(screw_name_to_pos.keys()))
```
tomsilver, line 248: depending on how the dictionary is built, this may be nondeterministic. let's change the `list` to `sorted`

tomsilver: also, you can remove the `.keys()`. so:
`goal_screw_name = rng.choice(sorted(screw_name_to_pos))`

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def _surface_xy_is_valid(self, x: float, y: float,
+                             existing_xys: Set[Tuple[float, float]]) -> bool:
+        # First, check that there isn't another screw at this location.
+        for (existing_x, existing_y) in existing_xys:
+            if (abs(existing_x - x) < self._screw_width
+                    and abs(existing_y - y) < self._screw_height):
+                return False
+        # Next, check that the location is valid.
+        y_is_valid = y == self.rz_y_lb + self._screw_height / 2.0
```
tomsilver, line 292: can you add parentheses around the right hand sides here for easier reading?

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+                next_state.set(screw, "held", 0.0)
+
+        return next_state
+
+    def _GripperCanPickScrew_holds(self, state: State,
+                                   objects: Sequence[Object]) -> bool:
+        gripper, screw = objects
+        gripper_y = state.get(gripper, "pose_y")
+        screw_y = state.get(screw, "pose_y")
+        screw_maxy = screw_y + self._screw_height / 2.0
```
tomsilver, line 374: nit: add underscores after the 'max' and 'min' here and below

tests/envs/test_screws.py
```diff
@@ -0,0 +1,72 @@
...
+        demagnetize_gripper_option
+    ])
+    traj = utils.run_policy_with_simulator(
+        policy,
+        env.simulate,
+        task.init,
+        lambda _: False,
+        max_num_steps=1000,
+        exceptions_to_break_on={utils.OptionExecutionFailure})
+    assert traj.states[-1].allclose(traj.states[-3])
```
tomsilver, line 72: how does this assertion check what's described in the comment?

tomsilver (commented):
looking good! I think we should be good after 1 more round

src/envs/screws.py
NishanthJKumar: yup - we can definitely do this and not lose anything! I'll go ahead and make this change

src/envs/screws.py
NishanthJKumar: good point! my bad; i missed this

tests/envs/test_screws.py
NishanthJKumar: added comment describing this

src/envs/screws.py
NishanthJKumar: wait - `possible_num_screws` is a list that's passed in from `settings.py` (e.g. `[15, 20]`). I was intending this line to select a particular number from this list, which i believe is what it currently does

tomsilver: oh sorry, my bad, ignore!

NishanthJKumar:
I ran `backchaining` and `cluster_and_intersect` strips learning just to verify. Results are that we solve `50/50` test tasks with backchaining and recover the ground truth operators. We solve `0/50` with cluster_and_intersect because planning always times out in grounding!

src/envs/screws.py
```diff
@@ -0,0 +1,482 @@
...
+                     action: Optional[Action] = None,
+                     caption: Optional[str] = None) -> List[Image]:
+        fig, ax = plt.subplots(1, 1)
+        plt.xlim([
+            self.rz_x_lb - 2 * self._screw_width,
+            self.rz_x_ub + 2 * self._screw_width
+        ])
+        plt.ylim([self.rz_y_lb, self.rz_y_ub])
+
+        # Draw receptacle
```
tomsilver, line 121: nit: add periods after this and all comments

tomsilver (approved):
LGTM, awesome work!

PR #896, opened 2022-05-15
simplify dataset filename for included options

tomsilver:
this change is due to very long filenames occurring in behavior when all options are included. this PR should fix that


wmcclinton (approved):
Cool! Looks good. Thx

PR #897, opened 2022-05-15
added Behavior state

wmcclinton:


tomsilver (approved):
two minor things, otherwise lgtm!

PR #898, opened 2022-05-16
minor change in requirements-dev to specify yapf version

NishanthJKumar:
Specifies the version of `yapf`, because @wmcclinton's system happened to `pip install` an older version, which caused all sorts of weird formatting changes. We want to avoid this for any new users in the future!


tomsilver (approved):
good catch!

PR #903, opened 2022-05-16
fix to doors action space

tomsilver:
fixes a crash with oracle options. the old action bounds didn't make sense, they were left over from a time where I was thinking about door handle rotations literally...


ronuchit (approved):
will this affect the other methods?

tomsilver:
yeah it might lead to improvements for the other methods too? we'll see

tomsilver:
partial supercloud results
doors_num_train_tasks_perc_solved

tomsilver:
partial supercloud results
doors_num_train_tasks_perc_solved

PR #909, opened 2022-05-16
simplifying changes to settings.py

### LINKED ISSUE #904: fix segmentation parameters in settings.py
we don't need this anymore:
```
            # Segmentation parameters.
            segmenter=defaultdict(
                lambda: "atom_changes",
                {
                    # When options are given, use them to segment instead.
                    "no_learning": "option_changes",
                })[args.get("option_learner", "no_learning")],
```

ronuchit:
closes #904

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -33,6 +33,9 @@ class GlobalSettings:
     pretty_print_when_loading = False
     # Used for random seeding in test environment.
     test_env_seed_offset = 10000
+    # The method to use for segmentation. By default, segment using options.
+    # If you are learning options, you should change this via the command line.
+    segmenter = "option_changes"
 
     # cover env parameters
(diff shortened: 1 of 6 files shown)


src/settings.py
```diff
@@ -33,6 +33,9 @@ class GlobalSettings:
     pretty_print_when_loading = False
     # Used for random seeding in test environment.
     test_env_seed_offset = 10000
+    # The method to use for segmentation. By default, segment using options.
```
tomsilver, line 36: @ashayathalye note this change (we should be using `--segmenter contacts`)

ronuchit: (it will crash if you don't make this change, so at least it won't silently do the wrong thing)

tomsilver (approved):

PR #910, opened 2022-05-16
separate online and offline transitions in metrics

### LINKED ISSUE #893: create or derive metric for number of online interaction transitions
NOT including demonstration transitions (which are included in "NUM_TRANSITIONS").

tomsilver:
closes #893 

@amburger66 this change is going to break the plotting script for any results that were previously generated. to plot old results, we could modify the script locally as needed. or, we can make this change backward compatible, but i'm worried then that we're going to get confused about what "num transitions" means... let me know what you think

### DIFF
--- a/scripts/plotting/create_num_demos_plots.py
+++ b/scripts/plotting/create_num_demos_plots.py
@@ -133,11 +133,7 @@ def _main() -> None:
                                 yerr=y_stds,
                                 label=label,
                                 marker=marker)
-                # Automatically make x ticks integers for certain X KEYS.
-                if x_key == "CYCLE":
-                    ax.xaxis.set_major_locator(MaxNLocator(integer=True))
-                elif x_key == "NUM_TRAIN_TASKS":
-                    ax.set_xticks(xs)
+                ax.set_xticks(xs)
                 ax.set_title(plot_title)
                 ax.set_xlabel(x_label)
                 ax.set_ylabel(y_label)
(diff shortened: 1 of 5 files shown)


src/main.py
```diff
@@ -147,7 +149,7 @@ def _run_pipeline(env: BaseEnv,
                 continue
             logging.info(f"\n\nONLINE LEARNING CYCLE {i}\n")
             logging.info("Getting interaction requests...")
-            if total_num_transitions > CFG.online_learning_max_transitions:
+            if num_online_transitions > CFG.online_learning_max_transitions:
```
tomsilver, line 152: @amburger66 note: we're capping _online_ transitions now, so we're going to be running the scripts for longer. I think this is more intuitive?

scripts/plotting/create_num_demos_plots.py
```diff
@@ -135,7 +134,7 @@ def _main() -> None:
                                 label=label,
                                 marker=marker)
                 # Automatically make x ticks integers for certain X KEYS.
-                if x_key in ("CYCLE", "NUM_TRANSITIONS"):
+                if x_key in ("CYCLE", "NUM_OFFLINE_TRANSITIONS"):
```
ronuchit, line 137: will this script still work correctly? it looks like you didn't add `NUM_OFFLINE_TRANSITIONS` to `X_KEY_AND_LABEL` above

tomsilver: fixed

src/main.py
amburger66: makes sense

amburger66 (approved):

amburger66:
> @amburger66 this change is going to break the plotting script for any results that were previously generated. to plot old results, we could modify the script locally as needed. or, we can make this change backward compatible, but i'm worried then that we're going to get confused about what "num transitions" means... let me know what you think

i think this is fine. thanks!

PR #911, opened 2022-05-17
remove input bias for sampler and option learning

### LINKED ISSUE #891: investigate removing the bias terms for sampler and option learning
some very preliminary evidence suggests removing those terms might help, or at least be unnecessary

tomsilver:
closes #891 

@ronuchit I guess we should also do a nightly on this?


ronuchit:
changes look good to me, and yeah we should do a nightly. should this be a draft PR unless it actually improves things?

tomsilver:
results look largely the same, maybe a smidge better in the nightly cover with bias, so we'll keep the code as it is now.

PR #914, opened 2022-05-17
nostdout() utility

ronuchit:
copied in from pddlgym


tomsilver (approved):
Bleh I hate this thing but ok. Can we at least say in the doc string that this should only be used when calling third party code? Otherwise it's just confusing

PR #921, opened 2022-05-17
use job arrays for supercloud

ronuchit:


tomsilver:
@amburger66 FYI, this addresses the supercloud email that you pointed out, and also affects your scripts

tomsilver (approved):
thank you v much for doing this 👼

PR #926, opened 2022-05-18
rewrite sampler dataset creation, cap amount of negative data generated

NishanthJKumar:
Verified that the command:

```
python src/main.py --env repeated_nextto_single_option --num_train_tasks 50 --experiment_id repeated_nextto_single_option_cluster_and_intersect_50demo --approach nsrt_learning --strips_learner cluster_and_intersect --disable_harmlessness_check True --seed 461
```

runs smoothly locally


src/nsrt_learning/sampler_learning.py
```diff
@@ -274,9 +296,14 @@ def _create_sampler_data(
                     continue
                 # Add this datapoint to the negative data.
                 negative_data.append((state, sub, option, goal))
-    logging.info(f"Generated {len(positive_data)} positive and "
-                 f"{len(negative_data)} negative examples")
-    assert len(positive_data) == len(datastores[datastore_idx])
+
+                if len(negative_data
+                       ) >= CFG.sampler_learning_max_negative_data:
```
ronuchit, line 301: nit: change >= to > here so that yapf doesn't make formatting weird

NishanthJKumar: can we keep the `>=`? my test kind depends on it...

src/nsrt_learning/sampler_learning.py
```diff
@@ -220,8 +222,36 @@ def _create_sampler_data(
     datastore_idx: int
 ) -> Tuple[List[SamplerDatapoint], List[SamplerDatapoint]]:
     """Generate positive and negative data for training a sampler."""
-    positive_data = []
-    negative_data = []
+    # Populate all positive data.
+    positive_data: List[SamplerDatapoint] = []
+    for (segment, var_to_obj) in datastores[datastore_idx]:
+        # Note: it should ALWAYS be the case that the segment has
```
ronuchit, line 228: let's remove this Note and the duplicated one below, they're pretty out of date

src/nsrt_learning/sampler_learning.py
```diff
@@ -220,8 +222,36 @@ def _create_sampler_data(
...
+        # an option here. If options are learned, then earlier calls
+        # to update_segment_from_option_spec() should have set the
+        # option correctly to a learned one.
+        option = segment.get_option()
+        state = segment.states[0]
+        if CFG.sampler_learning_use_goals:
+            # Right now, we're making the assumption that all data is
+            # demonstration data when we're learning samplers with goals.
+            # In the future, we may weaken this assumption.
+            assert segment.has_goal()
```
ronuchit, line 238: remove this line, it already will get checked by get_goal()

ronuchit: also remove it in the other place it appears

ronuchit (commented):
let @tomsilver look before merging

src/nsrt_learning/sampler_learning.py
```diff
@@ -220,8 +222,36 @@ def _create_sampler_data(
...
+            goal = segment.get_goal()
+        else:
+            goal = None
+        assert all(
```
tomsilver, line 242: add a similar assertion for the effects?

src/nsrt_learning/sampler_learning.py
```diff
@@ -245,22 +275,21 @@ def _create_sampler_data(
...
+                # we can continue.
                 if idx == datastore_idx:
-                    actual_grounding = [var_to_obj[var] for var in variables]
-                    if grounding == actual_grounding:
-                        assert all(
-                            pre.predicate.holds(
-                                state, [var_to_obj[v] for v in pre.variables])
-                            for pre in preconditions)
-                        positive_data.append((state, var_to_obj, option, goal))
+                    positive_grounding = [var_to_obj[var] for var in variables]
```
tomsilver, line 290: move this outside the for loop (one level)?

tomsilver (approved):
lgtm nice work!

tomsilver:
also we should do a nightly run after this

src/nsrt_learning/sampler_learning.py
NishanthJKumar: good catch!

src/nsrt_learning/sampler_learning.py
NishanthJKumar: you mean assert that all the add effects hold in the `segment.states[1]`?

src/nsrt_learning/sampler_learning.py
ronuchit: note: actually this is wrong, reverting the change. this line only makes sense if `idx == datastore_idx` so it needs to be here

tomsilver: I don't get it but okay 🙃

ronuchit: same, let's not ask too many questions

ronuchit: the real explanation is that `variables` is only associated with `datastores[datastore_idx]`, so other datastores may have entirely different variables, and so this line doesn't make sense for them

PR #927, opened 2022-05-18
updates to option plotting script

tomsilver:
* changes PickPlace1D to Cover (because it's not 1D)
* removes oracle and action policy from plots


ronuchit (commented):
cool, and do you still have a plan to run the action policy ones, or are we forgetting about that?

ronuchit (approved):

tomsilver:
yeah I am running them with num train tasks = 1000 only, and they won't appear in these plots

PR #928, opened 2022-05-18
add a strong timeout to prediction error hill climbing baseline

ronuchit:
we can try running this as is, but at least based on my local testing, i think we'll still have issues for rnt_painting, because the call to recompute_datastores_from_segments is still very expensive, and i don't know how to get around it.

works well on painting and screws, works decently on rnt_single_option (~40/50)


tomsilver (approved):

NishanthJKumar (approved):
LGTM! Nice work

tomsilver:
>we can try running this as is, but at least based on my local testing, i think we'll still have issues for rnt_painting, because the call to recompute_datastores_from_segments is still very expensive, and i don't know how to get around it.

should we resolve this first then?

ronuchit:
no, i want to let it run, it might be able to finish recompute_datastores_from_segments given enough time (like overnight maybe)

tomsilver (approved):

PR #936, opened 2022-05-20
coffee environment

### LINKED ISSUE #859: implement pour/shake/grind environment
table top environment with pouring, shaking a salt shaker, grinding pepper. to illustrate actually interesting skills

tomsilver:
closes #859 

NSRT learning looks good, but I'm not sure yet about option learning. we can do an initial review and tweak later if needed.

PyBulletification is coming soon. in the meantime, here are some crude renderings:
```python src/main.py --env coffee --approach oracle --seed 0 --coffee_render_dpi 150 --video_fps 10 --make_test_videos```
* https://user-images.githubusercontent.com/2816502/169549775-dcd4d013-e64c-4da7-98f4-87d3f3bcea75.mp4
* https://user-images.githubusercontent.com/2816502/169550071-4d328c0c-2a32-4840-a146-047f0b19d2ab.mp4
* https://user-images.githubusercontent.com/2816502/169550497-7f36c18d-e632-4591-af1a-882b9c3afaa2.mp4

### DIFF
--- a/src/envs/coffee.py
+++ b/src/envs/coffee.py
@@ -122,7 +122,31 @@ class CoffeeEnv(BaseEnv):
                      task: Task,
                      action: Optional[Action] = None,
                      caption: Optional[str] = None) -> List[Image]:
-        raise NotImplementedError("TODO")
+        del caption  # unused
+        # A crude top-down rendering.
+        figsize = (self.x_ub - self.x_lb, self.y_ub - self.y_lb)
+        fig, ax = plt.subplots(1, 1, figsize=figsize)
+        # Draw the cups.
+        for cup in state.get_objects(self._cup_type):
+            # TODO make color indicate filled level
+            color = "salmon"
+            x = state.get(cup, "x")
+            y = state.get(cup, "y")
+            circ = utils.Circle(x, y, self.cup_radius)
+            circ.plot(ax, facecolor=color, edgecolor="black", alpha=0.75)
+        # Draw the machine.
+        # TODO
+        # Draw the jug.
+        # TODO
+        # Draw the robot.
+        # TODO
+        ax.set_xlim(self.x_lb, self.x_ub)
+        ax.set_ylim(self.y_lb, self.y_ub)
+        ax.axis("off")
+        plt.tight_layout()
+        img = utils.fig2data(fig)
+        plt.close()
+        return [img]
 
     def _get_tasks(self, num: int, num_cups_lst: List[int],
                    rng: np.random.Generator) -> List[Task]:
--- a/src/envs/coffee.py
+++ b/src/envs/coffee.py
@@ -97,8 +97,23 @@ class CoffeeEnv(BaseEnv):
         self._InMachine = Predicate("InMachine",
                                     [self._jug_type, self._machine_type],
                                     self._InMachine_holds)
-        self._MachineIsOn = Predicate("MachineIsOn", [self._machine_type],
-                                      self._MachineIsOn_holds)
+        self._MachineOn = Predicate("MachineOn", [self._machine_type],
+                                    self._MachineOn_holds)
+        self._OnTable = Predicate("OnTable", [self._jug_type],
+                                  self._OnTable_holds)
+        self._HandEmpty = Predicate("HandEmpty", [self._robot_type],
+                                    self._HandEmpty_holds)
+        self._JugFilled = Predicate("JugFilled", [self._jug_type],
+                                    self._JugFilled_holds)
+        self._AboveCup = Predicate(
+            "AboveCup", [self._robot_type, self._jug_type, self._cup_type],
+            self._AboveCup_holds)
+        self._NotAboveCup = Predicate("NotAboveCup",
+                                      [self._robot_type, self._jug_type],
+                                      self._NotAboveCup_holds)
+        self._PressingButton = Predicate(
+            "PressingButton", [self._robot_type, self._machine_type],
+            self._PressingButton_holds)
         # TODO
 
         # Options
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: tests/envs/test_coffee.py [hunk 1]
```diff
@@ -0,0 +1,337 @@
...
+from predicators.src.envs.coffee import CoffeeEnv
+from predicators.src.structs import Action, GroundAtom
+
+
+def test_coffee():
+    """Tests for CoffeeEnv()."""
+    utils.reset_config({
+        "env": "coffee",
+        "coffee_num_cups_test": [4],  # used to assure 4 cups in custom state
+        "video_fps": 10,  # for faster debugging videos,
```
ronuchit, line 16: extra comma at the end?

src/envs/coffee.py
```diff
@@ -0,0 +1,756 @@
...
+            "cup",
+            ["x", "y", "capacity_liquid", "target_liquid", "current_liquid"])
+
+        # Predicates
+        self._CupFilled = Predicate("CupFilled", [self._cup_type],
+                                    self._CupFilled_holds)
+        self._Holding = Predicate("Holding",
+                                  [self._robot_type, self._jug_type],
+                                  self._Holding_holds)
+        self._InMachine = Predicate("InMachine",
```
ronuchit, line 104: kind of weird that this isn't called `JugInMachine` when the other one is called `JugFilled`

src/envs/coffee.py
```diff
@@ -0,0 +1,756 @@
...
+        y = np.clip(state.get(self._robot, "y") + dy, self.y_lb, self.y_ub)
+        z = np.clip(state.get(self._robot, "z") + dz, self.z_lb, self.z_ub)
+        current_tilt = state.get(self._robot, "tilt")
+        tilt = np.clip(current_tilt + dtilt, self.tilt_lb, self.tilt_ub)
+        current_fingers = state.get(self._robot, "fingers")
+        fingers = np.clip(current_fingers + dfingers, self.closed_fingers,
+                          self.open_fingers)
+        # The deltas may be outdated because of the clipping, so recompute
+        # or delete them.
+        dx = x - state.get(self._robot, "x")
```
ronuchit, line 189: we only need dx, dy here because we want to apply the same change to the jug if it is held? but then why don't we need dz too?

src/envs/coffee.py
```diff
@@ -0,0 +1,756 @@
...
+    cup_init_y_ub: ClassVar[float] = jug_init_y_ub
+    cup_capacity_lb: ClassVar[float] = 0.075 * (z_ub - z_lb)
+    cup_capacity_ub: ClassVar[float] = 0.15 * (z_ub - z_lb)
+    cup_target_frac: ClassVar[float] = 0.75  # fraction of the capacity
+    # Simulation settings.
+    pour_x_offset: ClassVar[float] = 1.5 * (cup_radius + jug_radius)
+    pour_y_offset: ClassVar[float] = cup_radius
+    pour_z_offset: ClassVar[float] = 1.1 * (cup_capacity_ub + jug_height - \
+                                            jug_handle_height)
+    pour_velocity: ClassVar[float] = cup_capacity_ub / 10.0
```
ronuchit, line 82: so this `/ 10.0` means it'll take on the order of 10 timesteps to finish pouring? (i know it's actually less because target is 3/4 the capacity)

src/envs/coffee.py
```diff
@@ -0,0 +1,756 @@
...
+            return self._get_move_action(button_pos, robot_pos)
+        # Move only in the z direction.
+        return self._get_move_action((x, y, self.button_z), robot_pos)
+
+    def _TurnMachineOn_terminal(self, state: State, memory: Dict,
+                                objects: Sequence[Object],
+                                params: Array) -> bool:
+        del memory, params  # unused
+        _, machine = objects
+        return self._MachineOn_holds(state, [machine])
```
ronuchit, line 660: why is the termination condition not PressingButton? actually in general i'm confused about the relationship between PressingButton and MachineOn

src/envs/coffee.py
```diff
@@ -0,0 +1,756 @@
...
+                              jug: Object) -> Tuple[float, float, float]:
+        target_x = state.get(jug, "x") + self.jug_handle_x_offset
+        target_y = state.get(jug, "y") + self.jug_handle_y_offset
+        target_z = self.jug_handle_height
+        return (target_x, target_y, target_z)
+
+    def _get_jug_z(self, state: State, jug: Object) -> float:
+        if state.get(jug, "is_held") > 0.5:
+            # Offset to account for handle.
+            return state.get(self._robot, "z") - self.jug_handle_height
```
ronuchit, line 718: i thought the jug's handle is in the y direction not the z direction

src/settings.py
```diff
@@ -153,6 +153,11 @@ class GlobalSettings:
     doors_birrt_smooth_amt = 50
     doors_draw_debug = False
 
+    # coffee env parameters
+    coffee_num_cups_train = [1, 2]
+    coffee_num_cups_test = [3, 4, 5, 6]
```
ronuchit, line 158: very cool that we're able to get up to 6 cups. how fast is oracle?

ronuchit (approved):
you are VERY FAST at making environments 🙇 🙇 🙇

src/envs/coffee.py
tomsilver: yeah

src/envs/coffee.py
tomsilver: PressingButton would work too. the button is pressed and the machine turns on at the same time. the difference is that MachineOn holds forever after, but PressingButton is deleted when the robot moves away.

src/envs/coffee.py
tomsilver: both!

src/settings.py
tomsilver: yup!
```python src/main.py --env coffee --approach oracle --seed 0 --coffee_num_cups_test "[6]"```
...
```
Tasks solved: 50 / 50
Average time for successes: 0.34843 seconds
```

PR #937, opened 2022-05-20
updates to coffee environment to make option learning work

tomsilver:


ronuchit (approved):

tomsilver:
the results for the env now are looking good for the main approach, but also good for some of the baselines. I'm going to merge this and then try another round of edits to try to make a gap.

PR #940, opened 2022-05-22
add eval experiment for GNN with train number of objects

tomsilver:


ronuchit (approved):
this is fine, but i'm wondering if we instead want to make a flag called `use_train_envs_for_eval`? (default to False obviously)

tomsilver:
> this is fine, but i'm wondering if we instead want to make a flag called `use_train_envs_for_eval`? (default to False obviously)

that would be tricky because I don't want to use literally the same train tasks (just the number of objects in the train tasks). there are ways we could do it, but I'll just stick with this PR for now

Issue #730, opened 2022-04-17
interface for collecting human demonstrations


tomsilver:

ronuchit:
what is this?

PR #942, opened 2022-05-23
new methods to expose plt in rendering functions

ronuchit:
now we have: `render_state_plt`, `render_state`, `render_plt`, and `render`, but most environments will only have to define the first, since `render_state` by default calls `utils.fig2data` on the return value of `render_state_plt`, and `render_plt` is to `render_state_plt` as `render` is to `render_state`

a step toward #730

### DIFF
--- a/src/envs/base_env.py
+++ b/src/envs/base_env.py
@@ -3,12 +3,15 @@
 import abc
 from typing import List, Optional, Set
 
+import matplotlib
+import matplotlib.pyplot as plt
 import numpy as np
 from gym.spaces import Box
 
 from predicators.src.settings import CFG
 from predicators.src.structs import Action, DefaultState, DefaultTask, Image, \
     ParameterizedOption, Predicate, State, Task, Type
+from predicators.src import utils
 
 
 class BaseEnv(abc.ABC):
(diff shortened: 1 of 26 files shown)


src/envs/base_env.py
```diff
@@ -87,22 +90,62 @@ def action_space(self) -> Box:
...
+        """Render a state and action into a Matplotlib figure.
+
+        Like simulate, this function is not meant to be part of the
+        "final system", where the environment is the real world. It is
+        just for convenience, e.g., in test coverage.
+
+        For environments which don't use Matplotlib for rendering, this
+        function should be overriden to simply crash.
+
+        NOTE: Users of this method must remember to call `plt.close()`,
```
tomsilver, line 108: if you wanna be fancy, we could try something like https://stackoverflow.com/questions/63508841/using-with-statement-in-python-to-auto-close-a-pyplot-figure

tests/test_utils.py
```diff
@@ -5,10 +5,10 @@
 from typing import Iterator, Tuple
 from typing import Type as TypingType
 
+import matplotlib.pyplot as plt
```
tomsilver, line 8: y?

tomsilver (approved):
good stuffs

tests/test_utils.py
ronuchit: just for consistency, i noticed we were doing it both ways

src/envs/base_env.py
ronuchit: ehhh

tomsilver: yeah lol nvm

PR #944, opened 2022-05-23
interface for collecting human demonstrations

### LINKED ISSUE #730: interface for collecting human demonstrations

ronuchit:
Note that there are no options in the data when doing human demos, so we
can only run approaches that do option learning!!!

`python src/main.py --env touch_point --approach nsrt_learning --seed 0
--option_learner direct_bc --segmenter contacts --demonstrator human`
after demonstrating all 50 tasks, gets:
```
Tasks solved: 44 / 50
Average time for successes: 0.12265 seconds
Test results: defaultdict(<class 'float'>, {'num_solved': 44,
'num_total': 50, 'avg_suc_time': 0.1226517070423473,
'min_skeletons_optimized': 1.0, 'max_skeletons_optimized': 1.0,
'num_execution_failures': 0, 'num_solve_failures': 6,
'avg_num_skeletons_optimized': 1.0, 'avg_num_nodes_expanded': 1.0,
'avg_num_nodes_created': 2.0, 'avg_num_nsrts': 1.0, 'avg_num_preds':
1.0, 'avg_plan_length': 1.0, 'avg_num_failures_discovered': 0.0,
'num_offline_transitions': 261, 'num_online_transitions': 0,
'query_cost': 0.0, 'learning_time': 6.754237174987793})
```

i also tried the same thing but with purposefully suboptimal demos,
leading to:
```
Tasks solved: 27 / 50
Average time for successes: 0.27913 seconds
Test results: defaultdict(<class 'float'>, {'num_solved': 27,
'num_total': 50, 'avg_suc_time': 0.2791292932298448,
'min_skeletons_optimized': 1.0, 'max_skeletons_optimized': 1.0,
'num_execution_failures': 0, 'num_solve_failures': 23,
'avg_num_skeletons_optimized': 1.0, 'avg_num_nodes_expanded': 1.0,
'avg_num_nodes_created': 2.0, 'avg_num_nsrts': 1.0, 'avg_num_preds':
1.0, 'avg_plan_length': 1.0, 'avg_num_failures_discovered': 0.0,
'num_offline_transitions': 441, 'num_online_transitions': 0,
'query_cost': 0.0, 'learning_time': 7.729219913482666})
```

note: with this PR, we can also collect demonstrations for stick_button,
but that one needs way more data to learn good options...

demos (using `--make_demo_videos`):

https://user-images.githubusercontent.com/3521150/169904526-dfd06480-af64-4bf6-9a63-3c347db52d23.mp4
https://user-images.githubusercontent.com/3521150/169904530-3ff73886-30d2-46a5-a3fb-49fa68db8a4a.mp4
https://user-images.githubusercontent.com/3521150/169904532-8ff6aabd-f3bb-4618-ab19-073f0ecc13e1.mp4
https://user-images.githubusercontent.com/3521150/169904534-17ac8eb8-07e3-4b41-97fc-9df8fe682e22.mp4

screen recording of myself collecting demos:
https://www.dropbox.com/s/rjeu4d81c62nk9y/human_demos.mov?dl=0

closes #730

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ requests
 slack_bolt
 pybullet>=3.2.0
 sklearn
+pyqt5
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -117,7 +117,9 @@ def _human_demonstrator_policy(env: BaseEnv, idx: int, num_tasks: int,
                                    [State, matplotlib.backend_bases.Event],
                                    Action],
                                state: State) -> Action:  # pragma: no cover
-    # Change the backend to one that supports a GUI.
+    # Temporarily change the backend to one that supports a GUI.
+    # We do this here because we don't want the rest of the codebase
+    # to use GUI-based Matplotlib.
     cur_backend = matplotlib.get_backend()
     matplotlib.use("Qt5Agg")
     # Render the state.
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -2,7 +2,7 @@
 
 import functools
 import logging
-from typing import List, Set
+from typing import Callable, List, Set
 
 import matplotlib
 import matplotlib.pyplot as plt
@@ -29,6 +29,14 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task],
             train_tasks,
             task_planning_heuristic=CFG.offline_data_task_planning_heuristic,
             max_skeletons_optimized=CFG.offline_data_max_skeletons_optimized)
+    else:
+        # Disable all built-in keyboard shortcuts.
+        keymaps = {k for k in plt.rcParams if k.startswith("keymap.")}
+        for k in keymaps:
+            plt.rcParams[k].clear()
+        # Create the environment-specific method for turning events into
+        # actions. This should also log instructions.
+        event_to_action = env.get_event_to_action_fn()
     trajectories = []
     num_tasks = min(len(train_tasks), CFG.max_initial_demos)
     for idx, task in enumerate(train_tasks):
@@ -53,7 +61,8 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task],
                 termination_function = lambda s: False
             else:
                 policy = functools.partial(_human_demonstrator_policy, env,
-                                           idx, num_tasks, task)
+                                           idx, num_tasks, task,
+                                           event_to_action)
                 termination_function = task.goal_holds
             if CFG.make_demo_videos:
                 monitor = utils.VideoMonitor(env.render)
@@ -104,7 +113,9 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task],
 
 
 def _human_demonstrator_policy(env: BaseEnv, idx: int, num_tasks: int,
-                               task: Task, state: State) -> Action:
+                               task: Task, event_to_action: Callable[
+                                   [State, matplotlib.backend_bases.Event],
+                                   Action], state: State) -> Action:
     # Change the backend to one that supports a GUI.
     cur_backend = matplotlib.get_backend()
     matplotlib.use("Qt5Agg")
@@ -115,7 +126,7 @@ def _human_demonstrator_policy(env: BaseEnv, idx: int, num_tasks: int,
     container = {}
 
     def _handler(event: matplotlib.backend_bases.Event) -> None:
-        container["action"] = env.event_to_action(state, event)
+        container["action"] = event_to_action(state, event)
 
     keyboard_cid = fig.canvas.mpl_connect("key_press_event", _handler)
     mouse_cid = fig.canvas.mpl_connect("button_press_event", _handler)
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -1,29 +1,36 @@
 """Create offline datasets by collecting demonstrations."""
 
+import functools
 import logging
 from typing import List, Set
+
+import matplotlib
+import matplotlib.pyplot as plt
 
 from predicators.src import utils
 from predicators.src.approaches import ApproachFailure, ApproachTimeout
 from predicators.src.approaches.oracle_approach import OracleApproach
 from predicators.src.envs import BaseEnv
 from predicators.src.settings import CFG
-from predicators.src.structs import Dataset, LowLevelTrajectory, \
-    ParameterizedOption, Task
+from predicators.src.structs import Action, Dataset, LowLevelTrajectory, \
+    ParameterizedOption, State, Task
 
 
 def create_demo_data(env: BaseEnv, train_tasks: List[Task],
                      known_options: Set[ParameterizedOption]) -> Dataset:
     """Create offline datasets by collecting demos."""
-    oracle_approach = OracleApproach(
-        env.predicates,
-        env.options,
-        env.types,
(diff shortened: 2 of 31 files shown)


requirements.txt
```diff
@@ -15,3 +15,4 @@ requests
 slack_bolt
 pybullet>=3.2.0
 sklearn
+pyqt5
```
ronuchit, line 18: because we never needed plt gui before, but now we do

src/utils.py
```diff
@@ -48,7 +48,11 @@
 if TYPE_CHECKING:
     from predicators.src.envs import BaseEnv
 
+# Switch to a non-GUI backend and disable all built-in keyboard shortcuts.
 matplotlib.use("Agg")
+keymaps = {keymap for keymap in plt.rcParams if keymap.startswith("keymap.")}
+for keymap in keymaps:
+    plt.rcParams[keymap].clear()
```
ronuchit, line 55: to stop keyboard buttons from doing things like opening the save menu or quitting

tomsilver:
😻 i'm excited. can you screen record yourself making a demonstration? and give a command in the PR description?

tomsilver:
@ronuchit can we add some user instructions, ideally in the GUI, but in the terminal would be okay?

src/datasets/demo_only.py
```diff
@@ -1,55 +1,79 @@
...
+        oracle_approach = OracleApproach(
+            env.predicates,
+            env.options,
+            env.types,
+            env.action_space,
+            train_tasks,
+            task_planning_heuristic=CFG.offline_data_task_planning_heuristic,
+            max_skeletons_optimized=CFG.offline_data_max_skeletons_optimized)
+    else:  # pragma: no cover
+        # Disable all built-in keyboard shortcuts.
```
tomsilver, line 33: you verified that everything still works after you moved this right? sometimes matplotlib is particular about the order in which you change settings vs importing stuff

src/datasets/demo_only.py
```diff
@@ -71,14 +98,46 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task],
...
             utils.save_video(outfile, video)
     return Dataset(trajectories)
+
+
+def _human_demonstrator_policy(env: BaseEnv, idx: int, num_tasks: int,
+                               task: Task, event_to_action: Callable[
+                                   [State, matplotlib.backend_bases.Event],
+                                   Action],
+                               state: State) -> Action:  # pragma: no cover
+    # Change the backend to one that supports a GUI.
```
tomsilver, line 120: could this be moved to be with the other matplotlib settings change?

src/datasets/demo_only.py
```diff
@@ -71,14 +98,46 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task],
...
+    def _handler(event: matplotlib.backend_bases.Event) -> None:
+        container["action"] = event_to_action(state, event)
+
+    keyboard_cid = fig.canvas.mpl_connect("key_press_event", _handler)
+    mouse_cid = fig.canvas.mpl_connect("button_press_event", _handler)
+    # Hang until either a mouse press or a keyboard press.
+    plt.waitforbuttonpress()
+    fig.canvas.mpl_disconnect(keyboard_cid)
+    fig.canvas.mpl_disconnect(mouse_cid)
+    plt.close()
```
tomsilver, line 138: you call close at each step of the policy? why is that?

src/envs/stick_button.py
```diff
@@ -576,3 +578,29 @@ def _AboveNoButton_holds(self, state: State,
...
+            tx = event.xdata
+            ty = event.ydata
+            assert tx is not None and ty is not None, "Out-of-bounds click"
+            # Move toward the target.
+            dx = np.clip(tx - rx, -self.max_speed, self.max_speed)
+            dy = np.clip(ty - ry, -self.max_speed, self.max_speed)
+            # Normalize.
+            dx = dx / self.max_speed
+            dy = dy / self.max_speed
+            # No need to rotate or press.
```
tomsilver, line 603: somewhere around here you should `assert CFG.stick_button_disable_angles`

src/envs/touch_point.py
```diff
@@ -177,3 +178,22 @@ def _Touched_holds(self, state: State, objects: Sequence[Object]) -> bool:
...
         dist = np.sqrt((rx - tx)**2 + (ry - ty)**2)
         return dist < self.action_magnitude * self.touch_multiplier
+
+    def get_event_to_action_fn(
+            self) -> Callable[[State, matplotlib.backend_bases.Event], Action]:
+        logging.info("Controls: mouse click to move")
+
+        def _event_to_action(state: State,
+                             event: matplotlib.backend_bases.Event) -> Action:
+            assert event.key is None, "Keyboard controls not allowed."
```
tomsilver, line 188: does this crash everything? or just wait for another action from the user? the latter is preferred

tomsilver (approved):
❤️‍🔥 ❤️‍🔥 ❤️‍🔥

src/datasets/demo_only.py
ronuchit: yup

src/datasets/demo_only.py
ronuchit: nope, it has to happen right before we call render(), and we revert this change back at the end of this function. this is because we don't want the rest of the codebase to use GUI-based matplotlib. i'll clarify the comment

src/datasets/demo_only.py
ronuchit: each call to `render_state_plt` creates a new plt, so i'm just closing it. that's what the docstring of `render_state_plt` tells me to do:
NOTE: Users of this method must remember to call `plt.close()`,
        because this method returns an active figure object!

src/envs/stick_button.py
ronuchit: oops thanks, had this before but it got lost somehow

src/envs/touch_point.py
ronuchit: it crashes everything, it's not super easy to wait for another action from the user because we're calling `plt.waitforbuttonpress()` which hangs until either a mouse click or a keystroke, but then we'd need to detect which one actually happened and receive information from the env about which of them is "legal". let's just leave it as is for now, we can address this later if we really care

ronuchit: btw for similar reasons, it will also crash everything (with an informative error message) if you accidentally click outside the bounds of the grid (but still within the figure window). try not to do that on task 999! 😛

tomsilver: You joke, but that would actually be tragic lol. Is there any way we can save as we go?

ronuchit: we can make that a standalone PR that saves each task separately during data collection and then deletes all of them when it reaches the end of data collection. thoughts? if that sounds good, i can make an issue

ronuchit: (this can happen regardless of the setting of `demonstrator`)

tomsilver: something like that sounds good, +1 to making it a separate issue and PR

ronuchit: hmm we would also need it to load from the previously saved tasks if there was partial progress

ronuchit: regardless, i'm merging this because whatever we choose to do can be a standalone PR

PR #945, opened 2022-05-23
fix edge case where option model returns num actions 0 during planning

tomsilver:


tests/test_planning.py
```diff
@@ -176,6 +176,33 @@ def test_sesame_plan_failures():
...
+        """A mock option model that raises an EnvironmentFailure."""
+
+        def __init__(self, simulator):
+            self._simulator = simulator
+
+        def get_next_state_and_num_actions(self, state, option):
+            return state, 0
+
+    option_model = _MockOptionModel(env.simulate)
+    with pytest.raises(PlanningFailure):
```
ronuchit, line 193: is there anything we can assert about the specific type of failure, using our usual `as e ... assert "blah" in str(e)` pattern?

tests/test_planning.py
```diff
@@ -176,6 +176,33 @@ def test_sesame_plan_failures():
             CFG.sesame_task_planning_heuristic,
             CFG.sesame_max_skeletons_optimized,
             max_horizon=0)
+
+    # Test that no plan is found when all of the options appear to terminate
+    # immediately, under the option model.
+
+    class _MockOptionModel(_OptionModelBase):
+        """A mock option model that raises an EnvironmentFailure."""
```
ronuchit, line 184: update docstring

ronuchit (approved):

tests/test_planning.py
tomsilver: it's going to be a max skeletons thing, but I don't think that's very illuminating. there's nothing particular about the num actions 0 case to check here

PR #948, opened 2022-05-24
Add planning time, execution time metrics per task. 

ashayathalye:
Want to keep track of planning time and policy execution time for each task so that we can plot a distribution of these times. 

Because metrics is a dictionary from str -> float, this is a hacky approach for now per Tom's suggestion: we just make keys like "task10_planning_time" for each test task id. We can improve this later. I actually don't think we needed to do this hacky approach because the analysis scripts take in a list of keys to operate on, so we could just not include "planning_times" in that list of keys. 

I will add the plotting script after we re-run the supercloud experiments with this enabled and get the data for it.


ronuchit:
what's the latest with this? can we close this PR if we're not looking to merge it in the short term?

PR #951, opened 2022-05-25
Add exception counts to metrics

ashayathalye:
Want the metrics dictionary to store how many times each kind of exception occurred so that we get a clearer picture about how and why an approach is failing.


ashayathalye:
@ Rohan's question, I don't think anything has to change on the analysis/plotting side of things, since those scripts all have a list of keys to index into the metrics dict with, and these new items just aren't in that list of keys so they are otherwise ignored.

PR #952, opened 2022-05-25
skip training tasks where human demonstrator gave a bad input

### LINKED ISSUE #946: temporarily save training tasks during data generation
To prevent a crash from ruining all our data collection so far, we can modify `demo_only.py` to temporarily save each demonstration as it's looping through the training tasks. At the end, we should delete all these temporary files.

On any run of the system, if one of these temporary files exists for a particular training task, we should _always_ load it. (Regardless of any setting of `--load_data` or whatever else.)

ronuchit:
i think this is easier and cleaner than the individual task saving idea in #946, and so we should only do that more complicated thing if we really find it necessary in the future.

closes #946

### DIFF
--- a/src/datasets/demo_only.py
+++ b/src/datasets/demo_only.py
@@ -75,7 +75,10 @@ def create_demo_data(env: BaseEnv, train_tasks: List[Task],
                 idx,
                 termination_function=termination_function,
                 max_num_steps=CFG.horizon,
-                exceptions_to_break_on={utils.OptionExecutionFailure},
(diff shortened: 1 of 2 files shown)


ronuchit:
i'm fixing that in a separate PR. my plan is to make changes here: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/main.py#L199

to search for other amounts of data first before giving up entirely.

tomsilver (approved):
LGTM

PR #963, opened 2022-06-01
NSRT objects expose an `op` property

ronuchit:
we should have done this long ago


tomsilver (approved):
lgtm modulo checks passing

PR #964, opened 2022-06-01
include offline learning metrics in saved results, so analysis scripts can read them

ronuchit:


tomsilver:
i think we were reluctant to do this before because approaches may have different metrics. but if the scripts are okay, it's okay

PR #965, opened 2022-06-01
flag to compute sidelining objective value

ronuchit:


src/approaches/nsrt_learning_approach.py
```diff
@@ -85,3 +89,61 @@ def load(self, online_learning_cycle: Optional[int]) -> None:
...
+            heuristic = utils.create_task_planning_heuristic(
+                CFG.sesame_task_planning_heuristic, init_atoms, task.goal,
+                ground_nsrts, preds, objects)
+            for skeleton, _, _ in task_plan(init_atoms,
+                                            task.goal,
+                                            ground_nsrts,
+                                            reachable_atoms,
+                                            heuristic,
+                                            CFG.seed,
+                                            timeout=10000000,
```
tomsilver, line 130: is this going to potentially kill us if we use this with baselines that have argument explosions?

src/approaches/nsrt_learning_approach.py
```diff
@@ -85,3 +89,61 @@ def load(self, online_learning_cycle: Optional[int]) -> None:
...
+                                            max_skeletons_optimized=10000000):
+                if len(skeleton) > len(segment_traj):
```
tomsilver, line 132: you should assert that the heuristic is lmcut, otherwise there's no guarantee that the order of the skeleton lengths will increase monotonically

src/approaches/nsrt_learning_approach.py
```diff
@@ -85,3 +89,61 @@ def load(self, online_learning_cycle: Optional[int]) -> None:
...
+                    break
+                num_plans_up_to_n += 1
+        # Calculate second term in objective. This is the complexity of the
+        # operator set. We measure this using the total number of atoms in all
+        # the operators.
+        complexity = 0
+        for op in strips_ops:
```
tomsilver, line 139: add a regularization weight for this part?

src/approaches/nsrt_learning_approach.py
```diff
@@ -61,6 +63,8 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
         save_path = utils.get_approach_save_path_str()
         with open(f"{save_path}_{online_learning_cycle}.NSRTs", "wb") as f:
             pkl.dump(self._nsrts, f)
+        if CFG.compute_sidelining_objective_value:
```
tomsilver, line 66: hopefully we can also use this to see whether the objective value improves with each iteration of backchaining

tomsilver (approved):

src/approaches/nsrt_learning_approach.py
ronuchit: i'm not summing this with the other term, just storing both into metrics. someone downstream can take a weighted sum

src/approaches/nsrt_learning_approach.py
ronuchit: that's, uhh... going to be annoying...
- if we want to do what you said, we can't put this method in `nsrt_learning_approach.py`, since the strips learners don't have access to the approach object
- we can't put this method in the strips learners since we need to update the metrics dictionary of the `NSRTLearningApproach` with this information so that we can analyze it in the scripts, but the `NSRTLearningApproach` only sees the top-level method `learn_nsrts_from_data()`, not any of the details about the strips learners
- we can't put this method in utils.py because it has dependencies on planner.py, which is circular

the only solution i can think of is to put it in planner.py.... thoughts on that?

tomsilver: Is there a version of the second bullet where we could pass the metrics dict into the helper function?

ronuchit: no because there are three layers: the Approach, `nsrt_learning_main.py`, and the strips learners. each one sees itself and the thing after it. so Approach can't see the strips learners.

tomsilver: aight then I vote to leave as is I guess

scripts/analyze_results_directory.py
```diff
@@ -44,6 +49,14 @@
     # ("QUERY_COST", "query_cost"),
 ]
 
+if args.sidelining:
```
ronuchit, line 52: note: usage is `python scripts/analyze_results_directory.py --sidelining`

PR #967, opened 2022-06-02
add Triangle Geom2D

ronuchit:
it's helpful for view cone stuff, which we need in the satellite domain

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -292,6 +292,44 @@ class Circle(_Geom2D):
 
 
 @dataclass(frozen=True)
+class Triangle(_Geom2D):
+    """A helper class for visualizing and collision checking triangles."""
+    x1: float
+    y1: float
+    x2: float
+    y2: float
+    x3: float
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
I <3 geoms!

src/utils.py
```diff
@@ -291,6 +291,44 @@ def contains_point(self, x: float, y: float) -> bool:
         return (x - self.x)**2 + (y - self.y)**2 <= self.radius**2
 
 
+@dataclass(frozen=True)
+class Triangle(_Geom2D):
+    """A helper class for visualizing and collision checking triangles."""
```
tomsilver, line 296: @ronuchit you're not going to be able to use this class for collision checking without implementing some collision checking code -- update docstring?

ronuchit: i was thinking of contains_point as doing some crude collision checking (and that's all i'm gonna use in satellites anyway). let's just leave it as is

PR #968, opened 2022-06-02
make render DPI a setting

ronuchit:


tomsilver (approved):
The following suggestion is a bit hacky but I think good: in reset config, can we set this new hyperparameter to something small by default?

tomsilver (approved):
lgtm thanks. did it get faster at all?

PR #981, opened 2022-06-21
make `offline_data_planning_timeout` default to `timeout`

### LINKED ISSUE #949: Ensure commands are deterministic for a given random seed.
If I run this command twice (so, the same random seed), I get two different results.
```python3 main.py --env doors --segmenter contacts --num_train_tasks 50 --timeout 300 --disable_harmlessness_check True --neural_gaus_regressor_max_itr 50000 --sampler_learner neural --approach nsrt_learning --option_learner direct_bc --num_test_tasks 50 --seed 0 --included_options MoveToDoor,MoveThroughDoor```.
(1) first time: 
```
Tasks solved: 15/ 50
Average time for successes: 4.87689 seconds
...
Main script terminated in 1594.71568 seconds
```
(2) second time: 
```
Tasks solved: 17 / 50
Average time for successes: 8.32305 seconds
...
Main script terminated in 2336.45988 seconds
```

I think given a random seed, we should be expecting determinism when the commands are run twice on the same computer. 

TODO: try running with fewer hyperparameters, and different environments.

ronuchit:
note: this will affect a couple of scripts slightly. in particular:
* stick_button experiments in the nightly require `--timeout 300`, and
so any training examples that fail due to the timeout will take 30x
longer to do so
* similar for ALL experiments in `run_option_learning_experiments.sh`

closes #949

### DIFF
--- a/src/args.py
+++ b/src/args.py
@@ -18,8 +18,9 @@ def create_arg_parser(env_required: bool = True,
     parser.add_argument("--included_options", default="", type=str)
     parser.add_argument("--seed", required=seed_required, type=int)
     parser.add_argument("--option_learner", type=str, default="no_learning")
-    # NOTE: this timeout is NOT used for planning when generating offline data.
-    # If you want to change that, modify offline_data_planning_timeout.
+    # NOTE: this timeout affects both data generation and evaluation.
+    # If you want to change only the data generation timeout,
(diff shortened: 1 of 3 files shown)


tomsilver (approved):
LGTM thanks (and re: your warnings, noted, I think it should be okay)

PR #982, opened 2022-06-21
unify two different places we were computing operator complexity, and change the computation

### LINKED ISSUE #979: change operator complexity to use arity and unify with get_operator_penalty

ronuchit:
Predicate invention and sidelining objective value computation were both
using (different) measures of operator complexity. Unify them into a
single method. Also, per discussion with @NishanthJKumar, make the computation only depend on arity since
that's what really affects planning. Note that this doesn't change
predicate invention functionally, since we have had
`grammar_search_operator_complexity_weight = 0.0` for a long time.

closes #979

### DIFF
--- a/src/approaches/nsrt_learning_approach.py
+++ b/src/approaches/nsrt_learning_approach.py
@@ -140,8 +140,7 @@ class NSRTLearningApproach(BilevelPlanningApproach):
                     break
                 num_plans_up_to_n += 1
         # Calculate second term in objective. This is the complexity of the
-        # operator set. We measure this using the total number of atoms in all
-        # the operators.
+        # operator set, measured as the sum of all operator complexities.
         complexity = 0.0
         for op in strips_ops:
             complexity += op.get_complexity()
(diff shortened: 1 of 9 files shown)


src/approaches/nsrt_learning_approach.py
```diff
@@ -142,11 +142,9 @@ def _compute_sidelining_objective_value(
         # Calculate second term in objective. This is the complexity of the
         # operator set. We measure this using the total number of atoms in all
         # the operators.
```
NishanthJKumar, line 144: update comment to reflect new computation

NishanthJKumar (approved):
LGTM after addressing minor comment!

src/structs.py
```diff
@@ -661,6 +662,15 @@ def effect_to_side_predicate(self, effect: LiftedAtom,
                               new_add_effects, new_delete_effects,
                               self.side_predicates | {effect.predicate})
 
+    def get_complexity(self) -> float:
+        """Get the complexity of this operator.
+
+        We only care about the arity of the operator, since that is what
+        affects grounding. We'll use the factorial of the arity as a
```
tomsilver, line 669: why is this factorial? remember that objects need not be unique

PR #985, opened 2022-06-21
[BEHAVIOR] Major Update

wmcclinton:
This PR should fully generate a trajectory using our behavior oracle NSRTs and learn NSRTs on that demonstration.


src/planning.py
```diff
@@ -417,6 +418,79 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
+    while cur_idx < len(plan):
+        if time.time() - start_time > timeout:
+            return LowLevelTrajectory([], [], False), False
+        state = traj[cur_idx]
+        option = plan[cur_idx]
+        cur_idx += 1
+        if option.initiable(state):
+            try:
+                next_state, num_actions = \
+                    option_model.get_next_state_and_num_actions(state, option)
```
ronuchit, line 448: is this function deterministic? if so, i don't see why we need to do backtracking in this function, because what would change between subsequent iterations?

src/planning.py
```diff
@@ -417,6 +418,79 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
+                        if task.goal_holds(traj[cur_idx]):
+                            return LowLevelTrajectory(
+                                traj, actions), True  # success!
+                        can_continue_on = False
+                        return LowLevelTrajectory([], [], False), False
+        else:
+            # The option is not initiable.
+            can_continue_on = False
+        if not can_continue_on:  # we got stuck, time to return False!
+            return LowLevelTrajectory([], [], False), False
```
ronuchit, line 485: oh, so we're not even doing backtracking, we're actually just returning failure here? in that case, we should be able to drastically simplify this function right? like we don't need `cur_idx` and stuff, all we need to do is loop `for option in plan`, and for each one, call the option model to get the next state, and build up the `traj` and `actions` as we go. if any option is not initiable, or if horizon is reached, or if the goal isn't achieved at the end, we can return failure

src/structs.py
```diff
@@ -665,7 +665,8 @@ def get_complexity(self) -> float:
         """Get the complexity of this operator.
 
         We only care about the arity of the operator, since that is what
-        affects grounding. We'll use 2^arity as a measure of grounding effort.
+        affects grounding. We'll use 2^arity as a measure of grounding
```
ronuchit, line 668: oops my bad will fix in master

ronuchit (commented):
i left some comments, but was this meant to be a draft PR? you can change it by clicking "Convert to draft" here: https://github.com/Learning-and-Intelligent-Systems/predicators/pull/985 (or, when creating the PR, choose the draft option). for draft PRs, we won't get emails from github requesting a review

src/planning.py
```diff
@@ -417,6 +418,79 @@ def _run_low_level_search(task: Task, option_model: _OptionModelBase,
...
+                action_option = ParameterizedOption(
+                    option.parent.name, option.parent.types,
+                    option.parent.params_space,
+                    lambda s, m, o, p: Action(np.array([0.0])),
+                    lambda s, m, o, p: False,
+                    lambda s, m, o, p: True).ground(option.objects,
+                                                    option.params)
+                action_option.name = option.name
+                action_option.memory = option.memory
+                actions[cur_idx - 1].set_option(action_option)
```
ronuchit, line 465: why can't you just do `actions[cur_idx - 1].set_option(option)`? why do you need a different thing called `action_option`? (please add a comment explaining this if there's a good reason, since it's a bit confusing!)

src/planning.py
tomsilver: I don't think we have such a util yet. Did you mean to say "one action per option"?

tomsilver: (Like why would there just be one action per option in general?)

ronuchit: (they use this in the context of generating demonstrations for behavior, since the actions in the demos don't really matter / aren't used)

tomsilver: ah right got it. so yeah I think we definitely would not have already implemented that util

wmcclinton: I made it simpler (no backchaining). The reason it is still in planning.py is because I got a bad circular-imports error when I tried to move it to utils @ronuchit .

ronuchit: ah lol ok i've been thru that before

src/datasets/demo_only.py
```diff
@@ -153,42 +168,69 @@ def _generate_demonstrations(
...
+                    #     curr_option = None
+                    #     print(pkld_plan[i][0])
+                    #     for option in env.options:
+                    #         if option.name == pkld_plan[i][0]:
+                    #             curr_option = option
+                    #     last_plan.append(curr_option.ground(pkld_plan[i][1], pkld_plan[i][2]))
+                    traj, success = _run_low_level_plan(
+                        task, oracle_approach._option_model, last_plan,
+                        CFG.offline_data_planning_timeout, CFG.horizon)
+                    if not success:
```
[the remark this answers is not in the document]
wmcclinton: I added a comment but basically `success` is False here if the execution of our plan on our option model fails the second time. If this is the case t will rerun `oracle_approach.solve()` to get a new `last_plan`, that will be different than the last `last_plan` and evaluate that.

NishanthJKumar:
to be superseded by a number of smaller, more self-contained PR's

PR #990, opened 2022-06-22
lifted decision list data structures

tomsilver:
starting to work towards reimplementing PG3 in this code base.

### DIFF
--- a/src/structs.py
+++ b/src/structs.py
@@ -1455,6 +1455,22 @@ class _GroundLDLRule:
         return str(self) > str(other)
 
 
+@dataclass(frozen=True)
+class LiftedDecisionList:
+    """A goal-conditioned policy from abstract states to ground NSRTs
+    implemented with a lifted decision list.
+
+    Given an abstract state and goal, the rules are grounded in order. The
+    first applicable ground rule is used to return a ground NSRT.
+
+    If no rule is applicable, returns None.
+
+    The logic described above is implemented in utils.query_ldl().
+    """
+    name: str
+    rules: Sequence[LDLRule]
+
+
 # Convenience higher-order types useful throughout the code
 OptionSpec = Tuple[ParameterizedOption, List[Variable]]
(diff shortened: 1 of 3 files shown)


src/structs.py
```diff
@@ -1333,6 +1333,144 @@ class PathToStateResponse(Response):
...
+        assert isinstance(other, _GroundLDLRule)
+        return str(self) > str(other)
+
+
+@dataclass(frozen=True)
+class LiftedDecisionList:
+    """A goal-conditioned policy from abstract states to ground NSRTs
+    implemented with a lifted decision list.
+
+    Given an abstract state and goal, the rules are grounded in order. The
```
ronuchit, line 1463: get rid of this para and the next? it's redundant with `utils.query_ldl` and it's also confusing bc this isn't a function

ronuchit (approved):

src/utils.py
```diff
@@ -1709,6 +1710,15 @@ def all_possible_ground_atoms(state: State,
     return sorted(ground_atoms)
 
 
+def all_ground_ldl_rules(
+        rule: LDLRule,
+        objects: Collection[Object]) -> Iterator[_GroundLDLRule]:
+    """Get all possible groundings of the given NSRT with the given objects."""
```
ronuchit, line 1716: update this comment ("NSRT")

src/utils.py
```diff
@@ -2439,3 +2449,21 @@ def nostdout() -> Generator[None, None, None]:
     sys.stdout = _DummyFile()
     yield
     sys.stdout = save_stdout
+
+
+def query_ldl(ldl: LiftedDecisionList, atoms: Set[GroundAtom],
+              goal: Set[GroundAtom]) -> Optional[_GroundNSRT]:
```
ronuchit, line 2455: why do we return the ground NSRT instead of the ground rule?

tomsilver: I'm thinking about this like querying a (partial) policy in the abstract space. also, I don't foresee the rule being useful except to get the ground NSRT from it

PR #996, opened 2022-06-23
Wester delivery approach

wester-is-not-real:
Onboarding assignment for MSRP.


tomsilver (commented):
excellent work! just a few small things :)

PR #997, opened 2022-06-23
[sidelining] Fix known issue with intermediate harmlessness in longrun tests

NishanthJKumar:
Gets the intermediate harmlessness check to pass for the longrun tests, resolving the known issue with these tests. 

Conceptually, we want to ensure at the end of our backchaining that:
- The segment.necessary_add_effects should be set correctly.
- We should be able to find a best-matching PNAD (unification) for this segment.
- *This segment should be in the particular PNAD’s datastore.*

We weren't previously explicitly checking this last condition (though it is important only rarely), which leads to intermediate harmlessness violations that get fixed at later iterations of the outer loop.

We now explicitly check this condition, and if it's false, we move the segment into the correct datastore to fix the problem. Here's a diagram of the logic:

image

It's possible we can accomplish the same thing in a nicer, more integrated way that makes the code cleaner/easier to understand.

**Supercloud results**:

```
a02d4fff0ca5e9cd9ea342e7b5079e7419b55d8e


AGGREGATED DATA OVER SEEDS:
                                                  NUM_TRAIN_TASKS     NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED  LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                                                            
painting_backchaining_10demo                         10.00 (0.00)   41.90 (3.91)  14.00 (0.00)   0.47 (0.20)   1995.81 (935.53)   53.19 (3.88)         10
painting_backchaining_25demo                         25.00 (0.00)   49.20 (1.17)  14.00 (0.00)   0.38 (0.08)   2378.04 (829.09)   56.16 (3.38)         10
painting_backchaining_50demo                         50.00 (0.00)   49.40 (0.66)  14.00 (0.00)   0.43 (0.12)  2748.98 (1084.43)   57.99 (2.40)         10
painting_backchaining_5demo                           5.00 (0.00)   12.00 (8.72)  14.00 (0.00)   1.20 (0.90)   1208.96 (882.06)   50.65 (2.30)         10
repeated_nextto_painting_backchaining_10demo         10.00 (0.00)  34.30 (10.00)  18.00 (0.00)   1.31 (0.46)  5763.46 (5495.06)   89.86 (6.76)         10
repeated_nextto_painting_backchaining_25demo         25.00 (0.00)   45.00 (3.71)  18.00 (0.00)   0.80 (0.23)  2544.65 (2748.56)   92.88 (3.36)         10
repeated_nextto_painting_backchaining_50demo         50.00 (0.00)   49.50 (0.67)  18.00 (0.00)   0.76 (0.06)   1829.65 (422.83)   96.13 (4.74)         10
repeated_nextto_painting_backchaining_5demo           5.00 (0.00)   20.40 (9.86)  18.00 (0.00)   1.58 (0.78)  7347.63 (6640.82)   83.10 (2.36)         10
repeated_nextto_single_option_backchaining_10demo    10.00 (0.00)   50.00 (0.00)   3.00 (0.00)   0.04 (0.01)       88.50 (2.35)   20.68 (2.67)         10
repeated_nextto_single_option_backchaining_25demo    25.00 (0.00)   50.00 (0.00)   3.00 (0.00)   0.03 (0.00)       88.50 (2.35)   17.13 (0.90)         10
repeated_nextto_single_option_backchaining_50demo    50.00 (0.00)   50.00 (0.00)   3.00 (0.00)   0.03 (0.00)       88.50 (2.35)   23.07 (1.15)         10
repeated_nextto_single_option_backchaining_5demo      5.00 (0.00)   49.80 (0.60)   3.00 (0.00)   0.04 (0.01)      83.81 (14.95)   22.77 (2.30)         10
satellites_backchaining_10demo                       10.00 (0.00)   23.20 (7.61)  13.00 (0.00)   2.00 (0.85)  5547.99 (3190.52)   10.40 (0.46)         10
satellites_backchaining_25demo                       25.00 (0.00)   38.60 (5.64)  13.00 (0.00)   1.34 (0.30)  5360.62 (2624.37)   14.48 (0.72)         10
satellites_backchaining_50demo                       50.00 (0.00)   47.30 (1.55)  13.00 (0.00)   1.10 (0.21)  6667.70 (2726.76)   66.22 (0.55)         10
satellites_backchaining_5demo                         5.00 (0.00)    6.90 (9.41)  13.00 (0.00)   1.33 (0.84)  6966.44 (9825.95)    9.38 (0.61)         10
satellites_simple_backchaining_10demo                10.00 (0.00)   33.50 (6.04)  13.00 (0.00)   0.85 (0.53)    402.45 (210.77)   10.11 (0.82)         10
satellites_simple_backchaining_25demo                25.00 (0.00)   44.00 (2.41)  13.00 (0.00)   0.37 (0.16)     288.60 (45.14)   11.39 (0.75)         10
satellites_simple_backchaining_50demo                50.00 (0.00)   49.20 (0.60)  13.00 (0.00)   0.16 (0.09)     278.21 (45.27)  33.98 (23.71)         10
satellites_simple_backchaining_5demo                  5.00 (0.00)    7.50 (8.15)  13.00 (0.00)   1.80 (2.47)    533.91 (345.06)    9.44 (0.57)         10
screws_backchaining_10demo                           10.00 (0.00)   50.00 (0.00)   4.00 (0.00)   0.09 (0.00)      117.32 (1.06)    0.12 (0.00)         10
screws_backchaining_25demo                           25.00 (0.00)   50.00 (0.00)   4.00 (0.00)   0.09 (0.00)      117.32 (1.06)    0.35 (0.01)         10
screws_backchaining_50demo                           50.00 (0.00)   50.00 (0.00)   4.00 (0.00)   0.09 (0.00)      117.32 (1.06)    0.61 (0.02)         10
screws_backchaining_5demo                             5.00 (0.00)   50.00 (0.00)   4.00 (0.00)   0.09 (0.00)      117.32 (1.06)    0.07 (0.00)         10
```

### DIFF
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -228,11 +228,11 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                     if segment not in segs_in_pnad:
                         # Find PNAD that the segment is currently in.
                         for seg_pnad in pnads_for_option:
-                            segs_in_seg_pnad = {
+                            segs_in_seg_pnad = [
                                 datapoint[0]
                                 for datapoint in seg_pnad.datastore
-                            }
-                            if segment in segs_in_seg_pnad:
+                            ]
+                            if segment in set(segs_in_seg_pnad):
                                 seg_idx = segs_in_seg_pnad.index(segment)
                                 seg_pnad.datastore.pop(seg_idx)
                                 break
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -227,9 +227,16 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                     segment, objects, pnads_for_option)
 
                 if pnad is not None:
-                    segs_in_pnad = [datapoint[0] for datapoint in pnad.datastore]
+                    segs_in_pnad = [datapoint[0] for datapoint in pnad.datastore]                        
+                    # In this case, we potentially want to recompute_datastores
+                    # to see if this will just resolve the issue.
+                    if segment not in segs_in_pnad:
+                        self._recompute_datastores_from_segments(pnads_for_option)
+                        # Since we're not changing anything aside from the PNADs'
+                        # datastores, we're guaranteed that the result of
+                        # _find_best_matching_pnad_and_sub will stay the same.
+                        segs_in_pnad = [datapoint[0] for datapoint in pnad.datastore]
 
-                # TODO: Make this much more beautiful.                
                 if pnad is not None and segment in segs_in_pnad:
                     assert var_to_obj is not None
                     obj_to_var = {v: k for k, v in var_to_obj.items()}
@@ -238,9 +245,9 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                         tuple(var_to_obj[var] for var in pnad.op.parameters))
                     if len(param_opt_to_nec_pnads[option.parent]) == 0:
                         param_opt_to_nec_pnads[option.parent].append(pnad)
-                    # if str(segment.necessary_add_effects) == "{A()}":
-                    #     import ipdb; ipdb.set_trace()
-                # If we weren't able to find a substitution (i.e, the above
+
+                # If we weren't able to find a substitution such that the given
+                # segment is in the particular PNAD's datastore (i.e, the above
                 # _find_best_matching call didn't yield a PNAD), we need to
                 # spawn a new PNAD from the most general PNAD to cover
                 # these necessary add effects.
@@ -281,7 +288,10 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                         segment, objects,
                         param_opt_to_nec_pnads[option.parent])
                     assert var_to_obj is not None
-                    assert best_score_pnad == pnad
+                    try:
+                        assert best_score_pnad == pnad
+                    except AssertionError:
+                        import ipdb; ipdb.set_trace()
                     obj_to_var = {v: k for k, v in var_to_obj.items()}
                     assert len(var_to_obj) == len(obj_to_var)
                     ground_op = pnad.op.ground(
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -97,12 +97,8 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
             self._induce_delete_side_keep(param_opt_to_nec_pnads)
 
             # Harmlessness should now hold, but it's slow to check.
-            try:
-                assert self._check_harmlessness(
-                    self._get_uniquely_named_nec_pnads(param_opt_to_nec_pnads))
-            except AssertionError:
-                import ipdb; ipdb.set_trace()
-
+            assert self._check_harmlessness(
+                self._get_uniquely_named_nec_pnads(param_opt_to_nec_pnads))
             # Recompute datastores and preconditions for all PNADs.
             # Filter out PNADs that don't have datastores.
             cur_itr_pnads_unfiltered = [
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -288,10 +288,7 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                         segment, objects,
                         param_opt_to_nec_pnads[option.parent])
                     assert var_to_obj is not None
-                    try:
-                        assert best_score_pnad == pnad
-                    except AssertionError:
-                        import ipdb; ipdb.set_trace()
+                    assert best_score_pnad == pnad
                     obj_to_var = {v: k for k, v in var_to_obj.items()}
                     assert len(var_to_obj) == len(obj_to_var)
                     ground_op = pnad.op.ground(
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -207,22 +207,6 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
                 pnad, var_to_obj = self._find_best_matching_pnad_and_sub(
                     segment, objects, pnads_for_option)
                 if pnad is not None:
-                    segs_in_pnad = [
-                        datapoint[0] for datapoint in pnad.datastore
-                    ]
-                    # In this case, we potentially want to recompute_datastores
-                    # to see if this will just resolve the issue.
-                    if segment not in segs_in_pnad:
-                        self._recompute_datastores_from_segments(
-                            pnads_for_option)
-                        # Since we're not changing anything aside from the
-                        # PNADs' datastores, we're guaranteed that the result
-                        # of _find_best_matching_pnad_and_sub will stay the
-                        # same.
-                        segs_in_pnad = [
-                            datapoint[0] for datapoint in pnad.datastore
-                        ]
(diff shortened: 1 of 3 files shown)


tomsilver:
i haven't looked at this at all yet but i'm immediately in love with the diagram

tomsilver:
okay now having actually processed the description... can you explain how it is possible that a segment does not end up in the best matching PNAD's datastore? (also happy to wait for tomorrow's meeting to go through this)

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -215,17 +215,35 @@ def _backchain_one_pass(
...
+                    # In this case, we potentially want to recompute_datastores
+                    # to see if this will just resolve the issue.
+                    if segment not in segs_in_pnad:
+                        self._recompute_datastores_from_segments(
+                            pnads_for_option)
+                        # Since we're not changing anything aside from the
+                        # PNADs' datastores, we're guaranteed that the result
+                        # of _find_best_matching_pnad_and_sub will stay the
+                        # same.
+                        segs_in_pnad = [
```
ronuchit, line 230: same thing, set instead of list

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -215,17 +215,35 @@ def _backchain_one_pass(
...
                     if len(param_opt_to_nec_pnads[option.parent]) == 0:
                         param_opt_to_nec_pnads[option.parent].append(pnad)
-                # If we weren't able to find a substitution (i.e, the above
-                # _find_best_matching call didn't yield a PNAD), we need to
-                # spawn a new PNAD from the most general PNAD to cover
-                # these necessary add effects.
+                # If we weren't able to find a substitution such that the given
+                # segment is in the particular PNAD's datastore (i.e, the above
+                # _find_best_matching call didn't yield a PNAD such that this
+                # PNAD's datastore contains the segment), we need to spawn a
```
ronuchit, line 244: wait... if we found a substitution with a PNAD that the segment is NOT in the datastore of, then why should the correct move be to spawn an entirely new PNAD? why wouldn't we just move the segment to the datastore of the PNAD it unified with?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -264,6 +282,12 @@ def _backchain_one_pass(
                         param_opt_to_nec_pnads[option.parent])
                     assert var_to_obj is not None
                     assert best_score_pnad == pnad
+                    # Also, since this segment caused us to induce the new
+                    # PNAD, it should appear in this new PNAD's datastore.
+                    segs_in_pnad = [
```
ronuchit, line 287: set instead of list

ronuchit (commented):
highest level comment: we should think about whether there's a different way to do this. particularly, what if we added a field to Segment class representing which PNAD this segment is in the datastore of?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -214,6 +214,31 @@ def _backchain_one_pass(
                         tuple(var_to_obj[var] for var in pnad.op.parameters))
                     if len(param_opt_to_nec_pnads[option.parent]) == 0:
                         param_opt_to_nec_pnads[option.parent].append(pnad)
+                    segs_in_pnad = [
```
ronuchit, line 217: remember to change to a set

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -214,6 +214,31 @@ def _backchain_one_pass(
...
+                    # another PNAD into the current PNAD. Note that
+                    # we don't have to recompute the PNAD's add
+                    # effects or preconditions because of the fact that
+                    # this PNAD was found by the _find_best_matching
+                    # function (which internally checks that the
+                    # preconditions and add effects are all correct).
+                    if segment not in segs_in_pnad:
+                        # Find PNAD that the segment is currently in.
+                        for seg_pnad in pnads_for_option:
+                            segs_in_seg_pnad = [
```
ronuchit, line 230: change to a set

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -214,6 +214,31 @@ def _backchain_one_pass(
...
+                                datapoint[0]
+                                for datapoint in seg_pnad.datastore
+                            ]
+                            if segment in segs_in_seg_pnad:
+                                seg_idx = segs_in_seg_pnad.index(segment)
+                                seg_pnad.datastore.pop(seg_idx)
+                                break
```
ronuchit, line 237: do we want to add an `else` to this `for seg_pnad in pnads_for_option:`? this would be the case where this segment wasn't in ANY pnad's datastore. that would be bad news bears, right?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -291,6 +318,19 @@ def _backchain_one_pass(
                 }
         return nec_pnad_set_changed
 
+    @staticmethod
+    def _remove_empty_datastore_pnads(param_opt_to_nec_pnads: Dict[
+        ParameterizedOption, List[PartialNSRTAndDatastore]],
+                                      param_opt: ParameterizedOption) -> None:
+        """Removes all PNADs with empty datastores from the input
```
ronuchit, line 325: "PNADs" -> "PNADs associated with the given param_opt"

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -214,6 +214,31 @@ def _backchain_one_pass(
...
+                            segs_in_seg_pnad = [
+                                datapoint[0]
+                                for datapoint in seg_pnad.datastore
+                            ]
+                            if segment in segs_in_seg_pnad:
+                                seg_idx = segs_in_seg_pnad.index(segment)
+                                seg_pnad.datastore.pop(seg_idx)
+                                break
+                        pnad.datastore.append((segment, var_to_obj))
+                        self._remove_empty_datastore_pnads(
```
ronuchit, line 239: is this because the one you just removed this segment from may now be empty?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -229,23 +254,19 @@ def _backchain_one_pass(
                     # matched to another PNAD.
                     self._recompute_datastores_from_segments(
                         param_opt_to_nec_pnads[option.parent])
+                    # Now that we have done this, certain PNADs may be
+                    # left with empty datastores. Remove these.
+                    self._remove_empty_datastore_pnads(param_opt_to_nec_pnads,
```
ronuchit, line 259: check whether there's any other places where you could make use of this new function -- i think we filter out empty datastore pnads in a couple of places?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
ronuchit: just use { } instead of [ ]

ronuchit (changes requested):

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
ronuchit: although coverage might be hard. maybe instead, just set some flag to true inside of `if segment in segs_in_seg_pnad:` and then assert that the flag is true after the for loop terminates?

NishanthJKumar: I considered this while implementing: if the segment is not in ANY pnad's datastore, then why do we really care? We can simply put it into the current PNAD without removing it from anywhere (which is what the current code does).

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
NishanthJKumar: exactly

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
NishanthJKumar: there's one other place (line 106 in this file), but using our function here seems like it'll be much uglier than the current code.

ronuchit (approved):

PR #999, opened 2022-06-23
[sidelining] simplifies code by removing general pnad dict

### LINKED ISSUE #989: remove `param_opt_to_general_pnad` in backchaining
hide away `param_opt_to_general_pnad`. concretely, rename `self._initialize_general_pnad_for_option` to `self._create_general_pnad_for_option` and then use a cache inside that function if we want.

NishanthJKumar:
Resolves #989

### DIFF
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -67,11 +67,10 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
         for param_opt in parameterized_options:
             pnad = self._create_general_pnad_for_option(param_opt)
             param_opt_to_nec_pnads[param_opt] = []
-        self._assert_all_data_in_exactly_one_datastore(
-            [
-                self._create_general_pnad_for_option(param_opt)
-                for param_opt in parameterized_options
-            ])
+        self._assert_all_data_in_exactly_one_datastore([
+            self._create_general_pnad_for_option(param_opt)
+            for param_opt in parameterized_options
+        ])
 
         prev_itr_ops: Set[STRIPSOperator] = set()
 
--- a/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
+++ b/src/nsrt_learning/strips_learning/gen_to_spec_learner.py
@@ -359,23 +359,19 @@ class BackchainingSTRIPSLearner(GeneralToSpecificSTRIPSLearner):
         return "backchaining"
 
     def _spawn_new_pnad(self, segment: Segment) -> PartialNSRTAndDatastore:
-        """Given a PNAD and some segment with necessary add effects that the
-        PNAD must achieve, create a new PNAD ("spawn" from the most general one
+        """Given some segment with necessary add effects that a new PNAD PNAD
+        must achieve, create such a PNAD ("spawn" from the most general one
         associated with the segment's option) so that it has the necessary add
-        effects contained in the given segment.
-
-        Returns the newly constructed PNAD, without modifying the
-        original.
-        """
+        effects contained in the given segment."""
         # Create a general PNAD for the segment's option.
         pnad = self._create_general_pnad_for_option(
             segment.get_option().parent)
-        # Assert that the segment contains necessary_add_effects.
-        necessary_add_effects = segment.necessary_add_effects
-        assert necessary_add_effects is not None
         # Assert that this really is a general PNAD.
         assert len(pnad.op.add_effects) == 0, \
             "Can't spawn from non-general PNAD"
+        # Assert that the segment contains necessary_add_effects.
+        necessary_add_effects = segment.necessary_add_effects
+        assert necessary_add_effects is not None
 
         # Get an arbitrary grounding of the PNAD's operator whose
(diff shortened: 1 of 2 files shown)


src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -229,7 +226,8 @@ def _backchain_one_pass(
                 else:
                     nec_pnad_set_changed = True
                     pnad = self._spawn_new_pnad(
-                        param_opt_to_general_pnad[option.parent], segment)
+                        self._create_general_pnad_for_option(option.parent),
```
tomsilver, line 229: do you want to just move this inside of `_spawn_new_pnad`? or are we ever calling `_spawn_new_pnad` with some pnad other than the most general one?

tomsilver (approved):
thanks!! 🤗

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -62,14 +63,14 @@ def _learn(self) -> List[PartialNSRTAndDatastore]:
             for segment in seg_traj:
                 parameterized_options.add(segment.get_option().parent)
 
-        # Set up the param_opt_to_general_pnad and param_opt_to_nec_pnads
-        # dictionaries.
+        # Set up the and param_opt_to_nec_pnads dictionary.
```
ronuchit, line 66: remove "and"

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -62,14 +63,14 @@ def _learn(self) -> List[PartialNSRTAndDatastore]:
...
         for param_opt in parameterized_options:
-            pnad = self._initialize_general_pnad_for_option(param_opt)
-            param_opt_to_general_pnad[param_opt] = pnad
+            pnad = self._create_general_pnad_for_option(param_opt)
             param_opt_to_nec_pnads[param_opt] = []
-        self._assert_all_data_in_exactly_one_datastore(
-            list(param_opt_to_general_pnad.values()))
+        self._assert_all_data_in_exactly_one_datastore([
```
ronuchit, line 70: we can remove this check, it's pretty useless since the datastores will just be grouped by option

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -62,14 +63,14 @@ def _learn(self) -> List[PartialNSRTAndDatastore]:
             for segment in seg_traj:
                 parameterized_options.add(segment.get_option().parent)
 
-        # Set up the param_opt_to_general_pnad and param_opt_to_nec_pnads
-        # dictionaries.
+        # Set up the and param_opt_to_nec_pnads dictionary.
         for param_opt in parameterized_options:
-            pnad = self._initialize_general_pnad_for_option(param_opt)
-            param_opt_to_general_pnad[param_opt] = pnad
+            pnad = self._create_general_pnad_for_option(param_opt)
```
ronuchit, line 68: why are you creating this `pnad` variable but then never using it? why do we need to call this function here at all? can we just remove all this initializing?

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
ronuchit: agreed, let's move this into `_spawn_new_pnad` since we are now understanding "spawn" to mean "spawn from the most general"

ronuchit (changes requested):
let's clean things up a bit more in this one!

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -367,16 +358,18 @@ def _get_uniquely_named_nec_pnads(
...
         return "backchaining"
 
-    def _spawn_new_pnad(self, pnad: PartialNSRTAndDatastore,
-                        segment: Segment) -> PartialNSRTAndDatastore:
-        """Given a general PNAD and some segment with necessary add effects
-        that the PNAD must achieve, create a new PNAD ("spawn" from the most
-        general one) so that it has the necessary add effects contained in the
-        given segment.
+    def _spawn_new_pnad(self, segment: Segment) -> PartialNSRTAndDatastore:
+        """Given a PNAD and some segment with necessary add effects that the
```
ronuchit, line 362: fix docstring: this function is not given a PNAD anymore

src/nsrt_learning/strips_learning/gen_to_spec_learner.py
```diff
@@ -367,16 +358,18 @@ def _get_uniquely_named_nec_pnads(
...
 
         Returns the newly constructed PNAD, without modifying the
         original.
         """
+        # Create a general PNAD for the segment's option.
+        pnad = self._create_general_pnad_for_option(
+            segment.get_option().parent)
         # Assert that the segment contains necessary_add_effects.
         necessary_add_effects = segment.necessary_add_effects
         assert necessary_add_effects is not None
```
ronuchit, line 375: move "# Assert that this really is a general PNAD." and the subsequent line a few lines up, to right after we create the pnad

ronuchit (approved):
lgtm minor things

PR #1000, opened 2022-06-23
[BEHAVIOR] type fix

NishanthJKumar:
generalizes behavior types so we don't have to learn a separate operator for every single book...


tomsilver:
PR 1000! 🎉 🎉 🎉 🎉 🎉 

@ronuchit

PR #1001, opened 2022-06-24
DDPG (Deep Deterministic Policy Gradient) RL Option Learner

ashayathalye:
RL option learner that updates each option's regressor via (essentially) DDPG. To review DDPG, see the algorithm definition in https://arxiv.org/abs/1509.02971. Currently, it runs. I need to do more testing to see if there are run-time bugs (e.g. by making sure rewards go up as we go through more online learning cycles.) 

There are a number of ways in which what we do is different than DDPG. The main thing is that DDPG updates the actor and critic after each single interaction with the environment. Meanwhile, we update the actor and critic only after a large number of interactions (after each online learning cycle). 

The main things this PR will add is: 
* Replay buffer class in (currently in `option_learner.py`)
* Noise class (Ornstein-Uhlenbeck Process) to be used for exploration by adding noise to the policy output via a `_NoisyNeuralParameterizedOption` class that subclasses `_LearnedNeuralParameterizedOption` (both currently in `option_learner.py`)
* A basic critic network (currently in `ml_models.py`)
* DDPG agent (in `option_learner.py`) that provides an update method that is used to update each option.

Example command: `python3 src/main.py --env cover_multistep_options --approach nsrt_rl --seed 32 --option_learner direct_bc --sampler_learner neural --skip_until_cycle 1 --segmenter contacts --num_train_tasks 30 --nsrt_rl_batch_size 200 --num_online_learning_cycles 100`.

### DIFF
--- a/src/ml_models.py
+++ b/src/ml_models.py
@@ -422,6 +422,17 @@ class MLPRegressor(PyTorchRegressor):
         return nn.MSELoss()
 
 
+class Critic(MLPRegressor):
+    """A critic network to be used in actor-critic RL methods."""
+
+    def forward(self, state: Tensor, action: Tensor) -> Tensor:
+        tensor_X = torch.cat([state, action], dim=1)
+        for _, linear in enumerate(self._linears[:-1]):
+            tensor_X = F.relu(linear(tensor_X))
+        tensor_X = self._linears[-1](tensor_X)
+        return tensor_X
+
+
 class ImplicitMLPRegressor(PyTorchRegressor):
     """A regressor implemented via an energy function.
 
--- a/src/ml_models.py
+++ b/src/ml_models.py
@@ -425,6 +425,13 @@ class MLPRegressor(PyTorchRegressor):
 class Critic(MLPRegressor):
     """A critic network to be used in actor-critic RL methods."""
 
+    def initialize_net(self, input_dim, output_dim) -> None:
+        self._linears.append(nn.Linear(input_dim, self._hid_sizes[0]))
+        for i in range(len(self._hid_sizes) - 1):
+            self._linears.append(
+                nn.Linear(self._hid_sizes[i], self._hid_sizes[i + 1]))
+        self._linears.append(nn.Linear(self._hid_sizes[-1], output_dim))
+
     def forward(self, state: Tensor, action: Tensor) -> Tensor:
         tensor_X = torch.cat([state, action], dim=1)
         for _, linear in enumerate(self._linears[:-1]):
--- a/src/ml_models.py
+++ b/src/ml_models.py
@@ -425,8 +425,7 @@ class MLPRegressor(PyTorchRegressor):
 class Critic(nn.Module):
     """A critic network to be used in actor-critic RL methods."""
 
-    def __init__(self, hid_sizes: List[int], input_dim: int,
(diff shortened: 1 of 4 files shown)


src/ml_models.py
```diff
@@ -422,6 +422,29 @@ def _create_loss_fn(self) -> Callable[[Tensor, Tensor], Tensor]:
         return nn.MSELoss()
 
 
+class Critic(nn.Module):
```
tomsilver, line 425: when we're making other kinds of models, like for policies and samplers, we perform concatenation "outside" the NN module, which makes it easier to reuse general classes like Regressors, Classifiers, etc. can we do the same pattern here, and concatenate the state and action outside, and then use one of our existing ML models? this will also make sure we don't have issues with normalization, etc.

src/nsrt_learning/option_learning.py
```diff
@@ -636,16 +685,215 @@ class _DummyRLOptionLearner(_RLOptionLearnerBase):
...
+        batch = self._rng.choice(filled, batch_size)
+        states = self._state_memory[batch]
+        actions = self._action_memory[batch]
+        rewards = self._reward_memory[batch]
+        next_states = self._next_state_memory[batch]
+        done = self._terminal_memory[batch]
+        return states, actions, rewards, next_states, done
+
+
+class OUNoise:
```
tomsilver, line 736: nice!

src/nsrt_learning/option_learning.py
```diff
@@ -636,16 +685,215 @@ class _DummyRLOptionLearner(_RLOptionLearnerBase):
...
+                self._batch_size)
+            states = torch.from_numpy(states_)
+            actions = torch.from_numpy(actions_)
+            rewards = torch.from_numpy(rewards_).view((self._batch_size, 1))
+            next_states = torch.from_numpy(next_states_)
+            dones = torch.from_numpy(done_)
+
+            # Update networks
+            # TODO(ashay): clip gradients?
+            self._critic.eval()
```
tomsilver, line 857: if we stick with the ML model APIs, I think the `eval()` should be taken care of inside `predict` already (but you can double check me)

src/nsrt_learning/option_learning.py
```diff
@@ -636,16 +685,215 @@ class _DummyRLOptionLearner(_RLOptionLearnerBase):
...
+            # Note that the actor has _x_dim and _y_dim set
+            # already.
+            self._actor = cast(MLPRegressor, option.regressor)
+            self._target_actor = copy.deepcopy(self._actor)
+            input_dim = self._actor._x_dim
+            output_dim = self._actor._y_dim
+            self._critic = Critic(CFG.nsrt_rl_critic_regressor_hid_sizes,
+                                  state_dim + action_dim)
+            self._critic_loss_fn = nn.MSELoss()
+            self._target_critic = copy.deepcopy(self._critic)
```
tomsilver, line 822: deepcopy is generally slow and discouraged, but maybe this is the way that other implementations do this?

tomsilver (commented):
looks like a great start! I'll do a closer look through as the PR continues. my biggest high-level reaction is that we should try to stick as close as possible to the `ml_model` APIs, so that we can reuse as much of that code as possible.

PR #1003, opened 2022-06-24
Wester delivery ldl

wester-is-not-real:
Implemented started code for ldl delivery approach.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,4 @@ requests
 slack_bolt
 pybullet>=3.2.0
 sklearn
-pyqt5
+#pyqt5
--- a/src/approaches/delivery_approach.py
+++ b/src/approaches/delivery_approach.py
@@ -8,13 +8,13 @@ Example command line:
 
 from typing import Callable, cast
 
-import numpy as np
-
-from predicators.src.approaches import ApproachFailure
-from predicators.src.approaches.nsrt_learning_approach import NSRTLearningApproach
-from predicators.src.envs.pddl_env import _PDDLEnvState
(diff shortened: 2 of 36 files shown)


requirements.txt
```diff
@@ -15,4 +15,4 @@ requests
 slack_bolt
 pybullet>=3.2.0
 sklearn
-pyqt5
+#pyqt5
```
tomsilver, line 18: now that you've got everything installed, you can undo the change to this file (since we don't want to merge it)

src/approaches/delivery_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        unpacked = predicates["unpacked"]
+
+        nsrts = {n.name: n for n in self._get_current_nsrts()}
+        deliver = nsrts["deliver"]
+        move = nsrts["move"]
+        pick_up = nsrts["pick-up"]
+
+        paper_var, loc_var = deliver.parameters
+        from_var, to_var = move.parameters
+        paperPick_var, homeBase_var = pick_up.parameters
```
tomsilver, line 46: minor: let's stick to [snake case](https://en.wikipedia.org/wiki/Snake_case) for all variable names

src/approaches/delivery_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        rules = [
+            LDLRule("Pick-up",
+                    parameters=[paperPick_var, homeBase_var, to_var],
```
tomsilver, line 49: I think you can get away without the `to_var` here and without the goal precondition. In general, it's a good idea to try to minimize the number of parameters in these rules, because that will then reduce the number of ground operator that we'll consider in each state.

tomsilver (commented):
excellent work!!

PR #1005, opened 2022-06-24
Abraham delivery approach

Abraham-Mitchell:
Thanks Tom and Wester :)


tomsilver:
@Abraham-Mitchell looks great!! Let's resolve the linting / type checking / auto formatting issues for practice, but otherwise this looks good to me.

PR #1006, opened 2022-06-24
Behavior initial state sampling

wmcclinton:
Fixes BEHAVIOR's initial state sampling (i.e. when `behavior_randomize_init_state = True`) by loading up to 20 pre-computed scenes for each house and task, using BehaviorEnv's `task_instance_id` parameter. This allows us to start collecting demos for both the `sorting_books` and `re-shelving_library_books` task. I also had to change the way we saved states in behavior by extending the path to prevent naming collisions when saving the pybullet state. 

Note, when `behavior_randomize_init_state = False` it will load the same default scene for every house and task.


src/envs/behavior_options.py
```diff
@@ -707,8 +707,15 @@ def graspObjectOptionModel(_state: State, env: "BehaviorEnv") -> None:
         else:
             grasp_obj_body_id = obj_to_grasp.body_id
         # 3.1 Call code that does assisted grasping
+        # bypass_force_check is basically a hack we should
+        # turn it off for the final system and use a real grasp
```
ronuchit, line 711: so what's the plan with this going forward? if we are not planning to revisit, maybe we shouldn't say that we are going to turn it off for the final system?

src/envs/behavior.py
```diff
@@ -30,9 +29,8 @@
 
     _BEHAVIOR_IMPORTED = True
     bddl.set_backend("iGibson")  # pylint: disable=no-member
-    if os.path.exists("tmp_behavior_states/"):
-        shutil.rmtree("tmp_behavior_states/")
-    os.makedirs("tmp_behavior_states/")
```
NishanthJKumar, line 35: why do we not need this anymore?

src/settings.py
```diff
@@ -466,6 +466,14 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
                     # For the tools environment, keep it much lower.
                     "tools": 1,
                 })[args.get("env", "")],
+
+            # Used to save nsrts as strings in pickle file.
```
ronuchit, line 470: NSRTs*

src/settings.py
```diff
@@ -466,6 +466,14 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
...
+            dump_nsrts_as_strings=defaultdict(
+                lambda: False,
+                {
+                    # We cannot pickle Behavior nsrts
```
ronuchit, line 474: NSRTs*

ronuchit (commented):
looks good! i left some minor comments, and then we can let tom take a pass

src/envs/behavior.py
```diff
@@ -55,14 +53,13 @@ def __init__(self) -> None:
...
-                         CFG.behavior_config_file), CFG.behavior_task_name,
-            CFG.behavior_scene_name, CFG.behavior_randomize_init_state)
+            os.path.join(igibson.root_path, CFG.behavior_config_file),
+            CFG.behavior_task_name, CFG.behavior_scene_name,
+            False)  # behavior_randomize_init_state = False in config file
 
         super().__init__()  # To ensure self._seed is defined.
         self._rng = np.random.default_rng(self._seed)
-        self.set_igibson_behavior_env(self._seed)
-        self.igibson_behavior_env.robots[0].initial_z_offset = 0.7
```
NishanthJKumar, line 65: why don't we need this anymore?

src/envs/behavior.py
```diff
@@ -55,14 +53,13 @@ def __init__(self) -> None:
         if not _BEHAVIOR_IMPORTED:
             raise ModuleNotFoundError("Behavior is not installed.")
         self._config_file = modify_config_file(
-            os.path.join(igibson.root_path,
-                         CFG.behavior_config_file), CFG.behavior_task_name,
-            CFG.behavior_scene_name, CFG.behavior_randomize_init_state)
+            os.path.join(igibson.root_path, CFG.behavior_config_file),
+            CFG.behavior_task_name, CFG.behavior_scene_name,
+            False)  # behavior_randomize_init_state = False in config file
```
NishanthJKumar, line 58: is the comment meant to say that this setting will _always_ be false in the config file?

src/envs/behavior.py
```diff
@@ -361,6 +371,7 @@ def set_igibson_behavior_env(self, seed: int) -> None:
             raise RuntimeError("ERROR: Failed to sample iGibson BEHAVIOR "
                                "environment that meets bddl initial "
                                "conditions!")
+        self.igibson_behavior_env.robots[0].initial_z_offset = 0.7
```
NishanthJKumar, line 374: oh - i see you just moved this line down here now. Does it need to be here? Can we just revert and keep it where it was?

src/envs/behavior.py
```diff
@@ -369,15 +380,14 @@ def _ig_object_to_object(self, ig_obj: "ArticulatedObject") -> Object:
         ig_obj_name = self._ig_object_name(ig_obj)
         return Object(ig_obj_name, obj_type)
 
-    @functools.lru_cache(maxsize=None)
     def object_to_ig_object(self, obj: Object) -> "ArticulatedObject":
         """Maintains a mapping of objects to underlying igibson objects."""
         return self._name_to_ig_object(obj.name)
 
-    @functools.lru_cache(maxsize=None)
```
NishanthJKumar, line 377: can we add comments here mentioning that it's really bad for anyone to _ever_ add caching to these functions?

NishanthJKumar (commented):
great work on this! just added a few suggestions to streamline/make things simpler

src/envs/behavior.py
wmcclinton: Yes it needs to be moved here. Since we are loading new scenes we need to set this after everytime we load a new scene not just on init.

src/envs/behavior.py
wmcclinton: Because these saved states are needed for loading the states to compute the abstracted states during NSRT learning when we `--load_data`.

src/envs/behavior_options.py
wmcclinton: We will eventually fix this

wmcclinton: Next year

src/envs/behavior.py
```diff
@@ -30,9 +29,8 @@
 
     _BEHAVIOR_IMPORTED = True
     bddl.set_backend("iGibson")  # pylint: disable=no-member
-    if os.path.exists("tmp_behavior_states/"):
-        shutil.rmtree("tmp_behavior_states/")
-    os.makedirs("tmp_behavior_states/")
+    if not os.path.exists("tmp_behavior_states/"):
```
tomsilver, line 32: with this change, our local `tmp_behavior_states/` directories are going to grow over time, and we'll have to periodically delete them. I guess that's okay, because the same can be said for `results/`, `videos/`, etc.

src/envs/behavior.py
```diff
@@ -54,15 +52,16 @@ class BehaviorEnv(BaseEnv):
...
+            CFG.behavior_task_name, CFG.behavior_scene_name, False)
 
         super().__init__()  # To ensure self._seed is defined.
         self._rng = np.random.default_rng(self._seed)
-        self.set_igibson_behavior_env(self._seed)
-        self.igibson_behavior_env.robots[0].initial_z_offset = 0.7
+        self.set_igibson_behavior_env(task_instance_id=0, seed=self._seed)
         self._type_name_to_type: Dict[str, Type] = {}
         # a unique id for saving and loading each task's state
         self.task_num: int = 0
```
tomsilver, line 67: i'm getting confused about `self.task_num`. is this supposed to be like a counter of all tasks created, or is it supposed to be the ID of the current environment task or something?

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -84,6 +85,10 @@ def _test_approach(env_name,
                                                    task.goal_holds,
                                                    max_num_steps=CFG.horizon)
             assert task.goal_holds(traj.states[-1])
+    # Do not load NSRTs from pickle file. This is because we cannot load
+    # NSRTs when they are saved as strings (Necessary for behavior)
+    if no_loading:
+        return approach
```
tomsilver, line 91: why is this returning approach?

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -117,6 +122,14 @@ def test_nsrt_learning_approach():
                               additional_settings={
                                   "compute_sidelining_objective_value": True,
                               })
+    approach = _test_approach(env_name="blocks",
+                              approach_name="nsrt_learning",
+                              try_solving=False,
+                              no_loading=True,
+                              additional_settings={
+                                  "compute_sidelining_objective_value": True,
```
tomsilver, line 130: remove `compute_sidelining_objective_value` True? it'll be slower as True, right? and it's covered by the test above

tomsilver (commented):
@wmcclinton good stuff! I have one question about `self.task_num` and then I'll take another quick look after that's answered

tests/approaches/test_nsrt_learning_approach.py
ronuchit: the function nominally returns approach anyway, this is just a flag that early terminates before we try loading

src/envs/behavior.py
wmcclinton: Yah I don't know a better way

src/envs/behavior.py
wmcclinton: `self.task_num` is always a counter of the tasks created. If `behavior_randomize_init_state = True` it will also act as the ID of the current environment instance which also goes from [0...19].

src/envs/behavior.py
```diff
@@ -166,19 +165,30 @@ def _generate_train_tasks(self) -> List[Task]:
...
+                               testing=True)
+
+    def _get_tasks(self,
+                   num: int,
+                   rng: np.random.Generator,
+                   testing: bool = False) -> List[Task]:
         tasks = []
+        assert num <= 10  # Max 10 train and test tasks for behavior
+        if testing:
+            self.task_num += 10
```
tomsilver, line 179: What's confusing me about task num is that it's incremented all at once here. If you're using it to identify different tasks, why does this work? I would have thought that you would increment it one by one after each task is created.

wmcclinton: Good catch I think it should be `self.task_num = 10`. So if `behavior_randomize_init_state = True` it will also need to act as the ID of the current environment instance which will go [0...N] for N train tasks and [10...10+M] for M train tasks

tomsilver: Two more questions based on that:
1. Can we set it to a CFG value instead of just 10? Like can it just be CFG.num_train_tasks instead?
2. Is there a way to make it not a class variable and to just have it be a variable in this function? Or where else is it gonna be importantly used?

wmcclinton: Yes we can do 1 where `self.task_num = CFG.max_num_train_tasks`. We use it when we load in different scenes. The task_num changes based on what scene is loaded.

tomsilver (approved):
lgtm after changing 10 to `CFG.num_train_tasks` (in two places)

PR #1007, opened 2022-06-25
factor out metacontroller approach 

tomsilver:


src/approaches/nsrt_metacontroller_approach.py
```diff
@@ -0,0 +1,72 @@
+"""An abstract approach implements a policy in the abstract transition space
```
ronuchit, line 1: that implements*

src/settings.py
```diff
@@ -215,8 +215,8 @@ class GlobalSettings:
     gnn_option_policy_shooting_variance = 0.1
     gnn_option_policy_shooting_max_samples = 100
 
-    # parameters for GNN metacontroller approach
-    gnn_metacontroller_max_samples = 100
+    # parameters for metacontroller approaches
```
ronuchit, line 218: looks like you already did but might as well:
git grep -i "gnn meta"
git grep -i "gnnmeta"
git grep -i "gnn_meta"
to make sure nowhere else needs to change

tests/approaches/test_gnn_metacontroller_approach.py
```diff
@@ -77,11 +75,11 @@ def test_gnn_metacontroller_approach_with_envs(env_name, num_epochs):
         policy = approach.solve(task, timeout=CFG.timeout)
         with pytest.raises(ApproachFailure) as e:
             policy(task.init)
-        assert "GNN metacontroller could not sample an option" in str(e)
+        assert "Metacontroller could not sample an option" in str(e)
```
ronuchit, line 78: do we need a separate test file for the new file? or no bc its abstract?

ronuchit (approved):

tests/approaches/test_gnn_metacontroller_approach.py
tomsilver: yeah nah

PR #1010, opened 2022-06-26
abstract task struct

tomsilver:


ronuchit:
why do we need this? add info to PR description. we already have PDDLEnvs which worked without this. also, many many things would need refactoring with this, such as task_plan only needing to take in an abstract task instead of an init_atoms and a goal (this pattern happens many other places too)

tomsilver:
it's convenient for forthcoming PG3 code. I agree that many things would be refactorable. is your feeling that we should just not do this then, and use tuples of `object, atoms, goal` where I need them?

ronuchit:
yeah i feel like that's better unless we're planning to properly refactor things to use this, otherwise it's confusing to have this while also having all the old code that uses those tuples (objects, atoms, goal)

PR #1011, opened 2022-06-26
a few updates to lifted decision list struct

tomsilver:


tests/test_structs.py
```diff
@@ -1038,3 +1052,15 @@ def test_lifted_decision_lists():
 
     atoms = set()
     assert utils.query_ldl(ldl, atoms, goal) is None
+
+    ldl2 = LiftedDecisionList(rules)
```
ronuchit, line 1056: add one more with rules[:] (i.e. a copy)

ronuchit (approved):

PR #1012, opened 2022-06-26
new utils for enumerating lifted variable combos

tomsilver:


src/structs.py
```diff
@@ -1494,6 +1494,7 @@ class LiftedDecisionList:
 NSRTOrSTRIPSOperator = TypeVar("NSRTOrSTRIPSOperator", NSRT, STRIPSOperator)
 GroundNSRTOrSTRIPSOperator = TypeVar("GroundNSRTOrSTRIPSOperator", _GroundNSRT,
                                      _GroundSTRIPSOperator)
+ObjectOrVariable = TypeVar("ObjectOrVariable", bound=_TypedEntity)
```
ronuchit, line 1497: what's the difference between this and _TypedEntity in practice? also, why not call it Entity?

ronuchit (approved):
lgtm

src/structs.py
tomsilver: I think we need to do it this way for `get_object_combinations` and `get_variable_combinations` to be able to ensure that their outputs are objects and variables respectively. I'm calling it `ObjectOrVariable` to try to make clear that it should be used as a placeholder for Object or Variable types specifically. /shrug

PR #1013, opened 2022-06-26
run Astar util

tomsilver:


ronuchit (approved):
if we cared, we could change skeleton search in planner.py to use this... but i guess we don't care? should we make an issue for it?

are there any other places we do search that isn't using these util functions? i can't think of any...i believe predicate invention uses run_hill_climbing

tomsilver:
https://github.com/Learning-and-Intelligent-Systems/predicators/issues/1014

yeah I don't know of any other searches that do their own thing

PR #1016, opened 2022-06-26
policy-guided A* search

tomsilver:


src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
+        heuristic: Callable[[_S], float],
+        policy: Callable[[_S], Optional[_A]],
+        num_rollout_steps: int,
+        rollout_step_cost: float = 0,  # by default, rollout steps are free
+        max_expansions: int = 10000000,
+        max_evals: int = 10000000,
+        timeout: int = 10000000,
+        lazy_expansion: bool = False) -> Tuple[List[_S], List[_A]]:
+    """Perform A* search, but at each node, roll out a given policy for a given
+    number of time steps, creating new successors at each step.
```
ronuchit, line 1532: timesteps* (for consistency)

src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
+
+    Stop the roll out prematurely if the policy returns None.
```
ronuchit, line 1534: rollout*

src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
+
+    Note that unlike the other search functions, which take get_successors as
+    input, this function takes get_valid_actions and get_next_state as two
+    separate inputs. This is necessary because we need to anticipate the next
+    state conditioned on the action output by the policy.
+
+    The get_valid_actions generates (action, cost) tuples. For policy-generated
+    transitions, the costs are ignored, and rollout_step_cost is used instead.
```
ronuchit, line 1542: why are rollouts free?

src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
 
 
+def run_policy_guided_astar(
+        initial_state: _S,
+        check_goal: Callable[[_S], bool],
+        get_valid_actions: Callable[[_S], Iterator[Tuple[_A, float]]],
+        get_next_state: Callable[[_S, _A], _S],
+        heuristic: Callable[[_S], float],
+        policy: Callable[[_S], Optional[_A]],
+        num_rollout_steps: int,
```
ronuchit, line 1525: what will this usually be in practice?

src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
+        # Get policy-based successors.
+        policy_state = state
+        policy_action_seq = []
+        for _ in range(num_rollout_steps):
+            action = policy(policy_state)
+            if action is None:
+                break
+            policy_state = get_next_state(policy_state, action)
+            policy_action_seq.append(action)
+            yield (list(policy_action_seq), policy_state, rollout_step_cost)
```
ronuchit, line 1556: i don't really understand the meaning of a successor here, since you're using `List[_A]` instead of the usual `_A`. is it supposed to mean, "from this state, if you take this sequence of actions in order, you'll end up at this final state"?

ronuchit, line 1556: why don't you have to accumulate the rollout cost in the same way you're accumulating actions into `policy_action_seq`?

src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
+
+        # Get primitive successors.
+        for action, cost in get_valid_actions(state):
+            next_state = get_next_state(state, action)
+            yield ([action], next_state, cost)
+
+    _, action_subseqs = run_astar(initial_state=initial_state,
```
ronuchit, line 1563: nice that our abstraction works out even though `List[_A]` instead of `_A`

src/utils.py
ronuchit, line 1556: why output every state encountered along this rollout instead of only the final state? don't you always want to use the policy all the way to its end? (but maybe not, if it's not super trustworthy)

src/utils.py
```diff
@@ -1515,6 +1515,75 @@ def run_hill_climbing(
...
+                                  check_goal=check_goal,
+                                  get_successors=get_successors,
+                                  heuristic=heuristic,
+                                  max_expansions=max_expansions,
+                                  max_evals=max_evals,
+                                  timeout=timeout,
+                                  lazy_expansion=lazy_expansion)
+
+    # The states are "jumpy", so we need to reconstruct the dense state
```
ronuchit, line 1572: it's a bit confusing that you're throwing out the state sequence returned by `run_astar` and then reconstructing it from scratch. but i guess you need to do this because you need `get_successors` to return a `_S`, not an entire `List[_S]` since that would change the API of `heuristic` and `check_goal`

tests/test_utils.py
```diff
@@ -2766,6 +2766,96 @@ def _local_minimum_grid_heuristic_fn(state: S) -> float:
         ]
 
 
+def test_run_policy_guided_astar():
+    """Tests for run_policy_guided_astar()."""
```
ronuchit, line 2770: nice testing

ronuchit (approved):

src/utils.py
ronuchit: hmm but actually yeah since rollouts are free, what's going to happen is that you roll the policy out all the way, then find the state along that trajectory which has the lowest heuristic value, then do regular A* from there.... is that right?

src/utils.py
tomsilver: one day I will get this right on the first try

src/utils.py
tomsilver: I think this one is actually right as is -- "rollout" is noun, "roll out" is verb

src/utils.py
tomsilver: this only makes sense in the context of PG3... I will change it to default 1

src/utils.py
tomsilver: we usually use 50 but it doesn't matter that much

src/utils.py
tomsilver: yeah, I'll add that as a comment

tomsilver: yes thanks

tomsilver: yeah what you said is right. and this is important when policies are not totally trustworthy

src/utils.py
tomsilver: yeah exactly

tests/test_utils.py
```diff
@@ -2766,6 +2766,96 @@ def _local_minimum_grid_heuristic_fn(state: S) -> float:
...
+        # Move right until we can't anymore.
+        _, c = state
+        if c >= arrival_costs.shape[1] - 1:
+            return None
+        return "right"
+
+    initial_state = (0, 0)
+    num_rollout_steps = 10
+
+    # The policy should bias toward the path that moves all the way right, then
```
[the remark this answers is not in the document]
tomsilver: i tried this out. I had to change the arrival cost for the bottom-middle state from 100 to 1 because otherwise the search wanted to enter that state from the right with the policy (to avoid the environment cost). but after that change, the result is
```
ipdb> action_sequence
['right', 'right', 'down', 'down', 'down', 'down']
```
as you expected

src/utils.py
ronuchit: it's "rollout" -- you're using it as a noun here. the rollout is the thing that you're stopping

PR #1017, opened 2022-06-26
partial re-implementation of PG3

tomsilver:
reference: https://arxiv.org/abs/2204.10420

not yet implemented:
* the bottom-up search operator (which can be extremely helpful)
* the delete rule / condition operators
* many possible optimizations
* domains other than Delivery

differences from the paper:
* defaulting to hill-climbing search instead of GBFS
* added a `pg3_plan_compare_inapplicable_cost`, which seems to make learning better in Delivery at least

example:
```python src/main.py --approach pg3 --seed 0  --env pddl_easy_delivery_procedural_tasks --strips_learner oracle --num_train_tasks 10```

```
Keeping best policy:
LiftedDecisionList[
LDLRule-pick-up:
    Parameters: [?paper:paper, ?loc:loc]
    Pos State Pre: [at(?loc:loc), ishomebase(?loc:loc), unpacked(?paper:paper)]
    Neg State Pre: []
    Goal Pre: []
    NSRT: pick-up(?paper:paper, ?loc:loc)
LDLRule-deliver:
    Parameters: [?loc:loc, ?paper:paper]
    Pos State Pre: [at(?loc:loc), carrying(?paper:paper)]
    Neg State Pre: [satisfied(?loc:loc)]
    Goal Pre: []
    NSRT: deliver(?paper:paper, ?loc:loc)
LDLRule-move:
    Parameters: [?from:loc, ?to:loc]
    Pos State Pre: [at(?from:loc), safe(?from:loc)]
    Neg State Pre: [satisfied(?to:loc)]
    Goal Pre: []
    NSRT: move(?from:loc, ?to:loc)
]
```

which gets 50/50.

### CODE REVIEW

File: src/approaches/pg3_approach.py [hunk 1]
```diff
@@ -0,0 +1,429 @@
...
+                                task_idx: int) -> Sequence[Set[GroundAtom]]:
+        del ldl  # unused
+        return self._get_demo_atom_plan_for_task(task_idx)
+
+    @functools.lru_cache(maxsize=None)
+    def _get_demo_atom_plan_for_task(
+            self, task_idx: int) -> Sequence[Set[GroundAtom]]:
+        demo_atom_plan = None
+        for ground_atom_traj in self._ground_atom_trajs:
+            traj, atom_seq = ground_atom_traj
```
ronuchit, line 359: generally in our codebase we use `atoms_seq` instead of `atom_seq` -- maybe do a find-and-replace?

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+                return True
+            ground_nsrt = utils.query_ldl(ldl, atoms, goal)
+            if ground_nsrt is None:
+                return False
+            atoms = utils.apply_operator(ground_nsrt, atoms)
+        return goal.issubset(atoms)
+
+
+class _PlanComparisonPG3Heuristic(_PG3Heuristic):
+    """Score a policy based on agreement with certain plans."""
```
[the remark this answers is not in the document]
ronuchit: oh, i see, this is what's left to subclasses. make that more clear.

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+from predicators.src.approaches.nsrt_metacontroller_approach import \
+    NSRTMetacontrollerApproach
+from predicators.src.settings import CFG
+from predicators.src.structs import NSRT, Box, Dataset, GroundAtom, \
+    GroundAtomTrajectory, LDLRule, LiftedAtom, LiftedDecisionList, \
+    LowLevelTrajectory, ParameterizedOption, Predicate, State, Task, Type, \
+    Variable, _GroundNSRT
+
+
+class PG3Approach(NSRTMetacontrollerApproach):
```
ronuchit, line 31: it's cool that this subclasses NSRTMetacontrollerApproach. that means we can try it with all envs, not just pddl envs, right? are there any conceptual blockades there?

ronuchit (commented):
very cool! i like the algorithm

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def _get_atom_plan_for_task(self, ldl: LiftedDecisionList,
+                                task_idx: int) -> Sequence[Set[GroundAtom]]:
+        del ldl  # unused
+        return self._get_demo_atom_plan_for_task(task_idx)
+
+    @functools.lru_cache(maxsize=None)
+    def _get_demo_atom_plan_for_task(
+            self, task_idx: int) -> Sequence[Set[GroundAtom]]:
+        demo_atom_plan = None
```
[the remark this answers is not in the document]
ronuchit: oh oops my `assert` there doesn't check for uniqueness, so needs to be fixed a bit

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+            traj, atom_seq = ground_atom_traj
+            if traj.train_task_idx == task_idx:
+                assert demo_atom_plan is None
+                demo_atom_plan = atom_seq
+        assert demo_atom_plan is not None
+        return demo_atom_plan
+
+
+class _PolicyGuidedPG3Heuristic(_PlanComparisonPG3Heuristic):
+    """Score a policy based on agreement with policy-guided plans."""
```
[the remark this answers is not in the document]
ronuchit: ignore this, in favor of below comment starting with "what is the TL;DR intuition"

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+                missed_steps += CFG.pg3_plan_compare_inapplicable_cost
+            else:
+                predicted_atoms = utils.apply_operator(ground_nsrt,
+                                                       atom_seq[t])
+                if predicted_atoms != atom_seq[t + 1]:
+                    missed_steps += 1
+        return missed_steps
+
+
+class _DemoPlanComparisonPG3Heuristic(_PlanComparisonPG3Heuristic):
```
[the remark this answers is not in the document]
ronuchit: okay but actually you're sorta generating those demos anyway

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+    def _create_heuristic(
+            self, trajectories: Sequence[LowLevelTrajectory]) -> _PG3Heuristic:
+        preds = self._get_current_predicates()
+        nsrts = self._get_current_nsrts()
+        ground_atom_trajs = utils.create_ground_atom_dataset(
+            trajectories, preds)
+        heuristic_name_to_cls: Dict[str, TypingType[_PG3Heuristic]] = {
+            "policy_guided": _PolicyGuidedPG3Heuristic,
+            "policy_evaluation": _PolicyEvaluationPG3Heuristic,
+            "plan_comparison": _DemoPlanComparisonPG3Heuristic,
```
ronuchit, line 131: rename this to `demo_plan_comparison` since PlanComparisonPG3Heuristic is its own thing (it's an abstract class)

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+                                  atoms: Set[GroundAtom],
+                                  goal: Set[GroundAtom]) -> bool:
+        for _ in range(CFG.horizon):
+            if goal.issubset(atoms):
+                return True
+            ground_nsrt = utils.query_ldl(ldl, atoms, goal)
+            if ground_nsrt is None:
+                return False
+            atoms = utils.apply_operator(ground_nsrt, atoms)
+        return goal.issubset(atoms)
```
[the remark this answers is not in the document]
tomsilver: it is unlikely but `return False` feels incorrect, and yeah `return True` -> `break` would just check the same thing twice. I hesitated about this too but I think this is best

src/approaches/pg3_approach.py
```diff
@@ -0,0 +1,429 @@
...
+
+    @staticmethod
+    def _count_missed_steps(ldl: LiftedDecisionList,
+                            atom_seq: Sequence[Set[GroundAtom]],
+                            goal: Set[GroundAtom]) -> float:
+        missed_steps = 0.0
+        for t in range(len(atom_seq) - 1):
+            ground_nsrt = utils.query_ldl(ldl, atom_seq[t], goal)
+            if ground_nsrt is None:
+                missed_steps += CFG.pg3_plan_compare_inapplicable_cost
```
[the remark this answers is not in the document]
tomsilver: yeah this is something we didn't do in the original implementation and I was playing around with here. a lower value seems to help because it means that if policy A is "confidently incorrect" and policy B is "not sure", we prefer policy B. this is maybe good because the policy search goes from general to specific, so there is kinda more hope for policy B than A in the future of search. but we will have to experiment at scale to see whether this actually helps.

src/approaches/pg3_approach.py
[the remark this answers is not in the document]
tomsilver: hopefully this is better now

tomsilver:

src/approaches/pg3_approach.py
```diff
@@ -354,36 +371,44 @@ def _get_atom_plan_for_task(self, ldl: LiftedDecisionList,
     @functools.lru_cache(maxsize=None)
     def _get_demo_atom_plan_for_task(
```
ronuchit, line 372: have you verified the caching works as intended?

src/approaches/pg3_approach.py
```diff
@@ -324,26 +342,25 @@ def _get_atom_plan_for_task(self, ldl: LiftedDecisionList,
 
     @staticmethod
     def _count_missed_steps(ldl: LiftedDecisionList,
-                            atom_seq: Sequence[Set[GroundAtom]],
+                            atoms_seq: Sequence[Set[GroundAtom]],
                             goal: Set[GroundAtom]) -> float:
         missed_steps = 0.0
-        for t in range(len(atom_seq) - 1):
-            ground_nsrt = utils.query_ldl(ldl, atom_seq[t], goal)
+        for t in range(len(atoms_seq) - 1):
```
ronuchit, line 348: i saw some stray uses of `atom_seq` in `test_datasets.py` as well, want to just lump that in here?

ronuchit: (or a new PR or just leave an issue for now)

tomsilver: #1020

src/approaches/pg3_approach.py
tomsilver: yeah it does

PR #1021, opened 2022-06-27
utils.py cleanup

ronuchit:
* remove several unused methods
* merge multiple methods that implement the same grounding functionality
* remove old note about disallowed duplicate arguments, update comment
to say that duplicate arguments ARE allowed
* add a small utility script for searching for unused functions, in case
it's helpful in the future again


scripts/find_unused_functions.py
```diff
@@ -0,0 +1,30 @@
+"""Report all unused top-level functions in a particular file.
```
tomsilver, line 1: fancy af

tomsilver (approved):
thanks for doing this! lmk if you need help with the amber stuff

PR #1027, opened 2022-06-27
[BEHAVIOR] Behavior type generalization

NishanthJKumar:
Generalizes behavior types so we don't have to learn a separate operator for every single book.

This change also allows us to get rid of a previously-complicated function!


ronuchit (approved):
nice!

PR #1028, opened 2022-06-28
[BEHAVIOR] Cuts down the number of 'reachable' predicates to only what's necessary

NishanthJKumar:
Prevents the creation of ground atoms like `reachable-hardback-hardback(book1, book1)`, which were previously making NSRT learning return complicated and hyper-specific operators.

### DIFF
--- a/src/envs/behavior.py
+++ b/src/envs/behavior.py
@@ -254,6 +254,11 @@ class BehaviorEnv(BaseEnv):
             for type_combo in itertools.product(types_lst, repeat=arity):
                 pred_name = self._create_type_combo_name(name, type_combo)
                 pred = Predicate(pred_name, list(type_combo), classifier)
+                # We only care about reachable when the agent is one of the
+                # types.
+                if name == "reachable" and not any(type_i.name == "agent"
+                                                   for type_i in type_combo):
+                    continue
                 predicates.add(pred)
 
         # Finally, add the reachable-nothing predicate, which only applies
@@ -453,6 +458,13 @@ class BehaviorEnv(BaseEnv):
         assert len(objs) == 2
         ig_obj = self.object_to_ig_object(objs[0])
         ig_other_obj = self.object_to_ig_object(objs[1])
+        # If the two objects are the same (i.e reachable(agent, agent)),
+        # we always want to return False so that when we learn
+        # operators, such predicates don't needlessly appear in preconditions.
+        # In the future, we probably want to prevent such atoms from
+        # ever being created.
+        if ig_obj == ig_other_obj:
+            return False
         return (np.linalg.norm(  # type: ignore
             np.array(ig_obj.get_position()) -
             np.array(ig_other_obj.get_position())) < 2)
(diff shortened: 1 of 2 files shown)


src/envs/behavior.py
```diff
@@ -254,6 +254,11 @@ def predicates(self) -> Set[Predicate]:
             for type_combo in itertools.product(types_lst, repeat=arity):
                 pred_name = self._create_type_combo_name(name, type_combo)
                 pred = Predicate(pred_name, list(type_combo), classifier)
+                # We only care about reachable when the agent is one of the
```
tomsilver, line 257: Move this up two lines?

src/envs/behavior.py
```diff
@@ -453,6 +458,13 @@ def _reachable_classifier(self, state: State,
         assert len(objs) == 2
         ig_obj = self.object_to_ig_object(objs[0])
         ig_other_obj = self.object_to_ig_object(objs[1])
+        # If the two objects are the same (i.e reachable(agent, agent)),
+        # we always want to return False so that when we learn
+        # operators, such predicates don't needlessly appear in preconditions.
+        # In the future, we probably want to prevent such atoms from
```
tomsilver, line 464: I don't get it -- why is this in the future? Won't this change prevent such atoms from ever getting created?

src/ground_truth_nsrts.py
```diff
@@ -2763,17 +2763,17 @@ def _get_predicate(base_pred_name: str,
...
 
-    # We start by creating reachable predicates for all possible type
-    # combinations. These predicates will be used as side predicates
-    # for navigateTo operators
+    agent_type = type_name_to_type["agent"]
+    agent_obj = Variable("?agent", agent_type)
+
+    # We start by creating reachable predicates for the agent and
+    # all possible other type combinations. These predicates will
+    # be used as side predicates for navigateTo operators
```
tomsilver, line 2771: nit: add period at the end

tomsilver (approved):

src/envs/behavior.py
ronuchit: i think nishanth means that we should never even call predicate classifiers with repeated arguments. but idk about that, you could totally imagine some settings where repeated arguments make sense. i think it's best to leave it to the classifier to decide what it wants to do wrt repeated argument, so like you're doing now already. i'm in favor of removing this sentence "In the future, ..."

src/ground_truth_nsrts.py
```diff
@@ -2763,17 +2763,17 @@ def _get_predicate(base_pred_name: str,
...
         return pred_name_to_pred[pred_name]
 
-    # We start by creating reachable predicates for all possible type
-    # combinations. These predicates will be used as side predicates
-    # for navigateTo operators
+    agent_type = type_name_to_type["agent"]
+    agent_obj = Variable("?agent", agent_type)
+
+    # We start by creating reachable predicates for the agent and
+    # all possible other type combinations. These predicates will
```
ronuchit, line 2770: do you mean "types" instead of "type combinations"? you're no longer doing a combination here -- that's precisely what this change is removing!

ronuchit (approved):

PR #1033, opened 2022-06-29
allow max training iterations in `ml_models.py` to depend on the dataset size

### LINKED ISSUE #659: allow `predicate_mlp_classifier_max_itr` to depend on the dataset size
we need this for the interactive learning approach. it will be good if we can implement this in a way that suggests a general pattern, in case we want to vary other hyperparameters later on too

ronuchit:
Example usage:
`python src/main.py --env cover --approach nsrt_learning --seed 0 --sampler_mlp_classifier_max_itr "lambda n: n * 100"`
```
Tasks solved: 50 / 50
Average time for successes: 0.04212 seconds
Test results: defaultdict(<class 'float'>, {'num_solved': 50, 'num_total': 50, 'avg_suc_time': 0.04211793422698975, 'min_skeletons_optimized': 1.0, 'max_skeletons_optimized': 2.0, 'num_solve_timeouts': 0, 'num_solve_failures': 0, 'num_execution_timeouts': 0, 'num_execution_failures': 0, 'avg_num_skeletons_optimized': 1.1, 'avg_num_nodes_expanded': 2.8, 'avg_num_nodes_created': 6.6, 'avg_num_nsrts': 2.0, 'avg_num_preds': 5.0, 'avg_plan_length': 2.46, 'avg_num_failures_discovered': 0.0, 'num_offline_transitions': 123, 'num_online_transitions': 0, 'query_cost': 0.0, 'learning_time': 22.879285097122192})
```

closes #659

### DIFF
--- a/src/ml_models.py
+++ b/src/ml_models.py
@@ -21,7 +21,7 @@ from sklearn.neighbors import \
 from torch import Tensor, nn, optim
 from torch.distributions.categorical import Categorical
 
-from predicators.src.structs import Array, Object, State
+from predicators.src.structs import Array, MaxTrainIters, Object, State
 
 torch.use_deterministic_algorithms(mode=True)  # type: ignore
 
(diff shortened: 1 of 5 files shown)


tomsilver (approved):
cool thank you! do you have a vision for how we should generalize this if we want to vary other hyperparameters over time?

ronuchit:
well max iters doesn't make sense to vary over time, but for something like learning rate we can easily do the same thing as here if we ever feel like it

ronuchit:
actually this might be better for learning rate https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate

PR #1035, opened 2022-06-29
slightly refactor `ml_models.py` to allow other ensemble types

### LINKED ISSUE #685: try different classifiers

ronuchit:
and add a setting for choosing the predicate classifier model

closes #685 (well, you would still need to actually test with supercloud, but we don't need to keep the issue open for that)


tomsilver (approved):
amber is going to be very grateful!

PR #1037, opened 2022-06-29
online nsrt learning

tomsilver:
@ronuchit can you comment on the design here before I add tests?

### DIFF
--- a/tests/approaches/test_online_nsrt_learning_approach.py
+++ b/tests/approaches/test_online_nsrt_learning_approach.py
@@ -1,9 +1,7 @@
 """Test cases for the online NSRT learning approach."""
 
 from contextlib import nullcontext as does_not_raise
-from typing import Dict, Sequence
 
-import numpy as np
 import pytest
 
 from predicators.src import utils
@@ -14,7 +12,6 @@ from predicators.src.datasets import create_dataset
 from predicators.src.envs.cover import CoverEnv
 from predicators.src.main import _generate_interaction_results
 from predicators.src.settings import CFG
-from predicators.src.structs import NSRT, Action, Array, Dataset, Object, State
 from predicators.src.teacher import Teacher
 
 
@@ -40,8 +37,8 @@ def test_online_nsrt_learning_approach(explorer, expectation):
     env = CoverEnv()
     train_tasks = env.get_train_tasks()
     approach = OnlineNSRTLearningApproach(env.predicates, env.options,
-                                           env.types, env.action_space,
-                                           train_tasks)
+                                          env.types, env.action_space,
+                                          train_tasks)
     teacher = Teacher(train_tasks)
     dataset = create_dataset(env, train_tasks, env.options)
     approach.learn_from_offline_dataset(dataset)


src/approaches/online_nsrt_learning_approach.py
```diff
@@ -0,0 +1,143 @@
+"""An NSRT learning approach that collects and learns from data online.
+
+Example command:
+    python src/main.py --approach online_nsrt_learning --seed 0 \
+        --env cover \
+        --max_initial_demos 1 \
+        --num_train_tasks 1000 \
+        --num_test_tasks 10 \
+        --max_num_steps_interaction_request 10 \
```
ronuchit, line 9: why is this so low?

src/approaches/online_nsrt_learning_approach.py
```diff
@@ -0,0 +1,143 @@
...
+        # explored at each time step is a hyperparameter. The train task
+        # is randomly selected.
+        explorer = self._get_explorer()
+        requests = []
+        for _ in range(CFG.online_nsrt_learning_tasks_per_request):
+            # Select a random task (with replacement).
+            task_idx = self._rng.choice(len(self._train_tasks))
+            task = self._train_tasks[task_idx]
+            # Set up the explorer policy and termination function.
+            policy, terminal = explorer.solve(task)
```
ronuchit, line 55: rename to `termination_function`

ronuchit (commented):
looks reasonable, but we'll see how things evolve when you add in GLIB. question: shouldn't we be unifying this with the stuff going on in `interactive_learning_approach.py`?

src/approaches/online_nsrt_learning_approach.py
```diff
@@ -0,0 +1,143 @@
...
+    def learn_from_offline_dataset(self, dataset: Dataset) -> None:
+        # Update the dataset with the offline data.
+        self._dataset = Dataset(dataset.trajectories)
+        super().learn_from_offline_dataset(dataset)
+
+    def learn_from_interaction_results(
+            self, results: Sequence[InteractionResult]) -> None:
+        # Add the new data to the cumulative dataset.
+        for result in results:
+            traj = LowLevelTrajectory(result.states, result.actions)
```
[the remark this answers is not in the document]
tomsilver: good point. I'm looking through `git grep '.is_demo'` now. I know that predicate learning will filter this out. NSRT learning should be fine I think. we will have to come back for predicate learning...

src/approaches/online_nsrt_learning_approach.py
```diff
@@ -0,0 +1,143 @@
...
+        self._dataset = Dataset([])
+        self._online_learning_cycle = 0
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "online_nsrt_learning"
+
+    def get_interaction_requests(self) -> List[InteractionRequest]:
+        # Explore in the train tasks. The number of train tasks that are
+        # explored at each time step is a hyperparameter. The train task
```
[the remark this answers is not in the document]
tomsilver: @&*#$*#@&2!

src/approaches/online_nsrt_learning_approach.py
tomsilver: just because each action is a pick or a place in cover

tomsilver:
the new tests add about 0.75 seconds

tomsilver:
>question: shouldn't we be unifying this with the stuff going on in interactive_learning_approach.py?
yeah but I would rather wait for Amber to get back before messing with that too much. I want her to be able to recognize her code lol. I will make an issue for the refactoring

tomsilver:
@ronuchit let's merge this first and then I'll add GLIB?

ronuchit:
i really think we should do that refactoring now (or not merge this) -- it's not going to change her code much unless i'm missing something. if you look at it, you basically re-implemented `create_random_interaction_strategy` here. all we need to do is pull that function out into a common place where both can use it (utils? or some new Explorer class in base_approach?). and then GLIB comes for free as well cause you can just use `create_glib_interaction_strategy`

ronuchit:
do you really need an Explorer class? what's wrong with the way it's done in `interactive_learning_approach.py`?

PR #1039, opened 2022-06-30
Covers domain rendering bug 

wester-is-not-real:
Fixed bug in the covers domain where the target square would get highlighted when a block was grasped


tomsilver (approved):
great job finding and fixing this! let's let Rohan take a quick look too.

ronuchit:
yes thank you! it was clearly a bug before. nice catch!

PR #1042, opened 2022-06-30
implement explorers

### LINKED ISSUE #1038: refactor online approaches to use a common explorers class
interactive learning approach especially, and maybe also the RL approach.

tomsilver:
closes #1038


tomsilver:
@ronuchit any comments before I continue?

src/args.py
```diff
@@ -18,6 +18,7 @@ def create_arg_parser(env_required: bool = True,
     parser.add_argument("--included_options", default="", type=str)
     parser.add_argument("--seed", required=seed_required, type=int)
     parser.add_argument("--option_learner", type=str, default="no_learning")
+    parser.add_argument("--explorer", type=str, default="no_explore")
```
ronuchit, line 21: does this make `CFG.interactive_action_strategy` removable?

ronuchit (commented):
i think you need NSRTs in the explorer as well

src/args.py
tomsilver: yeah, that will be part of the interactive approach refactoring

tomsilver:
pausing, remaining TODOs:
* add tests for GLIB failure cases
* add tests for greedy lookahead
* refactor interactive approach

tests/interaction/test_glib_explorer.py
```diff
@@ -0,0 +1,75 @@
...
+    option_model = _OracleOptionModel(env)
+    train_tasks = env.get_train_tasks()
+    score_fn = lambda atoms: 0.0
+    task_idx = 0
+    task = env.get_test_tasks()[task_idx]
+    # Test case where there are no possible goals.
+    explorer = create_explorer("glib", set(), env.options, env.types,
+                               env.action_space, train_tasks, nsrts,
+                               option_model, score_fn)
+    assert "No possible goals, falling back to random." not in caplog.text
```
tomsilver, line 66: @ronuchit `caplog`, cool pytest trick

tomsilver: ...which works for me locally but appears to not work here...

src/interaction/bilevel_planning_explorer.py
```diff
@@ -0,0 +1,58 @@
...
+                 option_model: _OptionModelBase) -> None:
+
+        super().__init__(predicates, options, types, action_space, train_tasks)
+        self._nsrts = nsrts
+        self._option_model = option_model
+        self._num_calls = 0
+
+    def _solve(
+        self, task: Task, timeout: int
+    ) -> Tuple[Callable[[State], Action], Callable[[State], bool]]:
```
ronuchit, line 36: consider giving this data type a name in structs.py?

src/interaction/bilevel_planning_explorer.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # Ensure random over successive calls.
+        seed = self._seed + self._num_calls
+        try:
```
ronuchit, line 40: sort of weird to replicate all this here, why not just pass approach.solve into the init?

src/approaches/interactive_learning_approach.py
```diff
@@ -227,6 +227,9 @@ def _create_glib_interaction_strategy(
         self, train_task_idx: int
     ) -> Tuple[Callable[[State], Action], Callable[[State], bool]]:
         """Find the most interesting reachable ground goal and plan to it."""
+        # TODO: when refactoring, give any goals that have only already-known
```
ronuchit, line 230: wait why? the trajectory that gets you to that goal could still be very useful from a score function perspective

src/interaction/random_options_explorer.py
```diff
@@ -0,0 +1,36 @@
...
+
+class RandomOptionsExplorer(BaseExplorer):
+    """Samples random options."""
+
+    def __init__(self, predicates: Set[Predicate],
+                 options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task]) -> None:
+        super().__init__(predicates, options, types, action_space, train_tasks)
+        # Set up a random options approach.
+        self._random_options_approach = RandomOptionsApproach(
```
ronuchit, line 22: okay so this is a third way to do it... so we have:
1) pass the approach.solve function itself into the init
2) instantiate an approach in the explorer class (what you're doing here)
3) copy-paste the logic of approach.solve into the explorer class (what you're doing in the bilevel explorer)

we should pick one

src/approaches/interactive_learning_approach.py
tomsilver: that's true... anyway, that would be a functional change

src/interaction/random_options_explorer.py
tomsilver: let's stick with (2), otherwise there's no guarantee that the bilevel planning explorer would actually be doing bilevel planning

tomsilver: oh I remember why I didn't do that. the bilevel planning approach needs the NSRTs. we could add something that `BilevelPlanningApproach` that sets the NSRTs, but that seems bad... thoughts?

src/interaction/bilevel_planning_explorer.py
tomsilver: see comment below

src/approaches/interactive_learning_approach.py
tomsilver: okay i'll make this functional change and we'll see if there is a noticeable difference in the supercloud

tomsilver: actually nvm, this breaks some things, so I'm gonna try to stay true to the original implementation

src/interaction/random_options_explorer.py
ronuchit: but doesn't the bilevel planning explorer get the nsrts in its init?

tomsilver: yes but `BilevelPlanningApproach` doesn't

tomsilver:
something is screwed up and I can't yet figure it out. the interactive approach unit tests are way slower, and one reason appears to be that GLIB is timing out after every attempt. I don't know why this is happening now and not before. I'm actually very surprised that GLIB was previously able to find plans consistently after training the samplers for such a short amount of time.

@ronuchit i'm timing out, if you want to take this over be my guest, otherwise i'll continue tomorrow afternoon

tomsilver:
urgh there also appears to be a nondeterministic cyclic import issue 😿

src/interaction/random_options_explorer.py
tomsilver: note that I removed this (2) case here to resolve circular import issues

tomsilver:
@ronuchit alright this is theoretically complete, but please suggest and/or implement improvements

src/explorers/glib_explorer.py
```diff
@@ -0,0 +1,87 @@
...
+from predicators.src.option_model import _OptionModelBase
+from predicators.src.settings import CFG
+from predicators.src.structs import NSRT, ExplorationStrategy, GroundAtom, \
+    ParameterizedOption, Predicate, Task, Type
+
+
+class GLIBExplorer(BilevelPlanningExplorer):
+    """GLIBExplorer implementation.
+
+    Sample a certain number of goals, plan to each of them, and execute
```
ronuchit, line 22: this is not what the code is doing. what the code is doing is sample goals, score each of them, and then try planning starting from the highest-scoring goal, terminating at the first goal for which we find a plan.

the current comment makes it sound like we're planning to every goal

ronuchit (approved):
very nice and useful refactoring! happy to see this common stuff getting factored out

src/approaches/interactive_learning_approach.py
```diff
@@ -172,16 +184,22 @@ def _score_atom_set(self, atom_set: Set[GroundAtom],
 
         Higher scores are better.
         """
+        # Filter out any atoms that are not being learned.
```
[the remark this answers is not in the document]
tomsilver: no, it should just be a refactor. before, these atoms were implicitly being filtered out before they were passed into the function.

scripts/supercloud/run_action_policy_experiments.sh
```diff
@@ -14,8 +14,8 @@ COMMON_ARGS="--env cover --approach interactive_learning --excluded_predicates C
     --min_data_for_nsrt $MIN_DATA --sampler_disable_classifier True --mlp_classifier_balance_data False"
 
 # glib
-python $FILE $COMMON_ARGS --experiment_id glib --interactive_action_strategy glib
```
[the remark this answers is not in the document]
tomsilver: I did

src/approaches/interactive_learning_approach.py
```diff
@@ -152,15 +150,29 @@ def _relearn_predicates_and_nsrts(
...
+            CFG.explorer,
+            self._get_current_predicates(),
+            self._initial_options,
+            self._types,
+            self._action_space,
+            self._train_tasks,
+            self._get_current_nsrts(),
+            self._option_model,
+            babble_predicates=self._predicates_to_learn,
+            atom_score_fn=self._score_atom_set_frequency,
```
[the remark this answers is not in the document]
tomsilver: it's more because we only have one such atom score function implemented right now. note the difference between glib and greedy lookahead score functions: glib can only score atoms (because it babbles an atomic goal, which doesn't have a full state associated with it) whereas the score function used by greedy lookahead can depend on the low-level state (e.g., entropy stuff).

tomsilver:
Here are supercloud results from before and after:
[explorer_ab_test.pdf](https://github.com/Learning-and-Intelligent-Systems/predicators/files/9034651/explorer_ab_test.pdf)

The trends are the same. There is a slight dip in evaluation performance of greedy lookahead -- I will look into it now, but it may just be a result of changes in arbitrary tiebreaking or numerical variation in the NNs, etc.

PR #1043, opened 2022-07-01
online nsrt learning approach 

tomsilver:


src/approaches/online_nsrt_learning_approach.py
```diff
@@ -0,0 +1,88 @@
+"""An NSRT learning approach that collects and learns from data online.
+
+Example command:
+    python src/main.py --approach online_nsrt_learning --seed 0 \
+        --env cover \
+        --explorer random_options \
+        --max_initial_demos 1 \
+        --num_train_tasks 1000 \
```
ronuchit, line 8: is this just a dummy large value, since the number of interaction cycles governs the speed of the pipeline?

src/approaches/online_nsrt_learning_approach.py
```diff
@@ -0,0 +1,88 @@
...
+    def learn_from_offline_dataset(self, dataset: Dataset) -> None:
+        # Update the dataset with the offline data.
+        self._dataset = Dataset(dataset.trajectories)
+        super().learn_from_offline_dataset(dataset)
+
+    def learn_from_interaction_results(
+            self, results: Sequence[InteractionResult]) -> None:
+        # Add the new data to the cumulative dataset.
+        for result in results:
+            traj = LowLevelTrajectory(result.states, result.actions)
```
ronuchit, line 82: as before, remember that not having `is_demo = True` here might screw things up

ronuchit (approved):
lgtm. i'd say merge explorers PR first before this one, so that they appear separately in master's commit history

src/approaches/online_nsrt_learning_approach.py
tomsilver: yeah

PR #1044, opened 2022-07-02
added weird test case

wmcclinton:
This is the case we talked about yesterday which generates two MoveTo operators instead of just the one we want.


tomsilver:
@wmcclinton should we close this for now?

NishanthJKumar:
It's still relevant in that it'll break our algorithm in general, so we should make an issue or something to tackle it later

tomsilver (commented):
I'm happy to merge this -- @ronuchit do you want to take a look?

wmcclinton:
> yea good to merge. do we understand why this is happening?

Yah because our algorithm goes from general to specific, so if you see a specific version of an operator (i.e. move from one object to two objects first [whom both are relevant]) it will learn that operator even though later it will see and induce the operator we actually want (i.e. move from one object to only one object).

PR #1048, opened 2022-07-05
raise error instead of falling back to random actions when...

tomsilver:
random option policy fails to sample.

justification: it's bad design to have actions returned by this policy sometimes have associated options, and sometimes not (and this was leading to a crash when exploring in pddl envs)

### DIFF
--- a/src/explorers/random_options_explorer.py
+++ b/src/explorers/random_options_explorer.py
@@ -17,6 +17,9 @@ class RandomOptionsExplorer(BaseExplorer):
         # Take random options, and raise an exception if no applicable option
         # can be found.
 
+        # Note that this fallback policy is different from the one in
+        # RandomOptionsApproach because explorers should raise
+        # RequestActPolicyFailure instead of ApproachFailure.
         def fallback_policy(state: State) -> Action:
             del state  # unused
             raise utils.RequestActPolicyFailure(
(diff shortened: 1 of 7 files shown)


src/utils.py
```diff
@@ -1082,8 +1082,8 @@ def _policy(state: State) -> Action:
                 if opt.initiable(state):
                     cur_option = opt
                     break
-            else:  # fall back to a random action
-                return Action(action_space.sample())
+            else:
+                raise OptionExecutionFailure("Random option sampling failed!")
```
ronuchit, line 1086: this isn't really the right exception to raise right? do we just not care?

ronuchit (approved):

src/utils.py
ronuchit: better would be to add a new exception for this

tomsilver:
how's this? @ronuchit

ronuchit (approved):
looks good now thanks

src/explorers/random_options_explorer.py
```diff
@@ -14,10 +14,16 @@ def get_name(cls) -> str:
...
-        # Take random options.
-        policy = utils.create_random_option_policy(self._options,
-                                                   self._action_space,
-                                                   self._rng)
+        # Take random options, and raise an exception if no applicable option
+        # can be found.
+
+        def fallback_policy(state: State) -> Action:
+            del state  # unused
+            raise utils.RequestActPolicyFailure(
```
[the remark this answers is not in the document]
ronuchit: random_options_approach.py*

PR #1049, opened 2022-07-05
pg3: handle case where plan comparison fails to find plans

tomsilver:


src/approaches/pg3_approach.py
```diff
@@ -328,7 +329,10 @@ def __init__(
 
     def _get_score_for_task(self, ldl: LiftedDecisionList,
                             task_idx: int) -> float:
-        atom_plan = self._get_atom_plan_for_task(ldl, task_idx)
+        try:
+            atom_plan = self._get_atom_plan_for_task(ldl, task_idx)
```
ronuchit, line 333: add a note to the docstring of `_get_atom_plan_for_task` explaining that it's allowed to raise a PlanningFailure

ronuchit (approved):

PR #1052, opened 2022-07-06
move class variables that use CFG into init

tomsilver:
I think we realized at some point that CFG should not be used in class variables, but there were still two cases where we had them (I checked all the environments).

the reason is that command line overrides will not work for those CFG settings, because by the time the overrides are processed, the class variables have already been set.

for example, `--blocks_num_blocks_test "[2]"` did not work as anticipated before this change.

### DIFF
--- a/src/envs/blocks.py
+++ b/src/envs/blocks.py
@@ -46,8 +46,6 @@ class BlocksEnv(BaseEnv):
     on_tol: ClassVar[float] = 0.01
     collision_padding: ClassVar[float] = 2.0
     assert pick_tol < block_size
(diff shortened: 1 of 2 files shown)


ronuchit (approved):
let's just make it private while we're at it? (prepend underscore)

tests/test_predicate_search_score_functions.py
```diff
@@ -316,8 +316,6 @@ def test_exact_energy_score_function():
         "env": "blocks",
         "offline_data_method": "demo+replay",
         "num_train_tasks": 2,
-        "blocks_num_blocks_train": [3],
```
tomsilver, line 319: lol these weren't getting used before, and in fact, using them broke the test

PR #1053, opened 2022-07-06
Added backtracking for low-level option search

wester-is-not-real:


src/approaches/pg3_approach.py
```diff
@@ -46,14 +47,39 @@ def __init__(self, initial_predicates: Set[Predicate],
     def get_name(cls) -> str:
         return "pg3"
 
-    def _predict(self, state: State, atoms: Set[GroundAtom],
-                 goal: Set[GroundAtom]) -> _GroundNSRT:
-        del state  # unused
+    def predict(self, atoms: Set[GroundAtom],
```
tomsilver, line 50: this can be a private method (add underscore). we can also make it more specific --  maybe `_predict_ground_nsrt`?

src/approaches/pg3_approach.py
```diff
@@ -46,14 +47,39 @@ def __init__(self, initial_predicates: Set[Predicate],
...
+                goal: Set[GroundAtom]) -> _GroundNSRT:
+        """Predicts next GroundNSRT to be deployed based on the PG3 generated
+        policy."""
         ground_nsrt = utils.query_ldl(self._current_ldl, atoms, goal)
         if ground_nsrt is None:
             raise ApproachFailure("PG3 policy was not applicable!")
         return ground_nsrt
 
+    def solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+        _S: TypeAlias = FrozenSet[GroundAtom]
```
tomsilver, line 60: it looks like `_S` and `_A` are no longer needed

src/approaches/pg3_approach.py
```diff
@@ -46,14 +47,39 @@ def __init__(self, initial_predicates: Set[Predicate],
...
+        atoms = utils.abstract(task.init, self._initial_predicates)
+        atoms_sequence.append(atoms)
+        start_time = time.time()
+
+        while not task.goal.issubset(atoms):
+            ground_nsrt = self.predict(atoms, task.goal)
+            atoms = utils.apply_operator(ground_nsrt, atoms)
+            skeleton.append(ground_nsrt)
+            atoms_sequence.append(atoms)
+            if (time.time() - start_time) >= timeout:
```
tomsilver, line 73: it doesn't matter much, but let's move this check up to the top of the loop, so that we terminate ASAP after time runs out

src/approaches/pg3_approach.py
```diff
@@ -46,14 +47,39 @@ def __init__(self, initial_predicates: Set[Predicate],
...
+    def predict(self, atoms: Set[GroundAtom],
+                goal: Set[GroundAtom]) -> _GroundNSRT:
+        """Predicts next GroundNSRT to be deployed based on the PG3 generated
+        policy."""
         ground_nsrt = utils.query_ldl(self._current_ldl, atoms, goal)
         if ground_nsrt is None:
             raise ApproachFailure("PG3 policy was not applicable!")
         return ground_nsrt
 
+    def solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
```
tomsilver, line 59: I think we actually want to be overriding `_solve`, not `solve` -- check the parent classes

src/approaches/pg3_approach.py
tomsilver, line 60: add a docstring explaining what's going on in this method at a high level

tomsilver (commented):
@wester-is-not-real great work! left some comments. please also update the PR description (at the top of this page) with an example command that shows the benefit of this PR (and the result).

src/approaches/pg3_approach.py
```diff
@@ -46,14 +47,39 @@ def __init__(self, initial_predicates: Set[Predicate],
...
+                raise ApproachFailure("Timeout exceeded")
+            ground_nsrt = self._predict_ground_nsrt(atoms, task.goal)
+            atoms = utils.apply_operator(ground_nsrt, atoms)
+            skeleton.append(ground_nsrt)
+            atoms_sequence.append(atoms)
+        option_list, succeeded = run_low_level_search(
+            task, self._option_model, skeleton, atoms_sequence, self._seed,
+            timeout - (time.time() - start_time), CFG.horizon)
+        if not succeeded:
+            raise ApproachFailure("Low Level Search Failed")
```
tomsilver, line 79: minor: change capitalization to be consistent with elsewhere -- so, "Low-level search failed"

tomsilver (approved):
one more tiny thing that I missed the first time, otherwise lgtm!

PR #1062, opened 2022-07-07
Objects are now obtained from low level state when querying a lifted decision list representing a goal-conditioned policy

wester-is-not-real:
Modified query_ldl so it takes as arguments the objects in the current low-level state.


src/approaches/pg3_approach.py
```diff
@@ -61,14 +62,18 @@ def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
         policy."""
         skeleton = []
         atoms_sequence = []
+        current_objects = set()
         atoms = utils.abstract(task.init, self._initial_predicates)
         atoms_sequence.append(atoms)
+        for key in task.init.data:
```
tomsilver, line 68: you can save a few lines and do:
`current_objects = set(task.init)`
to understand why this works, check out the `__iter__()` method of `State` in structs.py

src/approaches/pg3_approach.py
```diff
@@ -475,7 +480,8 @@ def get_next_state(atoms: _S, ground_nsrt: _A) -> _S:
         )
 
         def policy(atoms: _S) -> Optional[_A]:
-            return utils.query_ldl(ldl, set(atoms), goal)
+            objects = {o for a in atoms | goal for o in a.objects}
```
tomsilver, line 483: this looks like the same dangerous pattern that we want to fix right? can we use the already-defined `objects` variable instead? (so just remove this line). hopefully this doesn't diminish the results you were seeing in painting with box-only, let's check after we merge this

src/utils.py
```diff
@@ -1126,6 +1126,9 @@ def _get_entity_combinations(
         for ent in sorted_entities:
             if ent.is_instance(vt):
                 this_choices.append(ent)
+        # if len(this_choices) == 0:
```
tomsilver, line 1129: remove the commented out code

src/utils.py
```diff
@@ -1134,7 +1137,8 @@ def _get_entity_combinations(
 def get_object_combinations(objects: Collection[Object],
                             types: Sequence[Type]) -> Iterator[List[Object]]:
     """Get all combinations of objects satisfying the given types sequence."""
-    return _get_entity_combinations(objects, types)
+    penis = _get_entity_combinations(objects, types)
```
tomsilver, line 1140: 😛

src/utils.py
```diff
@@ -1326,9 +1330,8 @@ def _run_heuristic_search(
         lazy_expansion: bool = False) -> Tuple[List[_S], List[_A]]:
     """A generic heuristic search implementation.
 
-    Depending on get_priority, can implement A*, GBFS, or UCS.
-
-    If no goal is found, returns the state with the best priority.
+    Depending on get_priority, can implement A*, GBFS, or UCS. If no
```
tomsilver, line 1333: revert this change (just because we don't need it)

src/utils.py
```diff
@@ -1347,6 +1350,8 @@ def _run_heuristic_search(
 
     while len(queue) > 0 and time.time() - start_time < timeout and \
           num_expansions < max_expansions and num_evals < max_evals:
+        # if (method == "gbfs" and best_node_priority==36.0):
```
tomsilver, line 1353: remove

tests/test_structs.py
```diff
@@ -1041,17 +1041,20 @@ def test_lifted_decision_lists():
 
     atoms = {on_table([cup1]), hand_empty([robot])}
     goal = {on([cup1, plate1])}
+    objects = {o for a in atoms | goal for o in a.objects}
```
tomsilver, line 1044: so as to not suggest a bad pattern, let's just explicitly create the object set: `objects = {cup1, plate1, ...}`

tomsilver (commented):
excellent job finding and fixing this issue! please update the title and description of this PR too, so other readers have context

src/approaches/pg3_approach.py
wester-is-not-real: Oh thats pretty cool. Thanks!

tests/test_structs.py
wester-is-not-real: Good idea!

src/approaches/pg3_approach.py
```diff
@@ -61,14 +62,17 @@ def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
         policy."""
         skeleton = []
         atoms_sequence = []
+        current_objects = set()
```
tomsilver, line 65: you can remove this now

tomsilver (approved):
one small thing, otherwise LGTM!

PR #1063, opened 2022-07-07
forest domain from PG3 paper

tomsilver:
coauthored by @Abraham-Mitchell and based on original code by @ryangpeixu

### DIFF
--- a/src/envs/pddl_procedural_generation.py
+++ b/src/envs/pddl_procedural_generation.py
@@ -269,6 +269,10 @@ def _random_grid_walk(
         visited: Set[Tuple[int, int]], grid_height: int, grid_width: int,
         previous_coords: Optional[Tuple[int, int]],
         rng: np.random.Generator) -> Optional[List[Tuple[int, int]]]:
+    """Generates a random path through a grid.
+
+    For aesthetic reasons, the grid is not allowed to self-intersect.
+    """
     if curr_coords == goal_coords:
         return [curr_coords]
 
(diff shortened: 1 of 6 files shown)


src/envs/pddl_procedural_generation.py
```diff
@@ -110,6 +112,9 @@ def _blocks_piles_to_strs(
     return all_strs
 
 
+################################## Delivery ###################################
+
+
```
NishanthJKumar, line 117: nit: seems like other similar comments (Blocks and Forest) leave one newline after the comment, whereas this leaves 2. Can we standardize the number of newlines before and after this header comment?

src/envs/pddl_procedural_generation.py
```diff
@@ -186,3 +191,262 @@ def _generate_delivery_problem(num_locs: int, num_want_locs: int,
...
+        if grid_walk_from_child is not None:
+            return [curr_coords] + grid_walk_from_child
+
+    return None
+
+
+def _random_walk_reachable(curr_coords: Tuple[int,
+                                              int], goal_coords: Tuple[int,
+                                                                       int],
+                           prev_visited: Set[Tuple[int, int]],
```
NishanthJKumar, line 319: this function seems rather similar to the above and I'm struggling to understand how it's being used. Maybe add a comment here or above to explain what's happening just so people editing in the future can grok this easily?

NishanthJKumar (approved):
LGTM after addressing fairly minor comments. Great work!

PR #1066, opened 2022-07-09
add optional memory dict to nsrt metacontroller

tomsilver:
this will be used by a forthcoming LLM approach

### DIFF
--- a/src/approaches/gnn_approach.py
+++ b/src/approaches/gnn_approach.py
@@ -225,7 +225,8 @@ class GNNApproach(BaseApproach, Generic[_Output]):
         self._load_output_specific_fields_from_save_info(info)
 
     def _predict(self, state: State, atoms: Set[GroundAtom],
-                 goal: Set[GroundAtom]) -> _Output:
+                 goal: Set[GroundAtom], memory: Dict) -> _Output:
+        del memory  # unused
         # Get output graph.
         in_graph, object_to_node = self._graphify_single_input(
             state, atoms, goal)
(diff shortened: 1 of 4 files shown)


src/approaches/gnn_approach.py
```diff
@@ -225,7 +225,8 @@ def load(self, online_learning_cycle: Optional[int]) -> None:
         self._load_output_specific_fields_from_save_info(info)
 
     def _predict(self, state: State, atoms: Set[GroundAtom],
-                 goal: Set[GroundAtom]) -> _Output:
+                 goal: Set[GroundAtom], memory: Dict) -> _Output:
+        del memory  # unused
```
NishanthJKumar, line 229: is the idea that this will actually get used in the future?

NishanthJKumar (approved):
LGTM - just a minor question!

src/approaches/gnn_approach.py
tomsilver: not by the GNN, but the memory will be used by the open loop LLM approach (to store the plan)

ronuchit: this was confusing to me at first. but i see, you need to update this because gnn metacontroller approach will inherit it and also subclasses the nsrt metacontroller approach. that sucks

tomsilver: yeah this may not have been the best design after all. one day maybe we will refactor to not use the metacontroller at all. but for now this works

Issue #973, opened 2022-06-17
add pandas stubs for type checking


tomsilver:
https://github.com/pandas-dev/pandas-stubs

and fix a few type errors that come up afterwards in `scripts/`.

tomsilver:
reopening because `openai` uses `pandas-stubs`

PR #1069, opened 2022-07-09
exclude scripts from type checking

tomsilver:
this is a temporary measure until we have time to do the better thing...

- the scripts subdirectory uses `pandas` in a number of places
- the type annotations are not entirely correct, and it's nontrivial to fix them
- we've been getting away with this because the pandas library is mostly untyped
- but there's another library called `pandas-stubs` that adds type annotations for pandas
- the `openai` package has `pandas-stubs` as a dependency
- so in the short term, to add the `openai` package, we'll just stop type checking the entire scripts directory
- in the longer term, we need to close this issue https://github.com/Learning-and-Intelligent-Systems/predicators/issues/973

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,7 +2,7 @@
 strict_equality = True
 disallow_untyped_calls = True
 warn_unreachable = True
-exclude = (src/envs/assets|third_party|src/scripts)
+exclude = (src/envs/assets|third_party|scripts)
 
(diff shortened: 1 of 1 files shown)


NishanthJKumar (approved):
Damn - this is quite the annoying issue. Thanks for getting to it so quickly!

PR #1070, opened 2022-07-09
refactor core harmlessness check for later reuse

tomsilver:
specifically, we want to reuse the main logic in the open-loop LLM approach to turn the option plan output by the LLM into ground NSRT

@NishanthJKumar please check carefully that I didn't change the functionality


NishanthJKumar (approved):
LGTM! Thanks for doing this, this is much cleaner than what I did in #1015 ✨

PR #1073, opened 2022-07-10
Saving+Loading GroundAtomTrajectories

NishanthJKumar:
This PR allows us to save and load GroundAtomTrajectories with the `--load_atoms` flag. This functionality is extremely useful for BEHAVIOR, where running all the classifiers is slow even if we load trajectory data.

### DIFF
--- a/src/approaches/nsrt_learning_approach.py
+++ b/src/approaches/nsrt_learning_approach.py
@@ -5,6 +5,7 @@ or options.
 """
 
 import logging
+import os
 import time
 from typing import Dict, List, Optional, Set
 
@@ -19,7 +20,7 @@ from predicators.src.nsrt_learning.nsrt_learning_main import \
 from predicators.src.planning import task_plan, task_plan_grounding
 from predicators.src.settings import CFG
 from predicators.src.structs import NSRT, Dataset, LowLevelTrajectory, \
-    ParameterizedOption, Predicate, Segment, Task, Type
+    ParameterizedOption, Predicate, Segment, Task, Type, GroundAtom
 
 
 class NSRTLearningApproach(BilevelPlanningApproach):
@@ -53,7 +54,7 @@ class NSRTLearningApproach(BilevelPlanningApproach):
 
     def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
                      online_learning_cycle: Optional[int]) -> None:
-        dataset_fname = utils.create_dataset_filename_str(
+        dataset_fname, _ = utils.create_dataset_filename_str(
             saving_ground_atoms=True,
             online_learning_cycle=online_learning_cycle)
         # If CFG.load_atoms is set, then try to load the GroundAtomTrajectory
@@ -76,7 +77,7 @@ class NSRTLearningApproach(BilevelPlanningApproach):
         else:
             # Apply predicates to data, producing a dataset of abstract states.
             ground_atom_dataset = utils.create_ground_atom_dataset(
-                trajectories, predicates)
+                trajectories, self._get_current_predicates())
             # Save ground atoms dataset to file.
             if CFG.env == "behavior":  # pragma: no cover
                 # In the case of behavior, we cannot directly pickle the ground
--- a/src/approaches/nsrt_learning_approach.py
+++ b/src/approaches/nsrt_learning_approach.py
@@ -56,7 +56,7 @@ class NSRTLearningApproach(BilevelPlanningApproach):
                      online_learning_cycle: Optional[int]) -> None:
         dataset_fname, _ = utils.create_dataset_filename_str(
             saving_ground_atoms=True,
-            online_learning_cycle=str(online_learning_cycle))
+            online_learning_cycle=online_learning_cycle)
         # If CFG.load_atoms is set, then try to create a GroundAtomTrajectory
         # by loading sets of GroundAtoms directly from a saved file.
         if CFG.load_atoms:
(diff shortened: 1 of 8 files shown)


src/utils.py
```diff
@@ -1860,11 +1861,62 @@ def sample_subsets(universe: Sequence[_T], num_samples: int, min_set_size: int,
...
+            ground_atom_dataset.append((traj, atoms))
+        # Save ground atoms dataset to file.
+        ground_atom_dataset_to_pkl = []
+        for _, traj in enumerate(ground_atom_dataset):
+            trajectory = []
+            for i, ground_atom_seq in enumerate(traj[1]):
+                trajectory.append({
+                    GroundAtom(
+                        Predicate(atom.predicate.name, atom.predicate.types,
+                                  lambda s, o: None), atom.entities)
```
tomsilver, line 1914: major: you're stripping out the classifiers, presumably for behavior related reasons, but you're not putting them back in during loading. that seems risky to me -- can you fix that, and add a test to make sure that the loaded atoms have the right classifiers attached to them?

minor: add a comment explaining why we need to do this stripping

minor: I'm surprised mypy let's you get away with `None` here, shouldn't it be a bool?

tomsilver (changes requested):
looks like a step in a good direction, but some things that I think are important to change

src/utils.py
```diff
@@ -1860,11 +1861,62 @@ def sample_subsets(universe: Sequence[_T], num_samples: int, min_set_size: int,
 def create_ground_atom_dataset(
         trajectories: Sequence[LowLevelTrajectory],
         predicates: Set[Predicate]) -> List[GroundAtomTrajectory]:
-    """Apply all predicates to all trajectories in the dataset."""
-    ground_atom_dataset = []
-    for traj in trajectories:
-        atoms = [abstract(s, predicates) for s in traj.states]
-        ground_atom_dataset.append((traj, atoms))
+    """Generate or load a GroundAtomTrajectory by applying all predicates to
```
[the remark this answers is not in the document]
NishanthJKumar: this is a really good point, agreed! Any thoughts on which function we should add this to (I can't seem to find a good place for it)? My current inclination is that the right place is in `nsrt_learning_main.py`: instead of always creating a new dataset on line 41, we can instead load depending on the flag (here, we can be assured that the trajectories and predicates correspond to the offline dataset, I think)?

src/utils.py
NishanthJKumar: yup - i had to fix the `bool` thing. It didn't strike me that this stripping was always being done and that's potentially dangerous - will definitely make sure it's (1) only for BEHAVIOR and (2) will test classifiers after loading are correct!

src/utils.py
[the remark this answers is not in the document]
tomsilver: that would be a little better, but still has some of the same risks -- for example, if we run online learning, then `nsrt_learning_main.py` would get used multiple times, with different trajectories each time.

another idea (which is gonna look not pretty, but maybe is the safest bet): add an optional kwarg to `learn_nsrts_from_data` called `ground_atom_dataset`, which if provided, is used instead of running STEP 1. then do the saving and loading inside of `NSRTLearningApproach._learn_nsrts`, where importantly, we can use the `online_learning_cycle` in the cache file name.

src/approaches/nsrt_learning_approach.py
```diff
@@ -53,12 +54,90 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
+                assert len(trajectories) == len(ground_atom_dataset_atoms)
+                logging.info("\n\nLOADED GROUND ATOM DATASET")
+
+                if CFG.env == "behavior":  # pragma: no cover
+                    pred_name_to_pred = {
+                        pred.name: pred
+                        for pred in self._get_current_predicates()
+                    }
+                    new_ground_atom_dataset_atoms = []
+                    # Since we save ground atoms for behavior with dummy
```
ronuchit, line 78: if you're loading atoms from disk, why do you actually care if the classifiers are correct? where would they be used?

src/approaches/nsrt_learning_approach.py
```diff
@@ -53,12 +54,90 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
+            # Save ground atoms dataset to file. Note that a
+            # GroundAtomTrajectory contains a normal LowLevelTrajectory and a
+            # list of sets of GroundAtoms, so we only save the list of
+            # GroundAtoms (the LowLevelTrajectories are saved separately).
+            ground_atom_dataset_to_pkl = []
+            for gt_traj in ground_atom_dataset:
+                trajectory = []
+                for ground_atom_set in gt_traj[1]:
+                    if CFG.env == "behavior":  # pragma: no cover
+                        # In the case of behavior, we cannot directly pickle
```
ronuchit, line 116: if we're going to go this far, we might as well just always strip away the classifiers when saving atoms to disk? that seems reasonable to me given that in general the classifiers could be complicated non-pickleable things (and in fact they are, for behavior). but Tom might disagree on this (his preference is higher prio)

ronuchit (commented):
just some late night thoughts

src/approaches/nsrt_learning_approach.py
NishanthJKumar: So we don't actually care for the purposes of NSRT learning - I just wanted to set them back to be correct so that if in the future we use these atoms for something we don't get completely bamboozled by why they're horribly wrong

tomsilver: yeah, +1 to what Nishanth said

src/utils.py
```diff
@@ -1857,6 +1857,40 @@ def sample_subsets(universe: Sequence[_T], num_samples: int, min_set_size: int,
...
+
+    Returns a tuple of strings where the first element is the dataset
+    filename itself and the second is a template string used to generate
+    it. If saving_ground_atoms is True, then we will name the file with
+    a "_ground_atoms" suffix.
+    """
+    # Setup the dataset filename for saving/loading GroundAtoms.
+    regex = r"(\d+)"
+    suffix_str = ""
+    if online_learning_cycle is not None:
```
tomsilver, line 1874: the convention we've used thus far is to include online_learning_cycle regardless (so None appears in the str sometimes). this is helpful when parsing filenames, so we can assume the number of "__" separators is always the same.

tests/approaches/test_nsrt_learning_approach.py
```diff
@@ -154,6 +159,56 @@ def test_nsrt_learning_approach():
                            strips_learner=strips_learner)
 
 
+def test_saving_and_loading_atoms():
```
tomsilver, line 162: how long does this test take?

tomsilver (approved):
nice work!

tests/approaches/test_nsrt_learning_approach.py
NishanthJKumar: 1.19s. Interestingly, it's not even the most expensive test in this file.

PR #1074, opened 2022-07-11
BEHAVIOR Evaluation Pipeline

wmcclinton:
Here is the completed pipeline for BEHAVIOR which: (1) generates tasks with different scenes and seeds randomly, (2) learns NSRTs on those trajectories, and (3) evaluates those NSRTs on BEHAVIORs option models. Additionally, in this PR we added a flag that disables RRT when using our option models which speeds up execution time. Note, evaluating BEHAVIOR on option models only works with BilevelPlanningApproaches due to its current implementation. We will fix this by creating a `run_policy_with_option_model()` function.


src/envs/behavior.py
```diff
@@ -68,7 +68,14 @@ def __init__(self) -> None:
         # a map between task nums and the snapshot id for saving/loading
         # purposes
         self.task_num_to_igibson_seed: Dict[int, int] = {}
-
+        # Everytime we load a new scene in behavior we also need to set
+        # the vaild options again, because there might be new type combos
```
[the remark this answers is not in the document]
tomsilver: please apply the nit period comment throughout the PR too

src/envs/behavior.py
```diff
@@ -68,7 +68,14 @@ def __init__(self) -> None:
...
+        self.set_options()
+
+    def set_options(self) -> None:
+        """Sets the underlying options for this particular task using the
+        current type list."""
+        self._type_name_to_type = {}
```
tomsilver, line 78: this line seems unrelated to setting the options. if the idea is to clear this dict every time the scene is reset, let's add this line into a reset method?

src/settings.py
```diff
@@ -111,6 +111,8 @@ class GlobalSettings:
     behavior_task_name = "re-shelving_library_books"
     behavior_scene_name = "Pomaria_1_int"
     behavior_randomize_init_state = True
+    behavior_option_model_eval = True
+    behavior_option_model_RRT = False
```
tomsilver, line 115: lowercase rrt

tests/approaches/test_oracle_approach.py
```diff
@@ -170,6 +170,16 @@
         "pddl_easy_delivery_procedural_test_max_extra_newspapers": 1,
     },
 ]
+EXTRA_ARGS_ORACLE_APPROACH["behavior_tasks"] = [
```
tomsilver, line 173: i'm confused -- isn't behavior excluded from tests because we didn't install it on the github servers? is this change doing anything?

src/main.py
```diff
@@ -281,14 +284,31 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
-                env,
-                "test",
-                test_task_idx,
-                task.goal_holds,
-                max_num_steps=CFG.horizon,
-                monitor=monitor)
+            if CFG.env == "behavior" and \
+                CFG.behavior_option_model_eval:  # pragma: no cover
+                # To evaluate Behavior on our option model, we are going
+                # to run our approached plan on our option model.
```
tomsilver, line 290: approached plan?

src/main.py
```diff
@@ -281,14 +284,31 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+                # Note that if appraoch is not a BilevelPlanningApproach
```
tomsilver, line 291: appraoch -> approach

src/main.py
```diff
@@ -281,14 +284,31 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+                # We cannot use this method to evaluate and would need to
```
tomsilver, line 292: lowercase We

src/main.py
```diff
@@ -281,14 +284,31 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
-                policy,
-                env,
-                "test",
-                test_task_idx,
-                task.goal_holds,
-                max_num_steps=CFG.horizon,
-                monitor=monitor)
+            if CFG.env == "behavior" and \
+                CFG.behavior_option_model_eval:  # pragma: no cover
+                # To evaluate Behavior on our option model, we are going
```
tomsilver, line 289: be consistent about capitalization of behavior / Behavior / BEHAVIOR

src/main.py
```diff
@@ -281,14 +284,31 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
         try:
-            traj, execution_metrics = utils.run_policy(
-                policy,
-                env,
-                "test",
-                test_task_idx,
-                task.goal_holds,
-                max_num_steps=CFG.horizon,
-                monitor=monitor)
+            if CFG.env == "behavior" and \
```
tomsilver, line 287: this is okay for now given the impending deadline, but the accumulating behavior special-casing in the code base is making me sad face. after the deadline, let's talk about making a persistent branch or fork of predicators and moving all of the behavior stuff into it

src/envs/behavior.py
```diff
@@ -359,8 +370,10 @@ def set_igibson_behavior_env(self, task_instance_id: int,
                                "environment that meets bddl initial "
                                "conditions!")
         self.igibson_behavior_env.robots[0].initial_z_offset = 0.7
+        self.igibson_behavior_env.use_RRT = CFG.behavior_option_model_RRT
 
-    @functools.lru_cache(maxsize=None)
+    # Do not add @functools.lru_cache(maxsize=None) here this will
+    # lead to wrong mappings when we load a different scene
```
tomsilver, line 376: good catch!

src/approaches/bilevel_planning_approach.py
```diff
@@ -105,3 +105,19 @@ def _get_current_predicates(self) -> Set[Predicate]:
         Defaults to initial predicates.
         """
         return self._initial_predicates
+
+    def get_option_model(self) -> _OptionModelBase:
+        """For ONLY an oracle approach or BEHAVIOR, we allow the user to get
```
tomsilver, line 110: can you add assertions that we're in one of these two cases, both here and in the method below?

tomsilver (commented):
bunch of small things, mostly looks good

src/approaches/bilevel_planning_approach.py
```diff
@@ -105,3 +105,19 @@ def _get_current_predicates(self) -> Set[Predicate]:
         Defaults to initial predicates.
         """
         return self._initial_predicates
+
```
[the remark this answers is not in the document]
ronuchit: oh i see you're moving that API here well ok then +1 to tom's suggestion of adding an assertion but still :(

tests/approaches/test_oracle_approach.py
NishanthJKumar: "It fixes so coverage complaints" -Willie

tomsilver: Can you be specific?

src/main.py
NishanthJKumar: "We will do" -Willie

src/envs/behavior.py
```diff
@@ -174,22 +183,28 @@ def _get_tasks(self,
...
+            # BEHAVIOR uses np.random everywhere. This is a somewhat
             # hacky workaround for that.
             curr_env_seed = rng.integers(0, (2**32) - 1)
+            # ID used to generate scene in BEHAVIOR default scene is 0
+            self.task_instance_id = 0
             if CFG.behavior_randomize_init_state:
-                self.set_igibson_behavior_env(task_instance_id=self.task_num,
-                                              seed=curr_env_seed)
+                # Get random scene for BEHAVIOR between O-9 and 10-20
+                # if train or test, respectively
```
tomsilver, line 194: nit: period

src/envs/behavior.py
```diff
@@ -554,26 +572,33 @@ def _create_type_combo_name(original_name: str,
 def load_checkpoint_state(s: State, env: BehaviorEnv) -> None:
     """Sets the underlying iGibson environment to a particular saved state."""
     assert s.simulator_state is not None
-    # Get the task_num associated with this state from s.simulator_state
-    new_task_num = int(s.simulator_state.split("-")[0])
-    # If the task_num is new, then we need to load a new iGibson behavior
-    # env with our random seed saved in env.task_num_to_igibson_seed. Otherwise
+    # Get the new_task_num_task_instance_id associated with this state
+    # from s.simulator_state
```
tomsilver, line 576: nit: period

tomsilver (approved):
thanks! btw, you may already have tricks for doing this, but here's a bash function that I have in my bash profile for doing things like Behavior -> BEHAVIOR:
```
git-replace(){
    git grep -l "$1" | xargs sed -i '' -e "s/$1/$2/g"
}
```
so then I could do, for example, `git-replace Behavior BEHAVIOR` (with caution)

PR #1077, opened 2022-07-13
handle cases where oracle learner sees no data for nsrts

tomsilver:
before this PR, this command would crash:
```
python src/main.py --approach nsrt_learning --seed 0 --env painting --strips_learner oracle --sampler_learner oracle --num_train_tasks 10 --painting_lid_open_prob 1.0 --painting_goal_receptacles shelf
```
and in general, if we are doing oracle strips and sampler learning, and we add settings that lead to some NSRTs having no data, we will now raise a warning instead of crashing.

this is meant to be a more general solution to the problem that I papered over in PRs #1057 and #1065 , so those changes are reverted.

### DIFF
--- a/src/nsrt_learning/nsrt_learning_main.py
+++ b/src/nsrt_learning/nsrt_learning_main.py
@@ -106,7 +106,7 @@ def _learn_pnad_options(pnads: List[PartialNSRTAndDatastore],
     # same known parameterized option, or all actions should have no option.
     known_option_pnads, unknown_option_pnads = [], []
     for pnad in pnads:
-        if not pnad.datastore:
+        if not pnad.datastore:  # pragma: no cover
             raise Exception("No data found for learning an option.")
         example_segment, _ = pnad.datastore[0]
         example_action = example_segment.actions[0]
--- a/src/nsrt_learning/strips_learning/oracle_learner.py
+++ b/src/nsrt_learning/strips_learning/oracle_learner.py
@@ -1,5 +1,6 @@
 """Oracle for STRIPS learning."""
 
+import logging
 from typing import List
 
 from predicators.src.envs import get_or_create_env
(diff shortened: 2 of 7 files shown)


wester-is-not-real (approved):
Looks good!

src/nsrt_learning/strips_learning/oracle_learner.py
```diff
@@ -27,7 +28,18 @@ def _learn(self) -> List[PartialNSRTAndDatastore]:
             pnads.append(
                 PartialNSRTAndDatastore(nsrt.op, datastore, option_spec))
         self._recompute_datastores_from_segments(pnads)
-        return pnads
+        # Filter out any pnad that has an empty datastore. This can occur when
+        # using non-standard settings with environments that cause certain
+        # operators to be unnecessary. For example, in painting, when using
+        # --painting_goal_receptacles box, the operator for picking from
```
ronuchit, line 34: better to enclose in backquotes for readability: `--painting_goal_receptacles box`

src/nsrt_learning/nsrt_learning_main.py
```diff
@@ -106,7 +106,7 @@ def _learn_pnad_options(pnads: List[PartialNSRTAndDatastore],
     # same known parameterized option, or all actions should have no option.
     known_option_pnads, unknown_option_pnads = [], []
     for pnad in pnads:
-        if not pnad.datastore:
+        if not pnad.datastore:  # pragma: no cover
```
ronuchit, line 109: i'm confused by this... how can this arise in light of the change you made to `nsrt_learning_main.py`? can we change this to an assert instead?

ronuchit (commented):
looks like a reasonable thing to do! just left some minor comments

PR #1080, opened 2022-07-15
Instead of evaluating one completion, this approach evaluates multipl…

Alchemy5:
…e approaches in an effort to increase the accuracy of the overall prediction


src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +68,29 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+                                                       temperature=0.5,
+                                                       seed=CFG.seed,
+                                                       num_completions=10)
+        for llm_prediction in llm_predictions:
+            objects = set(state)
+            option_plan = self._llm_prediction_to_option_plan(
+                llm_prediction, objects)
+            if len(option_plan) == 0:
+                continue
+            nsrts = self._get_current_nsrts()
```
tomsilver, line 81: for these lines that don't depend on `llm_prediction`, let's move them above the for loop, so they only need to be called once

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +68,29 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+            option_specs = [(n.option, list(n.option_vars)) for n in nsrts]
+            ground_nsrt_plan = task_plan_with_option_plan_constraint(
+                objects, predicates, strips_ops, option_specs, atoms, goal,
+                option_plan)
+            if not ground_nsrt_plan:
+                continue
+            memory["abstract_plan"] = ground_nsrt_plan
+            return memory["abstract_plan"].pop(0)
+
+        raise ApproachFailure("LLM predicted plan does not achieve goal.")
```
ronuchit, line 93: it looks like a lot of the comments are getting dropped from the old code, but they were helpful. can we add those back in?

relatedly, is the idea of this PR basically that we are indenting the old code into a for loop? if so, a preferable pattern would be to move the old code into a new function (at the same level of indentation). suppose that new function is called `_process _single_completion` (just an example), then your loop would just be:
```
for llm_prediction in llm_predictions:
    self._process_single_prediction(llm_prediction)
```
this would make it easier to read this PR (https://github.com/Learning-and-Intelligent-Systems/predicators/pull/1080/files) and understand what is actually changing

src/approaches/open_loop_llm_approach.py
```diff
@@ -1,14 +1,13 @@
-"""Open-loop large language model (LLM) meta-controller approach.
+"""Open-loop large language model (LLM) meta-controller approach. Example
+command line: export OPENAI_API_KEY=<your API key> python src/main.py.
```
NishanthJKumar, line 2: can we move the `python src/main.py` to a newline so that it's obvious that the `export` command needs to be run separately?

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,50 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+        llm_predictions = self._llm.sample_completions(
+            prompt=prompt,
             temperature=CFG.open_loop_llm_temperature,
             seed=CFG.seed,
-            num_completions=1)[0]
+            num_completions=CFG.open_loop_llm_num_completions)
         # Try to convert the output into an abstract plan.
+
+        for llm_prediction in llm_predictions:
+
```
NishanthJKumar, line 80: nit: delete this blank line

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,50 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
-        # If we failed to find a nontrivial plan, give up.
-        if len(option_plan) == 0:
-            raise ApproachFailure("LLM did not predict an abstract plan.")
-        # Otherwise, we succeeded, so attempt to turn the plan into a
-        # sequence of ground NSRTs.
         nsrts = self._get_current_nsrts()
         predicates = self._initial_predicates
         strips_ops = [n.op for n in nsrts]
         option_specs = [(n.option, list(n.option_vars)) for n in nsrts]
+
```
NishanthJKumar, line 100: nit: delete unnecessary blank line

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,50 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+        option_plan = self._llm_prediction_to_option_plan(
+            llm_prediction, objects)
+        # If we failed to find a nontrivial plan with prediction,
+        # continue on to next prediction
+        if len(option_plan) == 0:
+            return None
+        #Attempt to turn plan into sequence of ground NSRTs
```
NishanthJKumar, line 107: nit: add space after `#`

NishanthJKumar (commented):
great work! Just a few mostly-minor things

src/approaches/open_loop_llm_approach.py
```diff
@@ -1,14 +1,13 @@
-"""Open-loop large language model (LLM) meta-controller approach.
+"""Open-loop large language model (LLM) meta-controller approach. Example
+command line: export OPENAI_API_KEY=<your API key> \
+python src/main.py.
```
tomsilver, line 3: this period here looks incorrect -- can we actually just revert all of the changes are the top of the file here?

tomsilver (approved):
nice! one more minor thing

src/approaches/open_loop_llm_approach.py
ronuchit: +1, we don't need any changes to the docstring of this file

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,46 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+            num_completions=CFG.open_loop_llm_num_completions)
         # Try to convert the output into an abstract plan.
+        for llm_prediction in llm_predictions:
+            ground_nsrt_plan = self._process_single_prediction(
+                llm_prediction, state, atoms, goal)
+            if ground_nsrt_plan is not None:
+                # If valid plan, add plan to memory so it can be refined!
+                memory["abstract_plan"] = ground_nsrt_plan
+                return memory["abstract_plan"].pop(0)
+        # Give up if none of the predictions work out
```
ronuchit, line 85: nit: we usually try to use punctuation for comments in our codebase, would be good to add a period at the end here

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,46 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
-        # Otherwise, we succeeded, so attempt to turn the plan into a
-        # sequence of ground NSRTs.
         nsrts = self._get_current_nsrts()
         predicates = self._initial_predicates
         strips_ops = [n.op for n in nsrts]
         option_specs = [(n.option, list(n.option_vars)) for n in nsrts]
+        option_plan = self._llm_prediction_to_option_plan(
+            llm_prediction, objects)
+        # If we failed to find a nontrivial plan with prediction,
+        # continue on to next prediction
```
ronuchit, line 100: same: period at end

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,46 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+        if len(option_plan) == 0:
+            return None
+        # Attempt to turn plan into sequence of ground NSRTs
```
ronuchit, line 103: same: period at end

src/approaches/open_loop_llm_approach.py
```diff
@@ -70,33 +69,46 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
         ground_nsrt_plan = task_plan_with_option_plan_constraint(
             objects, predicates, strips_ops, option_specs, atoms, goal,
             option_plan)
-        # If we can't find an NSRT plan that achieves the goal, give up.
+        # If we can't find an NSRT plan that achieves the goal,
+        # continue on to next prediction
```
ronuchit, line 108: same: period at end

src/settings.py
```diff
@@ -261,8 +261,9 @@ class GlobalSettings:
     llm_openai_max_response_tokens = 250
 
     # parameters for open loop LLM approach
-    open_loop_llm_model_name = "text-curie-001"
-    open_loop_llm_temperature = 0.0
+    open_loop_llm_model_name = "text-davinci-002"
```
ronuchit, line 264: is this change something you want to commit to master? @tomsilver

tomsilver: i'll revert it to be safe

src/settings.py
```diff
@@ -261,8 +261,9 @@ class GlobalSettings:
     llm_openai_max_response_tokens = 250
 
     # parameters for open loop LLM approach
-    open_loop_llm_model_name = "text-curie-001"
-    open_loop_llm_temperature = 0.0
+    open_loop_llm_model_name = "text-curie-001"  # "text-davinci-002"
+    open_loop_llm_temperature = 0.5
+    open_loop_llm_num_completions = 1
```
tomsilver, line 266: @Alchemy5 take note that I'm using small defaults here. we will override them in your script with the command line for the experiments

PR #1082, opened 2022-07-15
BEHAVIOR pipeline deterministic

wmcclinton:
This PR fixes the BEHAVIOR pipeline by reseting the BEHAVIOR environment in `load_checkpoint_state` to clear any behavior state information left over from previous action executions that might of happen during backtracking.


src/envs/behavior.py
```diff
@@ -583,22 +583,35 @@ def load_checkpoint_state(s: State, env: BehaviorEnv) -> None:
...
+        frame_count = env.igibson_behavior_env.simulator.frame_count
         env.set_igibson_behavior_env(
             task_instance_id=new_task_num_task_instance_id[1],
             seed=env.task_num_task_instance_id_to_igibson_seed[
                 new_task_num_task_instance_id])
+        env.igibson_behavior_env.simulator.frame_count = frame_count
         env.set_options()
         env.current_ig_state_to_state(
         )  # overwrite the old task_init checkpoint file!
+    env.igibson_behavior_env.reset()
```
tomsilver, line 607: this is called even if `reset` is False?

src/envs/behavior_options.py
```diff
@@ -644,7 +644,7 @@ def create_grasp_option_model(
 
     # NOTE: -26 because there are 25 timesteps that we move along the vector
```
tomsilver, line 645: fix this comment

src/main.py
```diff
@@ -284,34 +284,48 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
-                last_plan = approach.get_last_plan()
-                option_model_start_time = time.time()
-                traj, solved = _run_plan_with_option_model(
-                    task, test_task_idx, approach.get_option_model(),
-                    last_plan)
-                execution_metrics = {
-                    "policy_call_time": option_model_start_time - time.time()
-                }
+            # Will evaluate all approaches on our simulator with the options
+            # to (1) only evalutate based on ability to find a plan for
```
tomsilver, line 288: typo: evalutate

src/main.py
```diff
@@ -284,34 +284,48 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+            # BilevelPlanningApproaches and (2) for BEHAVIOR to evaluate on
+            # option models instead of the low-level simulator (with the
+            # plan_only_eval and behavior_option_model_eval flags, respectively)
+            if CFG.plan_only_eval and isinstance(
+                    approach, BilevelPlanningApproach):  # pragma: no cover
+                if approach.get_last_plan() != []:
```
tomsilver, line 294: what if the goal is satisfied in the initial state?

src/main.py
```diff
@@ -284,34 +284,48 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
-                execution_metrics = {
-                    "policy_call_time": option_model_start_time - time.time()
-                }
+            # Will evaluate all approaches on our simulator with the options
+            # to (1) only evalutate based on ability to find a plan for
+            # BilevelPlanningApproaches and (2) for BEHAVIOR to evaluate on
+            # option models instead of the low-level simulator (with the
+            # plan_only_eval and behavior_option_model_eval flags, respectively)
+            if CFG.plan_only_eval and isinstance(
+                    approach, BilevelPlanningApproach):  # pragma: no cover
```
tomsilver, line 293: if this is not behavior-specific, we should add a test for it. I'm not sure if we will want to keep it after we factor out all of the behavior code into a new branch, though. maybe @ronuchit has thoughts

src/settings.py
```diff
@@ -44,6 +44,11 @@ class GlobalSettings:
     # in unit tests, make sure to pass in a value for `render_state_dpi` into
     # your call to utils.reset_config().
     render_state_dpi = 150
+    # If this is True, then we will not execute our final plan in simulation
+    # and we will say a task is solved successfully if we were able to find
+    # a plan even if the plan would not necessarily work in simulation (this
+    # is especially true of option models that don't model the simulator well)
```
tomsilver, line 50: nit: period

tomsilver (commented):
please remove the `prof` file!

src/main.py
ronuchit: i guess i'll look after the other comments are addressed? but yeah just at a glance of these lines, they're hacky and we should find a better way to do it

src/main.py
```diff
@@ -284,34 +284,49 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+                if approach.get_last_plan() != [] or task.goal_holds(
+                        task.init):
+                    solved = True
+                    exec_time = 0.0
+                    metrics[
+                        f"PER_TASK_task{test_task_idx}_exec_time"] = exec_time
             else:
```
ronuchit, line 300: instead of:
```
if blah:
  blah
else:
  if blah:
    blah
  else:
    blah
```
change to:
```
if blah:
  blah
elif blah:
  blah
else:
  blah
   ```

File: src/main.py [hunk 8]
```diff
@@ -284,34 +284,49 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
-                    last_plan)
-                execution_metrics = {
-                    "policy_call_time": option_model_start_time - time.time()
-                }
+            # Will evaluate all approaches on our simulator with the options
+            # to (1) only evaluate based on ability to find a plan for
+            # BilevelPlanningApproaches and (2) for BEHAVIOR to evaluate on
+            # option models instead of the low-level simulator (with the
+            # plan_only_eval and behavior_option_model_eval flags, respectively)
+            if CFG.plan_only_eval and isinstance(
```
ronuchit, line 292: let's change this to:
```
if CFG.plan_only_eval:
  assert isinstance(approach, BilevelPlanningApproach)
  blah blah...
```

src/main.py
```diff
@@ -284,34 +284,49 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+                    approach, BilevelPlanningApproach):  # pragma: no cover
```
ronuchit, line 293: let's add unit tests to cover this

src/main.py
```diff
@@ -284,34 +284,49 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
-                    env,
-                    "test",
-                    test_task_idx,
-                    task.goal_holds,
-                    max_num_steps=CFG.horizon,
-                    monitor=monitor)
-            solved = task.goal_holds(traj.states[-1])
-            exec_time = execution_metrics["policy_call_time"]
-            metrics[f"PER_TASK_task{test_task_idx}_exec_time"] = exec_time
+                if CFG.env == "behavior" and \
```
ronuchit, line 301: similar pattern here:
```
if CFG.behavior_option_model_eval:
  assert CFG.env == "behavior"
  blah
```

src/settings.py
```diff
@@ -44,6 +44,11 @@ class GlobalSettings:
     # in unit tests, make sure to pass in a value for `render_state_dpi` into
     # your call to utils.reset_config().
     render_state_dpi = 150
+    # If this is True, then we will not execute our final plan in simulation
```
ronuchit, line 47: run-on sentence! please make this a bit more clear (grammatically) :)

src/main.py
```diff
@@ -284,34 +284,49 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
         else:
             monitor = None
         try:
```
ronuchit, line 286: let's separate these changes into a new PR

ronuchit:
hmm.. doesn't `plan_only_eval` make it so we don't need `behavior_option_model_eval` anymore?

src/main.py
wmcclinton: Moved to different PR

ronuchit (approved):
lgtm, hopefully the change of `behavior_option_model_eval` from True to False isn't an issue

PR #1085, opened 2022-07-15
Script for experimenting with PG3 on TAMP problems

wester-is-not-real:
Script for running experiments with PG3 on TAMP problems


scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Note: this script is too large to be run all at once. Comment things out
+# and run in multiple passes. For example, start with just 1000 train tasks.
+# If that looks good, launch the environments in separate runs.
+
+FILE="scripts/supercloud/submit_supercloud_job.py"
+# Note: this script is meant to be run first, to completion, with
```
tomsilver, line 8: you can remove this comment since it doesn't apply anymore

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,39 @@
...
+# RUN_LOAD_EXPERIMENTS=false, then rerun with RUN_LOAD_EXPERIMENTS=true.
+RUN_LOAD_EXPERIMENTS=false
```
tomsilver, line 10: and remove this variable

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,39 @@
...
+for ENV in ${ENVS[@]}; do
+    for NUM_TRAIN_TASKS in ${ALL_NUM_TRAIN_TASKS[@]}; do
+
+        COMMON_ARGS="--env $ENV  --sampler_learner oracle \
+                 --num_train_tasks $NUM_TRAIN_TASKS \"
+
+        # nsrt learning (oracle operators and options)
+        # note: $INCLUDED_OPTIONS excluded because all options are
+        # included for this oracle approach.
+        python $FILE $COMMON_ARGS --experiment_id ${ENV}_oracle_options_${NUM_TRAIN_TASKS} --approach pg3 --strips_learner oracle
```
tomsilver, line 36: let's move `--strips_learner oracle` up into the `COMMON_ARGS`

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,39 @@
...
+    "coffee"
+)
+
+for ENV in ${ENVS[@]}; do
+    for NUM_TRAIN_TASKS in ${ALL_NUM_TRAIN_TASKS[@]}; do
+
+        COMMON_ARGS="--env $ENV  --sampler_learner oracle \
+                 --num_train_tasks $NUM_TRAIN_TASKS \"
+
+        # nsrt learning (oracle operators and options)
```
tomsilver, line 33: this comment isn't applicable -- can you replace it with a different comment?

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Note: this script is too large to be run all at once. Comment things out
```
tomsilver, line 3: this comment can probably be removed (this file is currently small)

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,39 @@
...
+# and run in multiple passes. For example, start with just 1000 train tasks.
+# If that looks good, launch the environments in separate runs.
+
+FILE="scripts/supercloud/submit_supercloud_job.py"
+# Note: this script is meant to be run first, to completion, with
+# RUN_LOAD_EXPERIMENTS=false, then rerun with RUN_LOAD_EXPERIMENTS=true.
+RUN_LOAD_EXPERIMENTS=false
+ALL_NUM_TRAIN_TASKS=(
```
tomsilver, line 11: let's actually make the default here a few different values. maybe 10, 50, and 100? and leave them all uncommented.

scripts/supercloud/run_pg3_tamp_experiments.sh
tomsilver, line 36: change the name of the experiment_id to something more specific to what we're doing here

tomsilver (commented):
looking good! a number of small things. i'm also thinking about if and how we should run the variations of cover and painting that you've been running. do you have any thoughts about that? @wester-is-not-real

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,65 @@
...
+)
+
+for ENV in ${ENVS[@]}; do
+  for NUM_TRAIN_TASKS in ${ALL_NUM_TRAIN_TASKS[@]}; do
+
+      if [ $ENV = "painting" ] || [ $ENV = "cluttered_table" ] || \
+          [ $ENV = "coffee" ] then
+        COMMON_ARGS="--env $ENV  --strips_learner oracle \
+              --sampler_learner oracle --num_train_tasks $NUM_TRAIN_TASKS \
+              --pg3_hc_enforced_depth 1"
```
tomsilver, line 24: I'm thinking we should just use depth 1 for everything? It's generally better if we can use one set of hyperparameters for all environments and approaches. We would just expect it to take longer to finish, right? And if we want to later, we could even optimize this by terminating as soon as we find a policy with heuristic 0.

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,65 @@
...
+      else
+        COMMON_ARGS="--env $ENV  --strips_learner oracle \
+              --sampler_learner oracle --num_train_tasks $NUM_TRAIN_TASKS \"
+      fi
+
+      # If in painting enviroment override common arguments and included options
+      if [ $ENV = "painting" ]; then
+        INCLUDED_OPTIONS="--painting_lid_open_prob 1.0"
```
tomsilver, line 32: we can actually just add this flag and all of the env-specific ones to COMMON_ARGS. if we use the flag `--painting_lid_open_prob` while using the cover env or whatever it will just have no effect, which is fine. this should simplify this bash file a bit

scripts/supercloud/run_pg3_tamp_experiments.sh
```diff
@@ -0,0 +1,65 @@
...
+        python $FILE $COMMON_ARGS \
+              --experiment_id ${ENV}_pg3_policy_${NUM_TRAIN_TASKS} \
+              --approach pg3
+      else
+        # PG3 policy generation on TAMP problems using default options
+        # and included options
+        python $FILE $COMMON_ARGS \
+              --experiment_id ${ENV}_pg3_policy_${NUM_TRAIN_TASKS} \
+              --approach pg3
+        python $FILE $COMMON_ARGS $INCLUDED_OPTIONS \
```
tomsilver, line 61: after the change suggested above, you can remove INCLUDED_OPTIONS altogether and remove it from the comment above too

tomsilver (commented):
good stuff! a few more suggestions

scripts/supercloud/run_pg3_tamp_experiments.sh
wester-is-not-real: Oh wait you're right because for the easier environments they won't even need to look at that depth. I was just avoiding complexity when evaluating the easier environments, but yeah it won't matter.

scripts/supercloud/run_pg3_tamp_experiments.sh
wester-is-not-real: Oh, good to know! Thanks for the tip

tomsilver:
replaced by #1099

PR #1091, opened 2022-07-18
heuristic stop. tom

Abraham-Mitchell:
Added heuristic termination threshold, early_termination_heuristic_thresh


tomsilver (commented):
thank you! see comments below. also please update the pull request description so anyone else browsing has context :)

src/utils.py
```diff
@@ -1440,6 +1440,7 @@ def run_hill_climbing(
         check_goal: Callable[[_S], bool],
         get_successors: Callable[[_S], Iterator[Tuple[_A, _S, float]]],
         heuristic: Callable[[_S], float],
+        early_termination_heuristic_thresh: Optional[float] = 0,
```
tomsilver, line 1443: let's default this to None, rather than 0, but then change the PG3 approach to use 0 instead

src/utils.py
```diff
@@ -1461,6 +1462,11 @@ def run_hill_climbing(
     logging.info(f"\n\nStarting hill climbing at state {cur_node.state} "
                  f"with heuristic {last_heuristic}")
     while True:
+        
+        #Stops when heuristic reaches specified value.
```
tomsilver, line 1466: please add a space in between # and Stops

src/utils.py
```diff
@@ -1440,6 +1440,7 @@ def run_hill_climbing(
         check_goal: Callable[[_S], bool],
         get_successors: Callable[[_S], Iterator[Tuple[_A, _S, float]]],
         heuristic: Callable[[_S], float],
+        early_termination_heuristic_thresh: Optional[float] = None,
```
tomsilver, line 1443: please add a description of `early_termination_heuristic_thresh` in the docstring (Lines 1446-1454)

tomsilver (commented):
thanks for the quick turnaround! a few more things below, and then don't forget to update the pg3 approach.

finally, let's add a unit test. take a look at `test_run_hill_climbing` in `tests/test_utils.py`. at the end of that function, let's add an assertion that checks that if we use a very high value for `early_termination_heuristic_thresh`, then hill climbing terminates immediately with an empty plan. to test your test as you're developing it, you can do `pytest -s tests/test_utils.py -k test_run_hill_climbing`.

don't forget to run all the other checks too when you're done :) thanks again!

tests/test_utils.py
```diff
@@ -2718,8 +2718,17 @@ def _local_minimum_grid_heuristic_fn(state: S) -> float:
         assert heuristics == [
             8.0, float("inf"), 6.0, 5.0, 4.0, 3.0, 2.0, 1.0, 0.0
         ]
-
-
+        
+    #Test early_termination_heuristic_thresh with very high value
```
tomsilver, line 2722: make sure your comment format matches the standard!

tomsilver (approved):
great work!!

PR #1094, opened 2022-07-19
Implementation for policy-guided planning

wester-is-not-real:
When sesame generates a skeleton, it will first generate policy-based skeletons first.  If the policy-based skeletons fail, A* will just chose a different skeleton with both policy-based and primitive successors nodes. Therefore, PG4 uses a general policy to guide the agent, but will do bilevel planning where the general policy fails.  

To test this, run the following: 

 python3 src/main.py --approach pg4 --seed 0 --env cover --strips_learner oracle --sampler_learner oracle --num_train_tasks 50

In the cover domain, it is possible for the agent to start holding a block. Sometimes that block would not have any destination but a second block would. So, PG3 would place the block it was holding on the second block's target. Then when it would pick up the second block to place it on its target, it would fail because the first block was already occupying that target. Now, PG4 succeeds because it will do bilevel planning whenever it gets stuck.


src/approaches/pg3_approach.py
```diff
@@ -164,6 +165,55 @@ def _create_heuristic(self) -> _PG3Heuristic:
         return cls(preds, nsrts, self._train_tasks)
 
 
+class PG4Approach(PG3Approach):
+    """Policy-guided planning for generalized policy generation (PG4)."""
```
tomsilver, line 169: that's only 3 PG! 😛 add another "for planning guidance" or something

src/approaches/pg3_approach.py
```diff
@@ -164,6 +165,55 @@ def _create_heuristic(self) -> _PG3Heuristic:
...
+
+    def __init__(self, initial_predicates: Set[Predicate],
```
tomsilver, line 171: this looks identical to the PG3 init method, so we can remove it

src/approaches/pg3_approach.py
```diff
@@ -164,6 +165,55 @@ def _create_heuristic(self) -> _PG3Heuristic:
...
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self._current_ldl = LiftedDecisionList([])
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "pg4"
+
+    def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+        """Searches for a low level policy that satisfies PG3's abstract
```
tomsilver, line 183: update this docstring

src/planning.py
```diff
@@ -51,6 +51,7 @@ def sesame_plan(
     max_horizon: int,
     check_dr_reachable: bool = True,
     allow_noops: bool = False,
+    abstract_ldl: LiftedDecisionList = None,
```
tomsilver, line 54: let's change this back to an abstract policy, since there could be other kinds of policies that we would use here instead. inside of the PG4 approach, you can write something like:
```
abstract_policy = lambda a,o,g: utils.query_ldl(self._current_ldl, a, o, g)
```
and then inside of sesame, you can just use `abstract_policy(atoms, objects, goal)`

src/planning.py
```diff
@@ -217,9 +220,15 @@ def task_plan(
...
-    metrics: Metrics, max_skeletons_optimized: int
+    task: Task,
+    ground_nsrts: List[_GroundNSRT],
+    init_atoms: Set[GroundAtom],
+    heuristic: _TaskPlanningHeuristic,
+    seed: int,
+    timeout: float,
+    metrics: Metrics,
+    max_skeletons_optimized: int,
+    abstract_ldl: LiftedDecisionList = None
```
tomsilver, line 231: let's remove ` = None` here, since it will always be passed down from `sesame_plan`. also, let's change the type to `Optional[LiftedDecisionList]`, since it can be either an LDL or None. (I'm surprised mypy didn't complain about that)

src/planning.py
```diff
@@ -231,6 +240,8 @@ def _skeleton_generator(
     """
 
     start_time = time.time()
+    num_rollout_steps = CFG.pg3_max_policy_guided_rollout
```
tomsilver, line 243: this is now a general function that is not necessarily going to use pg3. so let's add another hyperparameter in settings called something like `sesame_max_policy_guided_rollout` and use that instead. we can keep the value at 50.

src/planning.py
```diff
@@ -240,6 +251,8 @@ def _skeleton_generator(
     rng_prio = np.random.default_rng(seed)
     hq.heappush(queue,
                 (heuristic(root_node.atoms), rng_prio.uniform(), root_node))
+    visited_skeletons: Set[Tuple[_GroundNSRT, ...]] = set(tuple())  #initialize
```
tomsilver, line 254: in general, please make a space between the hashtag and the first word in your comments

tomsilver: also, since this is split over two lines, let's just make it a comment above. so:
```
# Initialize with empty skeleton for root.
visited_skeletons: Set[Tuple[_GroundNSRT, ...]] = set(tuple())
```

tomsilver: also, can you add a comment explaining why we need to keep track of visited skeletons now? (this will also be a good opportunity to check your understanding!)

src/planning.py
```diff
@@ -258,11 +271,51 @@ def _skeleton_generator(
         else:
             # Generate successors.
             metrics["num_nodes_expanded"] += 1
+            #If an abstract ldl policy is provided, generate policy-based
```
tomsilver, line 274: as above, add a space at the start of the comments and a period at the end of the comments

tomsilver (commented):
excellent work! I can help with the test coverage once the minor comments are addressed. also, please update the description in the pull request to reflect the change in the method name to pg4. (thanks for a very thorough description!)

src/planning.py
```diff
@@ -240,6 +254,10 @@ def _skeleton_generator(
     rng_prio = np.random.default_rng(seed)
     hq.heappush(queue,
                 (heuristic(root_node.atoms), rng_prio.uniform(), root_node))
+    # Initialize with empty skeleton for root
```
tomsilver, line 257: sorry to be annoying about this, but please make the format of all your comments consistent :) This one should have a period at the end, and there's one below missing a space after the hashtag

src/planning.py
```diff
@@ -240,6 +254,10 @@ def _skeleton_generator(
...
+    # We want to keep track of the visited skeletons that way we avoid
```
tomsilver, line 258: minor: We want to keep track of the visited skeletons *so that we* avoid

tomsilver (approved):
looks really great! a few tiny complaints

src/planning.py
```diff
@@ -258,11 +276,51 @@ def _skeleton_generator(
         else:
             # Generate successors.
             metrics["num_nodes_expanded"] += 1
+            # If an abstract ldl policy is provided, generate policy-based
```
ronuchit, line 279: is this a repeat of the `run_policy_guided_astar` stuff in utils.py? is that just bc this entire `skeleton_generator` is really a repeat of `run_astar`?

src/approaches/pg3_approach.py
```diff
@@ -23,10 +23,11 @@
 from typing_extensions import TypeAlias
```
ronuchit, line 23: is the high-level idea of this PR that we want to use the LDL policies learned by PG3 to guide sesame rather than what we had before, which was to use it to guide `run_astar`?

ronuchit: so what would happen if we were to just use regular PG3 on an environment that requires bilevel planning? does that even make any sense at all?

src/planning.py
wester-is-not-real: Double yes! If we did this using run_astar it would break other stuff, so we just updated skeleton_generator instead.

src/planning.py
```diff
@@ -258,11 +276,51 @@ def _skeleton_generator(
...
+                    current_node = child_node
+                    if time.time() - start_time >= timeout:
+                        break
+            # Generate primitive successors
             for nsrt in utils.get_applicable_operators(ground_nsrts,
                                                        node.atoms):
                 child_atoms = utils.apply_operator(nsrt, set(node.atoms))
+                child_skeleton = node.skeleton + [nsrt]
+                child_skeleton_tup = tuple(child_skeleton)
+                if child_skeleton_tup in visited_skeletons:
```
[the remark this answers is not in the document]
wester-is-not-real: This is just for when you are generating policy-based skeletons. A* won't generate duplicate skeletons when doing bilevel planning but it is possible for the policy to generate skeletons identical to the ones generated in bilevel planning. So, those lines of code necessary only if we're using a policy. Otherwise, they're redundant. I can just put them inside an if, that was a really good catch. Thank you!

src/approaches/pg3_approach.py
```diff
@@ -165,6 +166,56 @@ def _create_heuristic(self) -> _PG3Heuristic:
         return cls(preds, nsrts, self._train_tasks)
 
 
+class PG4Approach(PG3Approach):
```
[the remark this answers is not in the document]
wester-is-not-real: Very very true

src/approaches/pg3_approach.py
wester-is-not-real: Exactly, we use the ldl policies to guide sesame. Initially, PG3 couldn't be used in TAMP problems, so we added low-level option search to it. That worked well for simple TAMP problems, but it would get stuck in environments like cover. To overcome this, we now use the ldl policy pg3 generates to guide bilevel planning.

PR #1097, opened 2022-07-20
LLM bilevel planning approach

Alchemy5:
An approach that uses the LLM predictions to initialize the queue in A*. This is a simple form of using the output of the LLM to guide planning.

should be merged after #1111


src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
+"""Open-loop large language model (LLM) meta-controller approach.
```
tomsilver, line 1: please update this comment

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+
+Example command line:
+    export OPENAI_API_KEY=<your API key>
+    python src/main.py --approach probe --seed 0 \
```
tomsilver, line 5: don't forget to update this after changing the approach name

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+                         action_space, train_tasks)
+        # Set up the LLM.
+        self._llm = OpenAILLM(CFG.open_loop_llm_model_name)
+        # Set after learning.
+        self._prompt_prefix = ""
+        self.last_char = ")"
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "probe"
```
tomsilver, line 67: let's make this a little more specific. how about: `llm_probe`

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+    ParameterizedOption, Predicate, State, Task, Type, _GroundNSRT, _Option
+from predicators.src.utils import apply_operator, \
+    create_task_planning_heuristic, get_applicable_operators, \
+    run_policy_guided_astar
+
+
+class Probe(OpenLoopLLMApproach):
+    """Probe definition."""
+
+    def __init__(self, initial_predicates: Set[Predicate],
```
tomsilver, line 54: this should be identical to the `__init__` method from the parent class, so we can remove it

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+from predicators.src.planning import task_plan_with_option_plan_constraint
+from predicators.src.settings import CFG
+from predicators.src.structs import Box, Dataset, GroundAtom, Object, \
+    ParameterizedOption, Predicate, State, Task, Type, _GroundNSRT, _Option
+from predicators.src.utils import apply_operator, \
+    create_task_planning_heuristic, get_applicable_operators, \
+    run_policy_guided_astar
+
+
+class Probe(OpenLoopLLMApproach):
```
tomsilver, line 51: let's also make this class name more specific, e.g., `LLMProbeApproach`

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+
+from predicators.src import utils
+from predicators.src.approaches.open_loop_llm_approach import \
+    OpenLoopLLMApproach
+from predicators.src.llm_interface import OpenAILLM
+from predicators.src.planning import task_plan_with_option_plan_constraint
+from predicators.src.settings import CFG
+from predicators.src.structs import Box, Dataset, GroundAtom, Object, \
+    ParameterizedOption, Predicate, State, Task, Type, _GroundNSRT, _Option
+from predicators.src.utils import apply_operator, \
```
tomsilver, line 46: for consistency with elsewhere, let's not import functions from utils, but rather just do `utils.apply_operator`, etc., in the code below

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+            num_rollout_steps=CFG.pg3_max_policy_guided_rollout,
+            rollout_step_cost=0)
+
+        if action_seq is not None:
+            # If valid plan, add plan to memory so it can be refined!
+            memory["abstract_plan"] = action_seq
+            return memory["abstract_plan"].pop(0)
+        raise "Approach does not work"
+
+    def _generate_states(self, atoms: Set[GroundAtom], goal: Set[GroundAtom],
```
tomsilver, line 156: let's name this method something more specific

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+                         state: State) -> Tuple[List, List]:
```
tomsilver, line 157: make the output type annotations more specific (lists of what?)

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+            prompt=prompt,
+            temperature=CFG.open_loop_llm_temperature,
+            seed=CFG.seed,
+            num_completions=5)
+
+        for llm_prediction in llm_predictions:
+            temp_states = []
+            temp_actions = []
+            temp_states.append(initial_atoms)
+            trimmed_prediction = ""
```
tomsilver, line 175: can you explain what's going on with this trimming? if we keep it, I'm thinking it probably belongs in `_llm_prediction_to_option_plan()`, but I want to understand what it's doing first

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+
+        if action_seq is not None:
+            # If valid plan, add plan to memory so it can be refined!
+            memory["abstract_plan"] = action_seq
+            return memory["abstract_plan"].pop(0)
+        raise "Approach does not work"
+
+    def _generate_states(self, atoms: Set[GroundAtom], goal: Set[GroundAtom],
+                         state: State) -> Tuple[List, List]:
+        states = []
```
tomsilver, line 158: I think the code above will become a little simpler if you collect a list of trajectories instead of two separate lists of states and actions. so:
```
trajectories = []
...
trajectories.append((temp_states, temp_actions))
```

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+        new_prompt = self._create_prompt(atoms, goal, [])
+        prompt = self._prompt_prefix + new_prompt
+        llm_predictions = self._llm.sample_completions(
+            prompt=prompt,
+            temperature=CFG.open_loop_llm_temperature,
+            seed=CFG.seed,
+            num_completions=5)
+
+        for llm_prediction in llm_predictions:
+            temp_states = []
```
tomsilver, line 172: after making the change I suggested above, we can change this variable name to just `states`, and `actions` below

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+            temp_actions = []
+            temp_states.append(initial_atoms)
```
tomsilver, line 174: move these lines down to right before you're about to add to the lists

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+            else:
+                trimmed_prediction = llm_prediction
+
+            objects = set(state)
+            option_plan = self._llm_prediction_to_option_plan(
+                trimmed_prediction, objects)
+            if len(option_plan) == 0:
+                continue
+            nsrts = self._get_current_nsrts()
+            ground_nsrt_plan = []
```
tomsilver, line 188: can you make this a separate helper method and add some comments explaining the assumption that operators and NSRTs are one and the same? (and that this would break for any non-PDDLEnv)

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+            for option in option_plan:
```
tomsilver, line 189: to improve clarity, write: `for (option, objects) in option_plan`:

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+                    if option[0].name == n.option.name:
+                        ground_nsrt_plan.append(n.ground(option[1]))
+                        break
+
+            for nsrt in ground_nsrt_plan:
+                atoms = apply_operator(nsrt, atoms)
+                temp_actions.append(nsrt)
+                temp_states.append(atoms)
+            states.append(temp_states)
+            actions.append(temp_actions)
```
tomsilver, line 200: this is where the new helper method should end (so it should return the trajectory)

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+
+        initial_atoms = atoms
+
+        new_prompt = self._create_prompt(atoms, goal, [])
+        prompt = self._prompt_prefix + new_prompt
+        llm_predictions = self._llm.sample_completions(
+            prompt=prompt,
+            temperature=CFG.open_loop_llm_temperature,
+            seed=CFG.seed,
+            num_completions=5)
```
tomsilver, line 169: change this to the CFG setting

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+        self._prompt_prefix = ""
+        self.last_char = ")"
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "probe"
+
+    def _predict(self, state: State, atoms: Set[GroundAtom],
+                 goal: Set[GroundAtom], memory: Dict) -> _GroundNSRT:
+        # If we already have an abstract plan, execute the next step.
```
tomsilver, line 71: let's refactor this and the parent class so that we can remove the redundant lines involving the memory. so, make a new helper function that gets called inside of _predict, and move the non-shared logic into that helper function.

src/approaches/probe_approach.py
```diff
@@ -0,0 +1,201 @@
...
+        if "abstract_plan" in memory and memory["abstract_plan"]:
+            return memory["abstract_plan"].pop(0)
+        # Otherwise, we need to make a new abstract plan.
```
tomsilver, line 74: I'll review the lines below after we do a first round of changes

tomsilver (commented):
great work! here's a first round of feedback

src/approaches/probe_approach.py
Alchemy5: Yeah, so if the response from the LLM is cut off abruptly then the trimming process basically gets rid of the last line that is incomplete.

tomsilver: I think that should case be handled already with the existing method -- it will fail to parse into an option, and then just terminate the plan there

ronuchit:
i'd also like to review this one before merging! looks like an interesting new direction and i'd like to understand it

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+    Sequence, State, _GroundNSRT
+
+
+class LLMProbeApproach(OpenLoopLLMApproach):
+    """LLMProbeApproach definition."""
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "llm_probe"
+    #
```
tomsilver, line 49: remove hashtag

src/approaches/probe_approach.py
tomsilver, line 1: this file should be deleted right?

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+    def _predict(self, state: State, atoms: Set[GroundAtom],
+                 goal: Set[GroundAtom], memory: Dict) -> _GroundNSRT:
+        # If we already have an abstract plan, execute the next step.
+        if "abstract_plan" in memory and memory["abstract_plan"]:
+            return memory["abstract_plan"].pop(0)
+        # Otherwise, we need to make a new abstract plan.
```
tomsilver, line 55: this looks better, thanks! to finish off the refactoring, please change the parent class (`OpenLoopLLMApproach`) as well, so that the only thing that differs between the parent class and this class is the implementation of `_run_planner` (and now this class doesn't need to override `predict` anymore). the objective is to remove the lines of code that are in common with the parent class.

tomsilver: and btw one of the reasons to remove duplicate code is that if we want to change it in the future, we only have to change it in one place, rather than two places

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+        action_seq = self._run_planner(state, atoms, goal)
+
+        if action_seq is not None:
+            # If valid plan, add plan to memory so it can be refined!
+            memory["abstract_plan"] = action_seq
+            return memory["abstract_plan"].pop(0)
+        raise ApproachFailure("Approach failed to find solution")
+
+    def _run_planner(self, state: State, atoms: Set[GroundAtom],
```
tomsilver, line 64: this is doing more than running the planner, so I'd like to give it a more descriptive name, something that makes it clear that the LLM querying is happening here. maybe `_get_llm_guided_plan`?

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+                     goal: Set[GroundAtom]) -> List[_GroundNSRT]:
```
tomsilver, line 65: can you add a docstring describing what's going on in this method? the goal is to make it clear enough that someone without context can understand (e.g. Rohan!)

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+        init = atoms
+        initial_atoms: FrozenSet[GroundAtom] = frozenset(atoms)
+        objects = set(state)
+        trajectories = self._get_llm_based_state_action_predictions(
```
tomsilver, line 69: this helper function helps a lot, thanks! let's keep going, and make two other helper functions: one for creating the policy from the trajectories, and one for setting up and running the policy-guided planning. then it will be very clear that this whole method is really doing three steps:
1. getting trajectories from the LLM
2. converting the trajectories into a policy
3. running policy-guided planning

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+            atoms, goal, state)
+        final_states = []
```
tomsilver, line 71: i'm again going to defer reviewing this part until the next round, after the helper method refactoring, but I will have some suggestions to improve the clarity and code cleanliness

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+            heuristic=heuristic,
+            policy=policy,
+            num_rollout_steps=CFG.pg3_max_policy_guided_rollout,
+            rollout_step_cost=0)
+
+        return action_seq
+
+    def _get_llm_based_state_action_predictions(
+            self, atoms: Set[GroundAtom], goal: Set[GroundAtom],
+            state: State) -> List[Tuple[List[Any], List[Any]]]:
```
tomsilver, line 140: these `Any` are suspicious -- it's rare that `Any` is the right choice, since it's the most general possible type. can't these be `Set[GroundAtom]` and `_GroundNSRT` respectively?

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+    def _get_nsrt_ground_plan(
+        self, initial_atoms: Set[GroundAtom],
+        option_plan: List[Tuple[ParameterizedOption, Sequence]]
+    ) -> Tuple[List, List]:
+
+        states = []
+        actions = []
+        states.append(initial_atoms)
+        nsrts = self._get_current_nsrts()
+        #We assume that operators and NSRTs are the same
```
tomsilver, line 173: please match the format of our other comments: spaces in between hashtags and the words, and periods at the end of the sentence

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,187 @@
...
+        #for our use case, so this function would break down
+        #for enviornments not written in PDDL
```
tomsilver, line 175: typo: environments

tomsilver (commented):
great progress! here are some more comments

tomsilver:
also, don't forget to do all of the checks that you can (yapf, pylint, mypy)! @Alchemy5

src/approaches/probe_approach.py
Alchemy5: yeah, i'm not sure how to do delete it from the pull request though

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+    @classmethod
+    def get_name(cls) -> str:
+        return "llm_probe"
+
+    def _get_llm_based_plan(self, state: State, atoms: Set[GroundAtom],
+                     goal: Set[GroundAtom]) -> List[_GroundNSRT]:
+        """In this function, we take predictions the llm outputs regarding 
+        the solution to our PDDL problem and use them to make a dictionary of corresponding 
+        states and actions. This dictionary is used to create a policy which guides 
+        the a star search algorithm in searching for a solution to the PDDL problem."""
```
tomsilver, line 57: great description! let's just change "a star" to A*

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+        init = atoms
+        initial_atoms: FrozenSet[GroundAtom] = frozenset(atoms)
+        objects = set(state)
+        # Dictionary used to create policy for search function.
```
tomsilver, line 61: I think it would be more clear to put the LLM querying helper function inside of this method, and then use the output of that method as input to the second helper function (currently called `get_dictionary`), matching the 3 steps that I described in my previous review

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+        dictionary = self._get_dictionary(initial_atoms, atoms, goal, state)
```
tomsilver, line 62: let's make these variable and method names more descriptive!

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+        ground_nsrts = [
```
tomsilver, line 63: since these ground NSRTs are only used within the next method, let's just move their creation into that method

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+                llm_prediction, objects)
+            if len(option_plan) == 0:
+                continue
+            states, actions = self._get_nsrt_ground_plan(
+                initial_atoms, option_plan)
+
+            trajectories.append((states, actions))
+        return trajectories
+
+    def _get_nsrt_ground_plan(
```
tomsilver, line 100: minor: `nsrt_ground` -> `ground_nsrt`

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+        self, initial_atoms: Set[GroundAtom],
+        option_plan: List[Tuple[ParameterizedOption, Sequence]]
+    ) -> Tuple[List, List]:
```
tomsilver, line 103: make these Lists more specific (apply my comment about this from the previous review)

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+            new_states = []
+            for s in final_states:
+                new_states.append(frozenset(s))
+            # Creating dictionary for state-action combinations.
+            dictionary = dict(zip(new_states, final_actions))
+        else:
+            dictionary = None
+        return dictionary
+    def _run_policy_guided_planning(
+        self, dictionary:dict,goal:Set[GroundAtom],ground_nsrts:Set[_GroundNSRT],
```
tomsilver, line 156: use `Dict` for annotating dictionary types instead of `dict`

tomsilver, line 156: don't forget to change the variable name `dictionary` here too

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def _get_dictionary(
+        self, initial_atoms: Set[GroundAtom], atoms: Set[GroundAtom],
+        goal: Set[GroundAtom],state:State
+    ) -> Optional[dict]:
+        
+        trajectories = self._get_llm_based_state_action_predictions(
+            atoms, goal, state)
+        final_states = []
+        final_actions = []
```
tomsilver, line 133: okay so the logic in this function is looking pretty complicated and I'm hoping we can simplify it. could we do something like the following? (not tested)
```python
abstract_policy_dict = {}
for (states, actions) in trajectories:
    assert len(states) == len(actions) + 1
    for (s, a) in zip(states, actions):
        abstract_policy_dict[s] = a
return abstract_policy_dict
```
does that miss any important cases that are handled by the existing code?

src/approaches/open_loop_llm_approach.py
```diff
@@ -67,6 +67,16 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
         if "abstract_plan" in memory and memory["abstract_plan"]:
             return memory["abstract_plan"].pop(0)
         # Otherwise, we need to make a new abstract plan.
+        action_seq = self._get_llm_based_plan(state, atoms, goal)
```
tomsilver, line 70: perfect!

tomsilver (commented):
great stuff! back to you :)

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,137 @@
+"""Large language model (LLM) based policy guided bilevel planning approach.
```
ronuchit, line 1: policy-guided*

src/approaches/open_loop_llm_approach.py
```diff
@@ -75,40 +100,22 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+            self, option_plan: List[Tuple[ParameterizedOption,
+                                          Sequence[Object]]],
+            atoms: Set[GroundAtom], objects: Set[Object],
             goal: Set[GroundAtom]) -> Optional[List[_GroundNSRT]]:
-        objects = set(state)
-        option_plan = self._llm_prediction_to_option_plan(
-            llm_prediction, objects)
-        # If we failed to find a nontrivial plan with this prediction,
-        # continue on to next prediction.
-        if len(option_plan) == 0:
```
ronuchit, line 97: where did this logic go? did it turn out to be automatically handled by the existing logic in `task_plan_with_option_plan_constraint`?

src/approaches/open_loop_llm_approach.py
```diff
@@ -75,40 +100,22 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+        for pred in llm_predictions:
+            option_plan = self._llm_prediction_to_option_plan(pred, objects)
+            yield option_plan
 
-    def _process_single_prediction(
-            self, llm_prediction: str, state: State, atoms: Set[GroundAtom],
+    def _option_plan_to_nsrt_plan(
+            self, option_plan: List[Tuple[ParameterizedOption,
+                                          Sequence[Object]]],
+            atoms: Set[GroundAtom], objects: Set[Object],
```
ronuchit, line 110: can we rename atoms to init_atoms here?

src/approaches/open_loop_llm_approach.py
```diff
@@ -67,6 +68,30 @@ def _predict(self, state: State, atoms: Set[GroundAtom],
...
+        if action_seq is not None:
+            # If valid plan, add plan to memory so it can be refined!
+            memory["abstract_plan"] = action_seq
+            return memory["abstract_plan"].pop(0)
+        raise ApproachFailure("No LLM predicted plan achieves the goal.")
+
+    def _get_llm_based_plan(
+            self, state: State, atoms: Set[GroundAtom],
+            goal: Set[GroundAtom]) -> Optional[List[_GroundNSRT]]:
+        # Try to convert each output into an abstract plan.
```
ronuchit, line 81: add: "Return the first abstract plan that is found this way."

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,137 @@
...
+            goal: Set[GroundAtom]) -> Optional[List[_GroundNSRT]]:
+        """Query the LLM and use the outputs to initialize the queue in A*."""
+        # To initialize the queue in A*, we will build up a partial policy
+        # and then run policy-guided A*.
+        start_time = time.time()
+        objects = set(state)
+        partial_policy_dict: Dict[FrozenSet[GroundAtom], _GroundNSRT] = {}
+        for option_plan in self._get_llm_based_option_plans(
+                atoms, objects, goal):
+            ground_nsrt_plan = self._option_plan_to_nsrt_plan(
```
ronuchit, line 63: add a comment clarifying that this is calling the overriden method in this class, not the parent class's method

src/approaches/open_loop_llm_approach.py
ronuchit: actually i think that's wrong... the use of `atoms` here comes from the `predict` abstract method in the nsrt_metacontroller_approach, which is run on every step, not just on the init_atoms

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,137 @@
...
+            # entries that get overridden. This is okay, and there's no
+            # reason to prefer one option plan over another, so we just let
+            # the overriding happen.
+            partial_policy_dict.update(single_partial_policy_dict)
+        # Run policy-guided A*. Note that we importantly want to use
+        # sesame_plan because we will want to report the metrics collected in
+        # sesame_plan.
+        abstract_policy = lambda a, o, g: partial_policy_dict.get(
+            frozenset(a), None)
+        max_policy_guided_rollout = CFG.horizon  # not important
```
ronuchit, line 87: wait..why is this not important? is it because your abstract policy only "works" on a few states?

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,137 @@
...
+    def _get_llm_based_plan(
+            self, state: State, atoms: Set[GroundAtom],
+            goal: Set[GroundAtom]) -> Optional[List[_GroundNSRT]]:
+        """Query the LLM and use the outputs to initialize the queue in A*."""
+        # To initialize the queue in A*, we will build up a partial policy
+        # and then run policy-guided A*.
+        start_time = time.time()
+        objects = set(state)
+        partial_policy_dict: Dict[FrozenSet[GroundAtom], _GroundNSRT] = {}
+        for option_plan in self._get_llm_based_option_plans(
```
ronuchit, line 61: i see, so all the option_plans produced by the LLM are merged into a single partial policy?

ronuchit, line 61: i think that all 3 settings in settings.py that start with "open_loop_llm" should be changed to just start with "llm", since all 3 are used in this new approach too, which is not open-loop

src/approaches/llm_probe_approach.py
ronuchit, line 1: why on earth is this approach called "llm_probe_approach"? what is a probe? can we rename it to something more descriptive, like "llm_bilevel_planning_approach"?

ronuchit, line 1: QOL thing: maybe let's rename "open_loop_llm" as "llm_open_loop" just so the prefix "llm" is consistently used for all approaches that are llm-based? (can just make an issue for this)

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,137 @@
...
+            max_policy_guided_rollout=max_policy_guided_rollout)
+        self._save_metrics(metrics, nsrts, preds)
+        # Now convert the options back into ground NSRTs (:facepalm:). This
+        # is very circuitous and we should refactor it later. There are two
+        # original sins that are converging and making this hard: (1) the
+        # fact that the skeleton generator implements its own search, because
+        # otherwise we could just run the search utility functions here; and
+        # (2) the fact that this and the parent class are implemented as
+        # metacontrollers, which do not use sesame plan.
+        option_plan = [(o.parent, o.objects) for o in options]
```
ronuchit, line 107: so this option_plan is the one that comes out of sesame, and therefore is guaranteed to be a valid solution (as long as our option model is accurate) -- right? (i'm contrasting this with the option_plans that you looped over above, `for option_plan in self._get_llm_based_option_plans()`, which have no guarantees)

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,137 @@
...
+        options, metrics = self._run_sesame_plan(
+            task,
+            nsrts,
+            preds,
+            CFG.timeout - (time.time() - start_time),
+            CFG.seed,
+            abstract_policy=abstract_policy,
+            max_policy_guided_rollout=max_policy_guided_rollout)
+        self._save_metrics(metrics, nsrts, preds)
+        # Now convert the options back into ground NSRTs (:facepalm:). This
```
ronuchit, line 100: i don't really get why this is raising alarm bells... aren't you just going from the `plan` returned by sesame back to the `skeleton`? i agree that it's non-invertible in general, but another way to do this would be to just change sesame_plan() to return the skeleton along with the plan and the metrics. do you prefer doing it that way?

src/approaches/llm_probe_approach.py
tomsilver: probe refers to https://www.aaai.org/ocs/index.php/ICAPS/ICAPS11/paper/viewFile/2706/3148, but it is a loose analogy. will change

src/approaches/llm_probe_approach.py
tomsilver: yeah

src/approaches/llm_probe_approach.py
tomsilver: yeah basically. adding a better comment

src/approaches/llm_probe_approach.py
tomsilver: it's raising alarm bells because after we call sesame plan, we have already figured out option parameters (the continuous ones), but here we're throwing those away and counting on the metacontroller sampling to fill them back in later. this doesn't matter for us right now only because there are no continuous parameters in the PDDL envs.

src/approaches/llm_probe_approach.py
tomsilver: right

src/approaches/open_loop_llm_approach.py
tomsilver: yeah, i'm not sure why we had it in the first place

ronuchit (approved):
LGTM!

PR #1098, opened 2022-07-20
refactor pg4 and bilevel planning

tomsilver:
this fixes the fact that pg4 was not previously logging metrics like we want it to, and also removes some redundant code between pg4 and bilevel planning approach.

### DIFF
--- a/src/approaches/bilevel_planning_approach.py
+++ b/src/approaches/bilevel_planning_approach.py
@@ -5,7 +5,7 @@ then Execution.
 """
 
 import abc
-from typing import Callable, List, Set
+from typing import Callable, List, Set, Tuple
 
 from gym.spaces import Box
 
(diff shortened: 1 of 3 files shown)


src/approaches/bilevel_planning_approach.py
```diff
@@ -89,6 +80,25 @@ def _policy(s: State) -> Action:
 
         return _policy
 
+    def _run_sesame_plan(self, task: Task, nsrts: Set[NSRT],
```
ronuchit, line 83: alternative (preferred??) implementation of what youre trying to do here: make this method have kwargs, this class doesnt use any, but the pg4 run_sesame_plan just calls super() with the extra kwarg of abstract_policy

ronuchit (approved):
lgtm

PR #1099, opened 2022-07-21
config for pg4 experiments

tomsilver:
```
                     NUM_TRAIN_TASKS     NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME AVG_NODES_CREATED  LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                              
cover-easy-pg3          50.00 (0.00)   50.00 (0.00)   0.00 (0.00)   0.01 (0.01)       0.00 (0.00)   36.32 (1.51)         10
cover-easy-pg4          50.00 (0.00)   50.00 (0.00)   5.00 (0.00)   0.01 (0.00)       6.15 (0.18)   35.89 (1.45)         10
cover-easy-plan-only    50.00 (0.00)   50.00 (0.00)   5.00 (0.00)   0.01 (0.00)       6.86 (0.11)    0.00 (0.00)         10
cover-pg3               50.00 (0.00)  38.60 (11.46)   0.00 (0.00)   0.01 (0.00)       0.00 (0.00)  49.05 (12.25)         10
cover-pg4               50.00 (0.00)   50.00 (0.00)   5.00 (0.00)   0.07 (0.02)       6.62 (1.19)  49.11 (12.87)         10
cover-plan-only         50.00 (0.00)   50.00 (0.00)   5.00 (0.00)   0.07 (0.02)       6.66 (0.32)    0.00 (0.00)         10
```


scripts/configs/pg4.yaml
```diff
@@ -0,0 +1,53 @@
+# PG4 experiments.
+---
+APPROACHES:
+  pg3:
+    NAME: "pg3"
+    FLAGS: []
```
tomsilver, line 6: this was my fault for doing this earlier -- can you actually change all instances of `FLAGS: []` to `FLAGS: {}`? since it's a dict not a list

scripts/configs/pg4.yaml
```diff
@@ -0,0 +1,53 @@
...
+ENVS:
+  cover-easy:  # downward refinable. PG3 should succeed
+    NAME: "cover"
+    FLAGS:
+      cover_initial_holding_prob: 0.0
+  cover:  # PG3 should sometimes fail, but PG4 should succeed
+    NAME: "cover"
+    FLAGS: []
+  painting-easy-no-holding:
```
tomsilver, line 21: do we know that this will succeed? should we do box-only?

tomsilver (commented):
exciting!

scripts/configs/pg4.yaml
ronuchit, line 6: note @wester-is-not-real : tom realized that all the "FLAGS: []" should actually be "FLAGS: {}" (see scripts/configs/example_basic.yaml)

wester-is-not-real: Sounds good!

scripts/configs/pg4.yaml
wester-is-not-real: PG3 fails sometimes when it is initially holding, PG4 never fails. What do you mean by box-only?

scripts/configs/pg4.yaml
wester-is-not-real: Thank you!

scripts/configs/pg4.yaml
tomsilver: I meant setting the goal to be box only. maybe let's add that as a separate env?

tomsilver:
I can't approve this because I opened the PR, but consider it approved! let's collect and put all results in the PR description above, and then merge. @wester-is-not-real

ronuchit (approved):

tomsilver:
actually I want to merge this now so we can run it on supercloud without changing the branch in the config. we can edit the PR description after it's merged.

PR #1110, opened 2022-07-22
add new `plan_only_eval` flag

wmcclinton:
This PR contains the main.py changes from the [Behavior Pipeline Deterministic PR](https://github.com/Learning-and-Intelligent-Systems/predicators/pull/1082). Here we add the `plan_only_eval` flag to allow for much quicker execution of BilevelPlanningApproaches in Behavior. When `plan_only_eval` is set True  we will not execute our final plan in simulation. In this case, we say a task is solved successfully if we were able to find a plan even if the plan would not have necessarily worked in simulation (this is especially true of option models that don't model the simulator well).


tests/test_main.py
```diff
@@ -221,3 +221,22 @@ def test_env_failure():
...
+        "plan_only_eval": True
+    })
+    env = _DummyCoverEnv()
+    train_tasks = env.get_train_tasks()
+    approach = create_approach("oracle", env.predicates, env.options,
+                               env.types, env.action_space, train_tasks)
+    assert not approach.is_learning_based
+    task = train_tasks[0]
+    approach.solve(task, timeout=500)
+    _run_testing(env, approach)
```
ronuchit, line 242: your comment says "Test plan_only_eval succeeds, even though simulation fails.". but i don't see where you're actually checking that it succeeds. do you mean to do something like...
```
metrics = _run_testing(env, approach)
assert metrics["num_solved"] == 1
```
?

tests/test_main.py
```diff
@@ -211,7 +211,7 @@ def test_env_failure():
         "timeout": 10,
         "make_test_videos": False,
         "cover_initial_holding_prob": 0.0,
-        "num_test_tasks": 1,
+        "num_test_tasks": 1
```
ronuchit, line 214: revert?

src/settings.py
```diff
@@ -44,6 +44,12 @@ class GlobalSettings:
     # in unit tests, make sure to pass in a value for `render_state_dpi` into
     # your call to utils.reset_config().
     render_state_dpi = 150
+    # If this is True, then we will not execute our final plan in simulation
+    # and we will say a task is solved successfully if we were able to find
+    # a plan. Note, we might solve the task even if the plan we found would
+    # not necessarily work in simulation (this is especially true of option
+    # models that don't model the simulator well).
```
ronuchit, line 51: Add: "This can only be True if the approach is a subclass of BilevelPlanningApproach."

src/main.py
```diff
@@ -309,7 +320,7 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
                     task.goal_holds,
                     max_num_steps=CFG.horizon,
                     monitor=monitor)
-            solved = task.goal_holds(traj.states[-1])
+                solved = task.goal_holds(traj.states[-1])
```
ronuchit, line 323: ah i see, this was a bug before, nice catch

src/main.py
```diff
@@ -284,14 +284,25 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
         else:
```
ronuchit, line 284: up above, right after the line "for test_task_idx, task in enumerate(test_tasks):" can you add the following comment:

Run the approach's solve() method to get a policy for this task.

src/main.py
```diff
@@ -284,14 +284,25 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
             monitor = None
         try:
-            if CFG.env == "behavior" and \
-                CFG.behavior_option_model_eval:  # pragma: no cover
+            # Will evaluate all approaches on our simulator with the options
```
ronuchit, line 287: given that EVERY settings configuration will get the code to this point, i feel like this comment should be more informative. i'll take a stab at rewriting it:

Now, measure success by running the policy in the environment. There are two special cases that we handle first. In the `if`, we consider the case where `plan_only_eval` is True, in which case we only check whether this BilevelPlanningApproach found a plan. In the `elif`, we consider the case where `behavior_option_model_eval` is True, in which case for BEHAVIOR we evaluate on option models instead of the low-level simulator. Finally, the `else` handles the default case, where we use `utils.run_policy` to roll out the policy in the environment.

ronuchit (commented):
lgtm as long as the checks pass. @tomsilver can give approval

tomsilver (approved):
nice work!

PR #1111, opened 2022-07-23
change cost of policy-generated actions to zero in sesame

tomsilver:
see the new comment for justification. this is important for @Alchemy5 's code and I suspect that it will also improve @wester-is-not-real 's results.

@wester-is-not-real  -- we should regenerate the pg4 results after this is merged.

### DIFF
--- a/tests/test_planning.py
+++ b/tests/test_planning.py
@@ -534,7 +534,7 @@ def test_policy_guided_sesame():
         block = unrealized_blocks[0]
         return pick_nsrt.ground([block])
 
-    plan, metrics = sesame_plan(
+    _, metrics = sesame_plan(
         task,
         option_model,
         nsrts,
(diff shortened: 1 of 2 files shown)


ronuchit (commented):
looks reasonable. is this what's already happening in the utils.py functions used by PG3, e.g. `run_policy_guided_astar`?

is this `cumulative_cost=0` the same functional thing as `rollout_step_cost` being 0 in that function?

tomsilver:
@ronuchit thanks for looking, yeah exactly

tests/test_planning.py
```diff
@@ -514,7 +514,7 @@ def test_policy_guided_sesame():
     pick_nsrt = nsrt_name_to_nsrt["Pick"]
     place_nsrt = nsrt_name_to_nsrt["Place"]
 
-    # When using a perfect policy, we should only expand nodes along the plan.
+    # When using a perfect policy, we should only expand the first node.
```
ronuchit, line 517: i don't get it, why did this change? how can you not expand all the nodes to the goal?

ronuchit (approved):

tests/test_planning.py
tomsilver: we create all the nodes to the goal, but we don't need to expand them

PR #1112, opened 2022-07-23
pull out metrics saving in bilevel_planning_approach.py

ronuchit:
makes #1097 a bit easier to review, and anyway is a good idea

### DIFF
--- a/src/approaches/bilevel_planning_approach.py
+++ b/src/approaches/bilevel_planning_approach.py
@@ -100,14 +100,14 @@ class BilevelPlanningApproach(BaseApproach):
                 "num_nodes_expanded", "num_nodes_created", "plan_length"
         ]:
             self._metrics[f"total_{metric}"] += metrics[metric]
-            self._metrics["total_num_nsrts"] += len(nsrts)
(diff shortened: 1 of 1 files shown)


src/approaches/bilevel_planning_approach.py
```diff
@@ -105,6 +93,22 @@ def reset_metrics(self) -> None:
         # Initialize min to inf (max gets initialized to 0 by default).
         self._metrics["min_num_skeletons_optimized"] = float("inf")
 
+    def _save_metrics(self, metrics: Metrics, nsrts: Set[NSRT],
+                      predicates: Set[Predicate]) -> None:
+        for metric in [
+                "num_skeletons_optimized", "num_failures_discovered",
+                "num_nodes_expanded", "num_nodes_created", "plan_length"
+        ]:
```
ronuchit, line 101: oh there's an indent problem here .... @tomsilver this was in #1097 too. is it a bug there?

tomsilver: oops yeah good catch

tomsilver (approved):
thank you!

PR #1114, opened 2022-07-23
safeguard for llm experiments where we expect to only use the cache

tomsilver:


ronuchit (approved):
would you prefer this to default True, to avoid wasting 💵 💰 💸 ?

tomsilver:
nah, the typical case will be that we need to query anew

PR #1120, opened 2022-07-23
catch sesame failures in llm bilevel approach 

tomsilver:
le sigh


ronuchit (approved):
lgtm, update PR name before merging

PR #1121, opened 2022-07-23
fixes for `visited_skeletons` in planning.py

### LINKED ISSUE #1119: fix erroneous usage of `set(tuple())` in planning.py
should be `{tuple()}`? why was it working fine up until now?

ronuchit:
closes #1119

### DIFF
--- a/src/planning.py
+++ b/src/planning.py
@@ -256,7 +256,8 @@ def _skeleton_generator(
     # Initialize with empty skeleton for root.
     # We want to keep track of the visited skeletons so that we avoid
     # repeatedly outputting the same faulty skeletons.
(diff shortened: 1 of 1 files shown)


src/planning.py
```diff
@@ -325,11 +325,10 @@ def _skeleton_generator(
                                                        node.atoms):
                 child_atoms = utils.apply_operator(nsrt, set(node.atoms))
                 child_skeleton = node.skeleton + [nsrt]
-                if abstract_policy is not None:
```
ronuchit, line 328: i don't get why this had to be in this `if`, going to try removing that...

tomsilver: we were erring on the side of caution in terms of impacting existing code. I agree it shouldn't matter in principle

tomsilver (approved):
lgtm

PR #1122, opened 2022-07-23
fix `set(tuple())`

### LINKED ISSUE #1119: fix erroneous usage of `set(tuple())` in planning.py

tomsilver:
closes #1119


tomsilver:
race conditions (#1121)

Issue #1109, opened 2022-07-22
supercloud + yaml


tomsilver:
it would be nice if we could use the YAML config files that we made for openstack on supercloud as well. that way we can define experiments once and then run them on either cluster.

PR #1127, opened 2022-07-24
refactor supercloud submission file

tomsilver:
part 2 of  #1109

I tried the nightly script with this change and the launch looks normal


ronuchit (approved):
i don't get it... what actually is changing here?

tomsilver:
hopefully nothing -- I just want to use the new function in a later PR

PR #1128, opened 2022-07-24
supercloud launch from config 

### LINKED ISSUE #1109: supercloud + yaml

tomsilver:
closes #1109 

can now run stuff like this from our local machines:
```
python scripts/supercloud/launch.py --config example.yaml --user tslvr
```


scripts/supercloud/launch.py
```diff
@@ -0,0 +1,81 @@
+"""Launch supercloud experiments defined by config files.
+
+Usage example:     python scripts/supercloud/launch.py --config
```
ronuchit, line 3: add a blank line after "Usage example:" so that yapf doesn't make this ugly

ronuchit (approved):
very very cool! is this the preferred pattern over our standard way of having `.sh` runfiles?

tomsilver:
thanks! yeah, I think we should use config files from now on when setting up new experiments. then we can use either cluster, and it's also prettier

PR #1130, opened 2022-07-24
download and kill-all scripts for supercloud

tomsilver:


scripts/openstack/kill_all.py
```diff
@@ -15,7 +15,7 @@
 import argparse
 import os
 
-from predicators.scripts.openstack.launch import run_cmds_on_machine
+from predicators.scripts.cluster_utils import run_cmds_on_machine
```
ronuchit, line 18: bugfix? git grep for any other stragglers?

scripts/supercloud/kill_all.py
```diff
@@ -0,0 +1,28 @@
+"""Script to kill all experiments running on supercloud for a given user.
+
+Runs scancel -u <user> on supercloud.
```
ronuchit, line 3: lol, i really don't think it's that hard to just log into supercloud and run scancel... do we really need a local script for it?

do you not already have aliases + ssh keys set up for logging into supercloud? i just run `supercloud`, `supercloudtom`, `supercloudnjk`, etc. takes 1 second to log in

ronuchit (approved):

scripts/supercloud/kill_all.py
tomsilver: yeah i have those aliases, i'm doing this so that
1. the scripts mirror the openstack ones
2. in a perfect world, users who are not me could use these scripts without ever having to directly ssh into the machines

PR #1131, opened 2022-07-24
allow non-cross product experiment configs 

tomsilver:


scripts/configs/example.yaml
```diff
@@ -3,15 +3,15 @@
 APPROACHES:
   my-oracle:  # used in constructing the experiment ID
     NAME: "oracle"
-    FLAGS: []
+    FLAGS: {}
```
tomsilver, line 6: it evidently didn't matter for the existing code, but i realized that the right yaml syntax for an empty dictionary is this

scripts/configs/example2.yaml
```diff
@@ -0,0 +1,36 @@
+# An example configuration file showing how to define multiple approach/env
```
ronuchit, line 1: give these files better names like `example_basic` and `example_multiple_combinations`

scripts/configs/example.yaml
```diff
@@ -3,15 +3,15 @@
...
   nsrt-learning:
     NAME: "nsrt_learning"
     FLAGS:
       disable_harmlessness_check: True  # just an example
 ENVS:
   my-cover:  # used in constructing the experiment ID
     NAME: "cover"
-    FLAGS: []
+    FLAGS: {}
   single-block-cover:
```
ronuchit, line 15: i was a bit confused by this naming and how it relates to "my-cover" above. it might be clearer if you change the names, like:
"my-cover" -> "cover-default-settings"
"single-block-cover" -> "cover-single-blocks"

ronuchit (approved):

PR #1134, opened 2022-07-24
purge `initial_predicates` from planning.py

ronuchit:
maybe we should run a nightly after this? @tomsilver


tomsilver (approved):
Yup! Do u wanna do the honors?

PR #1136, opened 2022-07-25
fix pg3 crash on discovered failures

tomsilver:
before this change, this crashed:
```python src/main.py --env cluttered_table --approach pg3 --seed 0 --experiment_id cluttered-table-pg3 --strips_learner oracle --sampler_learner oracle --num_train_tasks 5```

### DIFF
--- a/src/planning.py
+++ b/src/planning.py
@@ -814,6 +814,14 @@ def _sesame_plan_with_fast_downward(
     return plan, metrics
 
 
+class PlanningFailure(utils.ExceptionWithInfo):
+    """Raised when the planner fails."""
+
+
+class PlanningTimeout(utils.ExceptionWithInfo):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/planning.py [hunk 1]
```diff
@@ -823,7 +831,7 @@ class _DiscoveredFailure:
     failing_nsrt: _GroundNSRT
 
 
-class _DiscoveredFailureException(ExceptionWithInfo):
+class _DiscoveredFailureException(PlanningFailure):
```
ronuchit, line 834: in order for this change to be okay, you need to make sure that wherever we're catching DiscoveredFailureException, we aren't already catching PlanningFailure inside of there, because now that inside catch would activate, changing the behavior of the code

ronuchit (approved):

PR #1137, opened 2022-07-25
fix initialization in online pg3

tomsilver:
we were accidentally starting the search for online PG3 with the old policy, which could use outdated NSRTs

### DIFF
--- a/src/approaches/pg3_approach.py
+++ b/src/approaches/pg3_approach.py
@@ -100,6 +100,9 @@ class PG3Approach(NSRTLearningApproach):
         # The heuristic is what distinguishes PG3 from baseline approaches.
         heuristic = self._create_heuristic()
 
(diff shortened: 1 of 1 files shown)


Abraham-Mitchell (approved):

ronuchit (commented):
don't you instead want to reset self._current_ldl = LiftedDecisionList([]) at the start of learn_ldl?

tomsilver:
>don't you instead want to reset self._current_ldl = LiftedDecisionList([]) at the start of learn_ldl?
I considered that, but we do have a "current LDL" still technically, so I think that this is better

ronuchit:
why would you have a current LDL if you've just called learn_ldl()?

tomsilver:
you have it from the last learning cycle

PR #1140, opened 2022-07-25
Openstack progress

wester-is-not-real:
Added script that outputs the machines that are still running a python3.8 process. For this, I changed the run_cmds_on_machine function so it would output the response return code.

### DIFF
--- a/scripts/openstack/progress.py
+++ b/scripts/openstack/progress.py
@@ -1,0 +1,57 @@
+"""Script for getting status of active openstack predicator experiments.
+
+WARNING: Will return the status of any other python3.8 processes running on
+        each machine in machines.txt
+
+See launch.py for information about the format of machines.txt.
+
+Usage example:
+    python scripts/openstack/progress.py --machines machines.txt \
+        --sshkey ~/.ssh/cloud.key
+"""
+
+import argparse
+import os
+import subprocess
+
+from predicators.scripts.openstack.launch import run_cmds_on_machine
+
+
+def _main() -> None:
+    # Set up argparse.
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--machines", required=True, type=str)
+    parser.add_argument("--sshkey", required=True, type=str)
+    args = parser.parse_args()
+    openstack_dir = os.path.dirname(os.path.realpath(__file__))
+    # Load the machine IPs.
+    machine_file = os.path.join(openstack_dir, args.machines)
+    with open(machine_file, "r", encoding="utf-8") as f:
+        machines = f.read().splitlines()
+    # Make sure that the ssh key exists.
+    assert os.path.exists(args.sshkey)
+    # Loop through each machine and print those that have a python3.8 process.
+    progress_cmd = "pgrep python3.8"
+    print("The following machines are still running:\n")
+    for machine in machines:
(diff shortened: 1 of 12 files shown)


scripts/openstack/progress.py
```diff
@@ -0,0 +1,47 @@
...
+    openstack_dir = os.path.dirname(os.path.realpath(__file__))
+    # Load the machine IPs.
+    machine_file = os.path.join(openstack_dir, args.machines)
+    with open(machine_file, "r", encoding="utf-8") as f:
+        machines = f.read().splitlines()
+    # Make sure that the ssh key exists.
+    assert os.path.exists(args.sshkey)
+    # Loop through each machine and print those that have a python3.8 process.
+    progress_cmd = "pgrep python3.8"
+    print("The following machines are still running:\n")
```
tomsilver, line 34: can we print something different if no machines are running?

tomsilver (commented):
yay awesome, thanks for doing this 🙏

scripts/openstack/progress.py
wester-is-not-real: Great idea, on it.

scripts/openstack/progress.py
```diff
@@ -0,0 +1,47 @@
+"""Script for getting status of active openstack predicator experiments.
+
+WARNING: Will return the status of any other python3.8 processes running on
+        each machine in machines.txt
```
ronuchit, line 4: nit: indent by one more space to align with previous line

scripts/openstack/progress.py
```diff
@@ -0,0 +1,47 @@
...
+    args = parser.parse_args()
+    openstack_dir = os.path.dirname(os.path.realpath(__file__))
+    # Load the machine IPs.
+    machine_file = os.path.join(openstack_dir, args.machines)
+    with open(machine_file, "r", encoding="utf-8") as f:
+        machines = f.read().splitlines()
+    # Make sure that the ssh key exists.
+    assert os.path.exists(args.sshkey)
+    # Loop through each machine and print those that have a python3.8 process.
+    progress_cmd = "pgrep python3.8"
```
ronuchit, line 33: is there a reason you're only checking for python 3.8 in particular? what if someone is running 3.9 or 3.7? what would go wrong if you just do "pgrep python" here?

scripts/openstack/progress.py
```diff
@@ -0,0 +1,47 @@
...
+    print("The following machines are still running:\n")
+    for machine in machines:
+        # If return code is 0, print active machine.
+        returncode = run_cmds_on_machine([progress_cmd],
+                                         machine,
+                                         args.sshkey,
+                                         allowed_return_codes=(0, 1))
+
+        if returncode != 1:
+            print(f"{machine}")
```
ronuchit, line 43: nit: you can just do `print(machine)`

scripts/openstack/progress.py
wester-is-not-real: launch.py specifically chooses python3.8 for the machines, so it shouldn't be an issue.

tomsilver:
this command might work better: `ps -ef | grep python`

tomsilver:
gonna close this for now and reopen on an as-needed basis

PR #1142, opened 2022-07-25
Probe

Alchemy5:


src/approaches/analysis_approach.py
```diff
@@ -0,0 +1,274 @@
+"""Open-loop large language model (LLM) meta-controller approach with prompt modification where predicate names are randomly generated.
+Note: Predicate and Task classes must be set to frozen=False
+Example command line:
+    export OPENAI_API_KEY=<your API key>
+    python src/main.py --approach analysis --seed 0 \
```
NishanthJKumar, line 5: can we rename the approach to something more meaningful (e.g. `llm_predicate_renaming`)? Calling it `analysis` seems too vague and generic

src/approaches/analysis_approach.py
```diff
@@ -0,0 +1,274 @@
...
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        # Set up the LLM.
+        self._llm = OpenAILLM(CFG.open_loop_llm_model_name)
+        # Set after learning.
+        self._prompt_prefix = ""
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "analysis"
```
NishanthJKumar, line 66: See above comment about name

src/approaches/analysis_approach.py
```diff
@@ -0,0 +1,274 @@
...
+            return memory["abstract_plan"].pop(0)
+        raise ApproachFailure("Approach failed to find solution")
+
+    def _get_llm_based_plan(
+            self, state: State, atoms: Set[GroundAtom],
+            goal: Set[GroundAtom]) -> Optional[List[_GroundNSRT]]:
+        new_prompt = self._create_prompt(atoms, goal, [])
+        prompt = self._prompt_prefix + new_prompt
+        import pdb
+        pdb.set_trace()
```
NishanthJKumar, line 88: remove debug statement

src/approaches/analysis_approach.py
```diff
@@ -0,0 +1,274 @@
...
+from predicators.src.planning import task_plan_with_option_plan_constraint
+from predicators.src.settings import CFG
+from predicators.src.structs import Box, Dataset, GroundAtom, Object, \
+    ParameterizedOption, Predicate, State, Task, Type, _GroundNSRT, _Option
+
+existing = []
+existing_new = []
+
+
+class AnalysisApproach(NSRTMetacontrollerApproach):
```
NishanthJKumar, line 51: high-level comment: i don't really understand why this is subclassing `NSRTMetacontrollerApproach` and not `OpenLoopLLMApproach`? It seems like a lot of the methods here have the same code as `OpenLoopLLMApproach`. Even if subclassing is not possible for some reason, it really feels like we should think of a smart way to share code between these classes

src/approaches/analysis_approach.py
```diff
@@ -0,0 +1,274 @@
...
+from predicators.src.approaches.nsrt_metacontroller_approach import \
+    NSRTMetacontrollerApproach
+from predicators.src.llm_interface import OpenAILLM
+from predicators.src.planning import task_plan_with_option_plan_constraint
+from predicators.src.settings import CFG
+from predicators.src.structs import Box, Dataset, GroundAtom, Object, \
+    ParameterizedOption, Predicate, State, Task, Type, _GroundNSRT, _Option
+
+existing = []
+existing_new = []
```
NishanthJKumar, line 48: do these really need to be global variables? can we make them privately-accessed class variables instead?

src/approaches/analysis_approach.py
```diff
@@ -0,0 +1,274 @@
...
+                objs_list.append(obj)
+            if not malformed:
+                option_plan.append((option, objs_list))
+        return option_plan
+
+    def _generate_guess(self, sentence: str) -> str:
+        alphabet = [
+            'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
+            'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
+            ' '
```
NishanthJKumar, line 192: wait, if the `' '` character is part of this, that means a predicate like `On` could become `a b c`, which might be interpreted as 3 separate things right? I feel like we shouldn't have spaces here.

src/approaches/analysis_two_approach.py
```diff
@@ -0,0 +1,263 @@
+"""Open-loop large language model (LLM) meta-controller approach with prompt modification where operator names are randomly generated.
+Note: Predicate and Task classes must be set to frozen=False
+Example command line:
+    export OPENAI_API_KEY=<your API key>
+    python src/main.py --approach analysis_two --seed 0 \
```
NishanthJKumar, line 5: Similar to my comment on the `analysis_approach` script, can we use a more informative name than `analysis_two` (e.g. `llm_operator_renaming`)?

src/approaches/analysis_two_approach.py
```diff
@@ -0,0 +1,263 @@
...
+from predicators.src.approaches.nsrt_metacontroller_approach import \
+    NSRTMetacontrollerApproach
+from predicators.src.llm_interface import OpenAILLM
+from predicators.src.planning import task_plan_with_option_plan_constraint
+from predicators.src.settings import CFG
+from predicators.src.structs import Box, Dataset, GroundAtom, Object, \
+    ParameterizedOption, Predicate, State, Task, Type, _GroundNSRT, _Option
+
+original_strings = []
+random_strings = []
```
NishanthJKumar, line 47: same comment as with the other file, can we make these private class variables instead of global variables? Or is it necessary that they be global for some reason?

src/approaches/analysis_two_approach.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class AnalysisTwoApproach(NSRTMetacontrollerApproach):
```
NishanthJKumar, line 50: This seems to be sharing a lot of code with the `AnalysisOneApproach` (seems like almost all functions are the exact same except `_llm_prediction_to_option_plan` and `_create_prompt`?) and even the `LLMOpenLoopApproach`. I think things will be much cleaner if we share code amongst all of these in a nice way: e.g., maybe we should subclass `LLMOpenLoopApproach` into `LLMAnalysisApproach` and further subclass `AnalysisOneApproach` and `AnalysisTwoApproach` from this common superclass? Or maybe both these new analysis approaches should just directly subclass `LLMOpenLoopApproach`

src/approaches/analysis_two_approach.py
```diff
@@ -0,0 +1,263 @@
...
+        option_plan: List[Tuple[ParameterizedOption, Sequence[Object]]] = []
+        # Setup dictionaries enabling us to easily map names to specific
+        # Python objects during parsing.
+        option_name_to_option = {op.name: op for op in self._initial_options}
+
+        obj_name_to_obj = {o.name: o for o in objects}
+        # We assume the LLM's output is such that each line contains a
+        # option_name(obj0:type0, obj1:type1,...).
+        options_str_list = llm_prediction.split('\n')
+        print(options_str_list)
```
NishanthJKumar, line 142: can we get rid of this print?

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,185 @@
...
+        new_prompt = self._create_prompt(atoms, goal, [])
+        prompt = self._prompt_prefix + new_prompt
+        llm_predictions = self._llm.sample_completions(
+            prompt=prompt,
+            temperature=CFG.open_loop_llm_temperature,
+            seed=CFG.seed,
+            num_completions=CFG.open_loop_llm_num_completions)
+
+        for llm_prediction in llm_predictions:
+
```
NishanthJKumar, line 91: unnecessary new line, delete.

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,185 @@
...
+            option_plan = self._llm_prediction_to_option_plan(
+                llm_prediction, environment_objects)
+            if len(option_plan) == 0:
+                continue
+            states, actions = self._get_ground_nsrt_plan(
+                initial_atoms, option_plan)
+
```
NishanthJKumar, line 98: unnecessary new line, delete.

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,185 @@
...
+                                                      atoms, initial_atoms)
+        return action_seq
+
+    def _get_llm_based_state_action_predictions(
+        self, atoms: Set[GroundAtom], goal: Set[GroundAtom],
+        environment_objects: Collection[Object]
+    ) -> List[Tuple[List[Set[GroundAtom]], List[_GroundNSRT]]]:
+        trajectories = []
+        initial_atoms = atoms
+
```
NishanthJKumar, line 81: unnecessary new line, delete.

src/approaches/llm_probe_approach.py
```diff
@@ -0,0 +1,185 @@
...
+                initial_atoms, option_plan)
+
+            trajectories.append((states, actions))
+        return trajectories
+
+    def _get_ground_nsrt_plan(
+        self, initial_atoms: Set[GroundAtom],
+        option_plan: List[Tuple[ParameterizedOption, Sequence]]
+    ) -> Tuple[List[Set[GroundAtom]], List[_GroundNSRT]]:
+
```
NishanthJKumar, line 106: unnecessary new line, delete.

src/settings.py
```diff
@@ -175,13 +175,15 @@ class GlobalSettings:
...
-    pddl_spanner_procedural_test_max_extra_spanners = 10
-    pddl_spanner_procedural_test_min_locs = 20
-    pddl_spanner_procedural_test_max_locs = 30
-
+    pddl_spanner_procedural_test_max_extra_spanners = 5
+    pddl_spanner_procedural_test_min_locs = 5
+    pddl_spanner_procedural_test_max_locs = 6
+    #original pddl_spanner_procedural_test_min_locs = 20, pddl_spanner_procedural_test_max_locs=30
+    #at hyperparameters 1,3,0,2,2,4,5,6,0,5,5,6 there seems to be a bit of success with the oracle approach might work for easy difficulty, 1/2 accuracy, llm also seems to achieve 1/2 accuracy
+    #at
```
NishanthJKumar, line 186: is this comment necessary? If so, can you reformat it to be more easy-to-parse?

src/settings.py
```diff
@@ -261,9 +263,9 @@ class GlobalSettings:
     llm_openai_max_response_tokens = 700
 
     # parameters for open loop LLM approach
-    open_loop_llm_model_name = "text-curie-001"  # "text-davinci-002"
+    open_loop_llm_model_name = "text-davinci-002"  # "text-davinci-002"
     open_loop_llm_temperature = 0.5
-    open_loop_llm_num_completions = 1
+    open_loop_llm_num_completions = 2
```
NishanthJKumar, line 268: did we change this default? I thought we were planning to leave this at `1`?

NishanthJKumar (changes requested):
Good work implementing all this @Alchemy5 ! I didn't take too close of a look because I feel that there is some high-level design changes necessary in order to prevent substantial code duplication. In particular, I think we can figure out a better way to share most of the code amongst all the new classes you've made.

NishanthJKumar (commented):
Fantastic work @Alchemy5 ! Ping again after all the checks are passed and I think this PR should be ready to merge!

tomsilver:
I'd like to review too before we merge but after checks pass :)

NishanthJKumar:
Will definitely lyk then :100:

NishanthJKumar (commented):
LGTM! Ready for your eyes @tomsilver

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
+"""Open-loop large language model (LLM) meta-controller approach with
+ prompt modification where operator names are randomly generated.
+Note: Predicate and Task classes must be set to frozen=False
```
tomsilver, line 3: @Alchemy5 is this comment still true? that's definitely something we will want to change if so. we don't want to have to manually change stuff to run the code each time, and we also won't want to change the frozenness of these classes since that frozenness is assumed in many places

tomsilver, line 3: nit: can you add some new lines in between the sentences here?

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+
+class LLMOperatorRenamingApproach(LLMPredicateRenamingApproach):
+    """LLMOperatorRenamingApproach definition."""
+
+    def __init__(self, initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task]) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self.original_operators: List[str] = []
```
tomsilver, line 53: let's make these private attributes to it's clear that they shouldn't be used outside of this class (so just add underscores at the beginning of the names)

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        self.random_operators: List[str] = []
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "llm_operator_renaming"
+
+    def _llm_prediction_to_option_plan(
+        self, llm_prediction: str, objects: Collection[Object]
+    ) -> List[Tuple[ParameterizedOption, Sequence[Object]]]:
+        """Convert the output of the LLM into a sequence of
```
tomsilver, line 63: when overriding methods, we don't need to write docstrings, so you can remove this

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        ParameterizedOptions coupled with a list of objects that will be used
+        to ground the ParameterizedOption."""
+        # We assume the LLM's output is such that each line contains a
+        # option_name(obj0:type0, obj1:type1,...).
+        options_str_list = llm_prediction.split('\n')
+        for option_str in options_str_list:
+            for j in self.random_operators:
```
tomsilver, line 70: let's use a more informative variable name, like `random_operator`

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+
+from predicators.src.approaches.llm_open_loop_approach import \
+    LLMOpenLoopApproach
+from predicators.src.approaches.llm_predicate_renaming_approach import \
+    LLMPredicateRenamingApproach
+from predicators.src.structs import Box, GroundAtom, Object, \
+    ParameterizedOption, Predicate, Task, Type, _Option
+
+
+class LLMOperatorRenamingApproach(LLMPredicateRenamingApproach):
```
tomsilver, line 45: since the only thing in common between this and the parent class is the `_generate_guess` method, I think that method would be better as a new utility function in `utils.py` that gets used by both the operator and predicate renaming approaches

src/approaches/llm_operator_renaming_approach.py
tomsilver: please apply this to the other file too

src/approaches/llm_operator_renaming_approach.py
tomsilver: please apply this to the other file too

src/approaches/llm_operator_renaming_approach.py
tomsilver: please apply this to the other file too

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        to ground the ParameterizedOption."""
+        # We assume the LLM's output is such that each line contains a
+        # option_name(obj0:type0, obj1:type1,...).
+        options_str_list = llm_prediction.split('\n')
+        for option_str in options_str_list:
+            for j in self.random_operators:
+                if j in option_str:
+                    for i in option_str:
+                        if i != " ":
+                            startIndex = option_str.index(i)
```
tomsilver, line 74: let's stick with `snake_case` not `camelCase` :)

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        """Convert the output of the LLM into a sequence of
+        ParameterizedOptions coupled with a list of objects that will be used
+        to ground the ParameterizedOption."""
+        # We assume the LLM's output is such that each line contains a
+        # option_name(obj0:type0, obj1:type1,...).
+        options_str_list = llm_prediction.split('\n')
+        for option_str in options_str_list:
+            for j in self.random_operators:
+                if j in option_str:
+                    for i in option_str:
```
tomsilver, line 72: i don't actually understand what's going on here, can you explain?

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+                        if i != " ":
+                            startIndex = option_str.index(i)
+                            break
+                    sub = option_str[startIndex:option_str.index('(')]
```
tomsilver, line 76: i suspect that we can simplify / clean these lines here after I understand what's going on

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+            sub = option.name
+            if sub in self.original_operators:
+                option.name = self.random_operators[
+                    self.original_operators.index(sub)]
+            else:
+                self.original_operators.append(sub)
+                sub_random = ''.join(self._generate_guess(sub))
+                self.random_operators.append(sub_random)
+                option.name = sub_random
+        prompt = LLMOpenLoopApproach._create_prompt(self, init, goal, options)  # pylint: disable=protected-access
```
tomsilver, line 99: after changing the subclass, you can go back to using `super()` here

src/approaches/llm_operator_renaming_approach.py
```diff
@@ -0,0 +1,100 @@
...
+        unmodified_prediction = "\n".join(options_str_list)
+        option_plan = super()._llm_prediction_to_option_plan(
+            unmodified_prediction, objects)
+        return option_plan
+
+    def _create_prompt(self, init: Set[GroundAtom], goal: Set[GroundAtom],
+                       options: Sequence[_Option]) -> str:
+        for option in options:
+            sub = option.name
+            if sub in self.original_operators:
```
tomsilver, line 91: sorry I'm confused by this code too, can you explain? 🙃

src/approaches/llm_predicate_renaming_approach.py
```diff
@@ -0,0 +1,99 @@
...
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self.original_predicates: List[str] = []
+        self.random_predicates: List[str] = []
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "llm_predicate_renaming"
+
+    def _generate_guess(self, sentence: str) -> List[str]:
```
tomsilver, line 58: why is this called `_generate_guess`?

src/approaches/llm_predicate_renaming_approach.py
```diff
@@ -0,0 +1,99 @@
...
+        alphabet = [
```
tomsilver, line 59: you can `import string` and then use `string.ascii_lowercase` instead

src/approaches/llm_predicate_renaming_approach.py
```diff
@@ -0,0 +1,99 @@
...
+  {goal_str}
+)
+Solution:
+  {options_str}"""
+        return prompt
+
+    def _replace_with_random_predicate_names(
+            self, original_set: Set[GroundAtom]) -> List[str]:
+        modified_set = []
+        for i in original_set:
```
tomsilver, line 86: all the variable names in this method can be more descriptive (original set of what? `i` is what?)

src/approaches/llm_predicate_renaming_approach.py
```diff
@@ -0,0 +1,99 @@
...
+(:goal
+  {goal_str}
+)
+Solution:
+  {options_str}"""
+        return prompt
+
+    def _replace_with_random_predicate_names(
+            self, original_set: Set[GroundAtom]) -> List[str]:
+        modified_set = []
```
tomsilver, line 85: if this is a set, why is it actually a list? :)

src/approaches/llm_predicate_renaming_approach.py
```diff
@@ -0,0 +1,99 @@
...
+    def _generate_guess(self, sentence: str) -> List[str]:
+        alphabet = [
+            'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
+            'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'
+        ]
+        return [self._rng.choice(alphabet) for i in range(len(sentence))]
+
+    def _create_prompt(self, init: Set[GroundAtom], goal: Set[GroundAtom],
+                       options: Sequence[_Option]) -> str:
+        init_arr = self._replace_with_random_predicate_names(init)
```
tomsilver, line 67: the reason that we use `sorted` here https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/approaches/llm_open_loop_approach.py#L214 is to assure that this prompt generation will be deterministic in terms of the order of the atoms. let's do the same thing here too, because otherwise there's no guarantee about the order when we iterate through the sets.

tomsilver (commented):
great work! don't be scared by the number of comments, this is definitely close to merging :)

src/approaches/syntax_modifications_approach.py
```diff
@@ -0,0 +1,80 @@
+"""Open-loop large language model (LLM) meta-controller approach.
+
+Example command line:
+    export OPENAI_API_KEY=<your API key>
+    python src/main.py --approach llm_open_loop --seed 0 \
```
tomsilver, line 5: update the approach here and below

src/approaches/syntax_modifications_approach.py
```diff
@@ -0,0 +1,80 @@
...
+
+class SyntaxModifications(LLMOpenLoopApproach):
+    """SyntaxModifications definition."""
+
+    def __init__(self, initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task]) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks)
+        self.modification_one = "".join(self._generate_guess(")"))
```
tomsilver, line 50: make these private

tomsilver (commented):
@Alchemy5 I took a look at the syntax modification approach, overall it looks good! since we're on a deadline, I'm not going to insist on the following, but one thing for you to consider: how could we change the implementation so that it would be easier to extend to other sets of characters that we want to replace in the future? (other than the three "(", ")",":" that we're using now)

tomsilver:
after discussion, closing in favor of #1159

PR #1147, opened 2022-07-26
New tests for pg4 config 

wester-is-not-real:
Now we'll test painting, but we will run box as the only goal in a separate test case. Same with shelf as the only goal.


tomsilver (approved):
LGTM

PR #1152, opened 2022-07-26
allow branch override in cluster launch files

tomsilver:
because it's currently annoying to test new config files before merging them (because the new config files will have the 'master' branch in them, but they're not yet on master themselves)

### DIFF
--- a/scripts/local/launch.py
+++ b/scripts/local/launch.py
@@ -18,8 +18,13 @@ def _main() -> None:
     # Set up argparse.
     parser = argparse.ArgumentParser()
     parser.add_argument("--config", required=True, type=str)
+    # Optionally override the branch specified in the CFG.
+    parser.add_argument("--branch", required=False, type=str)
     args = parser.parse_args()
-    branch = config_file_to_branch(args.config)
+    if args.branch:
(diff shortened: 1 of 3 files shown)


scripts/local/launch.py
```diff
@@ -18,8 +18,13 @@ def _main() -> None:
     # Set up argparse.
     parser = argparse.ArgumentParser()
     parser.add_argument("--config", required=True, type=str)
+    # Optionally override the branch specified in the CFG.
```
ronuchit, line 21: i would suggest we don't use "CFG" here, this is a different config file than settings.py right?

ronuchit (approved):
i'm wondering if it would make sense to just not have a "branch" in these config files and instead either (1) have it always default to master (without needing to type it in every config file) or (2) force the user to pass in a `--branch` into the command line? either one sounds good to me, but as it stands, i don't see why we'd ever *not* have "BRANCH: master" in the config, especially after this PR is merged

tomsilver:
okay good point, let's do the first option -- defaulting to master but allowing command line overrides

tomsilver:
gonna do this differently and also some other time

PR #1153, opened 2022-07-27
rename some config files

ronuchit:
for consistency with the other files in this folder... the folder is already called "config" so the file names don't need "config" in them


tomsilver (approved):
thx

PR #1155, opened 2022-07-27
change the way branches are handled in launches

### LINKED ISSUE #1154: change the way branches are specified in config files
see discussion in https://github.com/Learning-and-Intelligent-Systems/predicators/pull/1152

tomsilver:
closes #1154

### DIFF
--- a/scripts/configs/pg3_offline.yaml
+++ b/scripts/configs/pg3_offline.yaml
@@ -11,7 +11,6 @@ APPROACHES:
   plan-only:
     NAME: "oracle"
     FLAGS: {}
-    
 ENVS:
   delivery:
     NAME: "pddl_easy_delivery_procedural_tasks"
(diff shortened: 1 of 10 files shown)


tomsilver:
@ronuchit should we handle START_SEED and NUM_SEEDS in the same way? the advantage would be reducing copied entries in the config files

ronuchit:
sure yeah but it's also fine to leave as is, start seed and num seeds have less obvious defaults and sometimes people want to remember that they only want to run 3 or 5 seeds

scripts/configs/pg3_offline.yaml
```diff
@@ -23,6 +23,6 @@ ENVS:
 ARGS: []
 FLAGS:
   num_train_tasks: 50
-BRANCH: master
+
```
ronuchit, line 26: added a blank line here? but no other one? standardize

PR #1158, opened 2022-07-27
Fixed bug where PG4 would keep generating policy nodes and never explore the primitive nodes

wester-is-not-real:
Before, when PG4 would generate skeletons, policy-based successors would have a cost of 1. So, it wasn't guaranteed that PG4 would try the policy skeleton first. To fix this, those policy based successors were given a score of 0; however, whenever the policy-based skeleton would fail, the skeleton generator would just repeatedly make a policy-based successor and choose that one first. That new skeleton was guaranteed to fail, so it would just get stuck. 

To remedy this, I changed the policy-based successor scores back to 1, but now it will immediately yield the policy-based skeleton once it reaches a goal. 

You can try it with this command:
Run this on master and it'll get a 37/50 in test tasks
python3 src/main.py --env coffee --approach pg4 --seed 0 --strips_learner oracle --sampler_learner oracle --num_train_tasks 50
Run the same command with this change and it'll get a perfect score :)

Here are some of the benchmark results between bilevel planning, PG3 with low-level option search, and PG4:

Screen Shot 2022-08-01 at 11 01 59 AM


src/approaches/bilevel_planning_approach.py
```diff
@@ -89,7 +89,6 @@ def _run_sesame_plan(self, task: Task, nsrts: Set[NSRT],
             raise ApproachFailure(e.args[0], e.info)
         except PlanningTimeout as e:
             raise ApproachTimeout(e.args[0], e.info)
-
```
tomsilver: remove this change (you can do `git checkout origin/master src/approaches/bilevel_planning_approach.py`)

src/planning.py
```diff
@@ -306,6 +306,7 @@ def _skeleton_generator(
     """
 
     start_time = time.time()
+    first_yield = True
```
tomsilver, line 309: remove right?

src/planning.py
```diff
@@ -353,6 +354,7 @@ def _skeleton_generator(
                 current_node = node
                 for _ in range(sesame_max_policy_guided_rollout):
                     if task.goal.issubset(current_node.atoms):
+                        yield current_node.skeleton,current_node.atoms_sequence
```
tomsilver, line 357: don't forget to autoformat!

src/planning.py
```diff
@@ -381,7 +383,7 @@ def _skeleton_generator(
                         atoms_sequence=current_node.atoms_sequence +
                         [child_atoms],
                         parent=current_node,
-                        cumulative_cost=child_cost)
+                        cumulative_cost=(1+child_cost))
```
tomsilver, line 386: let's actually put the +1 in the definition of the child cost rather than here

tomsilver (commented):
beautiful PR message and great job thinking this through! let's wait to see the results before we merge

src/planning.py
```diff
@@ -353,6 +353,7 @@ def _skeleton_generator(
                 current_node = node
                 for _ in range(sesame_max_policy_guided_rollout):
                     if task.goal.issubset(current_node.atoms):
+                        yield current_node.skeleton, current_node.atoms_sequence
```
ronuchit, line 356: the PR description doesn't make any mention of the goal, so i'm confused why this only happens inside this if

src/planning.py
```diff
@@ -374,7 +375,7 @@ def _skeleton_generator(
                     # also allows us to yield a policy-generated plan without
```
ronuchit, line 375: this comment needs to be updated with this change right?

src/planning.py
wester-is-not-real: I thought I had mentioned it lol, my bad. Thanks!

tomsilver (approved):

PR #1159, opened 2022-07-27
LLM renaming approaches for ablation studies

tomsilver:
these approaches are meant to test whether the LLM is picking up on English and/or PDDL syntax, or just pattern matching.


src/approaches/llm_option_renaming_approach.py
```diff
@@ -0,0 +1,34 @@
...
+        --env pddl_blocks_procedural_tasks \
+        --num_train_tasks 3 \
+        --num_test_tasks 1 \
+        --debug
+"""
+import string
+from typing import Dict
+
+from predicators.src import utils
+from predicators.src.approaches.llm_renaming_base import \
```
Alchemy5, line 17: should be predicators.src.approaches.llm_renaming_base_approach, same for other files as well

src/approaches/llm_base_renaming_approach.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import Collection, Dict, List, Sequence, Set, Tuple
+
+from predicators.src.approaches.llm_open_loop_approach import \
+    LLMOpenLoopApproach
+from predicators.src.structs import Box, GroundAtom, Object, \
+    ParameterizedOption, Predicate, Task, Type, _Option
+
+
+class LLMBaseRenamingApproach(LLMOpenLoopApproach):
+    """F definition."""
```
ronuchit, line 19: ?

src/approaches/llm_predicate_renaming_approach.py
```diff
@@ -0,0 +1,44 @@
...
+class LLMPredicateRenamingApproach(LLMBaseRenamingApproach):
+    """LLMPredicateRenamingApproach definition."""
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "llm_predicate_renaming"
+
+    @property
+    def _renaming_prefixes(self) -> List[str]:
+        # Predicates start with either a new line or a white space.
```
ronuchit, line 30: what happens if they don't? e.g. here but also all the other uses of this function + the suffix one

src/approaches/llm_syntax_renaming_approach.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @property
+    def _renaming_prefixes(self) -> List[str]:
+        # Since we're replacing single characters, we don't need to worry about
+        # the possibility that one string is a substring of another.
+        return [""]
+
+    @property
+    def _renaming_suffixes(self) -> List[str]:
+        return [""]
```
ronuchit, line 37: copy the comment above here too

src/approaches/llm_syntax_renaming_approach.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def _create_replacements(self) -> Dict[str, str]:
+        # Without replacement!
```
ronuchit, line 40: instead, explain _why_ you are doing without replacement here

ronuchit (approved):

src/approaches/llm_base_renaming_approach.py
NishanthJKumar, line 19: I'm really confused by this docstring? Does "F" stand for something here?

NishanthJKumar (approved):
LGTM aside from minor docstring thing.

Alchemy5 (approved):

src/approaches/llm_base_renaming_approach.py
tomsilver: lol F the definition

src/approaches/llm_predicate_renaming_approach.py
tomsilver: then we wouldn't perform the replacement. for predicates, there's no risk of this because we're constructing the prompts ourselves. for the options, if they're not formatted in the way that we expect, we wouldn't parse them into options anyway, so replacement doesn't matter.

src/approaches/llm_base_renaming_approach.py
tomsilver: just a typo (no clue how I did that)

PR #1167, opened 2022-08-02
Pybullet

williamshen-nz:
Initial PR to start getting comments.

Need to:
- Clean up ikfast/utils.py
- add test for all the pybullet utils
- check why autoformat is failing
- Refactor to work better with Rachel's robot setup?


src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,82 @@
+"""Pybullet helper class for geometry utilities."""
+from typing import NamedTuple, Sequence
+
+import numpy as np
+import pybullet as p
+from pybullet_utils.transformations import euler_from_quaternion, \
+    quaternion_from_euler
+
+from predicators.src.structs import Array, Pose3D, Quaternion, RollPitchYaw
```
tomsilver, line 9: should we move `Pose3D`, `Quaternion`, `RollPitchYaw` into this file?

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,82 @@
...
+    def orientation(self) -> Quaternion:
+        """The default quaternion representation is xyzw as followed by
+        pybullet."""
+        return self.quat_xyzw
+
+    @property
+    def quat_wxyz(self) -> Quaternion:
+        """Get the wxyz quaternion representation."""
+        return (
+            self.quat_xyzw[3],
```
tomsilver, line 37: what's with this transposition?

src/pybullet_helpers/ikfast/utils.py
```diff
@@ -0,0 +1,316 @@
+# pylint: skip-file
```
tomsilver, line 1: is this temporary? or do you think we'll need to keep it?

src/pybullet_helpers/ikfast/utils.py
```diff
@@ -0,0 +1,316 @@
...
+from predicators.src.settings import CFG
+from predicators.src.structs import JointsState
+
+"""
+Note: modified from pybullet-planning for our use cases. Minimal changes to the actual
+logic have been made.
+
+https://github.com/caelan/pybullet-planning/blob/master/pybullet_tools/ikfast/ikfast.py
+"""
+
```
tomsilver, line 34: I see what you mean about this IKFast stuff being tricky. my hunch is that Caelan accumulated a lot of bells and whistles to his implementation that he needed for different purposes, and we might not need all of them for our purposes. do you think that is right? if so, is there a way that we could simplify the code here and revisit it later as needed?

tomsilver (commented):
@williamshen-nz beautiful!! thank you so much!! I didn't dig into any details yet, but at a high level, everything looks really good.

the biggest thing to discuss is how we should proceed toward getting this merged. ideally, we would split this up into many small pull requests. I'm happy to help with that, and to add unit tests for coverage as needed.

do you have thoughts about how to split this up? I can suggest some if not. let's try to make a game plan tomorrow in terms of what the small PRs will be and our timeline for getting them merged.

also let me know if your tutorial on the real robot led to any new thoughts about how we should structure things for easier integration with the real. but we can revisit this too later after we merge all this.

mypy.ini
```diff
@@ -16,6 +16,10 @@ ignore_missing_imports = True
 [mypy-predicators.third_party.*]
 ignore_missing_imports = True
 
+# temporary as this module is a big mess
```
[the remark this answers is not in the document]
williamshen-nz: Done

src/envs/pybullet_blocks.py
```diff
@@ -31,8 +35,13 @@ class PyBulletBlocksEnv(PyBulletEnv, BlocksEnv):
...
 
     # Robot parameters.
-    _ee_orn: ClassVar[Sequence[float]] = p.getQuaternionFromEuler(
-        [0.0, np.pi / 2, -np.pi])
+    _ee_orn: ClassVar[Dict[str, Quaternion]] = {
+        # Fetch gripper down, since its thin we don't need to rotate 90 degrees.
+        "fetch": quaternion_from_euler(0.0, np.pi / 2, -np.pi),
+        # Panda gripper down and rotated 90 degrees as it's big and can cause
+        # collisions.
+        "panda": quaternion_from_euler(np.pi, 0, np.pi / 2)
```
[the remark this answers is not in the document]
williamshen-nz: No, using the same rotation with the fetch results in side rather than top down end-effector orientation

src/envs/pybullet_blocks.py
```diff
@@ -309,9 +318,18 @@ def _get_object_ids_for_held_check(self) -> List[int]:
         return sorted(self._block_id_to_block)
 
     def _get_expected_finger_normals(self) -> Dict[int, Array]:
+        if CFG.pybullet_robot == "panda":
+            # gripper rotated 90deg so parallel to x-axis
+            normal = np.array([1., 0., 0.], dtype=np.float32)
+        elif CFG.pybullet_robot == "fetch":
+            # gripper parallel to y-axis
```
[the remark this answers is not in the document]
williamshen-nz: Good idea, will give it a go. On first try IK is doing some funny stuff on the fetch

src/pybullet_helpers/controllers.py
```diff
@@ -0,0 +1,148 @@
+"""Generic controllers for the robots."""
```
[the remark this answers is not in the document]
williamshen-nz: Yep! I can breakdown the PR into smaller ones later on. I don't think I changed anything here apart from stuff to keep mypy/pylint happy

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,82 @@
...
+
+    We use a NamedTuple as it supports retrieving by integer indexing
+    and most closely follows the pybullet API.
+    """
+
+    position: Pose3D
+    quat_xyzw: Quaternion = (0.0, 0.0, 0.0, 1.0)
+
+    @classmethod
+    def from_rpy(cls, translation: Pose3D, rpy: RollPitchYaw) -> "Pose":
```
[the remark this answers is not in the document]
williamshen-nz: Good point - have added

src/pybullet_helpers/geometry.py
williamshen-nz: Just a different quaternion ordering. The default in pybullet is xyzw so I think we can remove this for now and re-add it when needed.

src/pybullet_helpers/ikfast/utils.py
williamshen-nz: Temporary so I could get the unit tests to run by telling pylint not to complain about this module

src/pybullet_helpers/ikfast/utils.py
williamshen-nz: Yep - I think some of the bells and whistles we do need - e.g. returning solution with minimum distance from current configuration but other stuff like the generators we could simplify a bit. 

So all the methods from `def get_length` to `def uniform_generator(d):` could be simplified.

williamshen-nz:
> @williamshen-nz beautiful!! thank you so much!! I didn't dig into any details yet, but at a high level, everything looks really good.
> 
> the biggest thing to discuss is how we should proceed toward getting this merged. ideally, we would split this up into many small pull requests. I'm happy to help with that, and to add unit tests for coverage as needed.
> 
> do you have thoughts about how to split this up? I can suggest some if not. let's try to make a game plan tomorrow in terms of what the small PRs will be and our timeline for getting them merged.

I think the easiest way is to start some PRs to split up this module which contains most of the existing pybullet helper logic on the main branch: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/src/envs/pybullet_robots.py

I can start working on that bit.

Then add those we can add our new functionality in, though they might be bigger PRs as things are interdependent between modules. I'll figure out best way to do this to minimize the diff.

> also let me know if your tutorial on the real robot led to any new thoughts about how we should structure things for easier integration with the real. but we can revisit this too later after we merge all this.

Looking through Rachel's Panda interface I think the only major thing is the gripper open and closing as we've discussed before. Apart from that we will either need to execute on the robot in real time (similar to simulating in pybullet step by step) or cumulating the entire plan and then executing that on the real robot in the end.

The latter probably would work better but may need refactoring depending on how we want to change things.

williamshen-nz:
This is an old leftover PR which was for the 1st iteration of Pybullet things.

This is the current PR to track Panda stuff in Pybullet: https://github.com/Learning-and-Intelligent-Systems/predicators/pull/1184

PR #1169, opened 2022-08-05
Pybullet Merging Part 1

williamshen-nz:
First PR to move single arm robot out to its own module.

I used Pycharm's move functionality so we can trust no code has been changed :)

Plan after this:
1. Move the other methods out of `src/env/pybullet_robots.py` 
2. Move the robots to the respective modules.
3. Apply my patches on top of everything iteratively


tomsilver (approved):
looks great! thanks for using an automated move tool so we can be sure that nothing changed :) at some point, we will want to update the directory structure in `tests/` to match the changes in directory structure in `src/` that are happening here. if you think it makes sense to do that as part of this PR, great, otherwise I'm fine to wait.

PR #1170, opened 2022-08-06
Pybullet Controllers + IK

williamshen-nz:
Move Pybullet Controllers + IK using the IDE.

Didn't move the motion planning and getting kinematic chain to keep the diff relatively small. So that will be the next PR and will be pretty quick to do.

Will fix up tests and moving into their respective places when patching in my commits from the `panda-willshen-cleanup` branch.


tomsilver (approved):
LGTM!

PR #1174, opened 2022-08-13
Add additional Pybullet utilities

williamshen-nz:
- Refactor Single Arm Robot to make it easier to extend to new robots
- Add new Pybullet functionality and data types
- Fix existing environments

```
FAIL Required test coverage of 100% not reached. Total coverage: 99.89%

Name                                        Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------
src/pybullet_helpers/geometry.py               40     14    65%   31, 37, 42, 51, 55-56, 61-66, 71, 76-78
src/pybullet_helpers/joint.py                  49      2    96%   33, 48
src/pybullet_helpers/link.py                   23      2    91%   23, 40
src/pybullet_helpers/robots/fetch.py           27      1    96%   24
src/pybullet_helpers/robots/single_arm.py     122      8    93%   326-339, 366
-------------------------------------------------------------------------
TOTAL                                       25540     27    99%

```

Will need to up coverage as some methods aren't used yet. Could add unit tests for them.


src/settings.py
```diff
@@ -122,7 +122,7 @@ class GlobalSettings:
     behavior_option_model_rrt = False
 
     # general pybullet parameters
-    pybullet_use_gui = False  # must be True to make videos
+    pybullet_use_gui = True  # must be True to make videos
```
williamshen-nz, line 125: Revert this line

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,78 @@
...
+    """Multiplies poses (which are essentially transforms) together."""
+    pose = poses[0]
+    for next_pose in poses[1:]:
+        pose = p.multiplyTransforms(pose[0], pose[1], next_pose[0],
+                                    next_pose[1])
+        pose = Pose(pose[0], pose[1])
+    return pose
+
+
+def matrix_from_quat(quat: Sequence[float]) -> Array:
```
williamshen-nz, line 69: Typing -> `Quaternion`

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+        get_joint_info(body, joint_id, physics_client_id)
+        for joint_id in joints
+    ]
+
+
+def get_joint_limits(
+        body: int, joints: List[int],
+        physics_client_id: int) -> Tuple[List[float], List[float]]:
+    """Get the joint limits for the given joints for a body. Circular joints do
+    not have limits (represented by ±np.pi).
```
williamshen-nz, line 76: Comment should be np.inf

src/pybullet_helpers/link.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def get_link_state(body: int, link: int, physics_client_id: int) -> LinkState:
+    """Get the state of a link in a given body."""
+    link_state = p.getLinkState(body, link, physicsClientId=physics_client_id)
+    return LinkState(*link_state)
+
+
+def get_link_pose(body: int, link: int, physics_client_id: int) -> Pose:
+    """Get the position and orientation for a link."""
```
williamshen-nz, line 38: Comment: Get the pose

src/pybullet_helpers/robots/__init__.py
```diff
@@ -1,16 +1,25 @@
...
+from predicators.src.pybullet_helpers.robots.fetch import FetchPyBulletRobot
 from predicators.src.pybullet_helpers.robots.single_arm import \
-    FetchPyBulletRobot, SingleArmPyBulletRobot
-from predicators.src.structs import Pose3D
+    SingleArmPyBulletRobot
+
+# Note: these are static base poses which suffice for the current environments.
+_ROBOT_TO_BASE_POSE: Dict[str, Pose] = {
+    "fetch": Pose(position=(0.75, 0.7441, 0.0)),
+}
```
williamshen-nz, line 12: This should be abstracted out eventually...

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,78 @@
...
+
+class Pose(NamedTuple):
+    """Pose which is a position (translation) and rotation.
+
+    We use a NamedTuple as it supports retrieving by integer indexing
+    and most closely follows the pybullet API.
+    """
+
+    position: Pose3D
+    quat_xyzw: Quaternion = (0.0, 0.0, 0.0, 1.0)
```
tomsilver, line 26: thoughts about just making this attribute called `orientation` and removing the orientation property?

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,78 @@
...
+        """Invert the pose (i.e., transform)."""
+        pos, quat = p.invertTransform(self.position, self.quat_xyzw)
+        return Pose(pos, quat)
+
+
+def multiply_poses(*poses: Pose) -> Pose:
+    """Multiplies poses (which are essentially transforms) together."""
+    pose = poses[0]
+    for next_pose in poses[1:]:
+        pose = p.multiplyTransforms(pose[0], pose[1], next_pose[0],
```
tomsilver, line 63: it's a little confusing that this `pose` is not a `Pose`, unlike the `pose` on L61 and L65 -- rename?

tomsilver, line 63: let's use `pose.position` and `pose.orientation` by default? (instead of indexing)

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import List, NamedTuple, Sequence, Tuple
+
+import numpy as np
+import pybullet as p
+
+from predicators.src.pybullet_helpers.geometry import Pose3D, Quaternion
+
+
+class JointInfo(NamedTuple):
+    """Joint Information from Pybullet."""
```
tomsilver, line 11: add the same rationale here for using NamedTuple

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+    jointIndex: int
```
tomsilver, line 12: is the reason for using camelCase here to match the names of pybullet attributes? if so, that's fine

tomsilver: but also let's point out the closest pybullet function(s) that these variable names are meant to match, for reference (`p.getJointInfo` probably)

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+    jointUpperLimit: float
+    jointMaxForce: float
+    jointMaxVelocity: float
+    linkName: str
+    jointAxis: Sequence[float]
+    parentFramePos: Pose3D
+    parentFrameOrn: Quaternion
+    parentIndex: int
+
+    def is_circular(self) -> bool:
```
tomsilver, line 30: any thoughts about making these properties instead?

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+        p.getJointInfo(body, joint, physicsClientId=physics_client_id))
+    # Decode the byte strings for joint name and link name
+    raw_joint_info[1] = raw_joint_info[1].decode("UTF-8")
+    raw_joint_info[12] = raw_joint_info[12].decode("UTF-8")
+
+    joint_info = JointInfo(*raw_joint_info)
+    return joint_info
+
+
+def get_joint_infos(body: int, joints: List[int],
```
tomsilver, line 63: since we've previously used `JointsState` instead of `JointStates`, should we call this `joints_info` instead of `joint_infos`?

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+                    physics_client_id: int) -> List[JointInfo]:
+    """Get the infos the given joints for a body."""
```
tomsilver, line 65: missing word

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,103 @@
...
+    ]
+
+
+def get_joint_limits(
+        body: int, joints: List[int],
+        physics_client_id: int) -> Tuple[List[float], List[float]]:
+    """Get the joint limits for the given joints for a body. Circular joints do
+    not have limits (represented by ±np.pi).
+
+    Returns
```
tomsilver, line 78: this comment format is probably left over from some very old code of ours -- you can just write what it returns in english.

src/pybullet_helpers/joint.py
tomsilver: similarly for the other functions below about joint limits, which makes even more sense to convert because "joint limits" sounds like the limits for a single joint

src/pybullet_helpers/link.py
```diff
@@ -0,0 +1,43 @@
...
+import pybullet as p
+
+from predicators.src.pybullet_helpers.geometry import Pose, Pose3D, \
+    Quaternion, get_pose
+
+BASE_LINK: int = -1
+
+
+class LinkState(NamedTuple):
+    """Link state from Pybullet."""
```
tomsilver, line 13: similarly to before, add the rationale for NamedTuple and mention the pybullet function that this mimics

src/pybullet_helpers/link.py
```diff
@@ -0,0 +1,43 @@
...
+    linkWorldPosition: Pose3D
+    linkWorldOrientation: Quaternion
+    localInertialFramePosition: Pose3D
+    localInertialFrameOrientation: Quaternion
+    worldLinkFramePosition: Pose3D
+    worldLinkFrameOrientation: Quaternion
+
+    def com_pose(self) -> Pose:
```
tomsilver, line 21: property? maybe a functools.cached_property?

src/pybullet_helpers/link.py
```diff
@@ -0,0 +1,43 @@
...
+        """Center of mass (COM) pose of link."""
+        return Pose(self.linkWorldPosition, self.linkWorldOrientation)
+
+    def pose(self) -> Pose:
```
tomsilver, line 25: ditto

src/pybullet_helpers/link.py
```diff
@@ -0,0 +1,43 @@
...
+    return LinkState(*link_state)
+
+
+def get_link_pose(body: int, link: int, physics_client_id: int) -> Pose:
+    """Get the position and orientation for a link."""
+    if link == BASE_LINK:
+        return get_pose(body, physics_client_id)
+
+    link_state = get_link_state(body, link, physics_client_id)
+    return link_state.pose()
```
tomsilver, line 43: love this nice clean structure overall!

src/pybullet_helpers/robots/__init__.py
```diff
@@ -1,16 +1,25 @@
...
 
 def create_single_arm_pybullet_robot(
-        robot_name: str, ee_home_pose: Pose3D, ee_orientation: Sequence[float],
+        robot_name: str, ee_home_pose: Pose3D, ee_orientation: Quaternion,
         physics_client_id: int) -> SingleArmPyBulletRobot:
     """Create a single-arm PyBullet robot."""
     if robot_name == "fetch":
-        return FetchPyBulletRobot(ee_home_pose, ee_orientation,
-                                  physics_client_id)
+        base_pose = _ROBOT_TO_BASE_POSE.get(robot_name, Pose.identity())
```
tomsilver, line 20: do we really want to have a default here? shouldn't all robots specify a base pose?

src/pybullet_helpers/robots/fetch.py
```diff
@@ -0,0 +1,40 @@
+"""Fetch Robotics Mobile Manipulator (Fetch)"""
```
tomsilver, line 1: add period

src/pybullet_helpers/robots/fetch.py
```diff
@@ -0,0 +1,40 @@
...
+from predicators.src import utils
+from predicators.src.pybullet_helpers.robots.single_arm import \
+    SingleArmPyBulletRobot
+
+
+class FetchPyBulletRobot(SingleArmPyBulletRobot):
```
tomsilver, line 7: can you make sure that we're consistent about `PyBullet` vs `Pybullet`? (find and replace `Pybullet` with `PyBullet` throughout the code base)

here's my bash function for find and replace in case it helps

```
# find and replace for git
git-replace(){
    git grep -l "$1" | xargs sed -i '' -e "s/$1/$2/g"
}
```

so `git-replace Pybullet PyBullet`

src/pybullet_helpers/robots/single_arm.py
```diff
@@ -1,107 +1,179 @@
...
+            physicsClientId=self.physics_client_id,
+        )
 
-        Represents position control of the arm and finger joints.
-        """
-        return Box(np.array(self.joint_lower_limits, dtype=np.float32),
-                   np.array(self.joint_upper_limits, dtype=np.float32),
-                   dtype=np.float32)
+        # Robot initially at home pose
+        self.go_home()
```
tomsilver, line 53: image

src/pybullet_helpers/robots/single_arm.py
```diff
@@ -1,107 +1,179 @@
...
+        )
+        # NOTE: pybullet tools assumes sorted arm joints.
+        joint_ids = sorted(joint_ids)
+        joint_ids.extend([self.left_finger_id, self.right_finger_id])
+        return joint_ids
+
+    @cached_property
+    def joint_infos(self) -> List[JointInfo]:
+        """Get the joint info for each joint of the robot."""
+        joint_ids = list(
```
tomsilver, line 115: use `get_joints` here?

src/pybullet_helpers/robots/single_arm.py
```diff
@@ -115,29 +187,76 @@ def closed_fingers(self) -> float:
...
+                              physicsClientId=self.physics_client_id)
 
-    @abc.abstractmethod
     def get_state(self) -> Array:
         """Get the robot state vector based on the current PyBullet state.
 
         This corresponds to the State vector for the robot object.
         """
-        raise NotImplementedError("Override me!")
+        ee_link_state = p.getLinkState(self.robot_id,
```
tomsilver, line 235: use the new utilities here?

src/pybullet_helpers/robots/single_arm.py
```diff
@@ -115,29 +187,76 @@ def closed_fingers(self) -> float:
...
+            physicsClientId=self.physics_client_id,
+        )[0]
+        # pose_x, pose_y, pose_z, fingers
+        return np.array([rx, ry, rz, rf], dtype=np.float32)
 
-    @abc.abstractmethod
     def get_joints(self) -> JointsState:
         """Get the joints state from the current PyBullet state."""
-        raise NotImplementedError("Override me!")
+        joints_state = [
```
tomsilver, line 251: use the new utilities here?

tomsilver (approved):
beautiful work!!

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,78 @@
...
+        pybullet."""
+        return self.quat_xyzw
+
+    @property
+    def rpy(self) -> RollPitchYaw:
+        """Get the Euler roll-pitch-yaw representation."""
+        return euler_from_quaternion(self.quat_xyzw)
+
+    @classmethod
+    def identity(cls) -> "Pose":
```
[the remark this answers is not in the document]
williamshen-nz: I will update - just need to from future import annotations

src/pybullet_helpers/geometry.py
williamshen-nz: We can do that, I just used indexing to keep it backwards compatible with the poses that come straight out of pybullet which are tuples.

williamshen-nz: Good point - will rename!

src/pybullet_helpers/geometry.py
williamshen-nz: I think that makes sense since it's what pybullet returns as well.

It may only get confusing when we use other simulators as well which use rpy, or wxyz quaternion representation. But let's worry about that when it happens to keep things cleaner pybullet side.

src/pybullet_helpers/joint.py
williamshen-nz: Yes to match pybullet, will update docstring

src/pybullet_helpers/robots/__init__.py
williamshen-nz: Yes at least for now let's constrain it to a specified one. Will update

src/pybullet_helpers/joint.py
williamshen-nz: I did consider it but can't remember why I didn't use it. I think it was so we didn't conflate between the fields from pybullet (in camelcase) and our custom functionality on top of that.

I'm fine with either way

src/pybullet_helpers/joint.py
williamshen-nz: I'm going to look through the codebase for these plural/singular words and get back to you.

Personally I prefer having stuff like JointStates, for some reason JointsState feels weird.

src/pybullet_helpers/link.py
williamshen-nz: Same comment as above for `is_circular` in JointInfo - I didn't want to confuse between pybullet data properties and functionality on top. But either way works

src/pybullet_helpers/robots/single_arm.py
williamshen-nz: There was some weird infinite recursion when I was trying to improve this late at night. Will figure out a better way

src/pybullet_helpers/robots/single_arm.py
williamshen-nz: We don't have utility for get joint states yet. Will come in future commits

src/pybullet_helpers/joint.py
williamshen-nz: Ok Pycharm will indicate these distinctly so I will turn into properties

src/pybullet_helpers/joint.py
williamshen-nz: I'll advocate for using:

- JointSingular when it refers to a single joint
- JointPlural when it refers to multiple joints

This is the convention at least when describing it in a textbook - e.g. joint positions, joint angles, joint limits

tomsilver: okay sounds good to me! i agree, jointsstate is kinda weird

williamshen-nz: Migrated to `JointPositions`

JointState (to match the pybullet definition) should include positions, velocities and forces which we don't use.

williamshen-nz:
More commits to come...

Final thing to clean up `def joint_infos` in `single_arm.py`

Also variable naming in unit tests to `JointPositions` instead of `JointsState`

src/envs/pybullet_env.py
```diff
@@ -371,16 +373,17 @@ def _fingers_opening(self, action: Action) -> bool:
...
 
-    def _get_finger_state(self, state: State) -> float:
+    def _get_finger_position(self, state: State) -> float:
         # Arbitrarily use the left finger as reference.
         state = cast(utils.PyBulletState, state)
-        joint_idx = self._pybullet_robot.left_finger_joint_idx
-        return state.joints_state[joint_idx]
+        joint_positions = state.joint_positions
+        finger_joint_idx = self._pybullet_robot.left_finger_joint_idx
+        return joint_positions[finger_joint_idx]
```
williamshen-nz, line 381: Can remove line 379

src/pybullet_helpers/controllers.py
```diff
@@ -61,13 +61,13 @@ def _policy(state: State, memory: Dict, objects: Sequence[Object],
             finger_delta = -finger_action_nudge_magnitude
         # Extract the current finger state.
         state = cast(utils.PyBulletState, state)
-        finger_state = state.joints_state[robot.left_finger_joint_idx]
+        finger_positions = state.joint_positions[robot.left_finger_joint_idx]
```
williamshen-nz, line 64: Use singular as its just a float

src/pybullet_helpers/link.py
```diff
@@ -10,34 +10,49 @@
...
 
-def get_link_state(body: int, link: int, physics_client_id: int) -> LinkState:
+def get_link_state(body: int,
+                   link: int,
+                   physics_client_id: int,
+                   compute_forward_kinematics: bool = True) -> LinkState:
     """Get the state of a link in a given body."""
-    link_state = p.getLinkState(body, link, physicsClientId=physics_client_id)
+    # Note: does computeForwardKinematics even do anything? Caelan and Rachel
+    # use it with False
```
williamshen-nz, line 43: Need to look into this

williamshen-nz (commented):
@tomsilver ready for 2nd round of review! I left a few comments myself as well.

williamshen-nz:
Here's the coverage info

```
---------- coverage: platform darwin, python 3.9.11-final-0 ----------
Name                                        Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------
src/pybullet_helpers/geometry.py               38     13    66%   33, 38, 47, 51-52, 57-63, 68, 73-75
src/pybullet_helpers/joint.py                  63      2    97%   43, 49
src/pybullet_helpers/link.py                   25      1    96%   55
src/pybullet_helpers/robots/__init__.py        12      1    92%   21
src/pybullet_helpers/robots/fetch.py           27      1    96%   24
src/pybullet_helpers/robots/single_arm.py     124      8    94%   330-343, 370
-------------------------------------------------------------------------
TOTAL                                       25549     26    99%

161 files skipped due to complete coverage.

FAIL Required test coverage of 100% not reached. Total coverage: 99.90%
```

src/pybullet_helpers/geometry.py
```diff
@@ -0,0 +1,75 @@
...
+    position: Pose3D
+    # Quaternion in (x, y, z, w) representation
+    orientation: Quaternion = (0.0, 0.0, 0.0, 1.0)
+
+    @classmethod
+    def from_rpy(cls, translation: Pose3D, rpy: RollPitchYaw) -> Pose:
+        """Create a Pose from translation and Euler roll-pitch-yaw angles."""
+        return cls(translation, quaternion_from_euler(*rpy))
+
+    @cached_property
```
tomsilver, line 35: it turns out that `cached_property` actually throws an error here, apparently because this class inherits from `NamedTuple`. the error is
```
TypeError: Cannot use cached_property instance without calling __set_name__ on it.
```

I'll revert to just `@property` for now, but if there are cases where cached properties are important, we may want to consider using dataclasses instead of named tuples (and implementing indexing ourselves if needed).

src/pybullet_helpers/joint.py
```diff
@@ -0,0 +1,133 @@
...
+    parentFramePos: Pose3D
+    parentFrameOrn: Quaternion
+    parentIndex: int
+
+    @property
+    def is_circular(self) -> bool:
+        """Whether the joint is circular or not."""
+        if self.is_fixed:
+            return False
+        return self.jointUpperLimit < self.jointLowerLimit
```
tomsilver, line 44: I know this is from Caelan -- do you understand the logic? (I don't!) Can you add a comment if so?

src/pybullet_helpers/link.py
tomsilver: I'm not sure if we ever verified that it does anything, but I believe the reason that we added it was so that we could reset joint states and then get the corresponding end effector poses during motion planning, _without_ calling stepSimulation().

src/pybullet_helpers/robots/single_arm.py
```diff
@@ -115,259 +192,190 @@ def closed_fingers(self) -> float:
...
+            self.robot_id,
+            self.end_effector_id,
+            physics_client_id=self.physics_client_id)
+        position = ee_link_state.worldLinkFramePosition
+        return position
 
-        If validate is True, guarantee that the returned joints state
-        would result in end_effector_pose if run through
-        forward_kinematics.
+    def _validate_joint_positions(self, joint_positions: JointPositions,
```
tomsilver, line 322: It looks like this method isn't used anywhere right now -- should we just remove it until we need it (if ever)?

tomsilver (approved):
looking good to me! I added tests -- the only uncovered thing now should be the validate method that I think we may remove (see comments above)

tomsilver:
before or shortly after we merge this, we should compare performance with our "nightly" script (https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/scripts/configs/nightly.yaml) to make sure there are no regressions. I'll launch it now so we have the latest results from "before".

williamshen-nz:
Thanks! I'll get updated version out by end of tomorrow

src/pybullet_helpers/joint.py
williamshen-nz: I think it's just the way Pybullet does it. Here's a continuous (circular) joint for the PR2 caster wheels: https://github.com/caelan/pybullet-planning/blob/master/models/pr2_description/pr2.urdf#L307-L317

Note you aren't supposed to specify joint limits in the URDF for continuous joints.

I debugged and on the getJointInfo and we get:

```
fl_caster_rotation_joint
lower limit: 0.0
upper limit: -1.0
```

It's probably some detail buried in the URDF parser that handles this, I'll add a comment about it

src/pybullet_helpers/geometry.py
williamshen-nz: Good point! I think this method is pretty cheap as well so we can keep as is for now.

src/pybullet_helpers/robots/single_arm.py
williamshen-nz: It'll come handy once we integrate IKFast as we need to validate the IK solutions.

tomsilver: makes sense -- can we remove it for the sake of merging this PR then, since it will probably be easier to test in a future PR where we add IKFast?

williamshen-nz: Sure - will do that

src/pybullet_helpers/link.py
williamshen-nz: So I just tried a little script to test things out, and it seems like it doesn't make any difference.

https://gist.github.com/williamshen-nz/9f287f23486d124ce664ba5b35edcc77

The only way to really tell is to probably dig into the Bullet source code, but I think that's more effort than it's worth.

Given that Caelan and Rachel don't use it and the PyBullet default is False, should we just set the default to False and add a comment that the behavior with computeForwardKinematics=True doesn't seem to change anything?

tomsilver: interesting, thanks for testing that! I agree it's not worth digging into Bullet.

why don't we add that comment, but just leave out the keyword argument altogether if it doesn't seem to make a difference?

we should also make a mental note to check this if we see any regression in the large-scale runs that we'll do after merging

williamshen-nz:
Just pushed an update - will rebase with master before merging but think we are good to go!

tomsilver:
@williamshen-nz it looks like there may be a small regression in pybullet-cover

before:
```
EXPERIMENT_ID                                                                                                               
pybullet_blocks-oracle    50.00 (0.00)  49.80 (0.40)   5.00 (0.00)   0.42 (0.13)  4259.69 (2224.80)   0.00 (0.00)         10
pybullet_cover-oracle     50.00 (0.00)  47.20 (1.40)   5.00 (0.00)   0.08 (0.01)        6.66 (0.32)   0.00 (0.00)         10
```

after:
EXPERIMENT_ID                                                                                                               
pybullet_blocks-oracle    50.00 (0.00)  49.80 (0.40)   5.00 (0.00)   0.43 (0.12)  4259.69 (2224.80)   0.00 (0.00)         10
pybullet_cover-oracle     50.00 (0.00)  45.90 (1.81)   5.00 (0.00)   0.09 (0.01)        6.66 (0.32)   0.00 (0.00)         10
```

we could check the computeForwardKinematics thing -- is there anything else that might have led to a functional change?

**Comment by williamshen-nz** [comment]
Interesting - I'll look into it. I would only think either the forward kinematics or stochasticity of IK would affect it.

I'll run the nightly script again to check if its there's a consistent decrease

**Comment by tomsilver** [comment]
with `computeForwardKinematics=True` by default:
```
AGGREGATED DATA OVER SEEDS:
EXPERIMENT_ID                                                                                                               
cover-nsrt_learning       50.00 (0.00)  49.50 (0.50)   5.00 (0.00)   0.06 (0.02)        6.66 (0.36)  21.10 (1.17)         10
cover-oracle              50.00 (0.00)  50.00 (0.00)   5.00 (0.00)   0.04 (0.01)        6.66 (0.32)   0.00 (0.00)         10
painting-oracle           50.00 (0.00)  50.00 (0.00)  15.00 (0.01)   0.39 (0.09)   2301.79 (738.99)   0.00 (0.00)         10
pybullet_blocks-oracle    50.00 (0.00)  47.30 (1.10)   5.00 (0.00)   0.39 (0.16)  4259.69 (2224.80)   0.00 (0.00)         10
pybullet_cover-oracle     50.00 (0.00)  47.00 (1.79)   5.00 (0.00)   0.09 (0.01)        6.66 (0.32)   0.00 (0.00)         10
```

**Comment by tomsilver** [comment]
@williamshen-nz should we make `computeForwardKinematics` True by default then? or do you think there's a reason to make it an optional kwarg and default to False, but set it to True sometimes (and if so, when)?

**Comment by williamshen-nz** [comment]
Over 100 seeds before our commit

```
304bb167d10f0afb673d94e590eeb06b8f5a7c55


AGGREGATED DATA OVER SEEDS:
EXPERIMENT_ID                                                                                                               
pybullet_blocks-oracle    50.00 (0.00)  49.55 (0.68)   5.00 (0.00)   0.37 (0.10)  3530.42 (1867.35)   0.00 (0.00)        100
pybullet_cover-oracle     50.00 (0.00)  46.43 (1.86)   5.00 (0.00)   0.08 (0.01)        6.69 (0.40)   0.00 (0.00)        100

```

**Comment by williamshen-nz** [comment]
```
3747608e1252cf0ca022c3f991b4e7680e07676c


AGGREGATED DATA OVER SEEDS:
EXPERIMENT_ID
pybullet_blocks-oracle    50.00 (0.00)  49.63 (0.58)   5.00 (0.00)   0.39 (0.10)  3530.42 (1867.35)   0.00 (0.00)        100
pybullet_cover-oracle     50.00 (0.00)  45.56 (2.01)   5.00 (0.00)   0.09 (0.01)        6.69 (0.40)   0.00 (0.00)        100
```

Current commit with computeFK = False

**Comment by williamshen-nz** [comment]
```
d2cf8aad97058e2f31b0b43454b2979948d709e9


AGGREGATED DATA OVER SEEDS:
EXPERIMENT_ID
pybullet_blocks-oracle    50.00 (0.00)  47.18 (1.61)   5.00 (0.00)   0.36 (0.10)  3530.42 (1867.35)   0.00 (0.00)        100
pybullet_cover-oracle     50.00 (0.00)  45.55 (2.03)   5.00 (0.00)   0.09 (0.01)        6.69 (0.40)   0.00 (0.00)        100
```

Updated local commit with computeFK=True

@tomsilver what do you think? The behavior isn't very consistent. I'll look through the code to see if there may be any other breaking changes.

tomsilver:
thanks very much for running these experiments! yeah, looks like computeFK doesn't explain the drop in cover after all, and it does seem to consistently hurt blocks to use computeFK=True. so yeah let me know if you see any other possible changes that could affect cover.

PR #1178, opened 2022-08-19
fix pddl domain creation with hierarchical types 

### LINKED ISSUE #1173: fix pddl domain string generation in case where there are hierarchical types
for example, the spanner domain is improperly constructed right now

tomsilver:
closes #1173


NishanthJKumar (approved):
LGTM! Glad we can finally write out PDDL files with type-inheritance!

PR #1185, opened 2022-08-25
Data re-ordering and operator pruning

wmcclinton:
This PR is made to help solve the weird cases that come up due to our backchaining approach not being data invariant. We do this by (1) re-ordering the dataset during nsrt_learning (settings.py now has parameter data_orderings_to_search = 1 (default), which represents how many orderings we will search over) to find the pnads set with the minimum complexity and (2) pruning harmless operators by increasing min_perc_data_for_nsrts incrementally until we find the smallest harmless operator set.


predicators/nsrt_learning/nsrt_learning_main.py
```diff
@@ -37,42 +39,68 @@ def learn_nsrts_from_data(
...
     logging.info(f"\nLearning NSRTs on {len(trajectories)} trajectories...")
 
-    # STEP 1: Segment each trajectory in the dataset based on changes in
-    #         either predicates or options. If we are doing option learning,
-    #         then the data will not contain options, so this segmenting
-    #         procedure only uses the predicates.
-    segmented_trajs = [
-        segment_trajectory(traj) for traj in ground_atom_dataset
-    ]
+    # Search over data orderings to find smallest PNAD set.
```
NishanthJKumar, line 42: change "smallest" to "least complex"

predicators/nsrt_learning/nsrt_learning_main.py
```diff
@@ -37,42 +39,68 @@ def learn_nsrts_from_data(
...
+    # If the strips learner is not Backchaining then it will
+    # only do one iteration, because all other appraoches are
+    # data invariant.
```
NishanthJKumar, line 45: change "data" to "data order"

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -288,7 +289,10 @@ def test_backchaining_strips_learner_order_dependence():
     # Weird Case: This case shows that our algorithm is not data invariant!
```
NishanthJKumar, line 289: change this comment to say "data order invariant!"

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -551,10 +570,28 @@ def test_backchaining_strips_learner_order_dependence():
...
+                                                      action_space,
+                                                      ground_atom_dataset,
+                                                      sampler_learner="random")
 
     # First, check that the two sets of PNADs have the same number of PNADs.
     # Uh oh, they don't
     assert len(natural_order_pnads) != len(reverse_order_pnads)
+    # Second, check that the two sets of NSRTs have the same number of NSRTs.
+    # They do! Because of dataset reordering and harmless operator pruning.
+    assert len(natural_order_nsrts) == len(reverse_order_nsrts)
```
NishanthJKumar, line 594: wait - I'm a little confused by this: why do the pnad's have different lengths, but the NSRT's work? Could you augment this comment to explain more about what exactly is going on here?

NishanthJKumar (commented):
LGTM after addressing comments; great work! I'll let @tomsilver take the next round of review and provide actual approval.

predicators/utils.py
```diff
@@ -166,6 +166,24 @@ def segment_trajectory_to_atoms_sequence(
     return atoms_seq
 
 
+def order_data(
```
tomsilver, line 169: can we name this something more specific?

predicators/utils.py
```diff
@@ -166,6 +166,24 @@ def segment_trajectory_to_atoms_sequence(
...
+    trajectories: List[LowLevelTrajectory],
+    ground_atom_dataset: List[GroundAtomTrajectory], rng: np.random.Generator
+) -> Tuple[List[LowLevelTrajectory], List[GroundAtomTrajectory]]:
+    """Re-orders a list of trajectories and matching list of ground atoms,
+    using the np.random.Generator provided.
+
+    This function returns another list of trajectories and matching list
+    of ground atoms in a new order.
+    """
+    rnd_seed = int(rng.random() * 1000000)
```
tomsilver, line 179: this way looks a little weird to me. can we instead use the pattern here? https://stackoverflow.com/questions/23289547/shuffle-two-list-at-once-with-same-order

tests/nsrt_learning/strips_learning/test_gen_to_spec_learner.py
```diff
@@ -285,10 +286,14 @@ def test_backchaining_strips_learner_order_dependence():
...
         assert str(reverse_order_pnads[i]) in correct_pnads
 
-    # Weird Case: This case shows that our algorithm is not data invariant!
+    # Weird Case: This case shows that our algorithm is not data order
+    # invariant!
     utils.reset_config({
         "approach": "nsrt_learning",
-        "strips_learner": "backchaining"
+        "strips_learner": "backchaining",
+        # Following are neccessary to solve this case.
```
tomsilver, line 294: necessary*

predicators/nsrt_learning/nsrt_learning_main.py
```diff
@@ -37,42 +39,68 @@ def learn_nsrts_from_data(
...
-    # STEP 1: Segment each trajectory in the dataset based on changes in
-    #         either predicates or options. If we are doing option learning,
-    #         then the data will not contain options, so this segmenting
-    #         procedure only uses the predicates.
-    segmented_trajs = [
-        segment_trajectory(traj) for traj in ground_atom_dataset
-    ]
+    # Search over data orderings to find least complex PNAD set.
+    # If the strips learner is not Backchaining then it will
+    # only do one iteration, because all other appraoches are
```
tomsilver, line 44: approaches*

predicators/utils.py
tomsilver: actually, now that I see where this is being used, maybe we don't really need a utility function for this and we can just move this code inside the loop? the fact that it's hard to come up with a specific name is good evidence that we probably won't be using this utility elsewhere. and since it's short, might as well just put it where it's used.

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -43,14 +43,43 @@ def learn(self) -> List[PartialNSRTAndDatastore]:
         filtering may break it.
         """
         learned_pnads = self._learn()
+        # Remove pnads by increasing min_data_perc until harmlessness breaks.
```
tomsilver, line 46: cool! I didn't know that this was how you implemented this specifically. I like it

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -43,14 +43,43 @@ def learn(self) -> List[PartialNSRTAndDatastore]:
...
+        if CFG.enable_harmless_op_pruning:
+            assert self._verify_harmlessness
+            assert not CFG.disable_harmlessness_check
+        # Keeps track of latest set of harmless pnads. Will interating over
```
tomsilver, line 50: iterate*?

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -43,14 +43,43 @@ def learn(self) -> List[PartialNSRTAndDatastore]:
...
+        # all min_perc_data_for_nsrts until it breaks.
+        min_harmless_pnads = learned_pnads
+        for min_perc_data_for_nsrt in range(CFG.min_perc_data_for_nsrt, 100):
```
tomsilver, line 53: it would be wiser to only iterate over thresholds that make a difference, right? and since checking harmlessness is slow, this actually seems like a change worth making. can we collect up all the numbers of data across the NSRTs and then only consider thresholding at those numbers?

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -43,14 +43,43 @@ def learn(self) -> List[PartialNSRTAndDatastore]:
...
+                pnad for pnad in learned_pnads
+                if len(pnad.datastore) >= min_data
+            ]
+            if self._verify_harmlessness and \
+                not CFG.disable_harmlessness_check:
+                logging.info("\nRunning harmlessness check...")
+                if not self._check_harmlessness(learned_pnads):
+                    break
+            else:
+                # If we are not verifying harmlessness than we assume that
```
tomsilver, line 67: the logic here is a bit confusing to follow for the cases we had before (where harmless op pruning is not enabled). could we refactor this a bit so that all that happens in this for loop is we determine a good `min_perc_data_for_nsrt` to use, and the for loop only gets run in the case where `enable_harmless_op_pruning` is True; and then outside of the for loop, we use the determined `min_perc_data_for_nsrt` to construct the final PNADs?

tomsilver (commented):
@wmcclinton really nice work! here are some suggestions

predicators/utils.py
tomsilver: actually, now that I see where this is being used, maybe we don't really need a utility function for this and we can just move this code inside the loop? the fact that it's hard to come up with a specific name is good evidence that we probably won't be using this utility elsewhere. and since it's short, might as well just put it where it's used.

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -43,14 +43,50 @@ def learn(self) -> List[PartialNSRTAndDatastore]:
...
+                ]
+                # Runs harmlessness check after we have pruned operators.
+                logging.info("\nRunning harmlessness check...")
+                if not self._check_harmlessness(learned_pnads):
+                    break
+                # We successfully verified harmlessness than we save this set of
+                # pnads and continue reducing min_perc_data_for_nsrts.
+                min_harmless_pnads = learned_pnads
+            learned_pnads = min_harmless_pnads
+        else:
```
tomsilver, line 77: the code below looks quite similar to the code above -- is it possible to remove the "else" here and just have the code once?

predicators/third_party/fast_downward_translator/axiom_rules.py
```diff
@@ -9,7 +9,6 @@
 
 
 class AxiomDependencies(object):
-
```
tomsilver, line 12: looks like the autoformatting went wild here -- can you revert all these third party files?

tomsilver (commented):
looking great! a few small things and then we should be good to merge

tomsilver:
also, let's rerun the sidelining experiments after we merge this

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -46,11 +46,45 @@ def learn(self) -> List[PartialNSRTAndDatastore]:
...
+            pnad_perc_data_low_to_high.sort()
+        else:
+            # If we are not doing harmless operator pruning, return
+            # PNADs at current min_perc_data_for_nsrts.
+            pnad_perc_data_low_to_high = [CFG.min_perc_data_for_nsrt / 100.0]
+        # Iterates over each PNAD in the learned PNADs removing the
+        # PNAD that uses the least amount of data.
+        for min_perc_data_for_nsrt in pnad_perc_data_low_to_high:
+            learned_pnads = self._learn()
+            min_data = max(CFG.min_data_for_nsrt,
```
tomsilver, line 70: is this going to work as expected if you set `--min_data_for_nsrt` to be something nontrivial, but then also `--enabled_harmless_op_pruning`?

wmcclinton: It will always chose the max between min_data_for_nsrt and the min_perc_data_for_nsrt. So for all min_perc_data_for_nsrts it is trying below the min_data_for_nsrt it will just use the min_data_for_nsrt instead. If min_data_for_nsrt causes us to break harmlessness it will return the original learned pnads.

tomsilver: kk thanks, I think that makes sense

tomsilver (approved):
very nice work!

PR #1188, opened 2022-08-27
Yapf fix

NishanthJKumar:
Changes CI yapf check to ignore the `third_party` folder and thus mirror what we're doing in `run_autoformat.sh`. I believe this should fix CI issues with #1185


tomsilver (approved):
makes sense, thanks!

PR #1195, opened 2022-09-04
Enables FD to use ignore effects

NishanthJKumar:
Previously, FD had no knowledge of operators' ignore effects. As a result, it wasn't able to solve even simple domains (such as `repeated_nextto_single_option`) where ignore effects are necessary.

This PR modifies the PDDL file creation code to parse ignore effects into universally-quantified delete effects that FD can operate on.

You can try this out locally by running the following command before and after this PR:
```
python predicators/main.py --env repeated_nextto_single_option --approach oracle --num_test_tasks 50 --seed 0 --sesame_task_planner fdopt
```

Before this PR, the above would result in slow execution and a large number of failed tasks. Afterwards, FD should be able to solve all 50 tasks.

### DIFF
--- a/tests/test_structs.py
+++ b/tests/test_structs.py
@@ -504,6 +504,15 @@ def test_operators_and_nsrts(state):
     Add Effects: [On(?cup:cup_type, ?plate:plate_type)]
     Delete Effects: [NotOn(?cup:cup_type, ?plate:plate_type)]
     Ignore Effects: [On]"""
+    assert strips_operator.pddl_str() == \
+        """(:action Pick
+    :parameters (?cup - cup_type ?plate - plate_type)
+    :precondition (and (NotOn ?cup ?plate))
+    :effect (and (On ?cup ?plate)
+        (not (NotOn ?cup ?plate))
+        (forall (?x0 - cup_type ?x1 - plate_type)(not (On ?x0 ?x1)))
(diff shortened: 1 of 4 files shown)


tomsilver (commented):
looking good, but can you also add a structs unit test? mainly I'd like to see an example of what the new pddl_str looks like in that test. thx!

tests/test_structs.py
```diff
@@ -504,6 +504,15 @@ def sampler(s, g, rng, objs):
     Add Effects: [On(?cup:cup_type, ?plate:plate_type)]
     Delete Effects: [NotOn(?cup:cup_type, ?plate:plate_type)]
     Ignore Effects: [On]"""
+    assert strips_operator.pddl_str() == \
+        """(:action Pick
+    :parameters (?cup - cup_type ?plate - plate_type)
+    :precondition (and (NotOn ?cup ?plate))
+    :effect (and (On ?cup ?plate)
+        (not (NotOn ?cup ?plate))
+        (forall (?x0 - cup_type ?x1 - plate_type)(not (On ?x0 ?x1)))
```
tomsilver, line 513: nit: can we add a space in between the `)(` parens here?

tomsilver (approved):
tiny thing, otherwise lgtm!

PR #1196, opened 2022-09-07
fixing github checks

tomsilver:
* pandas-stubs upgraded leading to mypy failures, changed the code to silence those
* pylint upgraded and broke in a very specific case. opened an issue here https://github.com/PyCQA/pylint/issues/7427
* we weren't previously autoformatting setup.py with `run_autoformat.sh`, fixed that too

### DIFF
--- a/scripts/plotting/create_bar_plots.py
+++ b/scripts/plotting/create_bar_plots.py
@@ -10,7 +10,7 @@ from scripts.analyze_results_directory import combine_selectors, \
     create_dataframes, get_df_for_entry, pd_create_equal_selector
 
 plt.style.use('ggplot')
-pd.options.mode.chained_assignment = None  # type: ignore
+pd.set_option('chained_assignment', None)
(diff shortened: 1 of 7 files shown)


scripts/plotting/create_bar_plots.py
```diff
@@ -10,7 +10,7 @@
     create_dataframes, get_df_for_entry, pd_create_equal_selector
 
 plt.style.use('ggplot')
-pd.options.mode.chained_assignment = None  # default='warn'
+pd.options.mode.chained_assignment = None  # type: ignore
```
NishanthJKumar, line 13: an alternative here could be to use `pd.set_option('chained_assignment',None)` as shown [here](https://stackoverflow.com/questions/21463589/pandas-chained-assignments)

NishanthJKumar (approved):
Thanks for the great, lightning-quick work on this confusing issue! Just left one suggestion to consider.

PR #1197, opened 2022-09-07
add optional log file flag

tomsilver:
usage example:

```python predicators/main.py --env cover --approach oracle --seed 0 --log_file ~/Desktop/logs_testing/test2.log```

I also verified that when the script crashes, the exception is logged.


NishanthJKumar (approved):
LGTM! Thanks so much; this'll be super helpful for BEHAVIOR!

PR #1198, opened 2022-09-08
Interactive learning config

amburger66:
- interactive learning yaml config
- remove old bash scripts


tomsilver:
@amburger66 can you split this into two? the removing of the old scripts will be quick to merge, but I'll review the new yaml more carefully

PR #1199, opened 2022-09-08
save successful evaluation trajectories to disk

tomsilver:
@williamshen-nz my plan here is to next make a separate script that loads our saved trajectories and converts them into the lisdf format. these saved trajectories should have all the info we need.


predicators/main.py
```diff
@@ -299,6 +302,17 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+            traj_file = f"{save_prefix}__task{test_task_idx+1}.traj"
+            traj_file_path = Path(CFG.eval_trajectories_dir) / traj_file
+            # Include the original task too so we know the goal.
+            traj_data = {
+                "task": task,
+                "trajectory": traj,
+                "pybullet_robot": CFG.pybullet_robot
+            }
+            with open(traj_file_path, "wb") as f:
+                pkl.dump(traj_data, f)
```
williamshen-nz, line 315: So the plan is to have a script, that given any task, trajectory, and pybullet robot will generate the LISDF plan right?

williamshen-nz: In the very long run we may want to output a LISDF input problem as well so we can load it dynamically on multiple simulators and its more self-contained.

williamshen-nz (approved):
LGTM, had some minor questions.

predicators/main.py
```diff
@@ -299,6 +302,17 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
             solved = task.goal_holds(traj.states[-1])
             exec_time = execution_metrics["policy_call_time"]
             metrics[f"PER_TASK_task{test_task_idx}_exec_time"] = exec_time
+            # Save the successful trajectory, e.g., for playback on a robot.
+            traj_file = f"{save_prefix}__task{test_task_idx+1}.traj"
+            traj_file_path = Path(CFG.eval_trajectories_dir) / traj_file
+            # Include the original task too so we know the goal.
+            traj_data = {
+                "task": task,
```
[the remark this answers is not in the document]
tomsilver: yeah, that's all in `task.init`

predicators/main.py
tomsilver: right, and yeah agreed!

PR #1201, opened 2022-09-08
Yaml config for interactive learning experiments

amburger66:


scripts/configs/interactive_learning.yaml
```diff
@@ -0,0 +1,65 @@
...
+    FLAGS:
+      explorer: "no_explore"
+ENVS:
+  cover:
+    NAME: "cover"
+    FLAGS: {}
+ARGS: {}
+FLAGS:  # common args
+  excluded_predicates: "Covers,Holding"
+  interactive_score_threshold: 0.05
```
tomsilver, line 55: I see that you removed `--interactive_score_function entropy` because it's the default in `settings.py` which makes sense. why don't we similarly change the defaults for the other key parameters here in settings.py, rather than overriding them here, now that we're converging on a final set of hyperparameters. so:
* `interactive_score_threshold`
* `predicate_mlp_classifier_max_itr`

I guess that's it

tomsilver (commented):
looks great! just one small thing

scripts/configs/interactive_learning.yaml
```diff
@@ -52,14 +52,11 @@ ENVS:
 ARGS: {}
 FLAGS:  # common args
   excluded_predicates: "Covers,Holding"
-  interactive_score_threshold: 0.05
-  num_online_learning_cycles: 1000  # way too many cycles
-  online_learning_max_transitions: 10000  # want this to be stop signal
-  interactive_num_requests_per_cycle: 10
+  num_online_learning_cycles: 100  # way too many cycles
```
tomsilver, line 55: did you intentionally reduce both of these?

amburger66: yeah, those were the original settings

tomsilver: Good catch!

tomsilver (approved):

PR #1202, opened 2022-09-08
script for converting saved eval traj to lisdf plan

tomsilver:
should be merged after #1199


scripts/eval_trajectory_to_lisdf.py
```diff
@@ -0,0 +1,72 @@
...
+    joint_arr = np.array([a.arr for a in ll_traj.actions], dtype=float)
+    robot_name = traj_data["pybullet_robot"]
+    # Create an instance of the robot with default position values so that we
+    # can extract the joint names.
+    physics_client_id = p.connect(p.DIRECT)
+    robot = create_single_arm_pybullet_robot(robot_name, (1.35, 0.6, 0.7),
+                                             (0.0, 0.0, 0.0, 1.0),
+                                             physics_client_id)
+    joint_names = robot.arm_joint_names
+    assert len(joint_names) == joint_arr.shape[1]
```
williamshen-nz, line 55: This is fine, but maybe we want to store this with the trajectory itself that way we don't need to load up pybullet again?

williamshen-nz (approved):
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

scripts/eval_trajectory_to_lisdf.py
tomsilver: hmmm yeah that would be nice, but since we're saving the trajectories right now in main.py, where there isn't necessarily any notion of a pybullet robot (because we could be in an environment without a robot), we would have to do something hacky like this anyway. I'd rather do the hacking here in this script for now rather than the main file, so I'll leave it

tomsilver:
@williamshen-nz do you have any idea why mypy is complaining with this?
```
scripts/eval_trajectory_to_lisdf.py:19: error: Cannot find implementation or library stub for module named "lisdf.planner_output.command"
scripts/eval_trajectory_to_lisdf.py:19: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
scripts/eval_trajectory_to_lisdf.py:20: error: Cannot find implementation or library stub for module named "lisdf.planner_output.plan"
```
I'm surprised because lisdf is all typed up, as far as I can tell

williamshen-nz:
> @williamshen-nz do you have any idea why mypy is complaining with this?
> ```
> scripts/eval_trajectory_to_lisdf.py:19: error: Cannot find implementation or library stub for module named "lisdf.planner_output.command"
> scripts/eval_trajectory_to_lisdf.py:19: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> scripts/eval_trajectory_to_lisdf.py:20: error: Cannot find implementation or library stub for module named "lisdf.planner_output.plan"
> ```
> I'm surprised because lisdf is all typed up, as far as I can tell



I think we might need to add a py.typed to the lisdf repo (and maybe include it with the setup.py, I'm not too familiar with it).

https://stackoverflow.com/a/68671774

williamshen-nz:
Do you want to merge this into the `pybullet-panda` branch before we merge that branch into main?

tomsilver:
nah let's merge your PR first

PR #1204, opened 2022-09-12
move argparse in analyze script

tomsilver:
this was leading to a tricky bug where

```
python scripts/plotting/create_classification_plots.py --env cover --approach interactive_learning --seed 123 --excluded_predicates Covers,Holding --experiment_id cover-main --num_online_learning_cycles 100 --load_approach --load_data
```

was leading to

```
usage: create_classification_plots.py [-h] [--sidelining]
create_classification_plots.py: error: unrecognized arguments: --env cover --approach interactive_learning --seed 123 --excluded_predicates Covers,Holding --experiment_id cover-main --num_online_learning_cycles 100 --load_approach --load_data
```

and it was because the argparse was getting created when the analysis script was getting imported, rather than inside of create_classification_plots.py itself.

@amburger66 (who found the bug) -- can you verify that this fixes things on your end?

### DIFF
--- a/scripts/analyze_results_directory.py
+++ b/scripts/analyze_results_directory.py
@@ -10,10 +10,6 @@ import pandas as pd
 
 from predicators.settings import CFG
 
-parser = argparse.ArgumentParser()
-parser.add_argument("--sidelining", action="store_true")
-args = parser.parse_args()
-
 GROUPS = [
     # "ENV",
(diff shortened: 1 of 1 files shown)


amburger66:
yep, the script works now, thanks!

amburger66 (approved):

PR #1209, opened 2022-09-20
Option to use GPU in YAML flags

### LINKED ISSUE #1206: make option to use GPU in config files

NishanthJKumar:
This PR introduces a new option `USE_GPU` to pass into configuration YAML files that toggles whether or not a GPU will be used to run the particular experiment on Supercloud. This flag is optional.

Resolves #1206 .


scripts/configs/example_basic_gpu.yaml
```diff
@@ -0,0 +1,27 @@
...
+    NAME: "cover"
+    FLAGS:
+      cover_num_blocks: 1
+      cover_block_widths: [0.1]
+ARGS:
+  - "debug"
+FLAGS:  # general flags
+  num_train_tasks: 20
+  num_test_tasks: 10
+  use_gpu: True
```
tomsilver, line 25: my preference would be to not make this a FLAG because if someone where to add the flag `--use_gpu` to a normal command line run, it would crash, so it kinda gives the wrong impression. I know that this will be more changes, but could we instead make `USE_GPU` a sibling to `START_SEED`, `NUM_SEEDS`, etc.?

tomsilver (commented):
thanks for doing this! one high-level thought

scripts/configs/example_basic_gpu.yaml
NishanthJKumar: makes a ton of sense (and in fact, the current setup leads to bugs!). Will do this!

scripts/cluster_utils.py
```diff
@@ -84,6 +85,7 @@ def generate_run_configs(config_filename: str,
         num_seeds = config["NUM_SEEDS"]
         args = config["ARGS"]
         flags = config["FLAGS"]
+        use_gpu = config["USE_GPU"]
```
tomsilver, line 88: looking great overall! one more medium thing: can we make USE_GPU optional and default to False so that we don't have to modify these existing files? I think we can accomplish this by just checking here if `USE_GPU` is in `config`, and if not, set `use_gpu` to False

tomsilver (approved):
LGTM great work!

tomsilver:
For posterity, please edit the part of the PR description saying the new thing is mandatory

PR #1211, opened 2022-09-21
fix supercloud GPU command string

NishanthJKumar:
When requesting a GPU on supercloud, the partition needs to be `--partition=xeon-g6-volta` and not `--partition=xeon-p8`. Also, the `--gres` argument cannot appear at the end of the string, but rather must appear before the `sh` filename.

Tested and verified this works on launching a BEHAVIOR experiment (in `predicators_behavior` repo).


tomsilver (approved):
thanks for finding and fixing!

PR #1214, opened 2022-09-24
New weight initialization option for `MLPBinaryClassifier`

amburger66:
- allows for linear layer weight parameters to be sampled from a Normal(0, 1) distribution whenever weights are reset
- this weight initialization benefits the interactive learning approach since it results in more areas of high variance as classified by the ensemble after training


predicators/ml_models.py
```diff
@@ -371,25 +375,51 @@ def _create_optimizer(self) -> optim.Optimizer:
         """Create an optimizer after the model is initialized."""
         return optim.Adam(self.parameters(), lr=self._learning_rate)
 
+    def _reset_weights(self) -> None:
+        """(Re-)initialize the network weights."""
+        self.apply(lambda m: self._weight_reset(m, self._weight_init))
+
+    @staticmethod
+    def _weight_reset(m: torch.nn.Module, weight_init: str) -> None:
+        if isinstance(m, nn.Linear):
```
tomsilver, line 384: what else could this be? `nn.ModuleList`? maybe we should assert that. I'm concerned that one day we'll add, e.g., `nn.Conv2D` somewhere, and we won't realize that the weights aren't getting reset

predicators/ml_models.py
```diff
@@ -371,25 +375,51 @@ def _create_optimizer(self) -> optim.Optimizer:
...
+            if weight_init == "default":
+                m.reset_parameters()
+            elif weight_init == "normal":
+                torch.nn.init.normal_(m.weight, std=1.0)
```
tomsilver, line 388: [it looks like](https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.normal_) `std=1.0` is the default, so you could just remove it if you want

predicators/ml_models.py
```diff
@@ -371,25 +375,51 @@ def _create_optimizer(self) -> optim.Optimizer:
...
+            # Run training.
+            best_loss = _train_pytorch_model(
+                self,
+                loss_fn,
+                optimizer,
+                batch_generator,
+                max_train_iters=self._max_train_iters,
+                dataset_size=X.shape[0],
+                n_iter_no_change=self._n_iter_no_change)
+            if best_loss < 1:
```
tomsilver, line 418: can you add a comment explaining this?

predicators/ml_models.py
```diff
@@ -939,7 +971,7 @@ def _train_pytorch_model(model: nn.Module,
                          print_every: int = 1000,
                          clip_gradients: bool = False,
                          clip_value: float = 5,
-                         n_iter_no_change: int = 10000000) -> None:
+                         n_iter_no_change: int = 10000000) -> float:
```
tomsilver, line 974: mention in the docstring what's now returned

tomsilver (approved):
excellent work! just a few tiny things otherwise LGTM

tomsilver (approved):
LGTM!

PR #1216, opened 2022-09-25
Fix supercloud launch command

amburger66:
tricky space


tomsilver (approved):
thanks, we need to test changes like this more carefully in the future!

PR #1220, opened 2022-09-26
fix pybullet rendering

### LINKED ISSUE #1181: fix pybullet rendering
it looks like the latest version of pybullet changed the return type for `getCameraImage`, so our rendering is broken in pybullet, e.g.,

```python src/main.py --approach oracle --env pybullet_blocks --seed 0 --pybullet_use_gui True --make_test_videos --num_test_tasks 1```

tomsilver:
closes #1181

at some point, pybullet changed the output of `getCameraImage` to be a flat vector instead of the RGBA image.


williamshen-nz (approved):
🚢

PR #1221, opened 2022-09-26
change all `time.time()` to `time.perf_counter()`

### LINKED ISSUE #1193: replace `time.time()` with `time.perf_counter()`
some references:
* https://github.com/scikit-learn/scikit-learn/issues/17316
* https://docs.python.org/3/library/time.html#time.time "Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second."

tomsilver:
closes #1193


williamshen-nz (approved):
:shipit:

PR #1222, opened 2022-09-26
move the base of the panda to be on the table

tomsilver:
needs testing


williamshen-nz:
I guess we should discard the changes here and use the `panda_real` branch?

PR #1226, opened 2022-09-28
updates to LISDF conversion script

tomsilver:
we now split the joint array into alternating arm joint space paths and gripper actuations.

### DIFF
--- a/scripts/eval_trajectory_to_lisdf.py
+++ b/scripts/eval_trajectory_to_lisdf.py
@@ -12,12 +12,15 @@ Then run this script:
         --output /tmp/pybullet_blocks__oracle__0________task1.json
 """
 import argparse
+from typing import List, Tuple
 
 import dill as pkl
 import numpy as np
 import pybullet as p
-from lisdf.planner_output.command import JointSpacePath
+from lisdf.planner_output.command import ActuateGripper, GripperPosition, \
+    JointSpacePath
 from lisdf.planner_output.plan import LISDFPlan
+from numpy.typing import NDArray
 
 from predicators import utils
 from predicators.pybullet_helpers.robots import \
@@ -45,7 +48,7 @@ def _main() -> None:
         traj_data = pkl.load(f)
 
     ll_traj = traj_data["trajectory"]
-    joint_arr = np.array([a.arr for a in ll_traj.actions], dtype=float)
+    joint_arr = np.array([a.arr for a in ll_traj.actions], dtype=np.float32)
     robot_name = traj_data["pybullet_robot"]
     # Create an instance of the robot with default position values so that we
     # can extract the joint names.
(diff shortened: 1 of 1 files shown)


scripts/eval_trajectory_to_lisdf.py
```diff
@@ -56,16 +59,76 @@ def _main() -> None:
...
-        joint_names,
-        duration=duration,
-        label=args.input,
-    )
-    lisdf_plan = LISDFPlan(commands=[command], lisdf_problem="no_problem")
+    # Convert the joint array into a list of (arm joint state, gripper state)
+    # where the gripper state is GripperPosition.open or GripperPosition.close
+    # and the arm joint state is the original joints with grippers removed.
+    gripper_idxs = [robot.left_finger_joint_idx, robot.right_finger_joint_idx]
+    arm_joint_names = list(np.delete(joint_names, gripper_idxs))
```
williamshen-nz, line 66: Just a general comment that doesn't need addressing.

The naming here isn't super satisfying given we have `robot.arm_joint_names` as well. We might want to think about renaming the latter so there are arm joints (for the actual arm) and gripper joints for the gripper, and they some name that encapsulates both.

scripts/eval_trajectory_to_lisdf.py
```diff
@@ -56,16 +59,76 @@ def _main() -> None:
...
+        dist_to_closed = abs(gripper_val - robot.closed_fingers)
+        if dist_to_closed < dist_to_open:
+            gripper_state = GripperPosition.close
+        else:
+            gripper_state = GripperPosition.open
+        # Extract the arm joint state.
+        arm_state = np.delete(joints, gripper_idxs)
+        return (arm_state, gripper_state)
+
+    arm_gripper_states = list(map(_joints_to_arm_gripper_vals, joint_arr))
```
williamshen-nz, line 84: Nice!

scripts/eval_trajectory_to_lisdf.py
```diff
@@ -56,16 +59,76 @@ def _main() -> None:
...
+
+    # Accumulate the arm-only actions until a change in the gripper state
+    # is observed, at which point we finalize the JointSpacePath command,
+    # add an ActuateGripper command, and then start a new JointSpacePath.
+    accum_arm_states: List[NDArray[np.float32]] = []
+
+    # Helper function for converting an accum_arm_states to a JointSpacePath.
+    def _create_path_command(
+            arm_states: List[NDArray[np.float32]]) -> JointSpacePath:
+        duration = args.time_per_conf * len(arm_states)
```
williamshen-nz, line 97: This `duration` is going to be crazy fast when we execute on the robot. Setting `time_per_conf` to 0.5 made it more reasonable.

I got the server to respect the duration of the command during execution, so we don't want the duration to be too short.

Also, just a note we should execute the LISDFPlan in pybullet before doing it on the actual robot. A script for that should be pretty straight forward. I can try that later.

scripts/eval_trajectory_to_lisdf.py
```diff
@@ -56,16 +59,76 @@ def _main() -> None:
...
+        arm_states_np = np.array(arm_states)
+        path_command = JointSpacePath.from_waypoints_np_array(
+            arm_states_np,
+            arm_joint_names,
+            duration=duration,
+            label=args.input,
```
williamshen-nz, line 103: Can we give a unique label for each command? That way it's easier to debug when it goes to the server and it executes each command.

scripts/eval_trajectory_to_lisdf.py
```diff
@@ -56,16 +59,76 @@ def _main() -> None:
...
+        # Check if a change in the gripper state has occurred.
+        if current_gripper_state != prev_gripper_state:
+            # Finalize the JointSpacePath until this point.
+            path_command = _create_path_command(accum_arm_states)
+            commands.append(path_command)
+            accum_arm_states = [arm_state]
+            # Create the ActuateGripper command.
+            gripper_command = ActuateGripper(
+                configurations={"panda_gripper": current_gripper_state},
+                label="gripper_action")
```
williamshen-nz, line 121: Unique label here would be helpful as well.

scripts/eval_trajectory_to_lisdf.py
```diff
@@ -56,16 +59,76 @@ def _main() -> None:
...
+            commands.append(gripper_command)
+            prev_gripper_state = current_gripper_state
+
+    # Finish the last JointSpacePath.
+    if len(accum_arm_states) > 1:
+        path_command = _create_path_command(accum_arm_states)
+        commands.append(path_command)
+
+    # Write out the commands to JSON.
+    lisdf_plan = LISDFPlan(commands=commands, lisdf_problem="no_problem")
```
williamshen-nz, line 131: Maybe let's just set the `lisdf_problem` to be the filename of the saved trajectory just so we can track what is what.

williamshen-nz (changes requested):
Looks great! Had some minor comments

scripts/eval_trajectory_to_lisdf.py
tomsilver: sounds good! I changed the default `time_per_conf` to 0.5

williamshen-nz (approved):
🚢  Thanks for the changes!

Issue #1182, opened 2022-08-20
fix pybullet cover regression


tomsilver:
see discussion in #1174

williamshen-nz:
Latest pybullet changes seems to have dropped coverage on cover again: https://github.com/Learning-and-Intelligent-Systems/predicators/commit/79ec87b98236d49e9bc59e82a0da4266a22c447e

```
Git commit hashes seen in results/:
810afd7a473be17fbe7ab92f5065a1ce8d84a1f2


AGGREGATED DATA OVER SEEDS:
                       NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                               
pybullet_blocks-oracle    50.00 (0.00)  47.18 (1.61)   5.00 (0.00)   0.35 (0.10)  3530.42 (1867.35)   0.00 (0.00)        100
pybullet_cover-oracle     50.00 (0.00)  43.41 (2.31)   5.00 (0.00)   0.07 (0.01)        6.69 (0.40)   0.00 (0.00)        100
```

williamshen-nz:
We can disregard all the IKFast and Panda-specific stuff, as fetch doesn't use it.

In `pybullet_covers.py`

```
    _move_to_pose_tol: ClassVar[float] = 1e-4
```

Pose tolerance was relaxed so it should only do better, if anything. Not sure what's happening, will rerun supercloud job and see if it's consistent.

tomsilver:
it is possible that the `_move_to_pose_tol` relaxation hurts performance. that parameter affects the controller, not the physics

williamshen-nz:
Yes, I see what you mean. I remember I changed it from the default as it didn't work otherwise.

Will experiment and run some regression tests

tomsilver:
I looked at this a little and it does seem like the `_move_to_pose_tol` change was responsible for the regression. We could make that hyperparameter robot-specific, but I think it's fine to just swallow the slight regression. Closing this for now.

PR #1227, opened 2022-09-28
update panda and blocks constants to be consistent with real

tomsilver:


tomsilver:
looks like `pytest -s tests/envs/test_pybullet_blocks.py` has a similar error (good job unit test!)

tomsilver:
supercloud results forthcoming

predicators/envs/pybullet_cover.py
```diff
@@ -193,6 +193,17 @@ def _reset_state(self, state: State) -> None:
                 assert self._held_obj_id == block_id
                 self._create_grasp_constraint()
 
+        # For any blocks not involved, put them out of view.
```
tomsilver, line 196: this was just a bug before (but it only affected the unit tests, because in normal circumstances, there are always exactly 2 blocks in the cover env)

tomsilver:
```
15dee8f5b6550cba1f8fb037094fc810905c807d


AGGREGATED DATA OVER SEEDS:
EXPERIMENT_ID                                                                                                               
pybullet_blocks-oracle    50.00 (0.00)  48.60 (1.02)   5.00 (0.00)   0.36 (0.08)  3670.01 (1584.52)   0.00 (0.00)         10
pybullet_cover-oracle     50.00 (0.00)  42.70 (2.49)   5.00 (0.00)   0.06 (0.01)        6.66 (0.32)   0.00 (0.00)         10
```

good enough I guess... (pybullet cover was around 43 as of https://github.com/Learning-and-Intelligent-Systems/predicators/issues/1182)

predicators/envs/blocks.py
```diff
@@ -28,7 +28,7 @@ class BlocksEnv(BaseEnv):
     """Blocks domain."""
     # Parameters that aren't important enough to need to clog up settings.py
     table_height: ClassVar[float] = 0.2
-    block_size: ClassVar[float] = 0.045
+    block_size: ClassVar[float] = 0.0505
```
williamshen-nz, line 31: The true block size is `0.05` but we added the buffer of 5mm to avoid the collision reflex on the robot. 

Should we just continue to use this buffer or add logic in the controller to place blocks marginally higher than the other ones?

predicators/envs/pybullet_cover.py
```diff
@@ -193,6 +193,17 @@ def _reset_state(self, state: State) -> None:
...
+        h = self._obj_len_hgt
+        oov_x, oov_y = self._out_of_view_xy
+        for i in range(len(block_objs), len(self._block_ids)):
```
williamshen-nz, line 199: So `block_objs` are the blocks involved while the others are not? Why are the unit tests creating environments with >2 blocks (maybe its intentional)?

predicators/pybullet_helpers/robots/single_arm.py
```diff
@@ -429,7 +433,11 @@ def inverse_kinematics(self, end_effector_pose: Pose3D,
...
             if validate:
-                self._validate_joints_state(joint_positions, end_effector_pose)
+                try:
+                    self._validate_joints_state(joint_positions,
+                                                end_effector_pose)
+                except ValueError as e:
+                    raise InverseKinematicsError(e)
             return joint_positions
 
         return pybullet_inverse_kinematics(
```
williamshen-nz, line 443: Pybullet IK may also fail (in practice it sometimes just returns nonsense), in which case it just raises an Exception. So should we change that to `InverseKinematicsError`?

https://github.com/Learning-and-Intelligent-Systems/predicators/blob/b69599a0bb9a0bd6d4051a4e163ab3867420fdbb/predicators/pybullet_helpers/inverse_kinematics.py#L83

tests/envs/test_pybullet_blocks.py
```diff
@@ -303,6 +304,27 @@ def test_pybullet_blocks_putontable(env):
     # Specifically, it should be at the center of the workspace.
     assert abs(state.get(block, "pose_x") - (env.x_lb + env.x_ub) / 2.) < 1e-3
     assert abs(state.get(block, "pose_y") - (env.y_lb + env.y_ub) / 2.) < 1e-3
+    # Test that when we attempt to put the block outside the workspace, an
+    # OptionExecutionFailure is raised. This is for the panda only because
+    # the fetch uses pybullet IK, which is not even smart enough to realize its
+    # own ineptitude.
+    if CFG.pybullet_robot == "panda":
```
williamshen-nz, line 311: Love the comment 🤣

williamshen-nz (approved):
Thanks for the changes! :shipit: 

Minor comments about block size, and updating Pybullet IK to maybe also raise `InverseKinematicsError`

predicators/envs/blocks.py
tomsilver: I'm gonna leave it as is for now because dropping the blocks from higher in simulation may lead to a little bouncing

predicators/envs/pybullet_cover.py
tomsilver: it's actually the opposite: unit tests are sometimes creating environments with 1 block, and the other pybullet block object was then left where ever it was in the previous task

Issue #1218, opened 2022-09-26
Real Panda Execution


williamshen-nz:
Just an issue to track todos for that so I can remember:

- [x] Move Panda in simulation to world origin (not completely necessary, just need table to be basically the same height as the robot)
- [x] LISDF Plan Executor server: https://github.com/williamshen-nz/panda-robot-server
- [x] Expose a panda client to predicators: https://github.com/williamshen-nz/panda-robot-client
- [x] Perception - determine best camera setup, probably need opencv if we just use aruco markers

williamshen-nz:
Panda Server and Client are implemented and working on the real robot. The server will also respect the duration in the commands of a LISDF Plan so we don't have to wait for it to take 2 minutes to go home.

So it seems like perception is the next thing to do, apart from making sure the plans we generate can be reliably executed.

williamshen-nz:
**Immediate next steps:**

- [ ] Recalibrate extrinsics of the two inward-facing cameras (to the robot)
- [ ] Capture multiple scenarios of the blocks, RGB + D
- [ ] Dump intrinsics, extrinsics and RGB-Ds into a folder
- [ ] Handover to Tom

Depending on how we want to process the images, we could use something like open3d (might be overkill) or just write our own methods.

tomsilver:
@williamshen-nz I guess we can close this?

williamshen-nz:
Yep! Will close

PR #1234, opened 2022-10-06
make rgb color part of block state

tomsilver:
as we move toward real world block stacking, we want to be able to consume a task specification that includes block colors (see #1218). the "right" way to handle this in our framework is to make color part of the block state. this PR investigates whether that will break things. in particular, we'll have to see whether any part of the learning pipeline overfits to color (predicate invention especially). if there is obvious performance degradation, we can hack around it by saving the correspondence between object and colors in the environment itself, but that is pretty ugly because objects vary per task.

### DIFF
--- a/predicators/envs/pybullet_blocks.py
+++ b/predicators/envs/pybullet_blocks.py
@@ -258,7 +258,7 @@ class PyBulletBlocksEnv(PyBulletEnv, BlocksEnv):
                 block_id, [bx, by, bz],
                 self._default_orn,
                 physicsClientId=self._physics_client_id)
-            # Update the block color.
+            # Update the block color. RGB values are between 0 and 1.
             r = state.get(block_obj, "color_r")
             g = state.get(block_obj, "color_g")
             b = state.get(block_obj, "color_b")
--- a/predicators/envs/pybullet_blocks.py
+++ b/predicators/envs/pybullet_blocks.py
@@ -258,6 +258,15 @@ class PyBulletBlocksEnv(PyBulletEnv, BlocksEnv):
                 block_id, [bx, by, bz],
                 self._default_orn,
                 physicsClientId=self._physics_client_id)
+            # Update the block color.
+            r = state.get(block_obj, "color_r")
+            g = state.get(block_obj, "color_g")
+            b = state.get(block_obj, "color_b")
+            color = (r, g, b, 1.0)  # alpha = 1.0
+            p.changeVisualShape(block_id,
+                                linkIndex=-1,
+                                rgbaColor=color,
+                                physicsClientId=self._physics_client_id)
 
         # For any blocks not involved, put them out of view.
         h = self.block_size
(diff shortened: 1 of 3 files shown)


tomsilver:
```
AGGREGATED DATA OVER SEEDS:
                                  NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED     LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                                              
blocks-invent_allexclude             50.00 (0.00)  48.20 (1.83)   7.40 (1.20)   0.38 (0.14)  4091.08 (1865.47)  2000.84 (454.45)         10
blocks-nsrt_learning                 50.00 (0.00)  48.30 (1.85)   5.00 (0.00)   0.31 (0.11)  4045.78 (1869.73)      10.27 (0.85)         10
blocks-oracle                        50.00 (0.00)  49.70 (0.46)   5.00 (0.00)   0.27 (0.09)  3752.81 (1797.57)       0.00 (0.00)         10
pybullet_blocks-invent_allexclude    50.00 (0.00)  46.60 (2.76)   7.10 (1.22)   0.52 (0.18)  4066.03 (2238.76)  3271.68 (886.61)         10
pybullet_blocks-nsrt_learning        50.00 (0.00)  46.70 (2.83)   5.00 (0.00)   0.44 (0.12)  3864.49 (1826.47)      20.59 (1.82)         10
pybullet_blocks-oracle               50.00 (0.00)  49.60 (0.66)   5.00 (0.00)   0.40 (0.09)  3752.81 (1797.57)       0.00 (0.00)         10
```

so there is a noticeable decline in sampler learning for pybullet blocks. I think it's marginal enough that we should just move forward with this change.

predicators/envs/pybullet_blocks.py
```diff
@@ -258,6 +258,15 @@ def _reset_state(self, state: State) -> None:
                 block_id, [bx, by, bz],
                 self._default_orn,
                 physicsClientId=self._physics_client_id)
+            # Update the block color.
+            r = state.get(block_obj, "color_r")
+            g = state.get(block_obj, "color_g")
+            b = state.get(block_obj, "color_b")
+            color = (r, g, b, 1.0)  # alpha = 1.0
```
williamshen-nz, line 265: Default color is black? Should we just randomly sample colors if not specified them? 

Maybe we should enforce either none or all of rgb is specified?

williamshen-nz (approved):
Looks good - had 2 minor comments

predicators/envs/pybullet_blocks.py
williamshen-nz: Oops I was on my phone and not looking clearly and it looked like `state` was some dict.

Looks in structs and it makes more sense now.

williamshen-nz (approved):
:shipit:

PR #1236, opened 2022-10-06
create blocks test tasks from json file

tomsilver:
see https://github.com/Learning-and-Intelligent-Systems/predicators/issues/1218

example:
```python predicators/main.py --env pybullet_blocks --approach oracle --seed 0 --num_test_tasks 1 --blocks_test_task_json_dir <path to json dir> --pybullet_use_gui True```

with example json:
```
{
    "problem_name": "blocks_2022-10-01_13-56-00",
    "blocks": {
        "red_block": {
            "position": [1.36409716, 1.0389289 , 0.2225],
            "color": [1, 0, 0]
        },
        "green_block": {
            "position": [1.36409716, 1.0389289 , 0.2675],
            "color": [0, 1, 0]
        },
        "blue_block": {
            "position": [1.35479861, 0.91064759, 0.2225],
            "color": [0, 0, 1]
        }
    },
    "block_size": 0.045,
    "goal": {
       "On": [
           ["red_block", "green_block"],
           ["green_block", "blue_block"]
       ],
      "OnTable": [
          ["blue_block"]
      ]
    }
}
```

### DIFF
--- a/predicators/envs/blocks.py
+++ b/predicators/envs/blocks.py
@@ -527,12 +527,11 @@ class BlocksEnv(BaseEnv):
         # One day, we can make the block size a feature of the blocks, but
         # for now, we'll just make sure that the block size in the real env
         # matches what we expect in sim.
-        assert abs(task_spec["block_size"] - self.block_size) < 1e-6
+        assert np.isclose(task_spec["block_size"], self.block_size)
         state_dict: Dict[Object, Dict[str, float]] = {}
-        id_to_obj: Dict[int, Object] = {}  # used in the goal construction
-        for block_spec in task_spec["blocks"]:
-            block_id: int = block_spec["block_id"]
-            block = Object(f"block{block_id}", self._block_type)
+        id_to_obj: Dict[str, Object] = {}  # used in the goal construction
+        for block_id, block_spec in task_spec["blocks"].items():
+            block = Object(block_id, self._block_type)
             id_to_obj[block_id] = block
             x, y, z = block_spec["position"]
             r, g, b = block_spec["color"]
--- a/predicators/envs/blocks.py
+++ b/predicators/envs/blocks.py
@@ -257,6 +257,7 @@ class BlocksEnv(BaseEnv):
             x = state.get(block, "pose_x")
             y = state.get(block, "pose_y")
             z = state.get(block, "pose_z")
+            # RGB values are between 0 and 1.
             color_r = state.get(block, "color_r")
             color_g = state.get(block, "color_g")
             color_b = state.get(block, "color_b")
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: predicators/envs/blocks.py [hunk 1]
```diff
@@ -509,3 +518,52 @@ def _get_highest_block_below(self, state: State, x: float, y: float,
...
         return max(blocks_here, key=lambda x: x[1])[0]  # highest z
+
+    def _load_task_from_json(self, json_file: Path) -> Task:
+        with open(json_file, "r", encoding="utf-8") as f:
+            task_spec = json.load(f)
+        # Create the initial state from the task spec.
+        # One day, we can make the block size a feature of the blocks, but
+        # for now, we'll just make sure that the block size in the real env
+        # matches what we expect in sim.
+        assert abs(task_spec["block_size"] - self.block_size) < 1e-6
```
williamshen-nz, line 529: Nit: I guess we could use `np.isclose` here

predicators/envs/blocks.py
```diff
@@ -509,3 +518,52 @@ def _get_highest_block_below(self, state: State, x: float, y: float,
...
+        on_args = goal_spec.get("On", [])
+        on_table_args = goal_spec.get("OnTable", [])
+        pred_to_args = {self._On: on_args, self._OnTable: on_table_args}
+        goal: Set[GroundAtom] = set()
+        for pred, args in pred_to_args.items():
+            for id_args in args:
+                obj_args = [id_to_obj[a] for a in id_args]
+                goal_atom = GroundAtom(pred, obj_args)
+                goal.add(goal_atom)
+        return Task(init_state, goal)
```
williamshen-nz, line 569: General question, but do we want to do validation of the JSONs at all?

e.g. if two blocks have the same `block_id` then things get weird, or are we assuming the JSONs will always be well formed? 

I guess in the `block_id` case we can just make the blocks map to a dict of `block_id` to the position/colors rather than a list

tests/envs/test_blocks.py
```diff
@@ -111,3 +115,66 @@ def test_blocks_failure_cases():
...
+
+
+def test_blocks_load_task_from_json():
+    """Tests for loading blocks test tasks from a JSON file."""
+    # Set up the JSON file.
+    task_spec = {
+        "problem_name":
+        "blocks_test_problem1",
+        "blocks": [{
+            "block_id": 0,
```
williamshen-nz, line 127: I was wondering but maybe we should give blocks more descriptive IDs so its easy for us to read and debug what's going on?

e.g. `block_id: "red_block"`

williamshen-nz (approved):
LGTM 🚀

predicators/envs/blocks.py
tomsilver: good point, I changed it to a dict as you suggested. I'll defer other validation

PR #1239, opened 2022-10-07
add pg3 with initialized policy approach

ryangpeixu:


predicators/approaches/initialized_pg3_approach.py
```diff
@@ -0,0 +1,35 @@
...
+    """Policy-guided planning for generalized policy generation (PG3) with
+    initialized policy."""
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "initialized_pg3"
+
+    @staticmethod
+    def _get_policy_search_initial_ldl() -> LiftedDecisionList:
+        # Initialize with intialized policy from file.
```
tomsilver, line 32: typo: initialized

tests/approaches/test_initialized_pg3_approach.py
```diff
@@ -0,0 +1,68 @@
+"""Test cases for the PG3 approach."""
```
tomsilver, line 1: PG3 initialized* approach

tests/approaches/test_initialized_pg3_approach.py
```diff
@@ -0,0 +1,68 @@
...
+
+from predicators import utils
+from predicators.approaches.initialized_pg3_approach import \
+    InitializedPG3Approach
+from predicators.envs import create_new_env
+from predicators.ground_truth_nsrts import get_gt_nsrts
+from predicators.structs import LDLRule, LiftedDecisionList
+
+
+@pytest.mark.parametrize("approach_name,approach_cls",
```
tomsilver, line 16: we use this feature when there are multiple arguments that we want to feed into the test (e.g., testing both pg3 and pg4 approaches). since we're only testing one thing here, you can remove this decorator and just define the approach name and approach cls within the test itself. for the latter, you can just do `approach = InitializedPG3Approach(env.predicates, ...)`

tests/approaches/test_initialized_pg3_approach.py
```diff
@@ -0,0 +1,68 @@
...
+        "env": env_name,
+        "approach": approach_name,
+    })
+
+    env = create_new_env(env_name)
+    nsrts = get_gt_nsrts(env.predicates, env.options)
+    train_tasks = env.get_train_tasks()
+
+    name_to_nsrt = {nsrt.name: nsrt for nsrt in nsrts}
+    approach = approach_cls(env.predicates, env.options, env.types,
```
tomsilver, line 31: I think you may be able to actually remove this and the following line, since it doesn't look like this approach gets used

tomsilver (commented):
nice work! here are few comments :)

tomsilver (approved):
LGTM, great work! :)

PR #1248, opened 2022-10-16
effect search operator learning

tomsilver:
New operator learning algorithm that performs a hill-climbing search over add effect sets. Successors are generated using backchaining and effect set deletion.

example:
```python predicators/main.py --env repeated_nextto_ambiguous --approach nsrt_learning --seed 0 --strips_learner effect_search```

### CODE REVIEW

File: predicators/nsrt_learning/strips_learning/effect_search_learner.py [hunk 1]
```diff
@@ -0,0 +1,424 @@
...
+                    yield (op, i), child, 1.0  # cost always 1
+
+        # Run hill-climbing search.
+        path, _, cost = utils.run_hill_climbing(initial_state=initial_state,
+                                                check_goal=lambda _: False,
+                                                get_successors=get_successors,
+                                                heuristic=heuristic)
+
+        # This effectively asserts harmlessness.
+        assert cost[-1] == 0
```
tomsilver, line 308: i'm not sure if we'll keep this long-term because we might add some kind of regularization to the cost to encourage simpler operator sets. but so long as we are doing this, we might as well add `early_termination_heuristic_thresh=0` to `run_hill_climbing`, which will terminate the search as soon as we find an effect set that has 0 cost.

predicators/nsrt_learning/strips_learning/effect_search_learner.py
```diff
@@ -0,0 +1,424 @@
...
+        return _EffectSets(param_option_to_groups)
+
+    @functools.lru_cache(maxsize=None)
+    def _effect_sets_to_pnads(
+            self, effect_sets: _EffectSets) -> List[PartialNSRTAndDatastore]:
+        # Start with the add effects and option specs.
+        pnads = []
+        for (option_spec, add_effects, keep_effects) in effect_sets:
+            parameterized_option, op_vars = option_spec
+            effect_vars = {v
```
tomsilver, line 345: maybe split this up into multiple lines to make less ugly

tests/nsrt_learning/strips_learning/test_effect_search_learner.py
```diff
@@ -0,0 +1,405 @@
...
+from predicators import utils
+from predicators.nsrt_learning.segmentation import segment_trajectory
+from predicators.nsrt_learning.strips_learning.effect_search_learner import \
+    EffectSearchSTRIPSLearner
+from predicators.structs import Action, LowLevelTrajectory, Predicate, \
+    Segment, State, Task, Type
+
+longrun = pytest.mark.skipif("not config.getoption('longrun')")
+
+
```
tomsilver, line 14: how much of these tests are directly copied? we should try to use parameterized testing to remove any copying

tomsilver (commented):
looking good! I can't give approval because I opened the PR, but after addressing the comments I'm happy

tests/nsrt_learning/strips_learning/test_effect_search_learner.py
NishanthJKumar: They're all directly copied (from `test_gen_to_spec_learner.py`). I'm not sure how to use parameterized testing to overcome this; could you maybe give an example of what you mean?

tomsilver: yeah here's one example: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/approaches/test_pg3_approach.py#L30 see also https://docs.pytest.org/en/6.2.x/parametrize.html

NishanthJKumar (approved):
Includes all Tom's changes; LGTM!

PR #1249, opened 2022-10-17
make ee arguments optional in pybullet robot creation

tomsilver:


predicators/pybullet_helpers/robots/__init__.py
```diff
@@ -20,9 +20,15 @@
...
-        physics_client_id: int) -> SingleArmPyBulletRobot:
+        robot_name: str,
+        physics_client_id: int,
+        ee_home_pose: Optional[Pose3D] = None,
+        ee_orientation: Optional[Quaternion] = None) -> SingleArmPyBulletRobot:
     """Create a single-arm PyBullet robot."""
+    if ee_home_pose is None:
+        ee_home_pose = (1.35, 0.6, 0.7)
+    if ee_orientation is None:
+        ee_orientation = (0.0, 0.0, 0.0, 1.0)
```
williamshen-nz, line 31: I guess you could set these as the defaults in the method signature and remove the Optional typing, but any specific reason to have an explicit None check?

williamshen-nz (approved):
LGTM 🚢 

Though I guess the defaults are specified for the current pybullet environments.

PR #1250, opened 2022-10-17
kinematic action space converter

tomsilver:
use FK and IK to convert actions in option learning

### DIFF
--- a/tests/nsrt_learning/test_option_learning.py
+++ b/tests/nsrt_learning/test_option_learning.py
@@ -3,6 +3,7 @@
 from unittest.mock import patch
 
 import numpy as np
+import pybullet as p
 import pytest
 
 import predicators.nsrt_learning.option_learning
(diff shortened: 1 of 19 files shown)


tests/nsrt_learning/test_option_learning.py
```diff
@@ -409,3 +412,21 @@ def test_create_action_converter():
...
+    physics_client_id = p.connect(p.DIRECT)
+    robot = create_single_arm_pybullet_robot("panda", physics_client_id)
+    robot.go_home()
+    env_action_arr = np.array(robot.get_joints())
+    converter = create_action_converter()
+    reduced_action_arr = converter.env_to_reduced(env_action_arr)
+    assert reduced_action_arr.shape == (4, )
+    env_action_arr2 = converter.reduced_to_env(reduced_action_arr)
+    reduced_action_arr2 = converter.env_to_reduced(env_action_arr2)
+    assert np.allclose(reduced_action_arr, reduced_action_arr2)
```
williamshen-nz, line 432: A check for `np.allclose(env_action_arr, env_action_arr2)` wouldn't hurt

williamshen-nz (approved):
:shipit:

tests/nsrt_learning/test_option_learning.py
tomsilver: isn't it possible that those won't match because FK isn't necessarily invertible?

williamshen-nz: Good point! I forgot about that.

Though I guess in this case, the robot is already at the given position (within the simulator) so IK should give something close by. That being said, we shouldn't check it as that's an implementation detail.

PR #1252, opened 2022-10-18
always create a new environment inside of option models 

tomsilver:


NishanthJKumar (commented):
I think this should be fine, since `predicators_behavior` does its own thing when the env is a BEHAVIOR env. But I'll let @wmcclinton provide the final approval.

wmcclinton (approved):
Yes, this is fine. Approved!

predicators/option_model.py
```diff
@@ -12,19 +12,19 @@
...
+from predicators.envs import BaseEnv, create_new_env
 from predicators.settings import CFG
 from predicators.structs import DefaultState, State, _Option
 
 
 def create_option_model(name: str) -> _OptionModelBase:
     """Create an option model given its name."""
     if name == "oracle":
-        env = get_or_create_env(CFG.env)
+        env = create_new_env(CFG.env)
```
tomsilver, line 23: note to self: also add `do_cache=False`

PR #1253, opened 2022-10-18
replace `pybullet_use_gui` setting with general `use_gui` flag...

tomsilver:
...which is then passed into `create_new_env`. The reason is that I want to be able to use the pybullet GUI for the "main" environment, but turn off the GUI for a secondary environment that will be used in the option model.

note that this should be merged before #1252 and that PR should be updated to pass `use_gui = False` to `create_new_env` in the option model code.

I will also make an announcement after this PR that `--pybullet_use_gui True` is deprecated and now `--use_gui` should be used.


NishanthJKumar (approved):
Mostly LGTM - nice catch! One minor comment.

PR #1255, opened 2022-10-19
add flag to disable repeated state checking in option model

tomsilver:
I did this because the oracle options for blocks repeat states, so you can't use the oracle option model with them. however, I'm now realizing that if we collect data from the oracle, we're going to have issues with the same state mapping to two different actions. so I probably need to change the oracle options themselves, rather than the option model. but I'm opening this PR anyway because it can't hurt to have this optional functionality and could be useful for debugging.


jorge-a-mendez (approved):
Looks good! Don't see a need to change anything.

PR #1257, opened 2022-10-19
a few fixes for the kinematic action space converter

tomsilver:


predicators/nsrt_learning/option_learning.py
```diff
@@ -286,12 +288,20 @@ class _KinematicActionConverter(_ActionConverter):
...
+    def _init(self) -> None:
         # Create a new PyBullet connection and robot.
         self._physics_client_id = p.connect(p.DIRECT)
         # Create the robot.
         self._robot = create_single_arm_pybullet_robot(CFG.pybullet_robot,
                                                        self._physics_client_id)
 
+    def __setstate__(self, state: Dict) -> None:
+        # Recreate the object to avoid issues with the PyBullet client.
+        del state  # unused
```
williamshen-nz, line 302: This feels like a hack, but I guess it's the best way to get around the problem for now

williamshen-nz (approved):
🚀 📦

predicators/nsrt_learning/option_learning.py
tomsilver: agreed

PR #1259, opened 2022-10-20
set the joint state at the end of IK

### LINKED ISSUE #1258: "real" version of pybullet environment leaking into sim
after we track down this issue, I may recommend purging `get_or_create_env` so we don't run into this again.

here's what I know so far:

```
def _get_from_env_by_names(env_name: str, names: Sequence[str],
                           env_attr: str) -> List:
    """Helper for loading types, predicates, and options by name."""

    # Changing this from == to != makes a big difference in performance.
    if env_attr == "options":
        env = get_or_create_env(env_name)
    else:
        env = create_new_env(env_name, do_cache=False)
```

see comment, while running:
```python predicators/main.py --env pybullet_blocks --approach oracle --seed 0 --num_test_tasks 1 --pybullet_robot panda```

on master.

I'm not sure yet exactly where in the options the leaking is occurring. maybe in IK? investigating now

### PR DESCRIPTION
**Pull request by tomsilver**
closes #1258

@williamshen-nz see comment for justification. let me know if it makes sense, and if you have any ideas for better workarounds. thanks!

### DIFF
--- a/predicators/pybullet_helpers/robots/single_arm.py
+++ b/predicators/pybullet_helpers/robots/single_arm.py
@@ -233,7 +233,7 @@ class SingleArmPyBulletRobot(abc.ABC):
     @cached_property
     def initial_joint_positions(self) -> JointPositions:
         """The joint values for the robot in its home pose."""
-        joint_positions = self.inverse_kinematics(self._ee_home_pose,
+        joint_positions = self.set_joints_with_ik(self._ee_home_pose,
                                                   validate=True)
         # The initial joint values for the fingers should be open. IK may
         # return anything for them.
@@ -260,8 +260,7 @@ class SingleArmPyBulletRobot(abc.ABC):
 
         # Now run IK to get to the actual starting rx, ry, rz. We use
         # validate=True to ensure that this initialization works.
-        joint_values = self.inverse_kinematics((rx, ry, rz), validate=True)
-        self.set_joints(joint_values)
+        self.set_joints_with_ik((rx, ry, rz), validate=True)
 
         # Handle setting the robot finger joints.
         for finger_id in [self.left_finger_id, self.right_finger_id]:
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: predicators/pybullet_helpers/robots/single_arm.py [hunk 1]
```diff
@@ -409,6 +409,21 @@ def _ikfast_inverse_kinematics(
...
+        # final_joint_state. This line of code was added to patch a difficult
+        # bug that arose when two versions of the environment were created: one
+        # that exposes the options used by the agent, and one that represents
+        # the "real" environment. In the former environment, IK was getting
+        # called, but the robot's state was not getting updated to match the
+        # state of the "real" environment, leading to subsequent IK failures.
+        # Ideally, none of this would be necessary, but the reality is that
+        # IK is very sensitive to initialization, so it is a stateful method
+        # masquerading as a stateless function.
+        self.set_joints(final_joint_state)
```
williamshen-nz, line 425: I understand the explanation, though I would think it should be up to the consumer to set the robot joints to whatever IK gives?

Like here: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/481c2a3f1f599fd49aedae3012332cffbe287898/predicators/pybullet_helpers/robots/single_arm.py#L261-L264

Also, if we modify this for the IKFast I think we need to do the same for the pybullet IK?

tomsilver: +1 to modifying pybullet IK too.

about the first part, I agree in principle, but in this case, the consumer is an option policy. it feels quite weird to let the option policy change the robot's state, since the option lives in the mind of the agent but the robot is in the environment...

also, if one day some magical IK solver comes along that is not sensitive to state, we would change this single_arm method, but the consumers shouldn't care...

on the other hand, it's possible that some future consumers would not follow the assumptions that I described in the comment, and so they should be responsible for dealing with that trickiness I guess.

I don't know. damnit IK, why are you so hard.

tomsilver: btw if we go the consumer route we'll have to change the kinematic action converter too

tomsilver: @williamshen-nz I'm leaning toward keeping this change in the robot itself, mostly to prevent changing multiple consumers. does the name change that I just made help?

williamshen-nz (approved):
I was going to suggest keeping `inverse_kinematics` and have a `move_to_pose` method that calls IK and then sets joints, but this does the same thing and we can separate them later if needed.

Looks good!

PR #1263, opened 2022-10-27
Interactive learning: Remove flaky classifier test

amburger66:
Remove flaky test from held out `Covers` classifier test states


tomsilver (approved):
lgtm!

PR #1266, opened 2022-10-28
Addressed long pytest in NSRT learning approaches

wmcclinton:
Fixed the long NSRT learning test by making it a longrun.


tomsilver (approved):
LGTM thanks! I confirmed the test speedup.

PR #1269, opened 2022-11-01
Fixed order invariance test

wmcclinton:
This PR fixes the order invariance test by making sure the NSRTs generated are the same as the correct ones.


NishanthJKumar (approved):
LGTM! Thanks for making this change!

PR #1270, opened 2022-11-03
Changes hillclimbing to do 2 successive steps of backchaining if 1 is not enough

NishanthJKumar:
Previously, we only ever called the `backchain` search operator during effect search once, expecting it to cover at least one additional transition. However, this might not happen if there is a keep effect issue. Thus, we explicitly detect such issues and run `backchain` twice in this case.

### DIFF
--- a/predicators/nsrt_learning/strips_learning/base_strips_learner.py
+++ b/predicators/nsrt_learning/strips_learning/base_strips_learner.py
@@ -260,6 +260,13 @@ class BaseSTRIPSLearner(abc.ABC):
                        not segment.necessary_add_effects.issubset(
                            ground_op.add_effects):
                         continue
+                    keep_effects = ground_op.preconditions & ground_op.add_effects
+                    # If the segment has a non-None necessary_image,
+                    # and the ground operator doesn't fit this.
+                    if segment.necessary_image is not None and \
+                        len(keep_effects) > 0 and \
+                        not segment.necessary_image.issubset(next_atoms):
+                        continue
                 else:
                     # If check_only_preconditions is True, we must be
                     # calling this from spawning during backchaining
--- a/predicators/nsrt_learning/strips_learning/base_strips_learner.py
+++ b/predicators/nsrt_learning/strips_learning/base_strips_learner.py
@@ -260,7 +260,8 @@ class BaseSTRIPSLearner(abc.ABC):
                        not segment.necessary_add_effects.issubset(
                            ground_op.add_effects):
                         continue
-                    keep_effects = ground_op.preconditions & ground_op.add_effects
+                    keep_effects = ground_op.preconditions & \
+                        ground_op.add_effects
                     # If the segment has a non-None necessary_image,
                     # and the ground operator doesn't fit this.
                     if segment.necessary_image is not None and \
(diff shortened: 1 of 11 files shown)


tomsilver (approved):
I'll trust that the finer details are correct here, but LGTM SWE-wise :)

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -260,6 +260,19 @@ def _find_best_matching_pnad_and_sub(
                        not segment.necessary_add_effects.issubset(
                            ground_op.add_effects):
                         continue
+                    keep_effects = ground_op.preconditions & \
```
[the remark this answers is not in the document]
NishanthJKumar: it shouldn't affect the old results, since in the `BackchainingSTRIPSLearner`, we don't change the `segment.necessary_image`, but you're right that this is probably a good thing to do.

predicators/nsrt_learning/strips_learning/effect_search_learner.py
```diff
@@ -120,17 +118,37 @@ class _BackChainingEffectSearchOperator(_EffectSearchOperator):
...
-                uncovered_transition
-            option_spec, lifted_add_effs, lifted_keep_effs = \
-                self._create_new_effect_set(param_option, option_objs, \
-                    add_effs, keep_effs)
-            new_effect_sets = effect_sets.add(option_spec, lifted_add_effs,
-                                              lifted_keep_effs)
-            yield new_effect_sets
+        initial_heuristic_val = self._associated_heuristic(effect_sets)
+
+        def _get_new_effect_sets_by_backchaining(
```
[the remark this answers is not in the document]
NishanthJKumar: we could; it's not used anywhere else, but it is definitely awkward to have it defined under another method. I'll move it out!

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -260,6 +260,19 @@ def _find_best_matching_pnad_and_sub(
...
+                        ground_op.add_effects
+                    # If the segment has a non-None necessary_image,
+                    # and there are keep effects, and the ground operator
+                    # doesn't fit this, skip.
+                    # NOTE: It's weird/janky to only be doing this check
+                    # when there are keep effects (we should always be
+                    # satisfying the necessary image I think?).
+                    # Will investigate in a future PR.
```
wmcclinton, line 271: Yes there should eventually be an assertion here.

predicators/nsrt_learning/strips_learning/effect_search_learner.py
```diff
@@ -118,19 +116,39 @@ def get_successors(self,
...
-        pnads = self._effect_sets_to_pnads(effect_sets)
+    def _get_new_effect_sets_by_backchaining(
+            self, curr_effect_sets: _EffectSets) -> _EffectSets:
+        new_effect_sets = curr_effect_sets
+        pnads = self._effect_sets_to_pnads(curr_effect_sets)
         uncovered_transition = self._get_first_uncovered_transition(pnads)
         if uncovered_transition is not None:
             param_option, option_objs, add_effs, keep_effs = \
-                uncovered_transition
+                    uncovered_transition
```
wmcclinton, line 126: Nit: Why tab?

wmcclinton (approved):
LGTM! I need to do some review of the hill-climbing code, but I didn't see anything wrong as far as I know.

predicators/nsrt_learning/strips_learning/effect_search_learner.py
NishanthJKumar: good catch; will revert!

PR #1272, opened 2022-11-04
Re-partitioning data in a way that's aware of unignorable effects

NishanthJKumar:
Previously, when converting from `_EffectSets` to `PNAD`s, we threw all predicates into the ignore effects. However, this is potentially problematic: consider the case where we realize that the transition `[A,D] -> [A, D, E, C]; Necessary Image=[A,D]` is uncovered. We want to induce an operator with null add effects (since `E` and `C` here are non-necessary add effects). However if we create a null effect set, then this will be converted into a PNAD with all predicates in its ignore effects. This PNAD does not achieve the required necessary image if applied from the `segment.init_atoms`, so this datapoint will not flow into this PNAD, and it will not cover the transition that it was literally induced to cover!

We now fix this by listing certain predicates that should not be ignored when converting `_EffectSets` into `PNAD`s. This requires some additional changes to keep effect induction. Now, we can assert that whenever a new operator is induced, it must cover the transition that caused it to be induced.

Merge only after merging #1270

### DIFF
--- a/predicators/envs/assets/pddl/gripper/domain.pddl
+++ b/predicators/envs/assets/pddl/gripper/domain.pddl
@@ -1,0 +1,34 @@
+
+(define (domain gripper)
+   (:predicates (room ?r)
+		(ball ?b)
+		(gripper ?g)
+		(at-robby ?r)
+		(at ?b ?r)
+		(free ?g)
+		(carry ?o ?g))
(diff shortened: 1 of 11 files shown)


NishanthJKumar:
Subsumed by #1288

PR #1274, opened 2022-11-04
Include operator complexity into objective term

NishanthJKumar:
Previously, our objective for hill-climbing during effect search operator learning was only data coverage. Now, we also include an operator complexity term.

This allows us to solve the `test_backchaining_strips_learner_order_dependence` test **without any randomized ordering search or harmlessness-based op pruning!!!** This was the original motivation for developing this effects search operator learner.

It still remains to be tested whether this new learner can (1) pass Rohan's randomly-generated test-suite (`test_backchaining_randomly_generated`), (2) handle all our domains from the original AAAI paper.

Merge only after merging #1272

### DIFF
--- a/predicators/envs/assets/pddl/gripper/domain.pddl
+++ b/predicators/envs/assets/pddl/gripper/domain.pddl
@@ -1,0 +1,34 @@
+
+(define (domain gripper)
+   (:predicates (room ?r)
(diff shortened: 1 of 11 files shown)


NishanthJKumar:
Subsumed by #1288

PR #1275, opened 2022-11-08
fix small bug in discovered failure predicate handling in planning

tomsilver:
before this, ```python predicators/main.py --env cluttered_table --approach oracle ``` was leading to a weird number of predicates in the metrics.


wmcclinton (approved):
LGTM! I verified it worked.

PR #1278, opened 2022-11-09
use env argument in get_gt_nsrts instead of CFG.env

tomsilver:
we should've always done this, and now I need it for another forthcoming PR


wmcclinton (approved):
LGTM! I don't understand why this is better, but I won't question the master.

PR #1286, opened 2022-11-12
make pybullet robot ee orns global settings 

tomsilver:
should fix a bug in the kinematic action space converter where a bad default (0.0, 0.0, 0.0, 0.1) end effector rotation was getting used during IK (and the wrist was facing upward). also tries to head off similar issues in the future by making the default end effector rotation a global setting.

### DIFF
--- a/predicators/settings.py
+++ b/predicators/settings.py
@@ -112,6 +112,21 @@ class GlobalSettings:
     pybullet_birrt_smooth_amt = 50
     pybullet_birrt_extend_num_interp = 10
     pybullet_control_mode = "position"
+    pybullet_robot_ee_orns = {
+        # Fetch and Panda gripper down and parallel to x-axis.
+        "pybullet_cover": {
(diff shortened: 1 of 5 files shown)


predicators/settings.py
```diff
@@ -112,21 +112,21 @@ class GlobalSettings:
     pybullet_birrt_smooth_amt = 50
     pybullet_birrt_extend_num_interp = 10
     pybullet_control_mode = "position"
-    pybullet_robot_ee_orns = {
-        # Fetch and Panda gripper down and parallel to x-axis.
-        "pybullet_cover": {
+    # env -> robot -> quaternion
+    pybullet_robot_ee_orns = defaultdict(
```
williamshen-nz, line 116: I guess specifying rpy is more verbose, if it wasn't the cause of the issue we were talking about

williamshen-nz (approved):

PR #1289, opened 2022-11-13
flag to use holding-only goals in blocks

tomsilver:
example:
```python predicators/main.py --env pybullet_blocks --approach oracle --seed 0 --use_gui --blocks_holding_goals True --blocks_num_blocks_test '[1]'```


NishanthJKumar (approved):
LGTM! Looks like a useful feature!

PR #1291, opened 2022-11-14
Interactive predicate learning plots: Add star marker

amburger66:
- add star marker so No Actions line is visible
- remove plot titles bc they're not needed

example:
action_baselines_in_coverenv_excluding_covers,holding_num_online_transitions_perc_solved


tomsilver (approved):
lgtm pending checks. can you also put an example output image in the PR description? thanks!

PR #1292, opened 2022-11-14
add tolerance to joint limit check 

tomsilver:


predicators/pybullet_helpers/joint.py
```diff
@@ -55,11 +55,12 @@ def is_fixed(self) -> bool:
         """Whether the joint is fixed or not."""
         return self.jointType == p.JOINT_FIXED
 
-    def violates_limit(self, value: float) -> bool:
```
williamshen-nz, line 58: I'm not sure whether we want to keep this here in the long term. 

We should instead support additional rules or scores on top of joint limits when calling IK. 

e.g. Encourage it to find solutions that are close to the current configuration but far from joint limits, solutions that are close to a given configuration, etc. 

For now it's probably ok though, but I imagine it may cause IK errors and drop coverage in things.

williamshen-nz (approved):
🚢  made some notes

predicators/pybullet_helpers/joint.py
tomsilver: yeah agreed. to be revisited!

PR #1300, opened 2022-11-16
allow PG3 init approach to suggest multiple initial LDLs

tomsilver:
this is because we want to consider multiple analogical mappings. for now we just use the "best" initial LDL according to the heuristic and discard the rest, but in the future we may instead want to initialize the search with all candidates. that would require a deeper code change because the heuristic search utility functions all assume a single initial state.

### DIFF
--- a/predicators/approaches/initialized_pg3_approach.py
+++ b/predicators/approaches/initialized_pg3_approach.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 
 import os
 from dataclasses import dataclass
-from typing import Dict, Set
+from typing import Dict, List, Set
 
 import dill as pkl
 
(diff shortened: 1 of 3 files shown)


predicators/approaches/initialized_pg3_approach.py
```diff
@@ -99,17 +101,17 @@ class _Analogy:
     types: Dict[Type, Type]
 
 
-def _find_env_analogy(base_env: BaseEnv, target_env: BaseEnv,
-                      base_nsrts: Set[NSRT],
-                      target_nsrts: Set[NSRT]) -> _Analogy:
+def _find_env_analogies(base_env: BaseEnv, target_env: BaseEnv,
```
sebdumancic, line 104: _find_env_analogy implementation in the other branch already supports outputting a list of mappings. It just needs one small modification. Let me know when should I do that.

sebdumancic (approved):

PR #1301, opened 2022-11-16
add flag for pg3 add condition operator fresh var creation

tomsilver:
setting this flag to False makes the branching factor much lower, and so the search over policies much faster, but I'm not sure if we can get away with it in general, so defaulting to True.


sebdumancic (approved):
This makes sense to have as an option.  As long as it is a single flag, we can see its impact when we progress more.

PR #1302, opened 2022-11-16
fixes to apply_analogy_to_ldl to handle partial analogies

tomsilver:


predicators/approaches/initialized_pg3_approach.py
```diff
@@ -116,18 +116,31 @@ def _apply_analogy_to_ldl(analogy: _Analogy,
                           ldl: LiftedDecisionList) -> LiftedDecisionList:
     new_rules = []
     for rule in ldl.rules:
+        # Can't create a rule if there is no NSRT match.
+        if rule.nsrt not in analogy.nsrts:
+            continue
+        # Can't create a rule if there is no type match.
+        if not all(p.type in analogy.types for p in rule.parameters):
```
sebdumancic, line 123: I think this is fine for now. In the future, we could derive a useful rule skeleton/initial rule even if we miss some of the types and, consequently, variables

sebdumancic (approved):

predicators/approaches/initialized_pg3_approach.py
tomsilver: true. this might change soon once we figure out how to remove types and variables from _Analogy. but I'll merge this anyway

PR #1305, opened 2022-11-16
Recomputes PNADs after 'delete' operator during hill-climbing

NishanthJKumar:
Previously, we always just deleted PNADs during hill-climbing to form a new PNAD set without recomputing them. Rather, we'd always recompute when we started to compute our heuristic. But it makes much more sense to recompute after deletion instead.

### DIFF
--- a/predicators/nsrt_learning/strips_learning/pnad_search_learner.py
+++ b/predicators/nsrt_learning/strips_learning/pnad_search_learner.py
@@ -144,9 +144,10 @@ class _PruningPNADSearchOperator(_PNADSearchOperator):
     def get_successors(
         self, pnads: FrozenSet[PartialNSRTAndDatastore]
     ) -> Iterator[FrozenSet[PartialNSRTAndDatastore]]:
-        for pnad_to_remove in sorted(pnads):
+        sorted_pnad_list = sorted(pnads)
+        for pnad_to_remove in sorted_pnad_list:
             pnads_after_removal = [
-                pnad for pnad in sorted(pnads) if pnad != pnad_to_remove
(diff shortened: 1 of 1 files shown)


predicators/nsrt_learning/strips_learning/pnad_search_learner.py
```diff
@@ -208,7 +208,7 @@ def __call__(self,
         # accurate measures that also take into account the add effects,
         # arity, etc. (though this might involve changing the weighting
         # of the coverage term).
-        complexity_term = len(recomp_curr_pnads)
+        complexity_term = len(sorted(curr_pnads))
```
tomsilver, line 211: no need to sort here

predicators/nsrt_learning/strips_learning/pnad_search_learner.py
```diff
@@ -144,10 +144,13 @@ class _PruningPNADSearchOperator(_PNADSearchOperator):
     def get_successors(
         self, pnads: FrozenSet[PartialNSRTAndDatastore]
     ) -> Iterator[FrozenSet[PartialNSRTAndDatastore]]:
-        # NOTE: Sorting done here to maintain determinism.
-        for pnad_to_remove in sorted(pnads,
-                                     key=lambda p: len(p.op.add_effects)):
-            yield frozenset([pnad for pnad in pnads if pnad != pnad_to_remove])
+        for pnad_to_remove in sorted(pnads):
+            pnads_after_removal = [
+                pnad for pnad in sorted(pnads) if pnad != pnad_to_remove
```
tomsilver, line 149: sort them once at the top?

tomsilver (approved):

PR #1308, opened 2022-11-17
fixes egregious pnad copying bug

NishanthJKumar:
Previously, our `recompute_pnads_from_effects` function would directly mutate underlying PNAD objects. This would change the state of previously-achieved nodes in hill-climbing, which is EXTREMELY unsafe (I frankly have no clue how our tests were passing...). This PR ensures that recomputing PNADs also *copies* them.


tomsilver (approved):
nice catch!

PR #1309, opened 2022-11-17
delete condition search operator

sebdumancic:
PG3 operators that creates candidate rules by deleting conditions

### DIFF
--- a/predicators/approaches/pg3_approach.py
+++ b/predicators/approaches/pg3_approach.py
@@ -299,76 +299,44 @@ class _DeleteConditionPG3SearchOperator(_PG3SearchOperator):
     def get_successors(
             self, ldl: LiftedDecisionList) -> Iterator[LiftedDecisionList]:
         for rule_idx, rule in enumerate(ldl.rules):
-            for destination in ("pos", "neg", "goal"):
-                if destination == "pos":
-                    dest_set = set(rule.pos_state_preconditions)
-                elif destination == "neg":
-                    dest_set = set(rule.neg_state_preconditions)
-                else:
-                    assert destination == "goal"
-                    dest_set = set(rule.goal_preconditions)
+            for condition in rule.pos_state_preconditions | \
+                rule.neg_state_preconditions | rule.goal_preconditions:
 
-                for condition in dest_set:
-                    # if the condition to be removed
-                    #    is a precondition of an nsrt, don't remove it
-                    if condition in rule.nsrt.preconditions:
-                        continue
+                # If the condition to be removed is a
+                # precondition of an nsrt, don't remove it.
+                if condition in rule.nsrt.preconditions:
+                    continue
 
-                    #rcreate new preconditions
-                    new_pos = set(rule.pos_state_preconditions)
-                    new_neg = set(rule.neg_state_preconditions)
-                    new_goal = set(rule.goal_preconditions)
+                # Recrete new preconditions.
+                # Assumes that a conditon can appear only in one set
+                new_pos = \
+                    set(rule.pos_state_preconditions).difference({condition})
+                new_neg = \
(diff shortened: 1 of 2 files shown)


predicators/approaches/pg3_approach.py
```diff
@@ -293,6 +293,84 @@ def _get_candidate_conditions(
...
+                    assert destination == "goal"
+                    dest_set = set(rule.goal_preconditions)
+
+                for condition in dest_set:
+                    # if the condition to be removed
+                    #    is a precondition of an nsrt, don't remove it
+                    if condition in rule.nsrt.preconditions:
+                        continue
+
+                    #rcreate new preconditions
```
tomsilver, line 317: typo: recreate*

predicators/approaches/pg3_approach.py
```diff
@@ -293,6 +293,84 @@ def _get_candidate_conditions(
...
+                if destination == "pos":
+                    dest_set = set(rule.pos_state_preconditions)
+                elif destination == "neg":
+                    dest_set = set(rule.neg_state_preconditions)
+                else:
+                    assert destination == "goal"
+                    dest_set = set(rule.goal_preconditions)
+
+                for condition in dest_set:
+                    # if the condition to be removed
```
tomsilver, line 312: nit: capitalize comments (and ideally end them with periods, but that one we're less sticklers about)

predicators/approaches/pg3_approach.py
```diff
@@ -293,6 +293,84 @@ def _get_candidate_conditions(
...
+            for destination in ("pos", "neg", "goal"):
+                if destination == "pos":
+                    dest_set = set(rule.pos_state_preconditions)
+                elif destination == "neg":
+                    dest_set = set(rule.neg_state_preconditions)
+                else:
+                    assert destination == "goal"
+                    dest_set = set(rule.goal_preconditions)
+
+                for condition in dest_set:
```
tomsilver, line 311: shouldn't this only be checked if the destination if "pos"?

tomsilver: also, it looks like `dest_set` isn't getting used below here, maybe can remove it?

predicators/approaches/pg3_approach.py
```diff
@@ -293,6 +293,84 @@ def _get_candidate_conditions(
...
+                    # remove the focus predicate
+                    if destination == "pos":
+                        new_pos.remove(condition)
+                    elif destination == "neg":
+                        new_neg.remove(condition)
+                    else:
+                        assert destination == "goal"
+                        new_goal.remove(condition)
+
+                    # check that the parameters of the removed
```
tomsilver, line 331: this looks reasonable, but it might be cleaner to just totally reconstruct the parameters of the rule like this: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/predicators/approaches/initialized_pg3_approach.py#L143

tomsilver: (in fact, maybe we should just write a helper function to do this and use it everywhere)

tests/approaches/test_pg3_approach.py
```diff
@@ -276,6 +277,77 @@ def test_pg3_search_operators():
         for rule in ldl.rules:
             assert set(rule.parameters).issubset(rule.nsrt.parameters)
 
+    # _DeleteConditionPG3SearchOperator
```
tomsilver, line 280: awesome tests!

tomsilver (commented):
@sebdumancic thanks so much for jumping through the github check hoops!! here are some small notes on the code -- if at any point you want to turn your attention back to more interesting things, just let me know and I can take care of these points.

tomsilver (approved):
looks great! thanks so much for doing this!

PR #1312, opened 2022-11-18
initialize PG3 by analogy via SME

tomsilver:


tomsilver:
```
---------- coverage: platform darwin, python 3.9.4-final-0 -----------
Name                                                 Stmts   Miss  Cover   Missing
----------------------------------------------------------------------------------
predicators/approaches/initialized_pg3_approach.py     147     58    61%   115-118, 125-138, 183-185, 206-231, 240-266, 271-273, 279, 286-293
```

predicators/approaches/initialized_pg3_approach.py
```diff
@@ -98,20 +107,58 @@ class _Analogy:
...
+    nsrt_variables: Dict[Tuple[NSRT, Variable], Tuple[NSRT, Variable]]
+
+    @cached_property
+    def types(self) -> Dict[Type, Type]:
+        """Infer type analogy from variables."""
+        type_map: Dict[Type, Type] = {}
+        for nsrt in self.nsrts:
+            var_map = self.base_nsrt_to_variable_analogy(nsrt)
+            for old_var, new_var in var_map.items():
+                type_map[old_var.type] = new_var.type
```
sebdumancic, line 121: Should we make sure that the SME result consistently matches the types, e.g., base variables of type A and matched exclusively to target variables of type B? I'm not sure if that would happen often, but SME is generally not type-aware. [I can make it, though, that wouldn't be too difficult]

sebdumancic: I think that suffices

sebdumancic (approved):

predicators/approaches/initialized_pg3_approach.py
tomsilver: I started to go down this road, but it's a bit annoying to test because we don't have domains that actually have different types yet (since ferry and gripper just have 'object'). and in fact, we might want to think about just converting all types to unary predicates.... so for now I'll leave it alone

PR #1314, opened 2022-11-21
CoverEnvHandEmpty

amburger66:
- this probably needs a better name
- makes `hand_empty` an attribute of robot so that interactive learning can learn all the Cover predicates

### DIFF
--- a/predicators/envs/cover.py
+++ b/predicators/envs/cover.py
@@ -364,6 +364,135 @@ class CoverEnv(BaseEnv):
         return False
 
 
+class CoverEnvHandEmpty(CoverEnv):
+    """Toy cover domain where the robot has a feature indicating whether its
+    hand is empty or not.
+
+    This allows us to learn all the predicates (Cover, Holding,
+    HandEmpty) through interactive learning.
+    """
+
+    def __init__(self, use_gui: bool = True) -> None:
+        super().__init__(use_gui)
+
+        # Add attribute.
+        self._robot_type = Type("robot",
+                                ["hand", "pose_x", "pose_z", "hand_empty"])
+        # Override HandEmpty predicate.
+        self._HandEmpty = Predicate("HandEmpty", [self._robot_type],
+                                    self._HandEmpty_holds)
+        # Create new robot because of new robot type
+        self._robot = Object("robby", self._robot_type)
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "cover_handempty"
+
+    def simulate(self, state: State, action: Action) -> State:
+        """This differs from CoverEnv because the hand_empty attribute of the
+        robot is set."""
+        assert self.action_space.contains(action.arr)
+        pose = action.arr.item()
+        next_state = state.copy()
+        hand_regions = self._get_hand_regions(state)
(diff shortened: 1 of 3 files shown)


predicators/envs/cover.py
```diff
@@ -364,6 +364,130 @@ def _any_intersection(self,
...
+        self._HandEmpty = Predicate("HandEmpty", [self._robot_type],
+                                    self._HandEmpty_holds)
+        # Create new robot because of new robot type
+        self._robot = Object("robby", self._robot_type)
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "cover_handempty"
+
+    def simulate(self, state: State, action: Action) -> State:
```
tomsilver, line 391: please refactor this and `_create_initial_state` to remove as much redundant code as possible 😇 the changes w.r.t. to the parent class should be <10 lines

predicators/envs/cover.py
```diff
@@ -86,6 +86,9 @@ def simulate(self, state: State, action: Action) -> State:
         if held_block is None and above_block is not None:
             grasp = pose - state.get(above_block, "pose")
             next_state.set(self._robot, "hand", pose)
+            if "hand_empty" in self._robot_type.feature_names:
```
amburger66, line 89: @tomsilver is this if stmt ok or is there a better way to do this?

tomsilver: hmmm yeah this seems fine

predicators/envs/cover.py
```diff
@@ -364,6 +378,36 @@ def _any_intersection(self,
         return False
 
 
+class CoverEnvHandEmpty(CoverEnv):
+    """Toy cover domain where the robot has a feature indicating whether its
+    hand is empty or not.
+
+    This allows us to learn all the predicates (Cover, Holding,
+    HandEmpty) through interactive learning.
```
tomsilver, line 386: instead of saying "through interactive learning" let's say "with the assumption that the predicates are a function of only their argument's states"

tests/envs/test_cover.py
```diff
@@ -117,6 +117,81 @@ def test_cover():
     assert "did not implement an interface for human demonstrations" in str(e)
 
 
+def test_cover_hand_empty():
```
tomsilver, line 120: a lot of this looks the same as the regular cover test. let's use parameterized testing to remove the redundant code. then if there are specific things about the CoverEnvHandEmpty environment that we should check, we can just include those things in this function.

to see what I mean by "parameterized test", here are a few examples:
* https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/explorers/test_glib_explorer.py#L12
* https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/approaches/test_gnn_option_policy_approach.py#L46
* https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/tests/pybullet_helpers/test_link.py#L31

tomsilver (commented):
main stuff looks good, but a suggestion for refactoring the tests

tomsilver (approved):
LGTM!

PR #1315, opened 2022-11-21
`--crash_on_failure` flag 

tomsilver:
will be useful for real robot pipeline where we don't want to continue if planning fails


williamshen-nz (approved):
:shipit:

PR #1317, opened 2022-11-23
convert blocks workspace assertion to warning

tomsilver:
the assertion crash is annoying if we're using the pybullet blocks simulator to plan, but it's still useful to get the warning, especially in the case where we're actually using the original blocks simulator to plan.


williamshen-nz (approved):
📦 🚚

PR #1319, opened 2022-11-23
use pybullet_robot_sim in options instead of "real" pybullet_robot

### LINKED ISSUE #1296: pybullet reset mode rendering leads to lagging block position
seems to only be a rendering issue. to reproduce:

```
python predicators/main.py --env pybullet_blocks --approach oracle --seed 0 --use_gui --pybullet_robot panda --make_test_videos --pybullet_camera_width 1674 --pybullet_camera_height 900 --pybullet_control_mode reset
```

example:

https://user-images.githubusercontent.com/2816502/202009350-92aa3619-79be-44c8-b433-fc022d03d56f.mp4

tomsilver:
closes #1296, but more importantly, fixes leaky abstraction. `pybullet_robot` is the real robot, in the environment. the agent should not be able to set that robot's state directly -- the agent can only give actions to the environment. `pybullet_robot_sim` is the agent's simulated robot, which we were already using for forward kinematics. now we'll also use it for inverse kinematics.


williamshen-nz (approved):
:shipit: Maybe giving `pybullet_robot_2` another name would help people avoid this issue. Since the planner is using that one to think in its head.

Not sure what a good name would be though...

PR #1321, opened 2022-11-28
save num_samples metrics for refinement learning

lilianluong:
**Changes:**

- `run_low_level_search` takes in a metrics parameter and saves the total `num_tries` as "num_samples"
- bilevel planning approach saves min, max, and total num_samples
- add min, max, and avg num_samples to `main.py`
- `analyze_results_directory.py` outputs min, max, avg num_samples (this is commented out for the PR)
- *minor change to NarrowPassageEnv: door and passage width padding are now config variables in `settings.py`

**Example use**

1. Uncomment metrics in `analyze_results_directory.py`
2. Clear results directory
3. Run `main.py` using the `oracle` approach
4. Run `analyze_results_directory.py`


tomsilver (commented):
looks really good. my one question is whether we should be accumulating the statistic over multiple skeletons. I think yes? but let me know if you disagree. if I understand right now, the number of samples is only recorded for the last skeleton optimized.

PR #1323, opened 2022-11-28
added refinement cost estimator classes and refinement_estimation approach

lilianluong:
###  Refinement cost estimator classes and RefinementEstimationApproach

**Changes:**

- Created `predicators/refinement_estimators/` which contains an abstract `BaseRefinementEstimator` and implemented `OracleRefinementEstimator` class (oracle implemented for `narrow_passage` env only)
- Added tests for the refinement cost estimators
- Created a new `RefinementEstimationApproach` which passes a refinement estimator to the planner, and tests based on the oracle approach tests
- `sesame_plan()` takes in an optional refinement estimator. If it is provided, during A* search we pre-generate N skeletons, sort them by estimated refinement cost, and attempt to refine them in that order
- New CFG parameters:
    - `refinement_estimator` is a choice of refinement estimator (default: `oracle`)
    - `refinement_estimation_num_skeletons_generated` is the number of skeletons to pre-generate

**Example usage**

`python main.py --env narrow_passage --approach refinement_estimation --seed 0`

**Tests**

`pytest tests/approaches/test_refinement_estimation_approach.py` (takes approx 4.1s)

`pytest tests/refinement_estimators/` (takes approx 1.4s total)

### DIFF
--- a/predicators/approaches/refinement_estimation_approach.py
+++ b/predicators/approaches/refinement_estimation_approach.py
@@ -1,0 +1,64 @@
+"""A bilevel planning approach that uses a refinement cost estimator.
+
+Generates N proposed skeletons and then ranks them based on a given
+refinement cost estimation function (e.g. a heuristic, learned model),
+attempting to refine them in this order.
+"""
+
+from typing import Any, List, Set, Tuple
+
+from gym.spaces import Box
+
+from predicators.approaches.bilevel_planning_approach import \
+    BilevelPlanningApproach
+from predicators.ground_truth_nsrts import get_gt_nsrts
+from predicators.refinement_estimators import create_refinement_estimator
+from predicators.settings import CFG
+from predicators.structs import NSRT, Metrics, ParameterizedOption, \
+    Predicate, Task, Type, _Option
+
+
+class RefinementEstimationApproach(BilevelPlanningApproach):
+    """A bilevel planning approach that uses a refinement cost estimator."""
+
+    def __init__(self,
+                 initial_predicates: Set[Predicate],
+                 initial_options: Set[ParameterizedOption],
+                 types: Set[Type],
+                 action_space: Box,
+                 train_tasks: List[Task],
+                 task_planning_heuristic: str = "default",
+                 max_skeletons_optimized: int = -1) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks, task_planning_heuristic,
+                         max_skeletons_optimized)
+        self._nsrts = get_gt_nsrts(CFG.env, self._initial_predicates,
+                                   self._initial_options)
+        self._refinement_estimator = create_refinement_estimator(
+            CFG.refinement_estimator)
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "refinement_estimation"
+
+    @property
+    def is_learning_based(self) -> bool:
+        return False
+
+    def _get_current_nsrts(self) -> Set[NSRT]:
+        return self._nsrts
+
+    def _run_sesame_plan(self, task: Task, nsrts: Set[NSRT],
+                         preds: Set[Predicate], timeout: float, seed: int,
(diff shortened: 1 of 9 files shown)


predicators/approaches/refinement_estimation_approach.py
```diff
@@ -0,0 +1,64 @@
...
+                 initial_options: Set[ParameterizedOption],
+                 types: Set[Type],
+                 action_space: Box,
+                 train_tasks: List[Task],
+                 task_planning_heuristic: str = "default",
+                 max_skeletons_optimized: int = -1) -> None:
+        super().__init__(initial_predicates, initial_options, types,
+                         action_space, train_tasks, task_planning_heuristic,
+                         max_skeletons_optimized)
+        self._nsrts = get_gt_nsrts(CFG.env, self._initial_predicates,
```
tomsilver, line 35: let's subclass `OracleApproach` if we're going to use the ground-truth NSRTs like this? that will also save you needing to implement `is_learning_based` and `_get_current_nsrts`

predicators/refinement_estimators/base_refinement_estimator.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def get_name(cls) -> str:
+        """Get the unique name of this refinement cost estimator, for future
+        use as the argument to `--refinement_estimator`."""
+        raise NotImplementedError("Override me!")
+
+    @abc.abstractmethod
+    def get_cost(self, skeleton: List[_GroundNSRT]) -> float:
```
tomsilver, line 21: I don't know if we'll end up using, this but why don't we also pass in the expected atoms sequence, in addition to the skeleton. that sequence accompanies each skeleton in `gen`

predicators/refinement_estimators/base_refinement_estimator.py
```diff
@@ -0,0 +1,28 @@
...
+        """Return an estimated cost for a proposed high-level skeleton."""
+        raise NotImplementedError("Override me!")
+
+
+class RefinementCostEstimationFailure(ExceptionWithInfo):
```
tomsilver, line 26: interesting -- what do you have in mind for this?

predicators/planning.py
```diff
@@ -175,6 +178,21 @@ def _sesame_plan_with_astar(
                 timeout - (time.perf_counter() - start_time), metrics,
                 max_skeletons_optimized, abstract_policy,
                 max_policy_guided_rollout, use_visited_state_set)
+            # If a refinement cost estimator is provided, generate a number of
```
tomsilver, line 181: i'm really pleased that this is so short, nice job

predicators/planning.py
```diff
@@ -175,6 +178,21 @@ def _sesame_plan_with_astar(
...
+            # skeletons first, then predict the refinement cost of each skeleton
+            # and attempt to refine them in this order.
+            if refinement_estimator is not None:
+                estimator: BaseRefinementEstimator = refinement_estimator
+                proposed_skeletons = []
+                for _ in range(
+                        CFG.refinement_estimation_num_skeletons_generated):
+                    try:
+                        proposed_skeletons.append(next(gen))
+                    except _MaxSkeletonsFailure:
```
tomsilver, line 191: maybe we should actually crash in this case (with a helpful error message). this can only happen if `refinement_estimation_num_skeletons_generated` is greater than `sesame_max_skeletons_optimized`, and it seems to me that we would never want to actually be in that situation

tomsilver (commented):
excellent work!! just a few comments

predicators/refinement_estimators/base_refinement_estimator.py
lilianluong: I think I had a plan for this and decided I didn't need it, then forgot to remove it. I'll remove it now haha

tomsilver (approved):
LGTM!

PR #1325, opened 2022-11-29
add image capture from realsense cameras to real blocks pipeline

tomsilver:
`realsense_helpers.py` is a pared down version of code by @williamshen-nz


williamshen-nz (approved):
:ship: Looks good, the realsense helpers are probably more complex than they need to be but it'll be needed if we move to multi-camera setup.

PR #1326, opened 2022-11-30
LISDF visualizer script

tomsilver:
this is getting embarrassingly hacky and redundant, but that's a problem for future Tom


williamshen-nz (approved):
:shipit: 

Still much cleaner than 99% of other research code!

We should move this to LISDF at some point, but that's going to require adding a bunch of pybullet dependencies which is more effort than it's worth

PR #1327, opened 2022-11-30
removes unnecessary call to learning

NishanthJKumar:
This slows us down somewhat substantially when learning takes a while because we're unnecessarily calling learning twice!

@wmcclinton - is it safe to take out this line entirely (seems so to me)? If not, can we at least move it so that it doesn't get called if `CFG.enable_harmless_op_pruning` is False?


tomsilver (approved):
good catch, I can't believe we didn't find this earlier!

wmcclinton (approved):
Damn. You are right good catch!

PR #1328, opened 2022-11-30
real panda execution final scripts

tomsilver:
the main new python code implements motion planning to get the robot from its current position to the initial task position


scripts/lisdf_plan_to_reset.py
```diff
@@ -0,0 +1,105 @@
...
+    ]
+    plan_init_joint_lst = [
+        plan_init_joints[joint_name] for joint_name in joint_names
+    ]
+    motion_plan = run_motion_planning(robot,
+                                      current_joint_lst,
+                                      plan_init_joint_lst,
+                                      collision_bodies={table_id},
+                                      seed=seed,
+                                      physics_client_id=physics_client_id)
```
williamshen-nz, line 84: I had a deep dive into this method. There might be an issue with self-collisions (I think maybe in general as well as we don't explicitly handle it)?

https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/predicators/pybullet_helpers/motion_planning.py#L27

Motion planning samples joint positions from the `action_space` of the robot, which is just between the lower and upper limits.

scripts/lisdf_plan_to_reset.py
```diff
@@ -0,0 +1,105 @@
...
+    assert motion_plan is not None
```
williamshen-nz, line 85: Could `motion_plan == []`? Would that happen when the robot is already at the start position (usually it would never exactly be at that start position but close I'd think as you can't get the joint positions to that level of precision).

In that case, we can just write the existing plan I guess.

williamshen-nz (approved):
Looks good in general - had some comments, mostly minor or some fundamental ones.

scripts/lisdf_plan_to_reset.py
```diff
@@ -0,0 +1,105 @@
...
+    first_path_command = cast(JointSpacePath, path_commands[0])
+    return first_path_command.waypoint(0)
+
+
+def _main(lisdf_filepath: Path, output_filepath: Path, seed: int = 0) -> None:
+    utils.reset_config({"seed": seed})
+    # Load the LISDF plan.
+    with open(lisdf_filepath, "r", encoding="utf-8") as f:
+        plan_json = f.read()
+        plan = cast(LISDFPlan, LISDFPlan.from_json(plan_json))
```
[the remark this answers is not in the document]
tomsilver: mypy thinks that `plan` is `OutputElement` when it needs to be `LISDFPlan`. is that not expected?

scripts/lisdf_plan_to_reset.py
tomsilver: yeah I think if it's empty then everything should work

scripts/lisdf_plan_to_reset.py
tomsilver: following up about this on slack

scripts/lisdf_plan_to_reset.py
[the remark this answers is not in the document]
williamshen-nz: I guess the typing returns `OutputElement` not `LISDFPlan` as its defined in the superclass. It's fine though, we can figure it out later.

Issue #1324, opened 2022-11-29
fix flakey pybullet cover test


tomsilver:
https://github.com/Learning-and-Intelligent-Systems/predicators/actions/runs/3575810051/jobs/6012821902

tomsilver:
I can't reproduce this. Closing until it comes up again.

tomsilver:
came up again: https://github.com/Learning-and-Intelligent-Systems/predicators/actions/runs/3588924526/jobs/6042167238

tomsilver:
still digging, but was able to reproduce by running
```
pytest -s tests/envs/test_pybullet_cover.py -k test_pybullet_cover_pick_workspace_bounds
```
~10 times.

it looks like the `MoveEndEffectorToPose` component of the `PickPlace` option is sometimes getting stuck. it proposes an action repeatedly
```
[-1.0354177 ,  0.6088901 ,  0.3550845 , -1.6462139 , -0.25037703,
        2.2100015 ,  1.7674462 ,  0.04      ,  0.04      ]
```

which is within the joint limits, but for some reason, the robot is not moving there. the robot's current state at the time is
```
[-1.0758204686993669, 0.5317730073010851, 0.3355580861897537, -1.7691650372884484, -0.1610936406097853, 2.446098853513991, 1.9120584051260467, 0.03985649745277354, 0.03988420413359203]
```

maybe it's too big of a leap?

I expect that using reset mode simulation would resolve it, but I don't want to do that necessarily

PR #1330, opened 2022-12-01
`PlayroomHardEnv`

amburger66:
the `LightOn` classifier is set to an arbitrary, piecewise, "hard" function instead of having a simple threshold at 0.5


tomsilver (commented):
Looking good! Did you verify that the distribution of initial states with on vs off is roughly half and half?

amburger66:
@tomsilver can you help a look at the failing test? i'm confused

tomsilver:
unrelated issue (#1324), sorry about that. it's a flaky test -- I'll try to fix soon. in the meantime, I'm rerunning the test here

predicators/envs/playroom.py
```diff
@@ -924,3 +924,37 @@ def _robot_can_move(self, state: State, action: Action) -> bool:
...
+        return "playroom_hard"
+
+    def _transition_dial(self, state: State) -> State:
+        next_state = state.copy()
+        level = state.get(self._dial, "level")
+        light_is_on = False
+        for (low, high) in self.dial_on_intervals:
+            if low <= level <= high:
+                light_is_on = True
+        if not light_is_on:
```
tomsilver, line 947: no action required, but to make sure I understand: we're only going to see a variety of dial values for states before the dial has been touched. after the dial is touched, it will either be set to 0.5 or 1.0 for the rest of the episode. if it was only ever 0.5 or 1.0, there's no way that we could learn the right predicate classifier, but since there's also variation in the initialization, we do have hope to learn it.

also, the reason for doing it this way is to avoid the need to continuously parameterize the dial turning option (although that would be pretty straightforward).

is that all correct?

tomsilver (approved):
LGTM

predicators/envs/playroom.py
amburger66: yeah, that is what's happening. I didn't realize this before though -- this is probably another reason why the experimental results, even using Ask All,  were not very good.

PR #1333, opened 2022-12-03
Classification plotting for cover_handempty

amburger66:
The plotting is exactly the same as for `CoverEnv`


tomsilver (approved):
Love a small change!

PR #1334, opened 2022-12-05
`PlayroomSimpleEnv`

amburger66:
Creates a simplest version of the boring room vs. playroom environment, which is like the old ones except there are no doors (and thus no regions). This becomes the parent class of `PlayroomEnv` and `PlayroomHardEnv`


tests/approaches/test_oracle_approach.py
```diff
@@ -455,6 +455,35 @@ def test_repeated_nextto_painting_get_gt_nsrts():
...
+    """Tests for the ground truth NSRTs in PlayroomSimpleEnv."""
+    utils.reset_config({
+        "env": "playroom_simple",
+        "num_train_tasks": 1,
+        "num_test_tasks": 1
+    })
+    env = PlayroomSimpleEnv()
+    # Test MoveTableToDial for coverage.
+    nsrts = get_gt_nsrts(env.get_name(), env.predicates, env.options)
+    mvoetabletodial = [nsrt for nsrt in nsrts \
```
tomsilver, line 468: typo

tomsilver (approved):
excellent work!

PR #1335, opened 2022-12-05
Rectangle contains point fix

jorge-a-mendez:
The check needed to flip the angle and apply the rotation to the point _relative_ to the bottom-left corner of the rectangle.


tomsilver (approved):
LGTM thanks!

PR #1337, opened 2022-12-13
Log and plot per-task nodes created/expanded

NishanthJKumar:
Saves the per-task nodes created/expanded. After collecting data across all our different AAAI domains, we get the following plots:
nodes_created_painting
nodes_expanded_painting


tomsilver (approved):
awesome work! approving to unblock, but see my question in predicators slack

PR #1341, opened 2022-12-23
Dev multi shot analogy

sebdumancic:
Change the `_find_env_analogy` function to perform multi-shot analogy finding so that it can match environments with a different number of actions.

The principle is simple:
 - Run SME on the two environments
 - If all actions are matched, return the analogical match
 - If  not all actions are matched
    - remove all matched actions from the *target* domain
    - perform analogical matching over this new set
    - merge the analogies if they are consistent


predicators/approaches/initialized_pg3_approach.py
```diff
@@ -135,19 +136,88 @@ def base_nsrt_to_variable_analogy(
...
+        analogy."""
+        return all(
+            self.predicates[p] == other.predicates[p]
+            for p in self.predicates if p in other.predicates) and \
+            all(
+                self.nsrts[n] == other.nsrts[n]
+                for n in self.nsrts if n in other.nsrts)
+
+    def extend_with(self, other: _Analogy) -> _Analogy:
+        """Creates a new analogy by exteding this one with another analogy.
```
tomsilver, line 150: extending*

predicators/approaches/initialized_pg3_approach.py
```diff
@@ -135,19 +136,88 @@ def base_nsrt_to_variable_analogy(
             old_to_new_var[old_v] = new_v
         return old_to_new_var
 
+    def is_consistent_with(self, other: _Analogy) -> bool:
+        """Checks whether a given analogy is consistent with a given
```
tomsilver, line 140: do we need to also check the nsrt_variables?

predicators/approaches/initialized_pg3_approach.py
```diff
@@ -135,19 +136,88 @@ def base_nsrt_to_variable_analogy(
...
+                        nsrt_variables={
+                            **self.nsrt_variables,
+                            **other.nsrt_variables
+                        })
+
 
 def _find_env_analogies(base_env: BaseEnv, target_env: BaseEnv,
                         base_nsrts: Set[NSRT],
                         target_nsrts: Set[NSRT]) -> List[_Analogy]:
+    """Multishot analogical matching intended to overcome 1-to-1 mapping
```
tomsilver, line 171: multi-shot*

predicators/approaches/initialized_pg3_approach.py
```diff
@@ -135,19 +136,88 @@ def base_nsrt_to_variable_analogy(
...
+
     analogies: List[_Analogy] = []
+
     for sme_mapping in _query_sme(base_sme_struct, target_sme_struct):
         analogy = _sme_mapping_to_analogy(sme_mapping, base_env, target_env,
                                           base_nsrts, target_nsrts)
-        analogies.append(analogy)
+
+        # Check which target nsrts are missing.
+        missing_nsrts = target_nsrts - {n for _, n in analogy.nsrts.items()}
```
tomsilver, line 193: could do `set(analogy.nsrts.values())`, maybe slightly simpler? doesn't matter though

predicators/approaches/initialized_pg3_approach.py
```diff
@@ -135,19 +136,88 @@ def base_nsrt_to_variable_analogy(
...
     target_sme_struct = _create_sme_inputs(target_env, target_nsrts)
+
     analogies: List[_Analogy] = []
+
     for sme_mapping in _query_sme(base_sme_struct, target_sme_struct):
         analogy = _sme_mapping_to_analogy(sme_mapping, base_env, target_env,
                                           base_nsrts, target_nsrts)
-        analogies.append(analogy)
+
+        # Check which target nsrts are missing.
```
tomsilver, line 192: (very nit-picky sorry!) please capitalize NSRT in comments

predicators/approaches/initialized_pg3_approach.py
```diff
@@ -135,19 +136,88 @@ def base_nsrt_to_variable_analogy(
...
+        missing_nsrts = target_nsrts - {n for _, n in analogy.nsrts.items()}
+
```
tomsilver, line 194: so this all will happen at most one time. do you anticipate that we would want to keep trying to extend the analogy later on? or should one extension be enough in general?

tomsilver: i.e. I'm wondering about doing this in a loop until the missing NSRTs or the target NSRTs are empty

tomsilver (approved):
@sebdumancic awesome!! just a few small things and a question. also, did you test this out with certain domains and see that it helps? if so, is there a way to reproduce those tests? (feel free to update the PR description if so)

predicators/approaches/initialized_pg3_approach.py
sebdumancic: ah, a matched between NSRTs doesn't imply that the variables all match correctly?
That was my assumption.

predicators/approaches/initialized_pg3_approach.py
sebdumancic: note that this is a recursive function. So it will keep extending the analogy as long as it can find some useful matches.

sebdumancic:
Re tests: yes, I tried it on Gripper -> Miconic and Gripper -> Logistics. It found a matched we hoped it would. I didn't include those tests here because they take a bit of time.

Should I add them?

predicators/approaches/initialized_pg3_approach.py
tomsilver: ahh right sorry I totally missed the recurison. makes sense!

predicators/approaches/initialized_pg3_approach.py
tomsilver: hmm you're probably right, maybe add an assertion just to be safe?

tomsilver:
>Re tests: yes, I tried it on Gripper -> Miconic and Gripper -> Logistics. It found a matched we hoped it would. I didn't include those tests here because they take a bit of time.

that's awesome! no need to add them, I was just curious.

tomsilver (approved):

tomsilver:
command for us to play with:
```
python predicators/main.py --approach initialized_pg3 --seed 0          --env pddl_miconic_procedural_tasks --strips_learner oracle         --num_train_tasks 20 --pg3_init_policy gripper_ldl_policy.txt         --pg3_init_base_env pddl_gripper_procedural_tasks         --pg3_add_condition_allow_new_vars False
```

tomsilver:
closing because we have a better implementation forthcoming, where the bulk of it will be in smepy instead of here

PR #1344, opened 2022-12-28
change glib novelty function to match original implementation

tomsilver:
in the original implementation of GLIB, we would only babble goals that had _never_ been seen before. in the predicators implementation, we were babbling goals in order of the _number_ of times that they had been seen, so the same goal could get babbled multiple times. this is actually bad because it's possible to get into situations where you babble the same goals that appear reachable over and over again, and never fall back to taking random actions, while not realizing that some other goals would be reachable if you corrected your faulty operators. I saw this happen sometimes while messing around with PDDL blocks.

this change may impact @amburger66 's baseline results, but the old results can be recovered by changing the new `online_interaction_max_novelty_count` flag to inf.


amburger66 (approved):
is this understanding correct: once some goal has been seen `online_learning_max_novelty_count` number of times, it is no longer considered "novel" and won't be babbled by the GLIB explorer anymore?

tomsilver:
yes exactly, I'll add that as a comment in the code. thanks!

PR #1346, opened 2023-01-05
Greedy Lookahead Explorer: NSRT resampling

amburger66:
- currently, NSRT preconditions being satisfied does not guarantee that the resulting option will be initiable, since predicates may be learned incorrectly
- the old assert statement led to experiments crashing, so now it is replaced with resampling (the NSRT)
- if resampling also fails, we simply terminate the current trajectory


tomsilver (approved):
LGTM, nice work!

PR #1349, opened 2023-01-06
[refinement estimation] data collection and tabular refinement cost estimator

lilianluong:
**Summary of changes:**
 - `predicators/train_refinement_estimator.py` data collection and training scripts
 - `TabularRefinementEstimator` memorizes (skeleton, atoms sequence) -> average refinement time as a dictionary
 - tests for the above
 - additions to `structs`, `settings`, `args` files to support the above
 - minor changes to refinement estimators in general and associated tests
 - refactored functions to use helper functions with given parser inputs: `reset_config()`, `update_config()`, and `parse_args()`

`train_refinement_estimator.py` can be run from the command line to generate training data, train refinement cost estimators on this data, or both (default). More detailed instructions are in the file docstring, but an example command to run would be:

`python predicators/train_refinement_estimator.py --env narrow_passage --approach refinement_estimation --refinement_estimator tabular --seed 0`

Additional flags:
- `--load_data` will skip the data generation and instead load the dataset from a file, defaulting to `CFG.data_dir` > config path string
- `--data_file_name <file_name>` is a new flag that will change the file path used either when writing or loading the dataset file, replacing the config path string part with whatever file name is given
- `--skip_training` is a new flag that will skip the training portion of the script (use if you just want to generate data)

The expected usage is to train a refinement cost estimator (i.e. `tabular`) and save the state of the estimator as a file. `RefinementEstimationApproach` automatically tries to load the state according to the default data file path when initializing an estimator that has `is_learning_based==True`.


predicators/approaches/refinement_estimation_approach.py
```diff
@@ -34,8 +38,22 @@ def __init__(self,
                 CFG.sesame_max_skeletons_optimized), \
                "refinement_estimation_num_skeletons_generated should not be" \
                "greater than sesame_max_skeletons_optimized"
+        estimator_name = CFG.refinement_estimator
         self._refinement_estimator = create_refinement_estimator(
-            CFG.refinement_estimator)
+            estimator_name)
+        # If the refinement estimator is learning based, try to load
+        # trained state if the file exists.
```
tomsilver, line 45: can we change the word "state" here to "model" or something like that to avoid confusion with `State`?

predicators/args.py
```diff
@@ -41,4 +41,7 @@ def create_arg_parser(env_required: bool = True,
                         const=logging.DEBUG,
                         default=logging.INFO)
     parser.add_argument("--crash_on_failure", action="store_true")
+    # Used when running train_refinement_estimator.py from the command line
```
tomsilver, line 44: we should either:
1. make this change inside of `train_refinement_estimator.py`, for example, by calling this function to get the initial `parser` and then called `add_argument(...)` on it after it's returned
2. change these flag names to make it clear that they're about refinement estimator training only

or both. my preference is both!

predicators/refinement_estimators/base_refinement_estimator.py
```diff
@@ -16,8 +17,32 @@ def get_name(cls) -> str:
...
+        """Does the estimator learn from training tasks?"""
+        raise NotImplementedError("Override me!")
+
+    @abc.abstractmethod
+    def get_cost(self, initial_state: State, skeleton: List[_GroundNSRT],
                  atoms_sequence: List[Set[GroundAtom]]) -> float:
         """Return an estimated cost for a proposed high-level skeleton."""
         raise NotImplementedError("Override me!")
+
+    def train(self, data: List) -> None:  # pragma: no cover
```
tomsilver, line 32: we should be able to remove this and the below pragma-no-covers after we add some tests for the new learning-based class

predicators/refinement_estimators/tabular_refinement_estimator.py
```diff
@@ -0,0 +1,72 @@
...
+    _GroundNSRT
+
+
+class TabularRefinementEstimator(BaseRefinementEstimator):
+    """A refinement cost estimator that memorizes refinement data using a
+    tabular method."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._cost_dict: Optional[Dict[Tuple[Tuple[_GroundNSRT, ...],
```
tomsilver, line 23: add a comment to help understand this monster type

predicators/refinement_estimators/tabular_refinement_estimator.py
```diff
@@ -0,0 +1,72 @@
...
+        return cost
+
+    def train(self, data: List[RefinementDatapoint]) -> None:
+        """Train the tabular refinement estimator on data by computing average
+        refinement time per (skeleton, atoms_sequence) pair."""
+        grouped_data = defaultdict(list)
+        # Go through data and group them by skeleton
+        for _, skeleton, atoms_sequence, succeeded, refinement_time in data:
+            # Convert skeleton and atoms_sequence into an immutable dict key
+            key = (tuple(skeleton),
```
tomsilver, line 52: want to make a small helper method for this since it's used both here and above?

predicators/refinement_estimators/tabular_refinement_estimator.py
```diff
@@ -0,0 +1,72 @@
...
+                value += CFG.refinement_data_failed_refinement_penalty
+            grouped_data[key].append(value)
+        # Compute average time for each (skeleton, atoms_sequence) key
+        processed_data = {
+            key: float(np.mean(times))
+            for key, times in grouped_data.items()
+        }
+        self._cost_dict = processed_data
+
+    def save_state(self, filepath: Path) -> None:  # pragma: no cover
```
tomsilver, line 66: ditto about these pragma no covers

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
+"""Scripts for collecting refinement data and for training learned refinement
+cost estimators.
+
+Example usage to generate data and train:
```
tomsilver, line 4: extremely helpful, thanks!

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
...
+from predicators.ground_truth_nsrts import get_gt_nsrts
+from predicators.option_model import _OptionModelBase, create_option_model
+from predicators.planning import PlanningFailure, PlanningTimeout, \
+    _MaxSkeletonsFailure, _skeleton_generator, _SkeletonSearchTimeout, \
+    run_low_level_search
+from predicators.settings import CFG
+from predicators.structs import NSRT, Metrics, ParameterizedOption, \
+    Predicate, RefinementDatapoint, Task, Type
+
+assert os.environ.get("PYTHONHASHSEED") == "0", \
```
tomsilver, line 54: we can remove this, unlikely that someone is going to use this script and not `main.py`

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
...
+    # env.action_space and env.options are often created during env __init__().
+    env.action_space.seed(CFG.seed)
+    for option in env.options:
+        option.params_space.seed(CFG.seed)
+    assert env.goal_predicates.issubset(env.predicates)
+    preds, _ = utils.parse_config_excluded_predicates(env)
+
+    # Create the train tasks.
+    train_tasks = env.get_train_tasks()
+    # If train tasks have goals that involve excluded predicates, strip those
```
tomsilver, line 92: as mentioned on slack, let's get rid of this here, which will resolve the pickling issue

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
...
+    refinement_estimator = approach.refinement_estimator
+    assert refinement_estimator.is_learning_based, \
+        "Refinement estimator (--refinement_estimator) must be learning-based"
+
+    # Train estimator
+    train_start_time = time.perf_counter()
+    refinement_estimator.train(dataset)
+    train_time = time.perf_counter() - train_start_time
+    logging.info(f"Finished training in {train_time:.5f} seconds")
+    # Save the training state to a file
```
tomsilver, line 139: same comment here about "state" -> "model"

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
...
+                                      option_model: _OptionModelBase,
+                                      nsrts: Set[NSRT],
+                                      predicates: Set[Predicate],
+                                      types: Set[Type], seed: int,
+                                      data: List[RefinementDatapoint]) -> None:
+    """Generate refinement data from given train task and append to data.
+
+    Adapted from _sesame_plan_with_astar() from predicators/planning.py.
+    """
+    ground_nsrt_timeout = CFG.timeout
```
tomsilver, line 199: this is quite a bit of copied code, so let's think about whether there's another way. if we're willing to make changes within `planning.py`, could we log the metrics we need inside of `metrics`?

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
...
+                atoms_sequence,
+                suc,
+                refinement_time,
+            ))
+    except _MaxSkeletonsFailure:
+        # Done finding skeletons
+        return
+
+
+def _get_data_file_path() -> Path:
```
tomsilver, line 269: in our main code, we just enforce the standard naming scheme. what's the reason to allow overrides?

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,285 @@
...
+    if len(CFG.data_file_name):
+        file_name = CFG.data_file_name
+    else:
+        config_path_str = utils.get_config_path_str()
+        file_name = f"refinement_data_{config_path_str}.data"
+    data_file_path = Path(CFG.data_dir) / file_name
+    return data_file_path
+
+
+if __name__ == "__main__":  # pragma: no cover
```
tomsilver, line 279: this pragma no cover is ok!

predicators/train_refinement_estimator.py
tomsilver: my brain is out of thinking juice for the week but I can return to this later if there isn't a clear answer

tomsilver (commented):
excellent work!

predicators/args.py
lilianluong: Change the utils.parse_args to use a helper function that takes in a parser as input.

predicators/planning.py
```diff
@@ -231,6 +204,63 @@ def _sesame_plan_with_astar(
...
+
+def filter_nsrts(
+    task: Task,
+    init_atoms: Set[GroundAtom],
+    ground_nsrts: List[_GroundNSRT],
+    check_dr_reachable: bool = True,
+    allow_noops: bool = False,
+) -> List[_GroundNSRT]:
+    """Helper function for _sesame_plan_with_astar(); optionally filter out
+    NSRTs with empty artifacts and/or those that are unreachable."""
```
tomsilver, line 248: empty effects*

predicators/refinement_estimators/oracle_refinement_estimator.py
```diff
@@ -15,8 +15,13 @@ class OracleRefinementEstimator(BaseRefinementEstimator):
...
         return "oracle"
 
-    def get_cost(self, skeleton: List[_GroundNSRT],
+    @property
+    def is_learning_based(self) -> bool:
+        return False
+
+    def get_cost(self, initial_state: State, skeleton: List[_GroundNSRT],
                  atoms_sequence: List[Set[GroundAtom]]) -> float:
+        del initial_state  # unused
```
tomsilver, line 24: I could be wrong, but I think that that this is unnecessary because it's overriding a parent class method -- I think that linting should be smart enough to realize that other subclasses might use that argument

predicators/refinement_estimators/tabular_refinement_estimator.py
```diff
@@ -0,0 +1,86 @@
...
+    def get_name(cls) -> str:
+        return "tabular"
+
+    @property
+    def is_learning_based(self) -> bool:
+        return True
+
+    def get_cost(self, initial_state: State, skeleton: List[_GroundNSRT],
+                 atoms_sequence: List[Set[GroundAtom]]) -> float:
+        del initial_state  # unused
```
tomsilver, line 42: ditto

predicators/refinement_estimators/tabular_refinement_estimator.py
```diff
@@ -0,0 +1,86 @@
...
+        assert self._cost_dict is not None, "Need to train"
+        key = TabularRefinementEstimator._immutable_cost_dict_key(
```
tomsilver, line 44: why not `self`?

predicators/refinement_estimators/tabular_refinement_estimator.py
```diff
@@ -0,0 +1,86 @@
...
+        return cost
+
+    def train(self, data: List[RefinementDatapoint]) -> None:
+        """Train the tabular refinement estimator on data by computing average
+        refinement time per (skeleton, atoms_sequence) pair."""
+        grouped_data = defaultdict(list)
+        # Go through data and group them by skeleton
+        for _, skeleton, atoms_sequence, succeeded, refinement_time in data:
+            # Convert skeleton and atoms_sequence into an immutable dict key
+            key = TabularRefinementEstimator._immutable_cost_dict_key(
```
tomsilver, line 57: ditto

predicators/settings.py
```diff
@@ -470,6 +470,12 @@ class GlobalSettings:
     refinement_estimator = "oracle"  # default refinement cost estimator
     refinement_estimation_num_skeletons_generated = 3
 
+    # refinement data collection parameters
+    refinement_data_num_skeletons = 3
+    refinement_data_skeleton_generator_timeout = 20
+    refinement_data_low_level_search_timeout = 5  # timeout for refinement try
+    refinement_data_failed_refinement_penalty = 5  # added time on failure (s)
```
tomsilver, line 477: all times are in seconds, so no need to indicate

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+from predicators.option_model import _OptionModelBase, create_option_model
+from predicators.planning import PlanningTimeout, _MaxSkeletonsFailure, \
+    _skeleton_generator, _SkeletonSearchTimeout, filter_nsrts, \
+    run_low_level_search, sesame_ground_nsrts
+from predicators.settings import CFG
+from predicators.structs import NSRT, Metrics, ParameterizedOption, \
+    Predicate, RefinementDatapoint, Task, Type
+
+
+def train_refinement_estimation_approach() -> None:
```
tomsilver, line 56: make private?

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+    model_file = f"{CFG.refinement_estimator}_{config_path_str}.estimator"
+    model_file_path = Path(CFG.approach_dir) / model_file
+    refinement_estimator.save_model(model_file_path)
+    logging.info(f"Saved trained estimator to {model_file_path}")
+
+    script_time = time.perf_counter() - script_start
+    logging.info(f"\n\nScript terminated in {script_time:.5f} seconds")
+
+
+def get_refinement_estimation_parser() -> ArgumentParser:
```
tomsilver, line 145: make private?

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+    env.action_space.seed(CFG.seed)
+    for option in env.options:
+        option.params_space.seed(CFG.seed)
+    assert env.goal_predicates.issubset(env.predicates)
+    preds, _ = utils.parse_config_excluded_predicates(env)
+
+    # Create the train tasks.
+    train_tasks = env.get_train_tasks()
+    # Assume we're not doing option learning, pass in all the environment's
+    # oracle options.
```
tomsilver, line 92: can you add
```python
assert CFG.option_learner == "no_learning"
```

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+                        handlers=handlers)
+    if CFG.log_file:
+        logging.info(f"Logging to {CFG.log_file}")
+    logging.info(f"Running command: python {str_args}")
+
+    # Create environment
+    env = create_new_env(CFG.env, do_cache=True, use_gui=CFG.use_gui)
+    # The action space and options need to be seeded externally, because
+    # env.action_space and env.options are often created during env __init__().
+    env.action_space.seed(CFG.seed)
```
tomsilver, line 83: can we put this into a new `utils.py` function called `seed_env`? (and refactor here and in main.py)

predicators/train_refinement_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+    # Terminate early if training should be skipped
+    if CFG.skip_refinement_estimator_training:
+        script_time = time.perf_counter() - script_start
+        logging.info(f"\n\nScript terminated in {script_time:.5f} seconds")
+        return
+
+    # Create approach
+    assert CFG.approach == "refinement_estimation", \
+        "Approach (--approach) must be set to refinement_estimation"
+    approach = cast(
```
tomsilver, line 119: any reason to not just do
```python
approach = RefinementEstimationApproach(preds, options, env.types,
                        env.action_space, train_tasks)
```
?

tomsilver (approved):
LGTM after minor stuff, great work!

predicators/train_refinement_estimator.py
lilianluong: Ended up not doing this to avoid annoying circular imports.

PR #1353, opened 2023-01-12
use unary static predicates to make LDL grounding faster

### LINKED ISSUE #1347: improve grounding of LDL rules using static preconditions
e.g. in gripper, there are no types, but there are static predicates which serve as types (like "gripper"), so we should use those to prevent nonsensical LDL rule groundings.

tomsilver:
closes #1347


NishanthJKumar (approved):
LGTM! Sorry for the delay with this review!!

PR #1357, opened 2023-01-19
`BlocksEnvClear`

amburger66:
A version of `BlocksEnv` where the block type has a feature name "clear" indicating whether the block is clear. This way we can learn all the predicates in the env with the assumption that the predicates are a function of only their argument's states.


tomsilver (approved):
nice work! please also test out oracle and nsrt_learning before merging

PR #1358, opened 2023-01-19
Interactive learning: glib novelty function 

amburger66:
set `online_learning_max_novelty_count` to inf so that the glib scoring works as intended (before #1344)

### DIFF
--- a/scripts/configs/interactive_learning.yaml
+++ b/scripts/configs/interactive_learning.yaml
@@ -79,5 +79,7 @@ FLAGS:  # common args
   min_data_for_nsrt: 10
   sampler_disable_classifier: True
   mlp_classifier_balance_data: False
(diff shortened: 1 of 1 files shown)


tomsilver (approved):
Assuming that this parses correctly (please try on command line before merging) lgtm!

amburger66:
wait actually @tomsilver i am failing to see how the change in #1344 would affect my experiments, since I use glib explorer with my own [frequency-based score function](https://github.com/Learning-and-Intelligent-Systems/predicators/blob/3b2980a89281042c2aac77379ca46b665695bf35/predicators/approaches/interactive_learning_approach.py#L296), which does not use `CFG.online_learning_max_novelty_count` and never returns -inf?

tomsilver:
Ah excellent point, my bad! So we can close this PR then.

PR #1361, opened 2023-01-31
Implements arity-2 classifiers

NishanthJKumar:
Changes necessary to get arity-2 classifiers. This is being implemented in anticipation of doing predicate invention in domains that are more complex than original AAAI paper domains.

To run, try:
```
python predicators/main.py --env cover --approach grammar_search_invention --excluded_predicates all --num_train_tasks 20 --seed 456 --grammar_search_grammar_use_diff_features True
```
Interestingly, this solves 46/50 eval tasks, while setting the `--grammar_search_grammar_use_diff_features True` arg to `False` only solves 43.

### DIFF
--- a/predicators/approaches/grammar_search_invention_approach.py
+++ b/predicators/approaches/grammar_search_invention_approach.py
@@ -199,7 +199,7 @@ class _DoubleAttributeCompareClassifier(_BinaryClassifier):
         name2 = CFG.grammar_search_classifier_pretty_str_names[
             self.object2_index]
         vars_str = (f"{name1}:{self.object1_type.name}, "
-                    "{name2}:{self.object2_type.name}")
+                    f"{name2}:{self.object2_type.name}")
         body_str = (f"(|{name1}.{self.attribute1_name} - "
                     f"{name2}.{self.attribute2_name}| "
                     f"{self.compare_str} {self.constant:.3})")
--- a/predicators/approaches/grammar_search_invention_approach.py
+++ b/predicators/approaches/grammar_search_invention_approach.py
@@ -32,11 +32,13 @@ from predicators.structs import Dataset, GroundAtom, GroundAtomTrajectory, \
 def _create_grammar(dataset: Dataset,
                     given_predicates: Set[Predicate]) -> _PredicateGrammar:
     # We start with considering various ways to split either single or
-    # double feature values across our dataset.
+    # two feature values across our dataset.
     grammar: _PredicateGrammar = _SingleFeatureInequalitiesPredicateGrammar(
         dataset)
-    if CFG.grammar_search_grammar_use_double_features:
-        grammar = _DoubleFeatureInequalitiesPredicateGrammar(dataset)
+    if CFG.grammar_search_grammar_use_diff_features:
+        diff_grammar = _FeatureDiffInequalitiesPredicateGrammar(dataset)
+        grammar = _ChainPredicateGrammar([grammar, diff_grammar],
+                                         alternate=True)
     # We next optionally add in the given predicates because we want to allow
     # negated and quantified versions of the given predicates, in
     # addition to negated and quantified versions of new predicates.
@@ -123,11 +125,12 @@ class _UnaryClassifier(_ProgrammaticClassifier):
 
 
 class _BinaryClassifier(_ProgrammaticClassifier):
-    """A classifier on two object."""
+    """A classifier on two objects."""
 
     def __call__(self, s: State, o: Sequence[Object]) -> bool:
         assert len(o) == 2
-        return self._classify_object(s, o[0], o[1])
+        o0, o1 = o
+        return self._classify_object(s, o0, o1)
 
     @abc.abstractmethod
     def _classify_object(self, s: State, obj1: Object, obj2: Object) -> bool:
@@ -165,7 +168,7 @@ class _SingleAttributeCompareClassifier(_UnaryClassifier):
 
 
 @dataclass(frozen=True, eq=False, repr=False)
-class _DoubleAttributeCompareClassifier(_BinaryClassifier):
+class _AttributeDiffCompareClassifier(_BinaryClassifier):
     """Compare a single feature value with a constant value."""
     object1_index: int
     object1_type: Type
@@ -499,7 +502,7 @@ class _SingleFeatureInequalitiesPredicateGrammar(_DataBasedPredicateGrammar):
 
 
 @dataclass(frozen=True, eq=False, repr=False)
-class _DoubleFeatureInequalitiesPredicateGrammar(
+class _FeatureDiffInequalitiesPredicateGrammar(
         _SingleFeatureInequalitiesPredicateGrammar):
     """Generates features of the form "|0.feature - 1.feature| >= c" or
     "|0.feature - 1.feature| <= c". This is also capable of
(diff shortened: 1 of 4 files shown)


predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -31,10 +31,12 @@
 
 def _create_grammar(dataset: Dataset,
                     given_predicates: Set[Predicate]) -> _PredicateGrammar:
-    # We start with considering various ways to split single feature values
-    # across our dataset.
+    # We start with considering various ways to split either single or
+    # double feature values across our dataset.
     grammar: _PredicateGrammar = _SingleFeatureInequalitiesPredicateGrammar(
```
tomsilver, line 36:
```python
if CFG.grammar_search_grammar_use_double_features:
    grammar = _DoubleFeatureInequalitiesPredicateGrammar(dataset)
else:
    grammar = _SingleFeatureInequalitiesPredicateGrammar(dataset)
```

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -120,6 +122,18 @@ def _classify_object(self, s: State, obj: Object) -> bool:
         raise NotImplementedError("Override me!")
 
 
+class _BinaryClassifier(_ProgrammaticClassifier):
+    """A classifier on two object."""
```
tomsilver, line 126: objects*

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -120,6 +122,18 @@ def _classify_object(self, s: State, obj: Object) -> bool:
...
+
+    def __call__(self, s: State, o: Sequence[Object]) -> bool:
+        assert len(o) == 2
```
tomsilver, line 129: alternative, not sure if actually better
```python
o0, o1 = o
return self._classify_object(s, o0, o1)
```

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -150,6 +164,48 @@ def pretty_str(self) -> Tuple[str, str]:
...
+    constant_idx: int
+    compare: Callable[[float, float], bool]
+    compare_str: str
+
+    def _classify_object(self, s: State, obj1: Object, obj2: Object) -> bool:
+        assert obj1.type == self.object1_type
+        assert obj2.type == self.object2_type
+        return self.compare(
+            abs(
+                s.get(obj1, self.attribute1_name) -
```
tomsilver, line 186: I wonder if we should name things in terms of "difference" instead of "double", since "double" suggests that we might compare two features in a more general way

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -442,6 +498,88 @@ def _get_feature_ranges(
         return feature_ranges
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class _DoubleFeatureInequalitiesPredicateGrammar(
+        _SingleFeatureInequalitiesPredicateGrammar):
+    """Generates features of the form "|0.feature - 1.feature| >= c" or
+    "|0.feature - 1.feature| <= c". This is also capable of
```
tomsilver, line 505: I'm finding the fact that this grammar also produces single feature inequalities a bit confusing, and redundant with the code we have for single feature inequalities -- it might be the best option, but let's talk through an alternative.

could we instead have this grammar only produce double (difference) feature predicates, and then "chain" together this grammar and the original single features grammar? we already have a `_ChainPredicateGrammar`, but as implemented right now, it won't work, because it enumerates all predicates from the first grammar before moving onto the second one, and both of these grammars are infinite. so instead, we could have the chaining alternate between the grammars, like this: https://stackoverflow.com/questions/2017895/alternating-between-iterators-in-python we could make "alternate" a kwarg for `_ChainPredicateGrammar`. would this work?

tomsilver (commented):
looking good! some questions

predicators/approaches/grammar_search_invention_approach.py
tomsilver: in the stackoverflow answers, I think the "roundrobin" one is probably the best

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: sure, I can change to this!

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: i agree that it's confusing this grammar produces single feature inequalities; let me look into this alternative and see if I can get it to work!

NishanthJKumar: got it working!

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -150,6 +167,48 @@ def pretty_str(self) -> Tuple[str, str]:
         return vars_str, body_str
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class _AttributeDiffCompareClassifier(_BinaryClassifier):
+    """Compare a single feature value with a constant value."""
```
tomsilver, line 172: update docstring

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -442,6 +501,66 @@ def _get_feature_ranges(
         return feature_ranges
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class _FeatureDiffInequalitiesPredicateGrammar(
+        _SingleFeatureInequalitiesPredicateGrammar):
+    """Generates features of the form "|0.feature - 1.feature| >= c" or
+    "|0.feature - 1.feature| <= c". This is also capable of
```
tomsilver, line 508: update docstring

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -442,6 +501,66 @@ def _get_feature_ranges(
...
+        if not any(r for r in feature_ranges.values()):
+            return
+        # 0.5, 0.25, 0.75, 0.125, 0.375, ...
+        constant_generator = _halving_constant_generator(0.0, 1.0)
+        for constant_idx, (constant, cost) in enumerate(constant_generator):
+            for (t1, t2) in itertools.combinations_with_replacement(
+                    sorted(self.types), 2):
+                for f1 in t1.feature_names:
+                    for f2 in t2.feature_names:
+                        # If the types and features are the same, then
```
tomsilver, line 525: why? isn't it possible that we would want to compare the same feature in two different objects of the same type?

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -442,6 +501,66 @@ def _get_feature_ranges(
...
+                            continue
+                        # In this case, we actually need a binary attribute
+                        # comparison classifier.
+                        lb1, ub1 = feature_ranges[t1][f1]
+                        if abs(lb1 - ub1) < 1e-6:
+                            continue
+                        lb2, ub2 = feature_ranges[t2][f2]
+                        if abs(lb2 - ub2) < 1e-6:
+                            continue
+                        if self.f_range_intersection(lb1, ub1, lb2, ub2):
```
tomsilver, line 537: I _think_ I get this but can you add some comments?

predicators/approaches/grammar_search_invention_approach.py
tomsilver: in addition to removing the single-feature comment, isn't it also the case that we're only generating `|0.feature - 1.feature| <= c` and not `|0.feature - 1.feature| >= c`? (which is good, because we can basically get the latter by applying negation later)

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -442,6 +501,66 @@ def _get_feature_ranges(
...
+                            0, t1, f1, 1, t2, f2, k, constant_idx, comp,
+                            comp_str)
+                        name = str(diff_classifier)
+                        types = [t1, t2]
+                        pred = Predicate(name, types, diff_classifier)
+                        assert pred.arity == 2
+                        yield (pred, 2 + cost
+                               )  # cost = arity + cost from constant
+
+    def f_range_intersection(self, lb1: float, ub1: float, lb2: float,
```
tomsilver, line 557: put this in utils.py?

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -456,10 +575,24 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
 class _ChainPredicateGrammar(_PredicateGrammar):
     """Chains together multiple predicate grammars in sequence."""
     base_grammars: Sequence[_PredicateGrammar]
+    alternate: bool = False
+
+    @staticmethod
+    def _roundrobin(iterables: Sequence[Iterator]) -> Iterator:
```
tomsilver, line 581: put this in utils.py?

tomsilver (commented):
very nice! a few more comments and then I think we should be good to go

tomsilver (approved):
LGTM!

PR #1363, opened 2023-02-02
generalize loading test tasks from json files

tomsilver:
this was previously implemented for blocks only. new environments will still need to implement `_load_task_from_json` but that's it.


NishanthJKumar (approved):
LGTM! Thanks for the ultra-fast turnaround with this!

PR #1366, opened 2023-02-04
Add flag for toggling usage of necessary atoms sequence in `planning.py`

amburger66:
`CFG.sesame_use_necessary_atoms` defaults to True, but should be set to False for interactive learning experiments as using the necessary atoms sequence was making the GLIB action baseline perform abnormally well


tomsilver (approved):
perf, thanks!

PR #1367, opened 2023-02-06
sandwich environment

tomsilver:
new environment where a fixed-base robot makes sandwiches.

* the main motivation for this environment is to work towards a real robot demo with the panda (cc @williamshen-nz ).
* the objects are inspired by [this sandwich making play set](https://a.co/d/gM2CzjS), which I'm hoping we can buy and use with the panda
* the environment may also be useful for active predicate learning (@amburger66 )

here are some renderings:
* https://user-images.githubusercontent.com/2816502/216854567-dcb3f030-c635-441d-9227-c029ec51d876.mp4
* https://user-images.githubusercontent.com/2816502/216854572-307aae9a-58a1-4b6d-aa3b-aabeb718bc73.mp4
* https://user-images.githubusercontent.com/2816502/216854578-ebf1a4ad-3113-4a7d-80d8-fdd98687acac.mp4

notes about the renderings:
* the left is a side view and the right is a top-down view
* the ingredients are initially upright to allow for easy grasping
* the robot will grasp an ingredient lengthwise so that it can subsequently place onto the board or table without one of its fingers getting stuck underneath. hopefully the gripper is wide enough and the ingredients are small enough to allow this

after merging this PR, I'm planning to:
1. implement pybullet version.
2. extend the environment to make it more interesting. for example, add squirt bottles (mayo, ketchup, etc.). maybe add tools, like a spatula. maybe require the robot to move the board after the sandwich is made.
3. play with our new LLM interface for goal specification (e.g., "make me a reuben sandwich with ketchup" or something)

### DIFF
--- a/predicators/envs/sandwich.py
+++ b/predicators/envs/sandwich.py
@@ -3,8 +3,10 @@
 import itertools
 import json
 import logging
+from collections import defaultdict
 from pathlib import Path
-from typing import ClassVar, Dict, List, Optional, Sequence, Set, Tuple
+from typing import ClassVar, DefaultDict, Dict, List, Optional, Sequence, \
+    Set, Tuple
 
 import matplotlib
 import matplotlib.pyplot as plt
@@ -324,9 +326,9 @@ class SandwichEnv(BaseEnv):
                 num_ing = rng.choice(possible_ing_nums)
                 ing_to_num[ing_name] = num_ing
             init_state = self._sample_initial_state(ing_to_num, rng)
-            goal = self._sample_goal(ing_to_num, rng)
-            # goal_holds = all(goal_atom.holds(init_state) for goal_atom in goal)
-            # assert not goal_holds
+            goal = self._sample_goal(init_state, rng)
+            goal_holds = all(goal_atom.holds(init_state) for goal_atom in goal)
+            assert not goal_holds
             tasks.append(Task(init_state, goal))
         return tasks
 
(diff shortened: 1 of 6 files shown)


predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+    board_x_ub: ClassVar[float] = x_ub - board_width / 2
+    board_y_lb: ClassVar[
+        float] = holder_y_lb + holder_length / 2 + board_length
+    board_y_ub: ClassVar[float] = board_y_lb + (y_ub - y_lb) * 0.05
+    board_color: ClassVar[RGBA] = (0.1, 0.1, 0.5, 0.8)
+    board_thickness: ClassVar[float] = 0.01
+    ingredient_thickness: ClassVar[float] = 0.02
+    ingredient_colors: ClassVar[Dict[str, Tuple[float, float, float]]] = {
+        "bread": (0.58, 0.29, 0.0),
+        "burger": (0.32, 0.15, 0.0),
```
NishanthJKumar, line 54: minor thing, but could we rename this to 'patty'? I was confused for a bit thinking that this object referred to the entire assembled thing (i.e, the bread + the patty + the veggies...)

predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+        return "sandwich"
+
+    def simulate(self, state: State, action: Action) -> State:
+        assert self.action_space.contains(action.arr)
+        x, y, z, fingers = action.arr
+        # Infer which transition function to follow
+        if fingers < self.held_tol:
+            return self._transition_pick(state, x, y, z)
+        thickness = self.ingredient_thickness + self.board_thickness
+        if z < self.table_height + thickness:
```
NishanthJKumar, line 172: I think I understand this, but could you leave a quick comment about why this logic is right here?

predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+            return next_state
+        # Check that both objects exist
+        ing = self._get_held_object(state)
+        assert ing is not None
+        other_ing = self._get_highest_object_below(state, x, y, z)
+        if other_ing is None:  # no object to stack onto
+            return next_state
+        # Can't stack onto yourself!
+        if ing == other_ing:
+            return next_state
```
NishanthJKumar, line 232: do we want to assert this? This should never happen right? Asserting sounds useful to avoid silent failures.

predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+        holder_ub = holder_y + self.holder_length / 2.
+        return holder_lb - 1e-5 <= obj_y <= holder_ub + 1e-5
+
+    def _OnBoard_holds(self, state: State, objects: Sequence[Object]) -> bool:
+        obj, board = objects
+        obj_y = state.get(obj, "pose_y")
+        board_y = state.get(board, "pose_y")
+        board_lb = board_y - self.board_length / 2.
+        board_ub = board_y + self.board_length / 2.
+        return board_lb - 1e-5 <= obj_y <= board_ub + 1e-5
```
NishanthJKumar, line 560: ditto to the above comment

predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+        return np.allclose([x1, y1, z1], [x2, y2, z2 + offset],
+                           atol=self.on_tol)
+
+    def _InHolder_holds(self, state: State, objects: Sequence[Object]) -> bool:
+        obj, holder = objects
+        obj_y = state.get(obj, "pose_y")
+        holder_y = state.get(holder, "pose_y")
+        holder_lb = holder_y - self.holder_length / 2.
+        holder_ub = holder_y + self.holder_length / 2.
+        return holder_lb - 1e-5 <= obj_y <= holder_ub + 1e-5
```
NishanthJKumar, line 552: hold on - do we not want to be testing the obj_x as well?

predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+        return self._obj_to_ingredient(obj, state) == ing_name
+
+    def _obj_to_ingredient(self, obj: Object, state: State) -> str:
+        obj_color = (state.get(obj, "color_r"), state.get(obj, "color_g"),
+                     state.get(obj, "color_b"))
+        affinities = {
+            n: np.sum(np.subtract(c, obj_color)**2)
+            for n, c in self.ingredient_colors.items()
+        }
+        closest = min(affinities, key=lambda o: affinities[o])
```
NishanthJKumar, line 775: if i'm understanding this right, you're using the object color to figure out what ingredient the object is here? Why is this necessary (e.g. why not just use the object name or subtype)? Could you leave a comment explaining this briefly?

predicators/envs/sandwich.py
```diff
@@ -0,0 +1,832 @@
...
+        return closest
+
+    def _state_to_ingredient_groups(self,
+                                    state: State) -> Dict[str, Set[Object]]:
+        ings = set(self.ingredient_colors)
+        ing_groups: Dict[str, Set[Object]] = {n: set() for n in ings}
+        for obj in state.get_objects(self._ingredient_type):
+            ing_name = self._obj_to_ingredient(obj, state)
+            ing_groups[ing_name].add(obj)
+        return dict(ing_groups)
```
NishanthJKumar, line 785: okay, this helps me see a bit why the `_obj_to_ingredient` function is written as above, but I'm still quite confused why you're using the color...

NishanthJKumar (commented):
Looks generally good, great work! Just left a couple questions about things I didn't understand.

predicators/envs/sandwich.py
tomsilver: it's possible that a random options approach, or some other naive approach, would attempt to use an option that would stack an ingredient onto itself

tomsilver:
@NishanthJKumar thanks for the great feedback!

PR #1368, opened 2023-02-06
Silence pnad_search hill-climbing during predicate invention.

NishanthJKumar:
Previously, during predicate invention using `pnad_search` as the operator learning method, we would log the hill-climbing of `pnad_search` operator learning **within each iteration of predicate learning**. This made it a nightmare to try to read the terminal to get a sense of the predicate learning progress. This PR introduces a flag that silences the operator-learning printouts when we're calling `pnad_search` as part of predicate invention.


tomsilver (approved):
Lgtm, but just to check, the issue you mentioned in the PR description is being addressed in some other branch, right?

NishanthJKumar:
yes, but I wanted to move it onto master as its own feature so it's not confusingly in some other PR :smile_cat:

PR #1370, opened 2023-02-09
Add new environment: a parameterized touch point with 2D action space.

ashayathalye:
Adds another touch point environment that is (1) parameterized and (2) has a 2D action space. The intention is to use is to debug the simplest cases for learning nsrts via reinforcement learning.

### DIFF
--- a/predicators/envs/touch_point.py
+++ b/predicators/envs/touch_point.py
@@ -253,8 +253,7 @@ class TouchPointEnvParam(TouchPointEnv):
         next_state.set(self._robot, "y", new_y)
         return next_state
 
-    @staticmethod
-    def _MoveTo_policy(state: State, memory: Dict, objects: Sequence[Object],
+    def _MoveTo_policy(self, state: State, memory: Dict, objects: Sequence[Object],
                        params: Array) -> Action:
         # Move in the direction of the target.
         del memory, params  # unused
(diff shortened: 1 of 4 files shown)


predicators/envs/touch_point.py
```diff
@@ -198,3 +198,71 @@ def _event_to_action(state: State,
...
+        next_state = state.copy()
+        next_state.set(self._robot, "x", new_x)
+        next_state.set(self._robot, "y", new_y)
+        return next_state
+
+    @staticmethod
+    def _MoveTo_policy(state: State, memory: Dict, objects: Sequence[Object],
+                       params: Array) -> Action:
+        # Move in the direction of the target.
+        del memory, params  # unused
```
tomsilver, line 260: should we make the policy actually use the params? for example, if we're learning to imitate this policy, then we may end up learning something that effectively ignores the params (and then might as well be not parameterized)

tomsilver: a bonus for using the params would be verifying that the sampler is working as intended

tomsilver (approved):
one question but otherwise LGTM!

predicators/envs/touch_point.py
ashayathalye: Oh yeah, definitely. I forgot to copy that part when I was merging my older stuff.

PR #1372, opened 2023-02-09
[narrow_passage] Randomly vary narrow_passage door/passage widths between task initializations

lilianluong:
**Summary of changes:**
 - `narrow_passage` environment randomly initializes the widths of the door and narrow passage for each task, as opposed to using a fixed width based on CFG parameters
 - `narrow_passage` environment can no longer maintain a static geom cache
 - CFG settings changed to take lower bounds and upper bounds for the door and passage width paddings

### DIFF
--- a/predicators/envs/narrow_passage.py
+++ b/predicators/envs/narrow_passage.py
@@ -198,25 +198,26 @@ class NarrowPassageEnv(BaseEnv):
         goal = {goal_atom}
 
         # The initial positions of the robot and target vary, while wall and
-        # door positions are fixed. The robot should be above the walls, while
+        # door y positions are fixed. The robot should be above the walls, while
         # the dot should be below the walls (y coordinate)
         y_mid = (self.y_ub - self.y_lb) / 2 + self.y_lb
         margin = self.wall_thickness_half + self.init_pos_margin
-        # Door width is generated randomly
-        door_width_padding = rng.uniform(
-            CFG.narrow_passage_door_width_padding_lb,
-            CFG.narrow_passage_door_width_padding_ub,
-        )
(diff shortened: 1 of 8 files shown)


predicators/envs/narrow_passage.py
```diff
@@ -205,10 +202,18 @@ def _get_tasks(self, num: int, rng: np.random.Generator) -> List[Task]:
         # the dot should be below the walls (y coordinate)
         y_mid = (self.y_ub - self.y_lb) / 2 + self.y_lb
         margin = self.wall_thickness_half + self.init_pos_margin
-        door_width = (self.robot_radius +
-                      CFG.narrow_passage_door_width_padding) * 2
-        passage_width = (self.robot_radius +
-                         CFG.narrow_passage_passage_width_padding) * 2
+        # Door width is generated randomly
+        door_width_padding = rng.uniform(
```
tomsilver, line 206: this should be happening once per task, right?

predicators/envs/narrow_passage.py
```diff
@@ -205,10 +202,18 @@ def _get_tasks(self, num: int, rng: np.random.Generator) -> List[Task]:
...
+            CFG.narrow_passage_door_width_padding_lb,
+            CFG.narrow_passage_door_width_padding_ub,
+        )
+        door_width = (self.robot_radius + door_width_padding) * 2
+        # Passage width is generated randomly
+        passage_width_padding = rng.uniform(
```
tomsilver, line 212: ditto

tomsilver (commented):
should we modify the oracle refinement cost estimator so that it remains approximately oracular?

tomsilver (approved):
Lgtm!

PR #1375, opened 2023-02-09
Add TouchOpenEnv toy domain for testing nsrt rl approach.

ashayathalye:
Adds a toy environment involving two options for testing the nsrt RL approach, one for moving to a door (just like moving to a point in TouchPointEnv), and another for opening a door (just like opening a door in DoorsEnv).


tomsilver (approved):
LGTM, nice work!

PR #1378, opened 2023-02-12
extend state of pybullet robot to include end effector orientation

### LINKED ISSUE #1377: extend state of `SingleArmPyBulletRobot` to include end effector orientation, not just position
needed for environments where the end effector actually changes orientation. I am hoping that the existing pybullet environments (cover, blocks) will be mostly unchanged. it's possible that this will impact option learning, though, in which case we may want to automatically detect action dimensions that are static or something.

tomsilver:
closes #1377

### DIFF
--- a/predicators/envs/pybullet_blocks.py
+++ b/predicators/envs/pybullet_blocks.py
@@ -13,7 +13,7 @@ from predicators.envs.blocks import BlocksEnv
 from predicators.envs.pybullet_env import PyBulletEnv, create_pybullet_block
 from predicators.pybullet_helpers.controllers import \
     create_change_fingers_option, create_move_end_effector_to_pose_option
-from predicators.pybullet_helpers.geometry import Pose3D, Quaternion
+from predicators.pybullet_helpers.geometry import Pose, Pose3D, Quaternion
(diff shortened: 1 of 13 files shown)


predicators/envs/pybullet_blocks.py
```diff
@@ -216,15 +216,24 @@ def _initialize_pybullet(self) -> None:
...
+        robot_ee_orn = self._robot_ee_home_orn
+        ee_home = Pose(
+            (self.robot_init_x, self.robot_init_y, self.robot_init_z),
+            robot_ee_orn)
         return create_single_arm_pybullet_robot(CFG.pybullet_robot,
                                                 physics_client_id, ee_home)
 
     def _extract_robot_state(self, state: State) -> Array:
+        # The orientation is fixed in this environment.
+        q0, q1, q2, q3 = self._robot_ee_home_orn
```
williamshen-nz, line 228: I think these are xyzw quaternions, so maybe better to call them `qx, qy, qz, qw`? Same comment throughout changes

williamshen-nz (approved):
LGTM, just had comment about naming of quaternions

PR #1381, opened 2023-02-13
`SandwichEnvClear`

amburger66:
Gives each ingredient obj and the board a "clear" feature so that all predicates can be learned with the assumption that they are a function of only their argument's states


tomsilver (approved):
LGTM

PR #1382, opened 2023-02-13
`PlayroomEnv` variant

amburger66:
another version of the simple playroom env where each block has a "clear" attribute so we can learn all predicates with the assumption that they are a function of only their argument's states


tomsilver (approved):
LGTM

PR #1386, opened 2023-02-17
refactor initialize pg3 into analogy base class and sme subclass

tomsilver:
to be merged after the other open PG3-related PRs


predicators/approaches/pg3_analogy_approach.py
```diff
@@ -0,0 +1,65 @@
...
+        base_env = get_or_create_env(base_env_name)
+        target_env = get_or_create_env(target_env_name)
+        base_options = get_gt_options(base_env.get_name())
+        target_options = get_gt_options(target_env.get_name())
+        base_nsrts = get_gt_nsrts(base_env.get_name(), base_env.predicates,
+                                  base_options)
+        target_nsrts = get_gt_nsrts(target_env.get_name(),
+                                    target_env.predicates, target_options)
+        # Initialize with initialized policy from file.
+        if CFG.pg3_init_policy is None:  # pragma: no cover
```
ryangpeixu, line 44: Is there still a way to just run an initialized policy? Just checking in case we want to test something we write for how it performs in search.

predicators/approaches/sme_pg3_analogy_approach.py
```diff
@@ -1,97 +1,38 @@
...
-                                                   base_nsrts)
+        return "sme_pg3"
+
+    def _induce_policies_by_analogy(
+            self, base_policy: LiftedDecisionList, base_env: BaseEnv,
+            target_env: BaseEnv, base_nsrts: Set[NSRT],
+            target_nsrts: Set[NSRT]) -> List[LiftedDecisionList]:
         # Determine analogical mappings between the current env and the
         # base env that the initialized policy originates from.
         analogies = _find_env_analogies(base_env, target_env, base_nsrts,
```
ryangpeixu, line 38: I think this is good right now - the only thing is maybe in the future we want to move from a very well defined analogy object to a function? (but this can/probably will be in a different approach) - just making a note.

ryangpeixu (approved):

predicators/approaches/pg3_analogy_approach.py
tomsilver: good question, there is not... I'll add an issue so we remember to add this

predicators/approaches/sme_pg3_analogy_approach.py
tomsilver: that's a good thought yeah. leaving for the future sounds good

Issue #1379, opened 2023-02-12
Make ground truth options file


tomsilver:
And move all options out of environments. Mainly because I want to force us to not use the pybullet simulator (implementing "reality") in the options.

| Environment      | Status |
| ----------- | ----------- |
| `blocks`        | ✅       |
| `cluttered_table`   | ✅         |
| `coffee`   | ✅         |
| `cover`   | ✅         |
| `doors`   | 🔜        |
| `narrow_passage`   | ✅       |
| `painting`   | ✅         |
| `pddl_env`   | ✅         |
| `playroom`   | ✅          |
| `pybullet_blocks`        | 🔜       |
| `pybullet_cover`        | 🔜        |
| `repeated_nextto`   | ✅          |
| `sandwich`   | ✅            |
| `satellites`   | ✅          |
| `screws`   | ✅          |
| `stick_button`   | 🔜          |
| `tools`   |  ✅           |
| `touch_open`   | ✅          |
| `touch_point`   | ✅          |

PR #1389, opened 2023-02-19
separate ground truth NSRTs and options per environment

tomsilver:
working towards #1379 

the objective of this and forthcoming PRs is to move all options out of environments, for the following reasons:
1. philosophically, options live in the agent, not in the environment*
2. practically, the current setup is starting to cause problems in our pybullet environments.**

*for this reason, we may also want to remove `env.predicates` in the future too, but this is lower priority because it's not currently bothering me 🙂 

**our pybullet environments currently create two robots: one that is meant to be "in the environment" and one that is meant to be "in simulation" (in the agent). the latter is used for kinematics checks during planning. but now that we're working on adding more sophisticated motion planning, we would need not just two robots, but two of everything. adding two of everything would be really gross. so what we will do instead is force the agent to construct its own pybullet simulation inside of the ground truth options. and it will be up to the agent to sync up pybullet to the current state. if all goes well, we may even be able to remove state setting from the pybullet environment.

I'm opening this PR to get feedback before going forward. once we converge, there will be a million changes to make in `tests/` to get rid of all of the `env.options`. at the moment, I've just removed most of the `env.options` in `predicators/`.


NishanthJKumar:
Interesting - I understand and like the rationale for this proposal! It occurs to me that we currently have a long and somewhat-complicated `ground_truth_nsrts.py` file floating around as well. Is the idea that our new `ground_truth_options.py` file will basically look like this (that's what it seems to me from this PR)? Maybe we can think about combining these two or splitting them up into environment-specific files or something just to make things easier to grok?

tomsilver:
@NishanthJKumar thanks for reviewing! yeah, the idea was that this would mirror `ground_truth_nsrts.py`.
>Maybe we can think about combining these two or splitting them up into environment-specific files or something just to make things easier to grok? 

that's a good idea. how about the following structure?
```
ground_truth_models/
    __init__.py    # implements get_gt_nsrts(), get_gt_options, and shared utilities
    blocks/
        nsrts.py
        options.py
    cover/
        nsrts.py
        options.py
    ...
```

NishanthJKumar:
This is great! I think this new organization would make things much cleaner!

ashayathalye:
Makes a lot of sense! 

One question: why have the options be in a separate file but not have the samplers be in a separate file, especially when the samplers can be similarly complex? I wonder, especially as we do more sampler learning as part of the "learn options/samplers without demos project", if it would help to also have a standalone file for the samplers, e.g. 
```
cover/
        nsrts.py
        samplers.py
        options.py
```
Just putting it out there.

predicators/datasets/demo_only.py
```diff
@@ -14,6 +14,7 @@
 from predicators.approaches import ApproachFailure, ApproachTimeout
 from predicators.approaches.oracle_approach import OracleApproach
 from predicators.envs import BaseEnv
+from predicators.ground_truth_models import get_gt_options
```
tomsilver, line 17: useful linux command:
```git grep -l 'env\.options' | xargs sed -i '' '2s/^/from predicators.ground_truth_models import get_gt_options\n/'```

tomsilver:
this PR now does the following:
1. creates `ground_truth_models/` as discussed above
2. creates `get_gt_options`
3. moves the NSRTs and options for the basic cover environments
4. fixes all the imports and tests to remove `env.options`

future PRs will move the NSRTs and options over for the other environments and eventually the `options` property in `BaseEnv` will be removed completely

NishanthJKumar (approved):
LGTM with a very minor comment! Thanks for the quick work on such a major re-org! 😸

PR #1394, opened 2023-02-21
Get blind predicate invention to work on plain RNT

NishanthJKumar:
This PR includes a couple changes to get blind predicate invention to succeed on `rnt_single_option`.
* adds a new setting that can vary the `nextto_thresh` in RNT to make experimentation easier.
* adds a debug grammar for the RNT environment.
* introduces a timeout (and corresponding command-line flag) for `pnad_search` operator learning. This was necessary after noticing that some (bad) predicate hypotheses in RNT cause PNAD search to take a stupid amount of time to learn (and even run out of memory on my local machine - which has 128GB of RAM :grimacing: ).

This below command should achieve ~45/50 test tasks after roughly 3000s of runtime. The majority (>90%) of this time is spent evaluating predicate hypotheses:
```
python predicators/main.py --env repeated_nextto_single_option --approach grammar_search_invention --excluded_predicates all --strips_learner pnad_search --num_train_tasks 20 --seed 456 --repeated_nextto_nextto_thresh 6.25 --grammar_search_grammar_use_diff_features True --grammar_search_predicate_cost_upper_bound 10.0 --grammar_search_max_predicates 302
```


predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -404,6 +407,7 @@ def generate(self, max_num: int) -> Dict[Predicate, float]:
     def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
         env_name = (CFG.env if not CFG.env.startswith("pybullet") else
                     CFG.env[CFG.env.index("_") + 1:])
+        # num_times_pred_encountered = 0
```
tomsilver, line 410: remove

predicators/envs/repeated_nextto.py
```diff
@@ -26,7 +26,7 @@ class RepeatedNextToEnv(BaseEnv):
     env_lb: ClassVar[float] = 0.0
     env_ub: ClassVar[float] = 100.0
     grasped_thresh: ClassVar[float] = 0.5
-    nextto_thresh: ClassVar[float] = 0.5
+    nextto_thresh: ClassVar[float] = CFG.repeated_nextto_nextto_thresh
```
tomsilver, line 29: we can run into trouble sometimes if we use `CFG` in defining a class variable, I believe particularly in testing. so instead, can you initialize this field inside of `__init__`?

predicators/nsrt_learning/strips_learning/pnad_search_learner.py
```diff
@@ -68,10 +69,16 @@ def _append_new_pnad_and_keep_effects(
...
         """
+        start_time = time.perf_counter()
         new_pnads = current_pnads + [new_pnad]
         # We first repartition data and ensure delete and ignore
         # effects for the newly-created PNAD are correct.
         new_pnads = self._learner.recompute_pnads_from_effects(new_pnads)
+        # For certain predicates, the above call can eat up significant
+        # time and memory. Thus, we check here if we exceed the preset
+        # timeout.
+        if time.perf_counter() - start_time > CFG.pnad_search_timeout:
```
tomsilver, line 80: should this be `self._timeout`? or what is the relationship between `self._timeout` and `CFG.pnad_search_timeout`?

tests/test_utils.py
```diff
@@ -2780,6 +2780,15 @@ def _local_minimum_grid_heuristic_fn(state: S) -> float:
...
         ]
+        # Test timeout.
+        with pytest.raises(TimeoutError):
+            utils.run_hill_climbing(initial_state,
+                                    _grid_check_goal_fn,
+                                    _grid_successor_fn,
+                                    _local_minimum_grid_heuristic_fn,
+                                    enforced_depth=1,
+                                    parallelize=parallelize,
+                                    timeout=1e-012)
```
tomsilver, line 2791: any reason to not just make this 0?

tomsilver (commented):
very nice! a few questions

predicators/envs/repeated_nextto.py
NishanthJKumar: is `__init__` allowed to take in different arguments than just `self` and `use_gui`?

tomsilver: I just meant like this: https://github.com/Learning-and-Intelligent-Systems/predicators/blob/master/predicators/envs/blocks.py#L94

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -32,6 +31,7 @@ def __init__(
         self._verbose = verbose
         self._num_segments = sum(len(t) for t in segmented_trajs)
         assert len(self._trajectories) == len(self._segmented_trajs)
+        self.timeout = timeout
```
tomsilver, line 34: from our slack conversation: having a `timeout` here that is separate from `CFG.pnad_search_timeout` is a little awkward, but may be useful if we want to use a different timeout during predicate learning versus normal operator learning.

I am leaning toward recommending that we still just use `CFG.pnad_search_timeout` but I could be persuaded otherwise.

if we do keep the two separate, then I think we should make this private, and pass the timeout through to the creation of the search operators, rather than having the search operators access `self.learner.timeout`, because the latter obfuscates what is being used by the operators.

tests/test_predicate_search_score_functions.py
```diff
@@ -575,3 +575,10 @@ def test_expected_nodes_score_function():
     })
     with pytest.raises(AssertionError):
         score_function.evaluate({Holding, HandEmpty})
+    # Try with extremely small timeout and pnad_search.
+    utils.update_config({
+        "pnad_search_timeout": 1e-12,
```
tomsilver, line 580: oops I missed this one before (could be 0 as well)

tomsilver (commented):
back to you!

predicators/nsrt_learning/strips_learning/__init__.py
```diff
@@ -22,14 +22,12 @@
...
-    segmented_trajs: List[List[Segment]],
-    verify_harmlessness: bool,
-    verbose: bool = True,
-) -> List[PNAD]:
+def learn_strips_operators(trajectories: List[LowLevelTrajectory],
+                           train_tasks: List[Task],
+                           predicates: Set[Predicate],
+                           segmented_trajs: List[List[Segment]],
+                           verify_harmlessness: bool,
+                           verbose: bool = True) -> List[PNAD]:
```
tomsilver, line 30: don't forget to `yapf`, I'm guessing this function will not change from master after that

predicators/nsrt_learning/strips_learning/base_strips_learner.py
```diff
@@ -15,15 +15,13 @@
...
-        verify_harmlessness: bool,
-        verbose: bool = True,
-    ) -> None:
+    def __init__(self,
+                 trajectories: List[LowLevelTrajectory],
+                 train_tasks: List[Task],
+                 predicates: Set[Predicate],
+                 segmented_trajs: List[List[Segment]],
+                 verify_harmlessness: bool,
+                 verbose: bool = True) -> None:
```
tomsilver, line 24: ditto about `yapf`

tomsilver (approved):
LGTM, thanks!

predicators/nsrt_learning/strips_learning/base_strips_learner.py
NishanthJKumar: yapf actually forces it back to this change, and idk why lol...

PR #1395, opened 2023-02-22
factor out common submodule loading code

tomsilver:


lilianluong (approved):
Seems reasonable enough.

PR #1396, opened 2023-02-23
[ml_models] Added a CNNRegressor

lilianluong:
**Summary of Changes:**
- Added a `CNNRegressor` class modeled off of `MLPRegressor`, which is a regression model composed of CNN-ReLU-MaxPool blocks, a flatten layer, and then linear-ReLU layers.
- Added tests for `CNNRegressor`
- Refactored `ml_models.py` so that all instances of `self._x_dim` are now tuples instead of single integers, to make it more flexible for inputs with more than 2 dimensions. Added some assertions in other model definitions that X is 2D in cases where it seemed necessary.


tomsilver (approved):
excellent work!

PR #1397, opened 2023-02-23
[refinement estimation] Added a CNNRefinementEstimator

lilianluong:
**Summary of Changes:**
- Created `CNNRefinementEstimator` which trains a `CNNRegressor` per skeleton, and added associated tests
- Created `PerSkeletonRefinementEstimator` as a parent abstract class for `CNNRefinementEstimator` and `TabularRefinementEstimator`
- Added parameters to `settings.py` related to CNN regressor model parameters and CNN estimator image pre-processing
- Refactored refinement estimation--related methods to accept `Task` instead of `State` as input into the estimators, so that `render_state` environment methods can be used
- `BaseRefinementEstimator` saves an initialization of the environment in `self._env` so that it can be used, e.g. by oracle, CNN estimators

Use/test by passing in `--approach refinement_estimation --refinement_estimator cnn`. (Need to run `train_refinement_estimator.py` with the same flags first to collect data/train.)


predicators/envs/narrow_passage.py
```diff
@@ -323,6 +324,8 @@ def _MoveAndOpenDoor_initiable(self, state: State, memory: Dict,
         # Append open door action to memory action plan
         memory["action_plan"].append(
             Action(np.array([0.0, 0.0, 1.0], dtype=np.float32)))
+        # Opening the door takes a little bit of time to plan, artificially
+        time.sleep(0.2)
```
tomsilver, line 328: does this slow down any unit tests significantly? (what's the total time for unit tests before and after this PR?)

predicators/refinement_estimators/base_refinement_estimator.py
```diff
@@ -4,12 +4,17 @@
...
+from predicators.envs import get_or_create_env
+from predicators.settings import CFG
+from predicators.structs import GroundAtom, Task, _GroundNSRT
 
 
 class BaseRefinementEstimator(abc.ABC):
     """Base refinement cost estimator."""
 
+    def __init__(self) -> None:
+        self._env = get_or_create_env(CFG.env)
```
tomsilver, line 16: this is only needed for rendering right? if so, can we make that clear by just storing the render function here instead of the whole environment?

predicators/refinement_estimators/cnn_refinement_estimator.py
```diff
@@ -0,0 +1,138 @@
...
+
+from predicators.ml_models import CNNRegressor
+from predicators.refinement_estimators import BaseRefinementEstimator
+from predicators.settings import CFG
+from predicators.structs import GroundAtom, ImageInput, RefinementDatapoint, \
+    Task, _GroundNSRT
+
+# Type of the (skeleton, atoms_sequence) key for model dictionary
+# which converts both of them to be immutable
+ModelDictKey = Tuple[Tuple[_GroundNSRT, ...],  # skeleton converted to tuple
```
tomsilver, line 21: this looks like `CostDictKey` from the tabular estimator -- can we unify them?

predicators/refinement_estimators/cnn_refinement_estimator.py
```diff
@@ -0,0 +1,138 @@
...
+            model = self._create_regressor()
+            logging.info(f"Training CNN for skeleton {i}/{total_num_keys} "
+                         f"using {X.shape[0]} data points...")
+            t0 = time.perf_counter()
+            model.fit(X, Y)
+            logging.info(f"Fit model in {time.perf_counter() - t0:.2f}s")
+            self._model_dict[key] = model
+
+    @staticmethod
+    def _immutable_model_dict_key(
```
tomsilver, line 92: building on my question about unifying `CostDictKey` and `ModelDictKey`, I'm wondering if we can remove this copied code as well. how about implementing a parent class for this and the tabular approach called `PerSkeletonRefinementEstimator`?

tomsilver (commented):
very nice! just a few small things

predicators/refinement_estimators/base_refinement_estimator.py
lilianluong: It's also used to get env.types in the oracle estimator

predicators/refinement_estimators/per_skeleton_estimator.py
```diff
@@ -0,0 +1,66 @@
...
+        if key not in self._model_dict:
+            return float('inf')
+        model = self._model_dict[key]
+        cost = self._model_predict(model, initial_task)
+        return cost
+
+    @abc.abstractmethod
+    def _model_predict(self, model: T, initial_task: Task) -> float:
+        """Get the cost prediction from a model using the initial task as
+        input."""
```
tomsilver, line 49: wait we don't need to `raise NotImplementedError`? lol why are we doing this everywhere else in the code... maybe i'll change it someday

tomsilver (approved):
LGTM, awesome work, and I learned something new!

PR #1402, opened 2023-02-27
Added online_learning_lifelong Flag

wmcclinton:
This PR adds an online_learning_use_last_state, to load the last_state during online exploration instead of sampling a random state. Try with command:

```python predicators/main.py --approach online_nsrt_learning --seed 0  --env cover --explorer glib --max_initial_demos 1 --num_train_tasks 1 --num_test_tasks 10 --max_num_steps_interaction_request 10000 --min_data_for_nsrt 10 --num_online_learning_cycles 1 --online_learning_lifelong True```

### DIFF
--- a/predicators/envs/base_env.py
+++ b/predicators/envs/base_env.py
@@ -345,4 +345,4 @@ class BaseEnv(abc.ABC):
 
     def get_state(self) -> State:
         """Get the current state of this environment."""
-        return self._current_state
+        return self._current_state.copy()
--- a/predicators/envs/base_env.py
+++ b/predicators/envs/base_env.py
@@ -342,3 +342,7 @@ class BaseEnv(abc.ABC):
         """
         raise NotImplementedError("This environment did not implement an "
                                   "interface for human demonstrations!")
+
+    def get_state(self) -> State:
+        """Get the current state of this environment."""
+        return self._current_state
--- a/predicators/main.py
+++ b/predicators/main.py
@@ -212,7 +212,7 @@ def _generate_interaction_results(
     logging.info("Generating interaction results...")
     results = []
     query_cost = 0.0
-    last_state = None
+    do_state_reset = True
     if CFG.make_interaction_videos:
         video = []
     for request in requests:
@@ -222,7 +222,6 @@ def _generate_interaction_results(
                                "if allow_interaction_in_demo_tasks is False.")
         monitor = TeacherInteractionMonitorWithVideo(env.render, request,
                                                      teacher)
-        init_state = last_state if CFG.online_learning_lifelong else None
         traj, _ = utils.run_policy(
             request.act_policy,
             env,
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: predicators/main.py [hunk 1]
```diff
@@ -228,11 +229,15 @@ def _generate_interaction_results(
             request.train_task_idx,
             request.termination_function,
             max_num_steps=CFG.max_num_steps_interaction_request,
+            init_state=last_state
```
tomsilver, line 232: can we move this out for better readability? so `init_state = CFG.online_learning_use_last_state else None` above `utils.run_policy` and then just `init_state=init_state` here

predicators/utils.py
```diff
@@ -905,6 +905,7 @@ def run_policy(
         task_idx: int,
         termination_function: Callable[[State], bool],
         max_num_steps: int,
+        init_state: State = None,
```
tomsilver, line 908: i'm surprised mypy doesn't complain but I think the type should be `Optional[State]`

predicators/utils.py
```diff
@@ -922,7 +923,8 @@ def run_policy(
     last action from the returned trajectory to maintain the invariant that
     the trajectory states are of length one greater than the actions.
     """
-    state = env.reset(train_or_test, task_idx)
+    state = init_state if init_state is not None else env.reset(
```
tomsilver, line 926: thinking out loud... if `env.reset` is not called, we're assuming that the environment's internal state here is equal to `init_state`, which should be true in the place that you're invoking it. however if this function was called by a different user who passed in an arbitrary `init_state`, then the environment would be out of sync with the `init_state` and there wouldn't be any error raised or anything, so it would get quite confusing.

ideally we would be able to do something like `assert env.get_state().allclose(init_state)` but we don't have a `get_state` method on environments. I suppose we could add one?

if we do add that method, then we might as well just call `state = env.get_state()` here instead of passing the `init_state` in as an argument

I can't immediately think of any better way, let me know if you have thoughts

predicators/settings.py
```diff
@@ -469,6 +469,7 @@ class GlobalSettings:
     # online NSRT learning parameters
     online_nsrt_learning_requests_per_cycle = 10
     online_learning_max_novelty_count = 0
+    online_learning_use_last_state = False
```
tomsilver, line 472: let's name this something with `reset_state` or `lifelong` in the name instead?

tomsilver (commented):
thanks for doing this! sorry for a number of comments, but this is one of those small changes with many implications

wmcclinton:
> thanks for doing this! sorry for a number of comments, but this is one of those small changes with many implications

No problem. Addressed.

predicators/envs/base_env.py
```diff
@@ -342,3 +342,7 @@ def get_event_to_action_fn(
         """
         raise NotImplementedError("This environment did not implement an "
                                   "interface for human demonstrations!")
+
+    def get_state(self) -> State:
+        """Get the current state of this environment."""
+        return self._current_state
```
tomsilver, line 348: let's do `self._current_state.copy()` to avoid any changes to the environment's internal state

predicators/main.py
```diff
@@ -212,6 +212,7 @@ def _generate_interaction_results(
     logging.info("Generating interaction results...")
     results = []
     query_cost = 0.0
+    last_state = None
```
tomsilver, line 215: thanks for making this change! can we give this a try?

>if we do add that method, then we might as well just call state = env.get_state() here instead of passing the init_state in as an argument

so in particular:
1. remove the current changes in `main.py`
2. rather than `init_state`, make the new argument to `run_policy` something like `do_state_reset: bool = False`
3. then the only change in `main.py` can be `do_state_reset=(not CFG.online_learning_lifelong)`
4. change the implementation of `run_policy` accordingly
5. add a comment that the `request.train_task_idx` is ignored if `do_state_reset` is True

tomsilver (commented):
thanks for the quick turnaround! two more requests

wmcclinton:
> thanks for the quick turnaround! two more requests

Done.

tomsilver (approved):
LGTM, thanks again!

PR #1404, opened 2023-02-28
Use feature changes to cut down on the number of generated predicates during grammar search

NishanthJKumar:
This PR implements the ability for the `_SingleFeatureInequalityGrammar` to only generate predicates that involve features that change in a particular transition. Specifically, we repeatedly:
* pick a particular transition
* pick some number of constants to generate from the `constant_generator` (this is the hyperparameter: `grammar_search_num_constants_each_iter`)
* generate predicates corresponding to these constants such that the features of these predicates change in the selected transition.

The PR also adds some timeouts to part of `pnad_search` to prevent it from taking a long time/running out of memory...

The below command achieves 50/50 test time performance on RNT in only ~2200s (down from ~3000s previously, but definitely still more to do!).
```
python predicators/main.py --env repeated_nextto_single_option --approach grammar_search_invention --excluded_predicates all --strips_learner pnad_search --num_train_tasks 20 --seed 456 --repeated_nextto_nextto_thresh 6.25 --grammar_search_grammar_use_diff_features True --grammar_search_predicate_cost_upper_bound 10.0 --grammar_search_max_predicates 135 --grammar_search_enumeration_approach "transition_changes"
```


predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -454,12 +455,58 @@ def _halving_constant_generator(
...
+    def _get_changing_feats_for_transition(
+            self, transition: Tuple[State, State]) -> Set[Tuple[Type, str]]:
+        """Given a transition, gets all the features that change during the
+        transition."""
+        changing_types_and_feats: Set[Tuple[Type, str]] = set()
+        for obj in transition[0].data:
+            for feat in obj.type.feature_names:
+                if np.sum(
+                        np.subtract(transition[0].data.get(obj, feat),
+                                    transition[1].data.get(obj, feat))) != 0:
```
tomsilver, line 485: checking for 0 seems strong, should we use `np.isclose`? (beyond noise, there could just be floating point issues)

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -471,20 +518,62 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+        constant_idx = 0
+        while True:
+            # First generate a set of constants that all predicates in
+            # this iteration will be evaluated on.
+            curr_iter_constants_and_costs = []
+            for _ in range(CFG.grammar_search_num_constants_each_iter):
+                try:
+                    curr_iter_constants_and_costs.append(
+                        next(constant_generator))
+                except StopIteration:  # pragma: no cover
```
tomsilver, line 538: if this is impossible, why do we need to catch it?

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -471,20 +518,62 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+    def _enumerate_transition_changes(
+        self, feature_ranges: Dict[Type, Dict[str, Tuple[float, float]]]
+    ) -> Iterator[Tuple[Predicate, float]]:
+        constant_generator = _halving_constant_generator(0.0, 1.0)
+        constant_idx = 0
+        while True:
+            # First generate a set of constants that all predicates in
+            # this iteration will be evaluated on.
+            curr_iter_constants_and_costs = []
+            for _ in range(CFG.grammar_search_num_constants_each_iter):
```
tomsilver, line 534: idk if this is better but you could also do

```python
curr_iter_constants_and_costs = list(itertools.islice(constant_generator, CFG.grammar_search_num_constants_each_iter))
```

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -471,20 +518,62 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
+                    for t in sorted(self.types):
+                        for f in t.feature_names:
+                            if (t, f) not in changing_types_and_feats:
+                                continue
+                            lb, ub = feature_ranges[t][f]
+                            # Optimization: if lb == ub, there is no variation
+                            # among this feature, so there's no point in
+                            # trying to learn a classifier with it. So, skip
+                            # the feature.
+                            if abs(lb - ub) < 1e-6:
```
tomsilver, line 555: I'm wondering how this line gets covered (it must get covered somehow because the coverage check passes). if a certain feature never changes, then how could it show up in `changing_types_and_feats`?

tomsilver: maybe the difference is that you're checking `<1e-6` here but checking `==0` before?

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -533,34 +649,63 @@ def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
...
-                        types = [t1, t2]
-                        pred = Predicate(name, types, diff_classifier)
-                        assert pred.arity == 2
+                        pred = self._construct_binary_pred(
+                            t1, t2, f1, f2, constant, ub1, ub2, lb1, lb2,
+                            constant_idx)
                         yield (pred, 2 + cost
                                )  # cost = arity + cost from constant
 
+    def _enumerate_transition_changes(
```
tomsilver, line 658: we can leave this for the future, but it feels like we should be able to refactor this and the other class to share some of this logic

tomsilver (commented):
nice work! tricky stuff but you made it very clear and easy to review :)

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -454,12 +455,58 @@ def _halving_constant_generator(
...
+                             "length greater than 2")
+        # We get transitions backwards thru the trajectory.
+        # We skip the last transition because we already have the goal
+        # predicates, so we don't need to 'inspect' this transition
+        # to invent new predicates that it is trying to achieve.
+        for idx_into_traj in range(-2, (-1 * max_traj_len) - 1, -1):
+            for traj in self.dataset.trajectories:
+                if len(traj.states) < (idx_into_traj * -1) + 1:
+                    continue
+                yield (traj.states[idx_into_traj - 1],
```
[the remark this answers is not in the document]
NishanthJKumar: ahh - good point! Yes - will modify to do that instead!

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: if we don't, `pylint` starts to complain

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: yeah probably actually! Good point, i'll update the previous check and then adjust this accordingly

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: agreed - i couldn't see an immediately-clean way to do it tho...

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -454,12 +455,58 @@ def _halving_constant_generator(
...
+                       traj.states[idx_into_traj])
+
+    def _get_changing_feats_for_transition(
+            self, transition: Tuple[State, State]) -> Set[Tuple[Type, str]]:
+        """Given a transition, gets all the features that change during the
+        transition."""
+        changing_types_and_feats: Set[Tuple[Type, str]] = set()
+        for obj in transition[0].data:
+            for feat in obj.type.feature_names:
+                if np.sum(
```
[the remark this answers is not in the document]
NishanthJKumar: nope - i just checked in the terminal that the result of `np.subtract` is itself an array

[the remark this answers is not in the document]
tomsilver: oh lol wait something is super weird... don't we want to be doing `state.get` instead of `state.data.get`? I actually have no idea what the latter will do, I'm surprised it doesn't crash...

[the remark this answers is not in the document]
tomsilver: okay I now understand why this doesn't crash I think, but this seems like a big bug, maybe fixing will lead to even bigger improvements. I would like to know why it led to any improvements in the first place

NishanthJKumar:
Closing since we realized this approach is flawed and decided to go in another direction!

PR #1408, opened 2023-03-02
A new explorer that takes random nsrts

ashayathalye:
Similar to RandomOptionsExplorer in that it chooses uniformly at random out of whichever ground nsrts are applicable in the current state, but different in that the continuous parameter of the parameterized option is generated by the nsrt's sampler rather than by sampling uniformly at random from the option's parameter space.

This explorer is intended to be used when learning options via reinforcement learning with oracle samplers. In this setting, planning (e.g. using BilevelPlanningExplorer) is not feasible because refinement is slow or fails when the learned options aren't good enough yet.

Note that the sampler requires the current low-level state as input. An alternative approach would generate a plan of options by planning towards the goal with the ground nsrts and avoid refinement by (1) inferring the terminal low-level state of each option (the initial low-level state of the subsequent option) from the option's sample, which is proposing a low-level subgoal, and (2) setting the simulator state directly to that state. Because applying dimensionality reduction techniques on the sample makes this approach infeasible, and this is something we may want to do, we avoid this approach for now.


predicators/explorers/random_nsrts_explorer.py
```diff
@@ -0,0 +1,108 @@
...
+from predicators.structs import NSRT, Action, DummyOption, \
+    ExplorationStrategy, ParameterizedOption, Predicate, State, Task, Type, \
+    _GroundNSRT, _Option
+
+
+class RandomNSRTsExplorer(BaseExplorer):
+    """RandomNSRTsExplorer implementation.
+
+    Similar to RandomOptionsExplorer in that it chooses
+    uniformly at random out of whichever ground nsrts are
```
tomsilver, line 18: nit: capitalize `nsrt` here and below

predicators/explorers/random_nsrts_explorer.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self, predicates: Set[Predicate],
+                 options: Set[ParameterizedOption], types: Set[Type],
+                 action_space: Box, train_tasks: List[Task],
+                 nsrts: Set[NSRT]) -> None:
+
+        super().__init__(predicates, options, types, action_space, train_tasks)
+        self._nsrts = nsrts
+
+        # Sequence of options generated during exploration.
+        self._option_seq: Dict[int, List[_Option]] = {}
```
tomsilver, line 52: add a comment explaining the key to this dict

predicators/explorers/random_nsrts_explorer.py
```diff
@@ -0,0 +1,108 @@
...
+        return "random_nsrts"
+
+    def get_option_sequence(self, train_task_idx: int) -> List[_Option]:
+        """Get the sequence of options produced by the exploration strategy for
+        a particular task."""
+        return self._option_seq[train_task_idx]
+
+    def get_exploration_strategy(self, train_task_idx: int,
+                                 timeout: int) -> ExplorationStrategy:
+        # Ensure random over successive calls.
```
tomsilver, line 65: is this comment left over from something? (I don't see what this is referring to)

predicators/explorers/random_nsrts_explorer.py
```diff
@@ -0,0 +1,108 @@
...
+        cur_option = DummyOption
+        task = self._train_tasks[train_task_idx]
+        self._option_seq[train_task_idx] = []
+
+        def fallback_policy(state: State) -> Action:
+            del state  # unused
+            raise utils.RequestActPolicyFailure(
+                "No applicable nsrt in this state!")
```
tomsilver, line 73: capitalize nsrt here too

predicators/explorers/random_nsrts_explorer.py
tomsilver: actually I am not clear on the need for this dict (the random options explorer doesn't need one, for example). if you need it for RL, is it possible to store what's needed inside of the RL approach itself?

predicators/explorers/random_nsrts_explorer.py
```diff
@@ -0,0 +1,108 @@
...
+                    return fallback_policy(state)
+                assert all(a.holds for a in ground_nsrt.preconditions)
+
+                # Sample an option.
+                option = ground_nsrt.sample_option(state,
+                                                   goal=task.goal,
+                                                   rng=self._rng)
+                cur_option = option
+
+                # Store option (so that we can attribute experience
```
tomsilver, line 98: yeah this part is still not clear to me. it probably makes more sense in the context of the RL approach, but that's a good reason to put this dict inside that approach instead

tomsilver (commented):
very nice, thanks! a few small things

predicators/explorers/random_nsrts_explorer.py
ashayathalye: There are two options:

(1) instantiate an explorer only once (e.g. in `init()` in `nsrt_rl_approach`), which then leads us to creating this dict that maps train task id -> list of options it generated when exploring on that task, because this explorer will keep track of the options its policy generates *for each task*. 

(2) instantiate a new explorer inside `get_interaction_requests()`, so that you have a new explorer for each task. In this case, we don't need the dictionary, but then we'd need to map train task id -> explorer in `nsrt_rl_approach`. 

In `_get_experience_from_result()`, we will ask the explorer what the list of options was for a particular training task. 

I'm fine with either (1) or (2). Do you have a different suggestion for how to do this?

ashayathalye: For each interaction_result we get back, we want to assign the tuples of (s, a, s') in that to whichever options generated that transition. To do that, we need an *ordered* list of options that we used to generated it. Could you do this from an unordered set of options? Not quite: because two options may be initiable on the same state s, so you wouldn't know which one of the two options generated (s, a, s'). You could figure out by running both options on that state, but that's extra work: since the explorer can easily keep track of which option sequence we used for that interaction result, we might as well use that information. 

So, we might want a map of interaction_request (which leads to an interaction_result) -> list of options used to generate that result. But each interaction_request corresponds to a train_task_idx, so an equivalent mapping is train_task_idx -> list of options. 

Now, *technically*, it would be ok for us to only maintain a SINGLE, *ordered* sequence of options (basically, think of concatenating the list of options from each train_task_idx, and just keep track of which index we are on as we loop through that entire list, when assigning (s, a, s') to an option). But this is more confusing, and prone to more bugs, I think. It's easier to keep track of it separately by train_task_idx (by interaction_result).

tomsilver: > we want to assign the tuples of (s, a, s') in that to whichever options generated that transition

Could we accomplish this using a.get_option()? (Note that the option is associated with the action inside of _Option.policy)

ashayathalye: Oh yeah, I forgot about that, that should work. Will remove the `get_option_sequence()` function, then.

tomsilver (approved):
LGTM!

PR #1411, opened 2023-03-04
move options out of doors environment 

tomsilver:
also move the caches into the simulator state (see `utils.StateWithCache`)


lilianluong (approved):
BB8 Thumbs Up

PR #1412, opened 2023-03-04
move options out of stick button environment

tomsilver:


lilianluong (approved):
Spongebob Many Thumbs

PR #1414, opened 2023-03-05
moving ground-truth options for a few more cover envs 

tomsilver:
I forgot about these. after these, hopefully only the pybullet ones remain


NishanthJKumar (approved):
LGTM!

PR #1415, opened 2023-03-07
move options out of pybullet environments

tomsilver:


NishanthJKumar (approved):
LGTM! Thanks for working thru all these intricate changes!

PR #1416, opened 2023-03-08
GPU training + weight decay in ml_models.py

lilianluong:
**Summary of changes:**
- Add weight decay parameters to `PyTorchRegressor`, `PyTorchBinaryClassifier`, and subclasses
- Add `use_torch_gpu` boolean parameters to the same classes, which converts models and data when fitting to the "cuda:0" device if it is available when set to True
- `CNNRefinementEstimator` sets `use_torch_gpu=True` when creating `CNNRegressor` models if the global `use_torch_gpu` flag is set to True

**Test:**

Manually tested using the following commands, since PyTorch with GPU isn't by default configured for the testing suite. (Note: to set up PyTorch for GPU usage, I removed the current PyTorch and reinstalled it using the command [here](https://pytorch.org/get-started/locally/))

1. `python predicators/train_refinement_estimator.py --env narrow_passage --approach refinement_estimation --refinement_estimator cnn --seed 0 --experiment_id cnn --cnn_regressor_max_itr 1 --num_train_tasks 20 --use_torch_gpu True`
2. `python predicators/main.py --env narrow_passage --approach refinement_estimation --refinement_estimator cnn --seed 0 --experiment_id cnn --num_test_tasks 5 --use_torch_gpu True`


predicators/ml_models.py
```diff
@@ -1041,7 +1135,8 @@ def _train_pytorch_model(model: nn.Module,
                          batch_generator: Iterator[Tuple[Tensor, Tensor]],
                          max_train_iters: MaxTrainIters,
                          dataset_size: int,
-                         print_every: int = 1000,
+                         device: torch.device,
+                         print_every: int = 10,
```
tomsilver, line 1139: can we make this a setting that gets passed through? and then make sure that when you run the command below before and after your change, the printouts and time taken look about the same:

`python predicators/main.py --approach nsrt_learning --env cover --seed 0`

tomsilver (approved):
one minor thing, but LGTM! thanks a lot for figuring all this out.

lilianluong:
Would be great if you could just review the new commit, since I figured I'd add weight_decay, use_torch_gpu, and train_print_every arguments to other usages of models throughout the repo as well.

tomsilver:
LGTM, thanks a lot for doing that!

PR #1418, opened 2023-03-08
initialize pg3 from file

### LINKED ISSUE #1409: pg3 initialize from policy file
we can probably implement this inside of the pg3 approach itself

tomsilver:
closes #1409


ryangpeixu (approved):
lgtm!

PR #1419, opened 2023-03-09
GNN refinement estimator

lilianluong:
**Summary of changes:**
 - `run_low_level_search()` optionally times refinement for each step of the skeleton, which is recorded separately in refinement cost training datasets
 - Implemented and added tests for `GNNRefinementEstimator`; this is based on the GNN option policy approach with some modifications
 - `compute_normalizers()` in `gnn_utils.py` optionally ignores a subset of nodes, edges, and globals instead of adding normalization to all of them

**Usage:** use `--refinement_estimator gnn` when running `predicators/train_refinement_estimator.py` and `predicators/main.py`

**Tests:** `pytest tests/refinement_estimators/test_gnn_refinement_estimator.py`

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -82,3 +82,9 @@ ignore_missing_imports = True
 
 [mypy-panda_robot_client.*]
 ignore_missing_imports = True
+
+[mypy-gym.*]
+ignore_missing_imports = True
+
+[mypy-gym_sokoban.*]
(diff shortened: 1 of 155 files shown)


tomsilver (approved):
LGTM, great work! Especially with the tests. Let's get some preliminary results before merging just to make sure nothing goes unexpectedly wrong.

PR #1422, opened 2023-03-09
pseudocode for perception modules

tomsilver:


tomsilver:
based on discussion: consider switching from `solve` to `step`, `reset`, etc.

PR #1424, opened 2023-03-10
add `bilevel_plan_without_sim` setting

### LINKED ISSUE #1420: implement flag for assuming downward refinability / perfect sampling
crucially, when this flag is on, we will never call `simulate` (so environments like `CozmoEnv` don't need to implement `simulate` or even `step`)

tomsilver:
with `--bilevel_plan_without_sim True`, it will now be possible to run bilevel planning-based approaches but with only task planning run, to avoid the need for a `simulate()` function in the environment.

this flag will only work in environments that are not "TAMP-interesting" in that the first skeleton is assumed to work and all samples are assumed to work.

it's useful for getting simpler environments off the ground (e.g., forthcoming sokoban environment, cozmo environment.)

closes #1420

### DIFF
--- a/predicators/approaches/bilevel_planning_approach.py
+++ b/predicators/approaches/bilevel_planning_approach.py
@@ -5,7 +5,8 @@ then Execution.
 """
 
 import abc
-from typing import Any, Callable, List, Set, Tuple
+import time
+from typing import Any, Callable, List, Optional, Sequence, Set, Tuple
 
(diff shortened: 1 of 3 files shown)


NishanthJKumar (approved):
LGTM! Tysm for the quick work on this!

wmcclinton:
Nice! Thanks!

PR #1426, opened 2023-03-10
gym sokoban environment

NishanthJKumar:
Implements a `sokoban` environment on top of the [gym_sokoban](https://github.com/mpSchrader/gym-sokoban) package. Also implements a task-planning heavy oracle approach that solves any problem instance!

Example command:
```
python predicators/main.py --approach oracle --env sokoban --seed 1 --num_train_tasks 0 --num_test_tasks 5 --make_test_videos --sesame_task_planner fdopt --bilevel_plan_without_sim True
```

Example videos:

https://user-images.githubusercontent.com/12446934/224580419-04906200-a2c5-484f-8ff0-338fa9804f2b.mp4

https://user-images.githubusercontent.com/12446934/224580420-f2ba437d-f0ac-4984-86fe-8d89276d804d.mp4

https://user-images.githubusercontent.com/12446934/224580421-d36d743e-d77c-42fa-8f9b-6c8cb972d57c.mp4

https://user-images.githubusercontent.com/12446934/224580424-08e6244e-da4a-4b7d-8194-d1a0f78db0f1.mp4

https://user-images.githubusercontent.com/12446934/224580425-9847271a-88d2-4123-86b4-0ac63328b335.mp4

PR #1427, opened 2023-03-12
implements FD task planning

NishanthJKumar:
Enables use of FD with the `--bilevel_plan_without_sim` flag set to `True`. Will be useful for environments like sokoban (#1426 ) that are challenging for task planning.

Example command:
```
python predicators/main.py --seed 0 --env pddl_blocks_procedural_tasks --approach oracle --num_train_tasks 0 --num_test_tasks 5 --bilevel_plan_without_sim True --sesame_task_planner fdopt
```


tomsilver (approved):
thanks for doing some refactoring to minimize redundant code! LGTM after the unit tests pass

PR #1428, opened 2023-03-13
collect demos without sim

tomsilver:
before this PR, the following would crash:

```python predicators/main.py --approach nsrt_learning --env pddl_easy_delivery_procedural_tasks --seed 0 --num_train_tasks 10 --bilevel_plan_without_sim True```

now it gets 50/50


NishanthJKumar (approved):
LGTM!

PR #1431, opened 2023-03-13
cogman

### LINKED ISSUE #1421: implement `Observation` data structure and closed loop interaction with environment
use Sokoban as example

`env.step` will return `Observation`

tomsilver:
for perception and execution monitoring

closes #1421


predicators/cogman.py
```diff
@@ -0,0 +1,50 @@
+"""Cognitive manager (CogMan).
+
+A wrapper around an approach that managers interaction with the environment.
```
ronuchit, line 3: manages*

predicators/cogman.py
```diff
@@ -0,0 +1,84 @@
...
+            assert self._current_goal is not None
+            task = Task(state, self._current_goal)
+            new_policy = self._approach.solve(task, timeout=CFG.timeout)
+            self._current_policy = new_policy
+            self._exec_monitor.reset(task)
+        assert self._current_policy is not None
+        act = self._current_policy(state)
+        return act
+
+    # The methods below provide an interface to the approach.
```
NishanthJKumar, line 53: I'd imagine in the future, we might want to have some methods here that allow us to interface with the perception module or exec monitor respectively? If so, maybe we can indicate this in this comment

predicators/envs/base_env.py
```diff
@@ -176,6 +181,19 @@ def get_test_tasks(self) -> List[EnvironmentTask]:
...
+        """Default for environments where states are observations."""
+        assert isinstance(self._current_observation, State)
+        return self._current_observation
+
+    def goal_reached(self) -> bool:
+        """Default implementation assumes environment tasks are tasks.
+
+        Subclasses may override.
+        """
+        return self._current_task.task.goal_holds(self._current_state)
```
NishanthJKumar, line 195: Following my comment on the `EnvironmentTask` PR, I still feel like CogMan should at least produce the state, if not the actual goal as well. Maybe we can move the dummy methods in `EnvironmentTask` into CogMan?

predicators/main.py
```diff
@@ -93,13 +97,13 @@ def main() -> None:
...
-    # and their initial states. We only make this assumption for the training
-    # tasks, we don't need to make it for the test tasks.
-    train_tasks = [t.task for t in env_train_tasks]
+    # We assume that a train Task can be constructed from a EnvironmentTask.
+    # In other words, the initial obs is assumed to contain enough information
+    # to determine all of the objects and their initial states. We only make
+    # this assumption for the training tasks, we don't need to make it for the
+    # test tasks. It is also something we can weaken later if needed.
+    perceiver = create_perceiver(CFG.perceiver)
+    train_tasks = [perceiver.reset(t) for t in env_train_tasks]
```
NishanthJKumar, line 106: Hold on - I'm pretty confused now, because this is exactly what I would've expected! If we're using the perceiver to generate train tasks for us, then why are we directly getting init state and goal from the `EnvironmentTask` in `base_env.py`. Don't we want to be using the perceiver there as well in exactly this way?

predicators/main.py
```diff
@@ -393,12 +397,82 @@ def _run_testing(env: BaseEnv, approach: BaseApproach) -> Metrics:
...
+        for _ in range(max_num_steps):
+            monitor_observed = False
+            exception_raised_in_step = False
+            try:
+                start_time = time.perf_counter()
+                act = cogman.step(obs)
+                metrics["policy_call_time"] += time.perf_counter() - start_time
+                # Note: it's important to call monitor.observe() before
+                # env.step(), because the monitor may use the environment's
+                # internal state.
```
NishanthJKumar, line 451: Isn't this 'cheating'? Are we planning to eventually move away from this? If so, would be good to note that here.

predicators/perception/sokoban_perceiver.py
```diff
@@ -0,0 +1,97 @@
...
+from predicators.envs.sokoban import SokobanEnv
+from predicators.perception.base_perceiver import BasePerceiver
+from predicators.structs import EnvironmentTask, GroundAtom, Object, \
+    Observation, State, Task
+
+# Each observation is a tuple of four 2D boolean masks (numpy arrays).
+# The order is: free, goals, boxes, player.
+
+
+class SokobanPerceiver(BasePerceiver):
```
NishanthJKumar, line 17: Great work on this! I LOVE how this perception is now separated from the env, and am very glad we decided to implement `Sokoban`! 😸

NishanthJKumar (commented):
Fantastic work 🎉 ! Just a few questions related to the `EnvironmentTask` PR as well

predicators/main.py
tomsilver: well we certainly don't want to be using the perceiver inside of the environment because that would break the environment/agent boundary, but you're right that using `EnvironmentTask.task` inside of the environment to check if the goal holds is not ideal. I'm only doing it out of convenience because all of the existing environment (except sokoban) have goal descriptions that are actually goals (ground atoms). but maybe a step in the right direction would be to change this implementation to assert that the environment task goal description is a set of ground atoms before checking the goal? and then I can also add a comment to clarify things?

predicators/main.py
tomsilver: kinda, but the monitor isn't part of the agent, it's just meant to collect statistics and stuff like that. so it doesn't really matter what it accesses

predicators/main.py
NishanthJKumar: I think that would help! iirc, you already check that somewhere, but explicitly checking it in `base_env.py` I think would help clarify things

predicators/main.py
NishanthJKumar: hmmm - that's fair. When we think of actually using a robot tho, wouldn't the monitor have to use a perceiver or something like that?

tomsilver: good point. okay after thinking about it more, I think the original comment is actually misleading, and I was wrong to say that the monitor is cheating. see the updated comment

NishanthJKumar: nice - i like it!

NishanthJKumar: wait - i'm kinda lost now; how come the `BaseExecutionMonitor` doesn't implement an `observe()` method?

NishanthJKumar (approved):
looks overall good to me, just a quick question/confusion!

predicators/main.py
tomsilver: ohhh I understand the confusion. this is talking about `Monitor` https://github.com/Learning-and-Intelligent-Systems/predicators/tree/master/predicators#L913 not the new execution monitor stuff 😅

NishanthJKumar: Wait - given this, can we do some renaming to avoid this confusion going forward? We could call the current `Monitor` something like `LoggingMonitor`? Or we could rename `Execution Monitor` to `ExMon` in keeping with the `CogMan` convention?

tomsilver: I changed `Monitor` to `LoggingMonitor`. I think `ExMon` would probably be more confusing, so I'll vote for keeping `ExecutionMonitor`

PR #1436, opened 2023-03-16
make sokoban seeds independent

### LINKED ISSUE #1432: change consecutive seeding of sokoban tasks
Otherwise when we do experiments over multiple seeds, many of the tasks will be the same between seeds

Cc @NishanthJKumar

tomsilver:
before this, we would be seeding the gym environment with 0, 1, 2, 3... in the first experiment, and then 1, 2, 3, 4... in the second experiment, etc., because we usually just increment seeds when we run multiple trials.

closes #1432

### DIFF
--- a/predicators/envs/sokoban.py
+++ b/predicators/envs/sokoban.py
@@ -74,11 +74,10 @@ class SokobanEnv(BaseEnv):
         self._box_loc_to_name: Dict[Tuple[int, int], str] = {}
 
     def _generate_train_tasks(self) -> List[Task]:
-        return self._get_tasks(num=CFG.num_train_tasks, seed_offset=0)
+        return self._get_tasks(num=CFG.num_train_tasks, train_or_test="train")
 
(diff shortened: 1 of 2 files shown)


predicators/envs/sokoban.py
```diff
@@ -364,3 +361,14 @@ def _is_static(cls, obj: Object, state: State) -> bool:
     @classmethod
     def _is_dynamic(cls, obj: Object, state: State) -> bool:
         return not cls._is_static(obj, state)
+
+    @staticmethod
+    def _get_task_seed(train_or_test: str, task_idx: int) -> int:
+        assert task_idx < CFG.test_env_seed_offset
+        entropy = CFG.seed
```
wmcclinton, line 368: Explain what this and the following lines are doing, so I don't have to refer to np.random.SeedSequence documentation.

wmcclinton (approved):
LGTM! There is a minor comment I would add

NishanthJKumar (approved):

Issue #1443, opened 2023-03-29
updates to `_generate_interaction_results`


tomsilver:
- [ ] integrate cogman
- [ ] allow the teacher to take actions and then give back control to the agent

tomsilver:
plan:
1. replace `run_policy` with a new function that takes in both a teacher and a policy
2. remove the teacher monitors and make everything in terms of just the teacher
3. change datasets/demo_only to use the teacher (and never call the agent policy)

tomsilver:
working toward integrating cogman. command to test:

```python predicators/main.py --env sokoban --seed 0 --approach online_nsrt_learning --num_train_tasks 3 --num_test_tasks 1 --sesame_task_planner fdopt --bilevel_plan_without_sim True --sokoban_gym_name Sokoban-small-v0 --strips_learner oracle --sampler_learner oracle --max_initial_demos 0```

PR #1444, opened 2023-03-29
human demo collection during online interaction

tomsilver:
this is hacky, but addressing root issues (#1443) will require a lot more work. a better way to implement this would be to have the teacher and the student interacting in the same environment. that's certainly what we will want to do in the real world.

I tested this through the online bridge policy learning branch.

### DIFF
--- a/predicators/teacher.py
+++ b/predicators/teacher.py
@@ -3,15 +3,18 @@ information to assist an agent during online learning."""
 
 from __future__ import annotations
 
+import functools
 from dataclasses import dataclass, field
 from typing import List, Optional, Sequence
 
 import numpy as np
+from matplotlib import pyplot as plt
 
 from predicators import utils
 from predicators.approaches import ApproachFailure, ApproachTimeout
 from predicators.approaches.oracle_approach import OracleApproach
-from predicators.envs import get_or_create_env
+from predicators.datasets.demo_only import human_demonstrator_policy
+from predicators.envs import BaseEnv, create_new_env
 from predicators.ground_truth_models import _get_options_by_names, \
     _get_types_by_names, get_gt_options
 from predicators.settings import CFG, get_allowed_query_type_names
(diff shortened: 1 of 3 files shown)


tests/approaches/test_online_nsrt_learning_approach.py
```diff
@@ -29,7 +29,6 @@ def test_online_nsrt_learning_approach():
         "num_test_tasks": 3,
         "explorer": "random_options",
         "online_learning_max_novelty_count": float("inf"),
-        "online_learning_lifelong": True
```
tomsilver, line 32: i'm removing this because I think there's a separate bug: #1445

predicators/teacher.py
```diff
@@ -187,9 +216,21 @@ class TeacherInteractionMonitor(utils.LoggingMonitor):
                                                  default_factory=list)
     _query_cost: float = field(init=False, default=0.0)
 
+    @property
+    def _teacher_env(self) -> BaseEnv:
+        """Exposes the teacher's environment to this monitor."""
+        # This is a temporary hack. When issue #1443 is resolved, this can be
+        # removed. That issue will require a fairly significant amount of work.
+        return self._teacher._env  # pylint: disable=protected-access
```
NishanthJKumar, line 224: Why did you change the `env` field of the Teacher class to be private (i.e, `_env`)? I don't see a great reason for it, especially given that you're accessing it like this

NishanthJKumar (commented):
Looks mostly good! Just left a minor question about a choice you made in the `Teacher` class; I'd like to understand before granting approval.

predicators/teacher.py
tomsilver: It wasn't a field at all before (it was just a variable). I think we should leave it private because other classes shouldn't rely on it

NishanthJKumar: i see - makes sense!

PR #1446, opened 2023-03-30
add segmentation strategy that segments at every step

tomsilver:
I need this for painting in online bridge learning from human demos
- segmenting from options doesn't work because we don't have options in human demos
- atom changes doesn't work because nothing changes when the box lid is opened


NishanthJKumar (approved):
LGTM!

PR #1452, opened 2023-04-11
a few new flags for exit garage environment

tomsilver:
because I want to use the environment for some bridge policy stuff. these changes should have no effect on the existing code unless you pass in new flags.

there are two new flags:
1. if `--exit_garage_raise_environment_failure True`, then an `EnvironmentFailure` is raised when the car collides with an object.
2. if `--exit_garage_motion_planning_ignore_obstacles True`, then the motion planner does not take into account the obstacles at all.

together, the idea is that we'll make an optimistic plan that ignores the obstacles and then use a bridge policy to store the obstacles when an EnvironmentFailure is encountered.

### DIFF
--- a/predicators/envs/exit_garage.py
+++ b/predicators/envs/exit_garage.py
@@ -125,7 +125,7 @@ class ExitGarageEnv(BaseEnv):
             if collision and CFG.exit_garage_raise_environment_failure:
                 raise utils.EnvironmentFailure(
                     "Collision", info={"offending_objects": {collision_obj}})
-            # tf there is a collision or out-of-bounds, revert the move
+            # If there is a collision or out-of-bounds, revert the move
             if collision or self.coords_out_of_bounds(cx + dx, cy + dy):
                 next_state.set(self._car, "x", cx)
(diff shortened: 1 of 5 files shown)


predicators/envs/exit_garage.py
```diff
@@ -119,9 +119,14 @@ def simulate(self, state: State, action: Action) -> State:
...
-            # If this causes a collision, revert the movement
-            if self.car_has_collision(next_state) or \
-                    self.coords_out_of_bounds(cx + dx, cy + dy):
+            # If this causes a collision, optionally raise failure
+            collision_obj = self.get_car_collision_object(next_state)
+            collision = collision_obj is not None
+            if collision and CFG.exit_garage_raise_environment_failure:
+                raise utils.EnvironmentFailure(
+                    "Collision", info={"offending_objects": {collision_obj}})
+            # tf there is a collision or out-of-bounds, revert the move
```
lilianluong, line 128: typo in comment?

lilianluong (approved):
Spongebob Many Thumbs

PR #1455, opened 2023-04-23
github action updates and other hot fixes 🔥 

tomsilver:
1. split the `autoformat` github action into three separate actions (and add `docformatter` for the first time).
2. remove dependence on third-party yapf action which was breaking because it wasn't properly pinned to our version of yapf
3. pin our version of yapf and other libraries
4. upgrade gym to fix issue with setuptools (a while ago, we downgraded gym because some unit tests had broke https://github.com/Learning-and-Intelligent-Systems/predicators/pull/482 but now that version of gym is causing other installation issues, so we're going to upgrade to a newer version and then fix our unit tests) also add some type hints in gym sokoban


.github/workflows/predicators.yml
```diff
@@ -80,9 +80,50 @@ jobs:
...
+      uses: actions/setup-python@v2
       with:
-        args: --verbose --style .style.yapf --exclude '**/third_party'
+        python-version: ${{ matrix.python-version }}
+        cache: 'pip'
+        cache-dependency-path: '**/setup.py'
     - name: Run isort to organize imports
-      uses: isort/isort-action@master
\ No newline at end of file
+      uses: isort/isort-action@master
```
NishanthJKumar, line 108: i don't think this action is configured to use the particular version we want, so we might wanna change this to run isort manually and check the diff like we do with docformatter?

NishanthJKumar (approved):
Mostly LGTM aside from a small thing with `isort`!

PR #1456, opened 2023-04-23
QoL changes to prevent errors when using sokoban

NishanthJKumar:


predicators/settings.py
```diff
@@ -574,6 +574,10 @@ def get_arg_specific_settings(args: Dict[str, Any]) -> Dict[str, Any]:
         experiment-specific args."""
 
         return dict(
+            # The method used for perception: now only "trivial" or "oracle"
```
tomsilver, line 577: is "oracle" a thing?

tomsilver (approved):
one q otherwise lgtm thx

predicators/settings.py
NishanthJKumar: rip my b

PR #1465, opened 2023-06-06
remove broken lifelong learning flag

tomsilver:
the flag is broken because the environment is never reset during training (not even once at the beginning). it's also not possible to periodically evaluate the agent if we use just one long training interaction.


NishanthJKumar (approved):
Sounds good!

PR #1466, opened 2023-06-07
bilevel plan no sim: differentiate offline data collection from main

tomsilver:
in environments where we do have a simulator implemented, we may still want to use bilevel planning to collect demonstrations for convenience, even if we otherwise will assume that no simulator exists during training and evaluation.


NishanthJKumar (approved):
Smart - makes sense!

PR #1467, opened 2023-06-07
make options and operators 1:1 in bumpy cover env

tomsilver:
in the sampler learning work, it is convenient implementation-wise to assume a 1:1 relationship between options and operators. we don't need this assumption conceptually; it is just a convenience.


NishanthJKumar (approved):
LGTM!

PR #1468, opened 2023-06-07
fix classifier predict_proba in case where there is 1 class

tomsilver:
before the change, the unit test would crash with an assertion error


NishanthJKumar (approved):
LGTM!

PR #1469, opened 2023-06-09
flag to offset target allowed hand region in bumpy cover

tomsilver:
the purpose of this flag is to test whether we can learn samplers that are aware of the task distribution. in particular, the pick grasp sampler should favor picking on the right side of blocks if it is only possible to place on the right side of targets.


NishanthJKumar (approved):
Interesting idea for a varaition; LGTM!

PR #1474, opened 2023-06-13
active sampler learning with fitted Q iteration

tomsilver:
the following command gets around 60/100, which isn't great, but I expect would be improved with hyperparameter changes (more data etc.). we're planning to use this as a baseline in the future, probably not a main approach, since RL is data inefficient.

```
python predicators/main.py --approach active_sampler_learning         --env bumpy_cover         --seed 0         --strips_learner oracle         --sampler_learner oracle         --sampler_disable_classifier True         --bilevel_plan_without_sim True         --offline_data_bilevel_plan_without_sim False         --explorer random_nsrts         --max_initial_demos 1         --num_train_tasks 1000         --num_test_tasks 100         --max_num_steps_interaction_request 4         --bumpy_cover_num_bumps 2         --bumpy_cover_spaces_per_bump 1         --sampler_mlp_classifier_max_itr 1000000         --pytorch_train_print_every 10000 --online_nsrt_learning_requests_per_cycle 25 --active_sampler_learning_model fitted_q
```


NishanthJKumar (approved):
LGTM! :heavy_check_mark:

PR #1475, opened 2023-06-13
Gets `ActiveSamplerLearningApproach` working for right_targets

NishanthJKumar:
Previously, we were creating datasets for active sampler learning for bumpy cover such that a sample is deemed 'successful' if the corresponding ground NSRT's effects are achieved. This works as long as `bumpy_cover_right_targets` is False. We now amend the sampler dataset creation to handle this.

Command:
```
python predicators/main.py --approach active_sampler_learning         --env bumpy_cover         --seed 0         --strips_learner oracle         --sampler_learner oracle         --sampler_disable_classifier True         --bilevel_plan_without_sim True         --offline_data_bilevel_plan_without_sim False         --explorer random_nsrts         --max_initial_demos 1         --num_train_tasks 1000         --num_test_tasks 100         --max_num_steps_interaction_request 15    --bumpy_cover_num_bumps 2         --bumpy_cover_spaces_per_bump 1         --sampler_mlp_classifier_max_itr 1000000         --pytorch_train_print_every 10000 --online_nsrt_learning_requests_per_cycle 10 --bumpy_cover_right_targets True
```

Previously, this command would get ~35% success rate. With these changes, it gets up to about 65%.

bumpy_cover_num_online_transitions_perc_solved

### DIFF
--- a/predicators/approaches/active_sampler_learning_approach.py
+++ b/predicators/approaches/active_sampler_learning_approach.py
@@ -107,6 +107,7 @@ class ActiveSamplerLearningApproach(OnlineNSRTLearningApproach):
 
     def _update_sampler_data(self) -> None:
         new_trajs = self._segmented_trajs[self._last_seen_segment_traj_idx:]
+        just_made_incorrect_pick = False
         for segmented_traj in new_trajs:
             self._last_seen_segment_traj_idx += 1
             for segment in segmented_traj:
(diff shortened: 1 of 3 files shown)


predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -107,6 +107,7 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
 
     def _update_sampler_data(self) -> None:
         new_trajs = self._segmented_trajs[self._last_seen_segment_traj_idx:]
+        just_made_incorrect_pick = False
```
tsilver-bdai, line 110: should this be reset one level down inside the for loop?

predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -115,6 +116,25 @@ def _update_sampler_data(self) -> None:
                 ns = segment.states[-1]
                 success = self._check_option_success(o, segment)
 
+                if CFG.env == "bumpy_cover" and CFG.bumpy_cover_right_targets:
```
tsilver-bdai, line 119: can we put this logic behind a flag so we can easily compare to the old version too? for example, rather than `myopic_classifier`, something like `teacher_classifier`?

tsilver-bdai: also behind this flag let's assert that the env is `bumpy_cover`

tsilver-bdai (approved):
nice work!

predicators/approaches/active_sampler_learning_approach.py
NishanthJKumar: yes! good catch!

tomsilver (approved):

PR #1478, opened 2023-06-14
implement approach wrapping with noisy button example

tomsilver:
introduces a new environment with a button on a 1D axis that is initially imperceptible and that can change positions  and become again imperceptible if it's pressed off-center. the point of this environment is to test the viability of an approach wrapper that always executes a "find button" action if the button is imperceptible and then otherwise defers to a base approach (e.g., an oracle approach that doesn't have any NSRT for "find").

this command gets 50/50:

```
python predicators/main.py --env noisy_button --approach 'noisy_button_wrapper[oracle]' --seed 0 --bilevel_plan_without_sim True
```

whereas this command gets 0/50:

```
python predicators/main.py --env noisy_button --approach oracle --seed 0  --bilevel_plan_without_sim 
```

example video:

https://github.com/Learning-and-Intelligent-Systems/predicators/assets/2816502/44ef0584-f525-48ed-abc4-9d2fa50086d2


predicators/approaches/noisy_button_wrapper_approach.py
```diff
@@ -0,0 +1,53 @@
...
+
+        def _policy(state: State) -> Action:
+            nonlocal base_approach_policy
+            # If the button state is unknown, execute find.
+            button, = list(state)
+            if state.get(button, "position_known") < 0.5:
+                # Reset the base approach policy.
+                base_approach_policy = None
+                return Action(np.array([0.0, 1.0], dtype=np.float32))
+            # Check if we need to re-solve.
```
NishanthJKumar, line 45: modify comment here to clarify that we're not actually re-solving!

NishanthJKumar (approved):
LGTM! (Comment addressed by in-person discussion, so feel free to merge!)

PR #1481, opened 2023-06-16
regional bumpy cover updates

tomsilver:
several changes that are meant to make 'picking from bumpy region' the only interesting aspect of the environment. prior to these changes, placing was also very interesting because of possible collisions with other blocks and because it's difficult to learn the right sampler for placing if we start from an uniformed base sampler.

### DIFF
--- a/predicators/ground_truth_models/cover/nsrts.py
+++ b/predicators/ground_truth_models/cover/nsrts.py
@@ -488,6 +488,12 @@ class RegionalBumpyCoverGroundTruthNSRTFactory(GroundTruthNSRTFactory):
             w = state.get(b, "width") / 2
             lb = CFG.bumpy_cover_bumpy_region_start + w
             ub = 1.0 - w
+            # TODO: rejection sample here, checking for possible collisions
+            # with other blocks that might be in the bumpy region. To avoid
+            # pathological cases, give up after some large number of tries,
+            # rather than using while True. Verify implementation by manually
+            # checking that the `_any_intersection` check in `CoverEnv.simulate`
+            # is never hit (except in pathological cases perhaps).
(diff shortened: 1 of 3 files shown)


predicators/ground_truth_models/cover/nsrts.py
```diff
@@ -461,9 +483,17 @@ def place_on_target_sampler(state: State, goal: Set[GroundAtom],
...
                                    objs: Sequence[Object]) -> Array:
-            del state, objs, goal  # unused
-            lb = CFG.bumpy_cover_bumpy_region_start
-            ub = 1.0
+            del goal  # unused
+            b, = objs
+            w = state.get(b, "width") / 2
+            lb = CFG.bumpy_cover_bumpy_region_start + w
+            ub = 1.0 - w
+            # TODO: rejection sample here, checking for possible collisions
```
tomsilver, line 491: @NishanthJKumar if you wanna tackle this, that would be helpful!

nkumar-bdai: yup - i got it!

predicators/ground_truth_models/cover/nsrts.py
```diff
@@ -461,10 +484,34 @@ def place_on_target_sampler(state: State, goal: Set[GroundAtom],
...
+            w = state.get(b, "width") / 2
+            lb = CFG.bumpy_cover_bumpy_region_start + w
+            ub = 1.0 - w
+            other_blocks = [
+                block for block in list(state)
+                if block.type.name == 'block' and block != b
+            ]
+            curr_pose_sample = rng.uniform(lb, ub, size=(1, ))
+
+            # import ipdb; ipdb.set_trace()
```
tomsilver, line 499: remove plz :)

PR #1482, opened 2023-06-16
updates to explorers in preparation for active sampler explorer

tomsilver:
changes include:
* add a wrapper around explorer's termination functions to allow early termination of an interaction request. this is used by the active sampler learning approach to gradually increase the length of "practice" sessions. early on, before it has seen many tasks from the distribution, it practices for shorter; later, it practices for longer.
* fixes a major bug in the active sampler learning approach that was causing the first learning cycle's data to get thrown out (off by one error in `self._last_seen_segment_traj_idx`)
* move some helper functions into `utils.py`

### DIFF
--- a/predicators/approaches/active_sampler_learning_approach.py
+++ b/predicators/approaches/active_sampler_learning_approach.py
@@ -10,39 +10,18 @@ so we're punting for now.
 Example commands
 ----------------
 
-Bumpy cover easy:
-    python predicators/main.py --approach active_sampler_learning \
-        --env bumpy_cover \
-        --seed 0 \
-        --strips_learner oracle \
-        --sampler_learner oracle \
-        --bilevel_plan_without_sim True \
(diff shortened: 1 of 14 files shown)


predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -106,7 +102,8 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
         self._learn_wrapped_samplers(online_learning_cycle)
 
     def _update_sampler_data(self) -> None:
-        new_trajs = self._segmented_trajs[self._last_seen_segment_traj_idx:]
+        start_idx = self._last_seen_segment_traj_idx + 1
+        new_trajs = self._segmented_trajs[start_idx:]
```
NishanthJKumar, line 106: good catch!

NishanthJKumar (approved):
LGTM!

PR #1486, opened 2023-06-27
variety of small fixes / changes for active sampler learning on robot

tomsilver:


predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -148,9 +148,10 @@ def _update_sampler_data(self) -> None:
 
     def _check_option_success(self, option: _Option, segment: Segment) -> bool:
         ground_nsrt = utils.option_to_ground_nsrt(option, self._nsrts)
-        return ground_nsrt.add_effects.issubset(
-            segment.final_atoms) and not ground_nsrt.delete_effects.issubset(
-                segment.final_atoms)
+        # Only the add effects are checked to determine option success. The
+        # philosophy is the same as pre-image backchaining (with positive
+        # preconditions).
+        return ground_nsrt.add_effects.issubset(segment.final_atoms)
```
NishanthJKumar, line 154: This isn't quite right (for instance, we could satisfy this check and still delete some important necessary atoms). I think this is fine for now, but let's update this comment to mention that this isn't really kosher in general.
I think the right thing is to call the preimage backchaining procedure to compute necessary atoms.

predicators/explorers/active_sampler_explorer.py
```diff
@@ -154,18 +165,19 @@ def _update_ground_op_hist(self, state: State) -> None:
         if nsrt is None:
             return
         atoms = utils.abstract(state, self._predicates)
-        success = nsrt.add_effects.issubset(atoms) and \
-            not nsrt.delete_effects & atoms
-        logging.debug(f"[Explorer] Last NSRT: {nsrt.name}{nsrt.objects}")
-        logging.debug(f"[Explorer]   outcome: {success}")
+        success = nsrt.add_effects.issubset(atoms)
```
NishanthJKumar, line 168: This isn't quite right as a check either! We should really be using the necessary atoms here. Probably want to update with a comment.

PR #1492, opened 2023-07-05
Refinement cost learning changes

lilianluong:
**Changes (most of them, anyways):**
- misc. `ExitGarage` changes, including replacing `PickupObstacle` and `StoreObstacle` options with one `ClearObstacle`
- fixes and improvements for GPU usage in PyTorch
- implemented "refinement cost" as a weighted sum of refinement time and execution cost based on low-level action count
- time the refinement time per step when collecting data so it can be predicted separately
- config and launch scripts to run refinement cost learning related data collection, training, and eval on supercloud


predicators/utils.py
```diff
@@ -1906,8 +1914,6 @@ def query_to_goal_fn(
 
         If none is found, returns None.
         """
-        if self._collision_fn(start):
```
tomsilver, line 1909: why this change?

tomsilver (approved):
awesome work, and really appreciate the post-grad PR!!

PR #1494, opened 2023-07-17
active sampler explorer: consider multiple goals in case one is not reachable

tomsilver:
and eventually give up if none are reachable


NishanthJKumar (approved):
LGTM!

PR #1495, opened 2023-07-18
Restart Online Learning from a Specific Cycle

NishanthJKumar:
This PR adds functionality that allows us to restart an online learning approach from a specific cycle without attempting to load *every* single cycle (rather, we only load the one specified by `--skip_until_cycle` and then continue learning from there). This is useful for settings (e.g. running on a real robot) where we often need to restart experiments from a particular point.

### DIFF
--- a/tests/explorers/test_online_learning.py
+++ b/tests/explorers/test_online_learning.py
@@ -90,6 +90,7 @@ class _MockApproach(BaseApproach):
         self._dummy_saved.append(next_saved)
 
     def load(self, online_learning_cycle):
+        # import ipdb; ipdb.set_trace()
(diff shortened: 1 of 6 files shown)


tests/explorers/test_online_learning.py
```diff
@@ -90,6 +90,7 @@ def learn_from_interaction_results(self, results):
         self._dummy_saved.append(next_saved)
 
     def load(self, online_learning_cycle):
+        # import ipdb; ipdb.set_trace()
```
tomsilver, line 93: remove

tomsilver (approved):
lgtm!

PR #1496, opened 2023-07-20
fix edge case in operator strips learning with empty segments

tomsilver:


NishanthJKumar (approved):
LGTM! thanks for the quick fix on this!

PR #1497, opened 2023-07-20
enforce max option steps in active sampler explorer

tomsilver:


NishanthJKumar (approved):
Huh - surprised we didn't have this/catch this before...

PR #1500, opened 2023-07-21
fix bug in active sampler explorer

tomsilver:
this is a pretty bad bug that was causing a bunch of nonsensical sampler data to get recorded (in a certain case described below).

the bug was in `self._last_executed_nsrt` and the other fields not getting properly reset to `None` after each interaction request. I mistakenly believed that `_get_exploration_strategy()` was getting called after each of the N interaction requests per online learning cycle, but that's not true -- it's called N times _before_ the online learning cycle. the whole "request" setup that we have is quite confusing and regrettable, but a larger refactor is not worth it right now.

the reason we didn't encounter this bug earlier is that we were almost always doing `--online_nsrt_learning_requests_per_cycle 1`, in which case the bug doesn't manifest.

### DIFF
--- a/predicators/explorers/active_sampler_explorer.py
+++ b/predicators/explorers/active_sampler_explorer.py
@@ -1,13 +1,8 @@
 """An explorer for active sampler learning."""
 
-import glob
 import logging
-import os
-import re
-import time
 from typing import Callable, Dict, Iterator, List, Optional, Set
 
-import dill as pkl
 import numpy as np
 from gym.spaces import Box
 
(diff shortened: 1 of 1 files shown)


predicators/explorers/active_sampler_explorer.py
```diff
@@ -162,15 +170,22 @@ def generate_goals() -> Iterator[Set[GroundAtom]]:
 
         # Wrap the option policy to keep track of the executed NSRTs and if
         # they succeeded, to update the ground_op_hist.
-        self._last_executed_nsrt = None
-
+        initialized = False
```
tomsilver, line 173: this is the only real change

NishanthJKumar (approved):

PR #1501, opened 2023-07-21
fix two bugs related to segmentation in active sampler learning

tomsilver:
first, the change I made in https://github.com/Learning-and-Intelligent-Systems/predicators/pull/1499 was just completely wrong. it was looking at the time step in the entire trajectory to determine whether the current option has exceeded the max horizon.

second, the ground op history (for sampler learning) inside the active sampler explorer was not recording data in the case where the episode is terminated due to the max option horizon.

before these changes, there were discrepancies between the number of data per operator reported by the active sampler explorer versus the number reported within the approach for sampler learning. the numbers are consistent after these fixes.

### DIFF
--- a/predicators/main.py
+++ b/predicators/main.py
@@ -505,7 +505,7 @@ def _run_episode(
                 observations.append(obs)
             except Exception as e:
                 if exceptions_to_break_on is not None and \
-                   type(e) in exceptions_to_break_on:
+                   any(issubclass(type(e), c) for c in exceptions_to_break_on):
                     if monitor_observed:
                         exception_raised_in_step = True
                     break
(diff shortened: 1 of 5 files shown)


predicators/main.py
```diff
@@ -505,7 +505,7 @@ def _run_episode(
                 observations.append(obs)
             except Exception as e:
                 if exceptions_to_break_on is not None and \
-                   type(e) in exceptions_to_break_on:
+                   any(issubclass(type(e), c) for c in exceptions_to_break_on):
```
NishanthJKumar, line 508: great catch!

NishanthJKumar (commented):
LGTM!

NishanthJKumar (approved):

PR #1510, opened 2023-07-24
add sliding cabinet to kitchen

tomsilver:
to merge after #1508

the current implementation of this new skill is not very stable. the command below gets 6/10. I tried to improve it for a while but I think I'm hitting diminishing returns. we may not bother including it among the goals later. I think it's more important that we implement "reverse" skills next so that we can get some reset-free exploration going.

```
python predicators/main.py --env kitchen --approach oracle --seed 0 --num_train_tasks 0 --num_test_tasks 10 --bilevel_plan_without_sim True --use_gui --kitchen_goals full --kitchen_use_perfect_samplers True
```


tomsilver:
closing until the back end stuff progresses

PR #1511, opened 2023-07-25
change kitchen backend

tomsilver:
this replaces the gym environment ("back end") for the kitchen environment to a more recent and stable version. the main advantage of the new version is that stable IK control of the end effector is possible, which massively reduces pain in designing options.

it's also possible that we could make the kitchen environment a regular dependency, but I'm punting on that because it requires a higher python version (3.9+) than our current minimum (3.8).

this PR recovers the performance of the "perfect" samplers. the next PRs will (1) recover performance of active learning; (2) recover supercloud-ability; (3) add more skills.

### DIFF
--- a/predicators/ground_truth_models/kitchen/options.py
+++ b/predicators/ground_truth_models/kitchen/options.py
@@ -39,7 +39,6 @@ class KitchenGroundTruthOptionFactory(GroundTruthOptionFactory):
         object_type = types["obj"]
 
         # Predicates
-        TurnedOn = predicates["TurnedOn"]
         OnTop = predicates["OnTop"]
 
         options: Set[ParameterizedOption] = set()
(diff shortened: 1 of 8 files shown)


predicators/ground_truth_models/kitchen/options.py
```diff
@@ -27,20 +39,23 @@ def get_env_names(cls) -> Set[str]:
     def get_options(cls, env_name: str, types: Dict[str, Type],
                     predicates: Dict[str, Predicate],
                     action_space: Box) -> Set[ParameterizedOption]:
-        assert _MJKITCHEN_IMPORTED
+
+        assert _MJKITCHEN_IMPORTED, "See kitchen.py"
+
+        # Need to define these here because uses may not have euler2quat.
```
NishanthJKumar, line 45: you mean 'users'?

NishanthJKumar (approved):
LGTM!

PR #1514, opened 2023-07-27
don't abstract intermediate states if not needed; only save ground atom trajectories when asked

tomsilver:
users who want to `--load_atoms` later will now need to `--save_atoms`. default is to not save atoms because the ground atom trajectories get quite large for environments that have long-horizon options (e.g. kitchen), leading to both memory and time issues.


NishanthJKumar (approved):
Nice! This is super useful to have!

PR #1515, opened 2023-07-27
kitchen: split out object types, add "turn off light" skill

tomsilver:


NishanthJKumar (approved):
lgtm! excited for the light skill that's soon to come!

PR #1520, opened 2023-08-01
plan with competence costs in active sampler learning approach

tomsilver:
to be merged after #1519

### DIFF
--- a/predicators/planning.py
+++ b/predicators/planning.py
@@ -1052,7 +1052,6 @@ def fd_plan_from_sas_file(
     """Given a SAS file, runs search on it to generate a plan."""
     cmd_str = (f"{timeout_cmd} {timeout} {exec_str} {alias_flag} {sas_file}")
     output = subprocess.getoutput(cmd_str)
-    print(output)
     cleanup_cmd_str = f"{exec_str} --cleanup"
     subprocess.getoutput(cleanup_cmd_str)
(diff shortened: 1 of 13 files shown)


predicators/planning.py
```diff
@@ -1156,9 +1213,17 @@ def run_task_plan_once(
...
             "Please follow instructions in the docstring of the" +\
             "_sesame_plan_with_fast_downward method in planning.py"
-        if CFG.sesame_task_planner == "fdopt":
+
+        sesame_task_planner = CFG.sesame_task_planner
+        if sesame_task_planner.endswith("-costs"):
+            use_costs = True
+            sesame_task_planner = sesame_task_planner[:-len("-costs")]
+        else:
+            use_costs = False
```
NishanthJKumar, line 1222: seems to me like currently cost-aware planning only works with FD. Shouldn't we assert that the `CFG.sesame_task_planner` is set to FD then to prevent any issues?

NishanthJKumar (approved):
looks mostly good! just left one minor comment

predicators/planning.py
tomsilver: we're already within `elif "fd" in CFG.sesame_task_planner:` here. if someone tries to do something like `--sesame-task-planner astar-costs`, it will just the usual NotImplementedError at the end of the outer conditional statement

PR #1523, opened 2023-08-01
Continuous Pushing and Pulling Test

wmcclinton:
Tweaks some tolerances and parameters to make Pushing and Pulling options work one after the other.


predicators/ground_truth_models/kitchen/options.py
```diff
@@ -238,13 +265,9 @@ def _PullKettle_terminal(state: State, memory: Dict,
                                  params: Array) -> bool:
             del memory, params  # unused
             _, obj, obj2 = objects
-            obj_y = state.get(obj, "y")
-            obj2_y = state.get(obj2, "y")
             if not GroundAtom(OnTop, [obj, obj2]).holds(state):
```
tomsilver, line 268: if the latter is no longer needed, let's change this to just `return GroundAtom(...`

PR #1526, opened 2023-08-03
Active exploration epsilon greedy

NishanthJKumar:


predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -548,6 +552,38 @@ def _sample(state: State, goal: Set[GroundAtom], rng: np.random.Generator,
...
 
 
+def _wrap_sampler_exploration(base_sampler: NSRTSampler,
+                              score_fn: _ScoreFn) -> NSRTSampler:
+    """Create a wrapped sampler that uses a score function to select among
+    candidates from a base sampler at exploration time."""
+
+    def _sample(state: State, goal: Set[GroundAtom], rng: np.random.Generator,
+                objects: Sequence[Object]) -> Array:
+        samples = [
```
tomsilver, line 562: let's remove the code that's redundant here with `_wrap_sampler` by just making one function that takes in a `sample_strategy` argument. then where you currently call `_wrap_sampler` you can call it with `sample_strategy=greedy`, and where you currently call `_wrap_sampler_exploration` you can call it with `sample_strategy=CFG.active_sampler_learning_num_samples`

tomsilver (approved):
one request, otherwise lgtm!

PR #1536, opened 2023-08-15
disable collision checking in regional bumpy cover

tomsilver:


tomsilver:
nvm, this leads to an assertion failure...

PR #1539, opened 2023-08-21
minor active sampler learning updates

tomsilver:
summary of changes:

* change the competence prior to be very optimistic (about 90% expectation). the failure case that led to this change was in the forthcoming grid row environment. with the old competence prior, the agent believed that the 'correct' plan of moving all the way to the light was high-cost (low probability) when it has limited data of moving in the beginning. the agent would therefore prefer to take the one-step shortcut for a very large number of steps until the probability of that individual step finally dipped below the combined probabilities of all of the move steps.
* use a window of 1 to look at possible skill competence changes during extrapolation. with a window of N, the agent wouldn't detect any change until N cycles were completed, and we don't want it to just be random for all that time.
* add a maximum horizon that we work on an assigned task during exploration before giving up. this was helpful for creating a wedge between us and baselines, because otherwise the baselines would repeatedly practice the hard skill many times.
* as an optimization for 'planning progress', return the worst possible score for skills that have perfect success rate. without this, the score computation is really painfully slow in grid row, where there are >250 ground skills to score, only 3 of which are actually non-perfect.
* increased epsilon for epsilon greedy parameter selection; otherwise was getting too stuck too quickly
* fix overly strict checks in cost-aware planning
* fix supercloud directory clearing (was crashing with directories that contain too many files)


NishanthJKumar (approved):
LGTM!

PR #1540, opened 2023-08-21
grid row environment

tomsilver:
simple 1D grid with a lightbulb that requires learning a sampler


NishanthJKumar (approved):
great work, and glad to have this new env implemented!

PR #1541, opened 2023-09-05
slippery table tricky floor environment

tomsilver:


predicators/ground_truth_models/sticky_table/options.py
```diff
@@ -30,19 +30,27 @@ def get_options(cls, env_name: str, types: Dict[str, Type],
...
+            params_space=Box(
+                np.array([StickyTableEnv.x_lb, StickyTableEnv.y_lb]),
+                np.array([StickyTableEnv.x_ub, StickyTableEnv.y_ub])),
             types=[cube_type])
 
         PlaceOnTable = utils.SingletonParameterizedOption(
             # variables: [cube, table]
             "PlaceOnTable",
-            cls._create_place_policy(action_space),
+            cls._create_pass_through_policy(action_space),
```
NishanthJKumar, line 53: wait - why did you rename `_create_place_policy` to `_create_pass_through_policy`? There appears to be no functional change?

predicators/envs/sticky_table.py
```diff
@@ -281,3 +304,35 @@ def _sample_floor_point_around_table(
...
+
+
+class StickyTableTrickyFloorEnv(StickyTableEnv):
+    """Variation where picking from the floor is the only thing that can be
+    improved through sampler learning.
+
+    Placing on the table is still noisy, but inherently so.
+    """
+
+    sticky_surface_mode = "whole"  # the 'sticky' table is sticky everywhere
```
NishanthJKumar, line 316: wait - this doesn't make sense? shouldn't the whole table be smooth?

NishanthJKumar: oh nvm - i get it

predicators/ground_truth_models/sticky_table/options.py
tomsilver: because it's being used now in the pick option also, not just the place option. you're right that it hasn't functionally changed, it was always just a "pass through" policy

PR #1546, opened 2023-09-08
maple Q approach

tomsilver:


predicators/execution_monitoring/expected_atoms_monitor.py
```diff
@@ -20,7 +19,6 @@ def get_name(cls) -> str:
     def step(self, state: State) -> bool:
         # This monitor only makes sense to use with an oracle
         # bilevel planning approach.
-        assert "oracle" in CFG.approach or "active_sampler" in CFG.approach
```
NishanthJKumar, line 23: are we actually fine taking out this assertion (I just took it out to get the experiments to run lol)? If so, we should remove the two lines of comments above it. If not, we should just add "maple_q" to the assertion

predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -470,7 +470,7 @@ def _learn_nsrt_sampler(self, nsrt_data: _OptionSamplerDataset,
                             nsrt: NSRT) -> Tuple[NSRTSampler, NSRTSampler]:
         # Build targets.
         gamma = CFG.active_sampler_learning_score_gamma
-        num_a_samp = CFG.active_sampler_learning_num_next_option_samples
+        num_a_samp = CFG.q_function_num_lookahead_samples
```
NishanthJKumar, line 473: hmmm - nit, but i don't like that this has 'q_function' in the name when other active sampler learning methods are going to be using this flag... Can we just keep the flag name as is?

predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -567,7 +567,7 @@ def _fit_regressor(self, nsrt_data: _OptionSamplerDataset) -> MLPRegressor:
             weight_decay=CFG.weight_decay,
             use_torch_gpu=CFG.use_torch_gpu,
             train_print_every=CFG.pytorch_train_print_every,
-            n_iter_no_change=CFG.active_sampler_learning_n_iter_no_change)
+            n_iter_no_change=CFG.q_function_n_iter_no_change)
```
NishanthJKumar, line 570: ditto

NishanthJKumar (approved):
LGTM aside from a few minor things!

NishanthJKumar (approved):

PR #1556, opened 2023-09-22
Clustering via reverse engineering

ashayathalye:
This PR implements a predicate invention approach that employs clustering to try to reverse engineer the clusters of segments that correspond to the oracle operators and select predicates that are add effects in those clusters, letting pnad_search downstream handle chainability and removing some extraneous predicates through backchaining that might cause overfitting issues.

The performance on the four tasks from the AAAI paper is illustrated below: 
The non-oracle clustering method:
```
                      NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED   LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID
blocks_main_200demo     200.00 (0.00)  48.70 (1.27)  14.20 (0.60)   0.66 (0.19)  3074.73 (1223.38)  149.81 (38.48)         10
cover_main_200demo      200.00 (0.00)  49.20 (0.75)   9.80 (3.25)   0.05 (0.02)        5.92 (0.30)    39.82 (2.82)         10
painting_main_200demo   200.00 (0.00)  49.70 (0.64)  41.90 (1.64)   1.06 (0.20)  4354.69 (1189.59)   253.45 (3.07)         10
tools_main_200demo      200.00 (0.00)  43.30 (1.68)  44.10 (0.30)   0.74 (0.16)   1357.64 (409.98)   227.94 (5.82)         10
```
v.s.
The AAAI method: 
```
                                              NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED        LEARNING_TIME  NUM_SEEDS
EXPERIMENT_ID                                                                                                                                             
pybullet_blocks_main_200demo                    200.00 (0.00)  49.00 (0.77)   7.60 (1.28)   0.54 (0.17)  3997.55 (2237.44)   13717.18 (2739.89)         10
cover_main_200demo                              200.00 (0.00)  49.30 (0.64)   5.90 (0.54)   0.01 (0.00)        4.94 (0.23)       549.25 (69.44)         10
painting_main_200demo                           200.00 (0.00)  49.90 (0.30)  20.80 (1.78)   0.18 (0.04)    402.88 (116.42)  89404.23 (13707.67)         10
tools_main_200demo                              200.00 (0.00)  45.90 (5.79)  28.00 (5.14)   0.44 (0.30)  1926.29 (1585.43)   19286.83 (3002.72)         10
```

The main takeaways from these results is that (1) it's much faster, (2) but it still learns a lot of extra predicates. Overall, this shows that the clustering idea, in how it leverages more structure in the demonstrations to be more efficient at predicate invention, has promise. 

The high-level outline of the algorithm is as follows:
1. cluster segments by option
2. further cluster by types of objects affected 
3. further cluster by # of objects affected 
4. If a parameter exists for the option involved in this cluster's segments...
If the parameters are described by a uniform distribution (via a KS test for now), don't cluster further. 
Else, further cluster by assignment from a GMM fit to the parameters. 
5. Get predicates from pure intersection of add effects within each cluster.
6. Remove inconsistent predicates except if removing them prevents us from disambiguating two or more clusters (i.e. their add effect sets are the same after removing inconsistent predicates). If it does, add those predicates back. 
7. Learn operators via pnad search

Some intuition and thoughts about the above:
- Overall, the procedure can be seen as figuring out clusters of segments for which we want a dedicated operator, and subselecting predicates from the initial candidate predicates based on those clusters. And then we let pnad search handle the chainability. 
- The idea behind step 6 is that HoldingTop and HoldingSide are inconsistent within the PlaceOnTable cluster in painting, but we don't want to remove them, since we had generated them specifically to disambiguate segments involving the Pick option into two clusters. 
- Step 4 (not clustering further if the parameters are described by a uniform) helps prevent overfitting. If the parameter data is uniform, a GMM might, e.g. still create 3 clusters. And then we might get some very specific predicates out of that, that end up being inconsistent. And then when we apply step 5, these clusters were quite similar (they should be one cluster!) so we won't disambiguate them, and we'll add the inconsistent predicates back (which we don't want, since they're the ones we get from overfitting). 
- This algorithm is a somewhat hacky, but also isn't that hacky (as evidenced by it working in RNT, etc). The next steps for this direction might be to create environments to break it, to understand how we can make it more general. But mainly it's been a great way through which to get a good understanding of what's going on with the clustering idea in general.

### DIFF
--- a/predicators/approaches/grammar_search_invention_approach.py
+++ b/predicators/approaches/grammar_search_invention_approach.py
@@ -974,14 +974,11 @@ class GrammarSearchInventionApproach(NSRTLearningApproach):
                     types_in_effects = [
                         set(a.predicate.types) for a in seg.add_effects
                     ]
-                    if len(types_in_effects) == 0 or len(
-                            set.union(*types_in_effects)) == 0:
-                        # If there are no add effects or if the object
-                        # arguments for all add effects are empty (which would
-                        # happen if the add effects only involved Forall
-                        # predicates that have no object arguments, e.g. in
-                        # repeated_nextto), then don't cluster further.
-                        continue
+                    # To cluster on type, there must be types. That is, there
+                    # must be add effects in the segment and the object
+                    # arguments for at least one add effect must be nonempty.
+                    assert len(types_in_effects) > 0 and len(
+                        set.union(*types_in_effects)) > 0
                     types = tuple(sorted(list(set.union(*types_in_effects))))
                     types_to_segments.setdefault(types, []).append(seg)
                 logging.info(
--- a/predicators/approaches/grammar_search_invention_approach.py
+++ b/predicators/approaches/grammar_search_invention_approach.py
@@ -13,6 +13,9 @@ from typing import Callable, Dict, FrozenSet, Iterator, List, Sequence, Set, \
     Tuple
 
 from gym.spaces import Box
+import numpy as np
+from sklearn.mixture import GaussianMixture as GMM
+from scipy.stats import kstest
 
 from predicators import utils
 from predicators.approaches.nsrt_learning_approach import NSRTLearningApproach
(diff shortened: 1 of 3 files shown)


ashayathalye:
As of right now, the coverage is:
`predicators/approaches/grammar_search_invention_approach.py     579      1    99%   984`. 

This one line of code was motivated by something that comes up in RNT, but if I test on RNT it's producing this error:
`../../predicators/venv/lib/python3.9/site-packages/pyperplan/heuristics/lm_cut.py:188: AssertionError`, which I have to investigate further.

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -887,13 +890,290 @@ def _get_successors(
...
+
+            # Step 6:
+            # Remove inconsistent predicates except if removing them prevents us
+            # from disambiguating two or more clusters (i.e. their add effect
+            # sets are the same after removing the inconsistent predicates). The
+            # idea here is that HoldingTop and HoldingSide are inconsistent
+            # within the PlaceOnTable cluster in painting, but we don't want to
+            # remove them, since we had generated them specifically to
+            # disambiguate segments in the cluster with the Pick option.
+            # A consistent predicate is is either an add effect, a delete
```
NishanthJKumar, line 1138: Nit: delete one of the 'is' here

NishanthJKumar (approved):
Incredible work @ashayathalye ! This is a well-written and well-documented PR that was super easy to review despite the conceptual changes being somewhat complex. Once the tests pass (messaged about this on Slack), and a minor grammatical issue is addressed, I think we should be good to go!

ashayathalye:
Note: the failed unit test is `tests/approaches/test_oracle_approach.py` - not sure what's going on with it.

NishanthJKumar:
Fixed! Just needed to rerun the tests :smile: . Phenomenal work!

PR #1563, opened 2023-10-04
freeze pytorch version to fix mypy crash

tomsilver:


NishanthJKumar (approved):
I wonder if the pytorch team did this simply to spite us ;)

PR #1565, opened 2023-10-10
Implement infinite-horizon for exploration

NishanthJKumar:
During active sampler learning, instead of measuring success rate under a particular timeout, we might want to measure success rate under a particular horizon. We can do this by setting the `--horizon` argument to whatever particular number we want when running our experiments. However, this causes an issue with approaches that call the planner during exploration: in particular, these will sometimes come up with plans they want to practice that might be longer than the given horizon. If this happens, an error gets thrown that causes a crash...

This PR sets the horizon to `np.inf` during exploration, thereby disabling the horizon check during exploration, but keeping it around for use at evaluation time.


tomsilver (approved):
nice catch!

PR #1566, opened 2023-10-10
Approach for the Delivery PDDL environment

barci2:
The approach is a simple domain-specific strategy. It assumes the robot is at one location at the same time and that all the drop-off and home base locations are safe.

### DIFF
--- a/predicators/approaches/delivery_approach.py
+++ b/predicators/approaches/delivery_approach.py
@@ -1,7 +1,7 @@
 """An approach that implements a delivery-specific policy.
 
 Example command line:
-    python src/main.py --approach delivery_policy --seed 0 \
+    python predicators/main.py --approach delivery_policy --seed 0 \
         --env pddl_easy_delivery_procedural_tasks
 """
 
--- a/predicators/approaches/delivery_approach.py
+++ b/predicators/approaches/delivery_approach.py
@@ -28,6 +28,7 @@ class DeliverySpecificApproach(BaseApproach):
     def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
 
         def _policy(state: State) -> Action:
+            # Extract the predicators and options from the state.
             options = {o.name: o for o in self._initial_options}
             predicates = {p.name: p for p in self._initial_predicates}
             types = {t.name: t for t in self._types}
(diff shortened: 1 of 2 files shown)


predicators/approaches/delivery_approach.py
```diff
@@ -0,0 +1,197 @@
...
+
+import numpy as np
+
+from predicators.approaches import ApproachFailure, BaseApproach
+from predicators.envs.pddl_env import _PDDLEnvState
+from predicators.structs import Action, GroundAtom, Object, Predicate, State, \
+    Task, _Option
+
+
+def filter_predicated_objs(objs: Iterable[Object], pred: Predicate,
```
tomsilver, line 18: nice helper function!

predicators/approaches/delivery_approach.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "delivery_policy"
+
+    @property
+    def is_learning_based(self) -> bool:
+        return False
+
+    def _extract_at_loc_optional(
```
tomsilver, line 70: I think this logic could be included inside `_extract_at_loc`, but that's just my personal preference :)

predicators/approaches/delivery_approach.py
```diff
@@ -0,0 +1,197 @@
...
+    def _deliver(self, state: _PDDLEnvState, paper: Object,
+                 loc: Object) -> Action:
+        ground_option = self.opt_deliver.ground([paper, loc],
+                                                np.empty(0, dtype=np.float32))
+        check_option(ground_option, state)
+        return ground_option.policy(state)
+
+    def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+        # Extracting task-specific information
+        init_state = cast(_PDDLEnvState, task.init)
```
tomsilver, line 109: slightly better than using `cast` here is `assert isinstance(_PDDLEnvState, task.init)` which should satisfy mypy and has the added benefit that the code will actually crash (at runtime) if the assumption is violated

predicators/approaches/delivery_approach.py
```diff
@@ -0,0 +1,197 @@
...
+        goal_unpacked = filter_predicated_objs(obj_papers, self.pred_unpacked,
+                                               task.goal)
+        goal_carrying = filter_predicated_objs(obj_papers, self.pred_carrying,
+                                               task.goal)
+
+        # Joint object extraction
+        obj_available_papers = obj_papers - goal_unpacked
+        obj_satisfying_papers = obj_available_papers - goal_carrying
+
+        # Sanity checks (satisfies all the "safe" and "home_base" goals)
```
tomsilver, line 138: love all the careful checks!

predicators/approaches/delivery_approach.py
```diff
@@ -0,0 +1,197 @@
...
+                    iter(obj_satisfying_papers - obj_carried_papers), None)
+                if obj_satisfying_paper is None:
+                    raise ApproachFailure(
+                        "Too little paper that can satisfy locations")
+                return self._pick_up(state, obj_satisfying_paper,
+                                     obj_home_base)
+
+            # If already at a location to satisfy, satisfy it
+            if obj_at_loc in obj_locs_to_satisfy:
+                carried_paper = obj_carried_papers.pop()
```
tomsilver, line 173: another very minor thing: we generally try to make sure that our code is fully deterministic so that we can more easily reproduce and debug things. popping from a set in Python is not necessarily deterministic, so we would prefer to do this another way (e.g., maintaining a list instead of a set)

predicators/approaches/delivery_approach.py
```diff
@@ -0,0 +1,197 @@
...
+            # Finally go to the desired goal location (satisfies the "at_loc" goal)
+            if goal_at_loc and goal_at_loc != obj_at_loc:
+                return self._move(state, obj_at_loc, goal_at_loc)
+
+            ApproachFailure("Ran out of things to do")
+
+        return _policy
+
+
+__all__ = [DeliverySpecificApproach]
```
tomsilver, line 197: a good practice that we don't usually follow in this repo!

tomsilver (commented):
great work! very thorough. I left a few small suggestions, no need to actually implement them if you're ready to move on to other things. we won't actually merge this PR because we like to re-use this onboarding assignment.

predicators/approaches/delivery_approach.py
barci2: I separated it mostly for the sake of typing

tomsilver: if you assert something isn't None, mypy will be smart enough to incorporate that

barci2: yeah but I wanted a nice error message printed

Issue #1567, opened 2023-10-11
consider upgrading mypy


tomsilver:
the original issue causing us to peg to an earlier version of mypy was: https://github.com/python/mypy/issues/5485 it seems to be resolved now, so we could potentially update the version

PR #1568, opened 2023-10-11
MyPy version bump

barci2:
Version bump raised in issue #1567


tomsilver (approved):
wow awesome, so glad this was small, thanks a lot for doing it!

PR #1578, opened 2023-10-28
allow third party users to define their own oracle NSRTs

tomsilver:
minor change, most of the lines are just tests


NishanthJKumar (approved):
LGTM! Thanks for such a quick turnaround!

PR #1579, opened 2023-10-29
Split fail focus into UCB and non-UCB baselines

NishanthJKumar:
Something we discussed a while ago. To be merged after #1577 

Graph (Note: I accidentally ran the non-ucb version for longer than all the others :/)

grid_1d_environment_num_online_transitions_perc_solved


tomsilver (approved):
lgtm

PR #1582, opened 2023-11-20
add info to FD crashes

tomsilver:


NishanthJKumar (approved):
Nice! I've been meaning to do this myself for a while; thanks for getting around to it!

PR #1583, opened 2023-11-21
remove dead email and add NJK email in README

tomsilver:
with Rohan's blessing


NishanthJKumar (approved):
Awwww - I'm honored! ❤️

PR #1584, opened 2023-11-21
handle planning failures within task planning in active sampler explorer

tomsilver:


NishanthJKumar (approved):
Makes sense to me - but I'm assuming we only expect this to happen during prototyping/dev right? Seems like if this happens for a finalized env, it means the planning graph isn't reversible/we didn't fully design the domain and the goal proposers correctly?

tomsilver:
> Makes sense to me - but I'm assuming we only expect this to happen during prototyping/dev right? Seems like if this happens for a finalized env, it means the planning graph isn't reversible/we didn't fully design the domain and the goal proposers correctly?

Yeah that's right

PR #1585, opened 2023-11-21
add separate flag for approach wrapper

tomsilver:
so that we can include approach wrappers as flags in YAML configs


NishanthJKumar (approved):
Nice!

PR #1597, opened 2023-12-09
fix rotation bug in rectangle random sampling

tomsilver:
before:
rectangle_unit_test_before

after:
rectangle_unit_test_after


NishanthJKumar (approved):
woops my bad! Good catch!

PR #1599, opened 2023-12-10
Fixes competence estimation for when we do epsilon-greedy exploration sampling

NishanthJKumar:
Previously, we always updated the competence estimator with new data. Now, we only do so if we actually sampled from the learned policy; we don't let exploration samples influence our competence estimate.

### DIFF
--- a/predicators/approaches/active_sampler_learning_approach.py
+++ b/predicators/approaches/active_sampler_learning_approach.py
@@ -315,6 +315,9 @@ class ActiveSamplerLearningApproach(OnlineNSRTLearningApproach):
         for old_nsrt in self._nsrts:
             if old_nsrt.option not in new_nsrt_options:
                 new_test_nsrts.add(old_nsrt)
+                # Since we don't have a learned score function, just make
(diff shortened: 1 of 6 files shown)


predicators/approaches/active_sampler_learning_approach.py
```diff
@@ -313,7 +315,11 @@ def _learn_wrapped_samplers(self,
         for old_nsrt in self._nsrts:
             if old_nsrt.option not in new_nsrt_options:
                 new_test_nsrts.add(old_nsrt)
-                self._nsrt_to_explorer_sampler[old_nsrt] = old_nsrt._sampler  # pylint: disable=protected-access
+                self._nsrt_to_explorer_sampler[
+                    old_nsrt] = _wrap_sampler_exploration(
+                        old_nsrt._sampler,  # pylint: disable=protected-access
+                        lambda o, _, params: [1.0] * len(params),
```
tomsilver, line 321: I think I understand what's happening here now but I was initially confused -- can you add a comment?

tomsilver (approved):
elegant, nice work!

PR #1603, opened 2023-12-15
object-specific active sampler learning

tomsilver:
The results are worse, so we're disabling this by default.


NishanthJKumar (approved):
elegant!

PR #1606, opened 2023-12-29
Add competence gradient and skill diversity baselines

NishanthJKumar:
Adds two new important baselines to finish up all approaches we want to compare our approach against for our final experiments.

Also, normalizes the UCB bonus by dividing by the number of test tasks.

image

image

PR #1608, opened 2023-12-30
disable ucb for planning progress

tomsilver:
image
image


NishanthJKumar:
useful to have these results towards getting #1606 to work, but we never really intended to merge this directly

PR #1614, opened 2024-02-19
Simple variant of the RNT env

NishanthJKumar:
This PR introduces a variant of RNT where the objects are spaced such that it's generally not possible to be nextTo more than one at a time. This serves as a useful (simple) test environment for seeing if we can efficiently invent predicates.

Result of running OG predicate invention approach with `pnad_search` strips learner:
```
AGGREGATED DATA OVER SEEDS:
                                                NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME AVG_NODES_CREATED       LEARNING_TIME AVG_REF_COST  NUM_SEEDS
EXPERIMENT_ID                             CYCLE                                                                                                                        
rnt_simple-invent_allexclude_backchaining None     50.00 (0.00)  50.00 (0.00)   6.10 (0.30)   0.01 (0.01)       6.12 (0.71)  13711.14 (3965.85)  0.16 (0.00)         10
```


ashayathalye:
From discussion with Nishanth:

@ `dot_x = d_num * self._nextto_thresh * 2.1`, depending on what's in the grammar, the corresponding "nextto" predicates in the grammar may still turn on/off in a way that requires us to have them as ignore effects -- so this environment doesn't let us not have to deal with ignore effects during predicate invention. It just has the ground truth predicates not lead to ignore effects in learned operators. But, it's important to note that this happens in any environment when doing predicate invention with a grammar -- we always have random things turning on/off. 

The use of this environment then might be that you can run `cluster_and_intersect` and still get a proper theory -- so it can be useful as a debugging tool.

ashayathalye (approved):
From discussion with Nishanth, Qs about this PR:

@ `dot_x = d_num * self._nextto_thresh * 2.1`, depending on what's in the grammar, the corresponding "nextto" predicates in the grammar may still turn on/off in a way that requires us to have them as ignore effects -- so this environment doesn't let us not have to deal with ignore effects during predicate invention. It just has the ground truth predicates not lead to ignore effects in learned operators. But, it's important to note that this happens in any environment when doing predicate invention with a grammar -- we always have random things turning on/off that we have to figure out how to include or not include in learned operators. 

The use of this environment then might be that you can run `cluster_and_intersect` and still get a proper theory -- so it can be useful as a debugging tool.

predicators/envs/repeated_nextto.py
```diff
@@ -236,3 +236,56 @@ def _get_tasks_ambiguous(self, num: int, rng: np.random.Generator,
...
+        goal2 = {
+            GroundAtom(self._Grasped, [self._robot, dots[0]]),
+            GroundAtom(self._Grasped, [self._robot, dots[1]]),
+        }
+        goal3 = {
+            GroundAtom(self._Grasped, [self._robot, dots[0]]),
+            GroundAtom(self._Grasped, [self._robot, dots[1]]),
+            GroundAtom(self._Grasped, [self._robot, dots[2]]),
+        }
+        goals = [goal3, goal2, goal1]
```
tomsilver, line 278: why this order?

tomsilver (approved):
approving to unblock, but it looks like there's a significant amount of repeated code w.r.t. `RepeatedNextToAmbiguousEnv`, so consider refactoring :)

predicators/envs/repeated_nextto.py
NishanthJKumar: arbitrary!

PR #1616, opened 2024-03-01
Enable saving and loading of predicate grammar atom datasets

NishanthJKumar:
As @ashayathalye and I are starting to do predicate invention with increasingly larger grammars, the time taken to just apply the grammar predicates to a dataset is getting significant and slowing down iteration time. This PR makes the `--load_atoms` and `--save_atoms` flags apply to the predicate grammar as well!

### DIFF
--- a/predicators/approaches/grammar_search_invention_approach.py
+++ b/predicators/approaches/grammar_search_invention_approach.py
@@ -9,8 +9,8 @@ import logging
 from dataclasses import dataclass, field
 from functools import cached_property
 from operator import le
-from typing import Any, Callable, Dict, FrozenSet, Iterator, List, Sequence, \
-    Set, Tuple
+from typing import Any, Callable, Dict, FrozenSet, Iterator, List, Optional, \
+    Sequence, Set, Tuple
 
 import numpy as np
 from gym.spaces import Box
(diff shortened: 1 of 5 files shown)


predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -764,10 +764,34 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
...
         logging.info("Applying predicates to data...")
-        atom_dataset = utils.create_ground_atom_dataset(
-            dataset.trajectories,
-            set(candidates) | self._initial_predicates)
+
+        # Get the template str for the dataset filename for saving
+        # a ground atom dataset.
+        dataset_fname, _ = utils.create_dataset_filename_str(True)
+        # Add a bunch of things relevant to grammar search to the
+        # dataset filename string.
```
tomsilver, line 772: looks fine, but beware too-long filenames can lead to crashes

predicators/utils.py
```diff
@@ -2571,6 +2572,51 @@ def prune_ground_atom_dataset(
     return new_ground_atom_dataset
 
 
+def load_ground_atom_dataset(
+        dataset_fname: str,
+        trajectories: List[LowLevelTrajectory]) -> List[GroundAtomTrajectory]:
+    """Load a previously-saved ground atom dataset."""
```
tomsilver, line 2578: can you add some info about what `trajectories` is?

PR #1617, opened 2024-03-02
Update a paper's title in README.md

yichao-liang:


tomsilver:
thanks for opening a PR! you're right that the README is outdated and confusing about what papers are implemented in this repo. it turns out that many papers are implemented in this repo (not just the predicate invention one). I opened #1618 to clarify.

PR #1618, opened 2024-03-03
update README papers

tomsilver:
@NishanthJKumar after we merge this, we can also update the BDAI fork and add the planning to practice reference in that README


NishanthJKumar (approved):
LGTM!

PR #1619, opened 2024-03-15
Adds new variant of stick button + flag for pruning grammar

NishanthJKumar:
Introduces a new minor variant of the `StickButton` env with movement options. This requires predicate invention to have an expanded grammar that can handle euclidean distances between objects. Develops a debug grammar for predicate invention in this environment which gets ~50/50 success rate via the following command:
```
python predicators/main.py --excluded_predicates all --approach grammar_search_invention --env stick_button_move --seed 0 --grammar_search_grammar_use_diff_features True --grammar_search_grammar_use_euclidean_dist True --grammar_search_max_predicates 2000 --grammar_search_predicate_cost_upper_bound 25 --grammar_search_use_handcoded_debug_grammar True --num_train_tasks 50
```


predicators/utils.py
```diff
@@ -3532,6 +3532,28 @@ def f_range_intersection(lb1: float, ub1: float, lb2: float,
     return (lb1 <= lb2 <= ub1) or (lb2 <= lb1 <= ub2)
 
 
+def compute_abs_bounds_given_frange(lb1: float, ub1: float, lb2: float,
```
tomsilver, line 3535: it seems like this function doesn't necessarily involve "features", so can we change the name to remove the 'f'? let's also do so for `f_range_intersection` above, while we're at it

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -513,56 +572,135 @@ class _FeatureDiffInequalitiesPredicateGrammar(
...
     def enumerate(self) -> Iterator[Tuple[Predicate, float]]:
         # Get ranges of feature values from data.
         feature_ranges = self._get_feature_ranges()
         # Edge case: if there are no features at all, return immediately.
         if not any(r for r in feature_ranges.values()):
             return
+        # Start by generating predicates such that the two features are
+        # very close together.
+        for ret_val in self._yield_pred_given_const(feature_ranges, 0,
+                                                    (1 / 60.0), 4.0):
```
tomsilver, line 616: where are these numbers (1/60 and 4) coming from?

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -513,56 +572,135 @@ class _FeatureDiffInequalitiesPredicateGrammar(
...
+                    feature_ranges, constant_idx, constant, cost):
+                yield ret_val
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class _EuclideanDistancePredicateGrammar(
+        _SingleFeatureInequalitiesPredicateGrammar):
+    """Generates predicates of the form "|0.x - 1.x|^2 + |0.y - 1.y|^2 <= c^2".
+    Importantly, this only operates over types that have features
+    named "x" and "y".
```
tomsilver, line 631: wait, literally 'x' and 'y'? that seems weird to me... can we at least pull out the names 'x' and 'y', even if those are the feature names that are ultimately used at the moment?

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -513,56 +572,135 @@ class _FeatureDiffInequalitiesPredicateGrammar(
...
+        # Compute the upper and lower bounds of each feature range.
+        lbx, ubx = utils.compute_abs_bounds_given_frange(
+            lbx1, ubx1, lbx2, ubx2)
+        lby, uby = utils.compute_abs_bounds_given_frange(
+            lby1, uby1, lby2, uby2)
+        # Now, use these to compute the upper and lower bounds of
+        # the squared expression of interest.
+        lb = lbx**2 + lby**2
+        ub = ubx**2 + uby**2
+        return (lb, ub)
```
tomsilver, line 653: fun math times

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -513,56 +572,135 @@ class _FeatureDiffInequalitiesPredicateGrammar(
...
+        # Start by generating predicates with a very small constant,
+        # to indicate that the objects are touching/overlapped.
+        for (t1, t2) in itertools.combinations_with_replacement(
+                sorted(self.types), 2):
+            if t1 == t2:
+                continue
+            if ("x" in t1.feature_names and "x" in t2.feature_names
+                    and "y" in t1.feature_names and "y" in t2.feature_names):
+                lb, ub = self._compute_xy_bounds(feature_ranges, t1, t2)
+                constant = ((ub - lb) / 500) + lb
```
tomsilver, line 682: 500 also looks magic to me

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -513,56 +572,135 @@ class _FeatureDiffInequalitiesPredicateGrammar(
...
+        feature_ranges = self._get_feature_ranges()
+        # Edge case: if there are no features at all, return immediately.
+        if not any(r for r in feature_ranges.values()):
+            return
+
+        # Start by generating predicates with a very small constant,
+        # to indicate that the objects are touching/overlapped.
+        for (t1, t2) in itertools.combinations_with_replacement(
+                sorted(self.types), 2):
+            if t1 == t2:
```
tomsilver, line 677: how come?

predicators/envs/stick_button.py
```diff
@@ -416,7 +416,11 @@ def _Pressed_holds(state: State, objects: Sequence[Object]) -> bool:
     @classmethod
     def Above_holds(cls, state: State, objects: Sequence[Object]) -> bool:
         """Public for use by oracle options."""
-        obj1, obj2 = objects
+        if len(objects) == 3:
```
tomsilver, line 419: what is the meaning of Above(x, y, z)?

predicators/ground_truth_models/stick_button/options.py
```diff
@@ -242,3 +242,163 @@ def _get_stick_grasp_loc(cls, state: State, stick: Object,
...
+                   params: Array) -> Action:
+            del memory, params  # unused
+            _, stick, button = objects
+            button_circ = StickButtonEnv.object_to_geom(button, state)
+            stick_rect = StickButtonEnv.object_to_geom(stick, state)
+            assert isinstance(stick_rect, utils.Rectangle)
+            tip_rect = StickButtonEnv.stick_rect_to_tip_rect(stick_rect)
+            # If the stick tip is pressing the button, press.
+            if tip_rect.intersects(button_circ):
+                # if (state.get(button, "y") <= 2.96):
```
tomsilver, line 397: remove

tomsilver (approved):
cool stuff! please sanity check that our old predicate invention results are unchanged before merging

predicators/approaches/grammar_search_invention_approach.py
ashayathalye: maybe have the small constant `epsilon=1/60.0` come from `settings.py`?

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: it is! will pull these out and make any magic numbers global constants

predicators/approaches/grammar_search_invention_approach.py
tomsilver: Pulling them into settings is good, but I also want to understand for myself why the magic is needed. If the intention is to choose a value that measures "closeness", why wouldn't something like 1e-6 work?

predicators/envs/stick_button.py
NishanthJKumar: woops - this was a vestigial hack; my b!

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: oh - good question! It's because objects have some amount of "size", and so even when they're touching, it's not like their centers overlap. E.g. in stick button, when the robot touches the button, the center of the robot and the object might still be offset by a bit

predicators/approaches/grammar_search_invention_approach.py
NishanthJKumar: you're asking why we want to continue if `t1 == t2`?

predicators/approaches/grammar_search_invention_approach.py
tomsilver: Got it, thanks! Let's leave a note somewhere for future users. That would be easy to get confused about when trying this grammar in a new environment.

predicators/approaches/grammar_search_invention_approach.py
tomsilver: Yeah

NishanthJKumar: so the form of the predicate will be `(t1.x - t2.x)^2 + (t1.y - t2.y)^2 <= c^2`, and if `t1 == t2`, then this is trivially true

tomsilver: But couldn't the predicate be ground with two different objects of the same type?

NishanthJKumar: oh wait you're so right lol. Will fix that!

tomsilver: 👍 I think there's at least one other case of this below too

ashayathalye:
Potential issues I'm coming across: 
- AboveNoButton() should not be a delete effect of StickPressButton
- AboveNoButton() should not be an add effect of PlaceStickFromButton
- AboveNoButton() should not be a delete effect of StickMoveToButtonFromNothing

NishanthJKumar:
Results from running original method experiments on supercloud:
```
AGGREGATED DATA OVER SEEDS:
                                   NUM_TRAIN_TASKS    NUM_SOLVED AVG_NUM_PREDS AVG_TEST_TIME  AVG_NODES_CREATED         LEARNING_TIME  AVG_REF_COST  NUM_SEEDS
EXPERIMENT_ID                CYCLE                                                                                                                            
cover_main_200demo           None    200.00 (0.00)  49.30 (0.64)   5.90 (0.54)   0.01 (0.00)        4.94 (0.23)        737.06 (92.45)   0.13 (0.01)         10
painting_main_200demo        None    200.00 (0.00)  49.89 (0.31)  21.67 (1.76)   0.26 (0.04)    433.27 (112.58)  115521.74 (18152.33)   0.99 (0.06)          9
pybullet_blocks_main_200demo None    200.00 (0.00)  49.70 (0.46)   7.60 (1.28)   0.75 (0.19)  4044.15 (1784.72)    16956.32 (3235.66)  13.61 (0.83)         10
tools_main_200demo           None    200.00 (0.00)  48.30 (1.95)  29.30 (5.57)   0.57 (0.43)  1858.28 (1776.73)    26487.71 (4595.49)   1.06 (0.43)         10
```

Note that one seed in painting died due to an out-of-memory error that probably had something to do with supercloud and not with the experiment itself

PR #1625, opened 2024-04-03
Fix PlaceStickFromButton NSRT

yichao-liang:
Add `StickAboveButton` precondition to `PlaceStickFromButton` NSRT in the `stick_button` domain


tomsilver (approved):
LGTM assuming tests pass!

PR #1627, opened 2024-04-12
Fix some conceptual issues with StickButtonMovementEnv

### LINKED ISSUE #1626: StickButtonEnv issues
- operator RobotPressButton (and its option) has a stick as an argument, but the stick is never used or referenced by anything, so it seems unnecessary
- operators RobotMoveToButtonFromNothing and RobotMoveToButtonFromButton don't have stick as an argument, but access the state of the stick inside the evaluation of the predicate HandEmpty
- "press" action dimension affects both pressing a button, and placing a stick. This is can cause unintended behavior when a button and holder are both under the stick, in which case the stick is dropped into the holder when only the button was meant to be pressed (e.g. when executing option StickPressButton). The press and place action dimensions should be different.
- PlaceStick option accesses holder state but does not have it as an option argument. 

solution:
- should be able to remove stick as an argument from RobotPressButton without issue
- should introduce an attribute "grip" or something like that to the robot, and then calculate HandEmpty that way
- add action dimension for place, separate from press
- add holder object as parameter to the PlaceStick option and corresponding nsrt

ashayathalye:
There were a few conceptual issues in the environment and definitions of nsrts/options for stick button that I wanted to resolve to minimize confusing results when iterating on predicate invention. This PR closes #1626. 

**Summary of changes:**

- Operator RobotPressButton (and its option) had stick as an argument -- but the stick is never used or referenced by the operator or option, so I removed this. 
- Operators RobotMoveToButtonFromNothing and RobotMoveToButtonFromButton don't have stick as an argument, but access its state inside the evaluation of the predicate HandEmpty (which in turn calls the Grasped predicate's classifier). I added a new attribute for the robot (fingers) like we see in PaintingEnv and ToolsEnv and made HandEmpty's classifier look at that attribute. 
- Sometimes when running the option StickPressButton, the holder happens to be nearby and you also place the stick down. To prevent this unwanted behavior, I made the action space have a separate dimension for "press" and "pickplace". 
- The PlaceStick option accesses the holder's state but didn't have it as an option argument, so I added it as an argument to the option and the corresponding operator. 

**Notes about code:**
There is some amount of code duplication in `StickButtonMovementEnv.simulate()` but I thought it's fine because `StickButtonMovementEnv._get_tasks()` already has a lot of duplicated code. 

I only made the changes to StickButtonMovementEnv and not StickButtonEnv because results from previous or other ongoing work may rely on the way stick button currently works, and we're only making use of StickButtonMovementEnv for our predicate invention work -- but if reviewers would like, I can make these changes to StickButtonEnv. 

**Tests:**
`python predicators/main.py --env stick_button_move --approach oracle --seed [n]` gets 28/50, 30/50, 32/50, and 34/50 solved for seeds 0 through 3, for an average of 31/50 solved. The same command before the changes gets 35/50, 29/50, 33/50, and 32/50 solved for seeds 0 through 3, for an average of 32.25/50 solved. I'm not exactly sure why it does a bit worse after these changes (probably just randomness), but I think the difference is small enough to give us confidence that we don't have any significant errors in the code.

### DIFF
--- a/predicators/envs/stick_button.py
+++ b/predicators/envs/stick_button.py
@@ -537,7 +537,8 @@ class StickButtonMovementEnv(StickButtonEnv):
                 theta = np.pi / 2
             else:
                 theta = rng.uniform(self.theta_lb, self.theta_ub)
-            state_dict[self._robot] = {"x": x, "y": y, "theta": theta}
+            # Initialize the robot with open fingers.
+            state_dict[self._robot] = {"x": x, "y": y, "theta": theta, "fingers": 1.0}
             # Sample the stick, making sure that the origin is in the
             # reachable zone, and that the stick doesn't collide with anything.
             radius = self.robot_radius + self.init_padding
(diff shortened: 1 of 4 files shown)


predicators/envs/stick_button.py
```diff
@@ -611,16 +625,118 @@ def _get_tasks(self, num: int, num_button_lst: List[int],
...
 
+    # def simulate(self, state: State, action: Action) -> State:
+    #     """Run simulation and update tip_x and tip_y."""
+    #     next_state = super().simulate(state, action)
+    #     stick_rect = self.object_to_geom(self._stick, next_state)
+    #     assert isinstance(stick_rect, utils.Rectangle)
+    #     tip_rect = self.stick_rect_to_tip_rect(stick_rect)
+    #     next_state.set(self._stick, "tip_x", tip_rect.x)
+    #     next_state.set(self._stick, "tip_y", tip_rect.y)
+    #     return next_state
```
NishanthJKumar, line 738: Remove

NishanthJKumar (approved):
Looks generally good to me! Thanks for doing this 😄 . I think it's good to leave OG stick button as is, and only change the new variant of the env. 
I see in your above comment that you checked that oracle still runs on cover, but don't we want it to run on stick button? I think the tests check that, but do manually check to make sure before merging

ashayathalye:
> but don't we want it to run on stick button? I think the tests check that, but do manually check to make sure before merging
Lol yeah I did that my brain is just fried right now so I wrote the incorrect thing in the PR description -- will fix.

predicators/envs/stick_button.py
```diff
@@ -611,14 +625,106 @@ def _get_tasks(self, num: int, num_button_lst: List[int],
...
+                next_state.set(self._robot, "fingers", 1.0)
+
+            # Check if the stick is now held for the first time.
+            if not stick_held and stick_rect.intersects(robot_circ):
+                # Check for a collision with the stick holder. The reason that
+                # we only check for a collision here, as opposed to every
+                # timestep, is that we imagine the robot moving down in the z
+                # direction to pick up the stick, at which button it may
+                # collide with the stick holder. On other timesteps, the robot
+                # would be high enough above the holder to avoid collisions.
```
NishanthJKumar, line 716: the "at which button" part of this comment makes no sense to me... Could you update?

NishanthJKumar (commented):
requested changes on one of the comments

predicators/envs/stick_button.py
ashayathalye: Done.

PR #1629, opened 2024-04-25
Enable saving an image sequence of each demonstration

ashayathalye:
For predicate invention with a video-language model, we need to pass in images of the demonstrations to the VLM. This PR adds the flag `--make_demo_images`, which puts image sequences of each demonstration inside `CFG.image_dir` for debugging purposes, analagous to how `--make_demo_videos` puts videos in `CFG.video_dir`.

### DIFF
--- a/predicators/args.py
+++ b/predicators/args.py
@@ -28,6 +28,7 @@ def create_arg_parser(env_required: bool = True,
     parser.add_argument("--make_failure_videos", action="store_true")
     parser.add_argument("--make_interaction_videos", action="store_true")
     parser.add_argument("--make_demo_videos", action="store_true")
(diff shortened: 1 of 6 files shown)


NishanthJKumar (approved):
LGTM! One question tho: we only want one frame after every skill execution for VLM invention, so perhaps you'll need to modify/add utilities to accomplish that

ashayathalye:
How many frames we pass into the VLM per skill execution is an open question. We may want to do more than just one frame from the start/end if the frames in the middle give lots of useful information on what is changing. It may be easier to figure out what changed from something happening in the middle of a trajectory than just looking at the start and end. 

Agree that we can add functions to utililities.py to give us different options for how many / which images from a trajectory we pass into the VLM.

PR #1630, opened 2024-04-27
bump python version to 3.10

tomsilver:


README.md
```diff
@@ -27,7 +27,7 @@ Methods for predicate learning are implemented as Approaches (e.g., `predicators
 A simple implementation of search-then-sample bilevel planning is provided in `predicators/planning.py`. This implementation uses the "SeSamE" strategy: SEarch-and-SAMple planning, then Execution.
 
 ## Installation
-* This repository uses Python versions 3.8+.
+* This repository uses Python versions 3.10+. We recommend 3.10.14.
```
NishanthJKumar, line 30: yichao and others have pointed out issues with using python 3.12+ iirc (i think this is because of the specific package version numbers we freeze in `setup.py`). So maybe we should note that here?

NishanthJKumar (approved):
LGTM! Tysm for helping with this!

PR #1631, opened 2024-04-27
Adds the ability to use VLM's in addition to LLM's

NishanthJKumar:
Main things in this PR
* Makes a common superclass `PretrainedModel` to consolidate LLM's, VLM's and any future large models we might try to use as part of decision-making pipelines
* Implements a VLM class, and integrates with Google's Gemini API (also changes `setup.py` to add necessary requirements).


predicators/pretrained_model_interface.py
```diff
@@ -0,0 +1,247 @@
...
+        # If the temperature is 0, the seed does not matter.
+        if temperature == 0.0:
+            config_id = f"most_likely_{num_completions}_{stop_token}"
+        cache_foldername = f"{model_id}_{config_id}_{prompt_id}"
+        if imgs is not None:
+            # We also need to hash all the images in the prompt.
+            img_hash_list: List[str] = []
+            for img in imgs:
+                img_hash_list.append(str(imagehash.phash(img)))
+            imgs_id = "".join(img_hash_list)
```
tomsilver, line 87: is it possible that this folder name will get so long that we'll have issues?

tomsilver: there might be ways to combine the hashes into one

setup.py
```diff
@@ -33,7 +33,9 @@
         "seaborn==0.12.1",
         "smepy@git+https://github.com/sebdumancic/structure_mapping.git",
         "pg3@git+https://github.com/tomsilver/pg3.git",
-        "gym_sokoban@git+https://github.com/Learning-and-Intelligent-Systems/gym-sokoban.git"  # pylint: disable=line-too-long
+        "gym_sokoban@git+https://github.com/Learning-and-Intelligent-Systems/gym-sokoban.git",  # pylint: disable=line-too-long
+        "ImageHash",
```
tomsilver, line 37: nit: change to `imagehash`

tomsilver (approved):
nice, thanks for splitting, much easier to review!

predicators/pretrained_model_interface.py
NishanthJKumar: yes - good point. It seems like I accidentally auto-merged this PR, but let me think + do something about this!

NishanthJKumar: actually, upon looking into this a bit, I think let's leave this alone until we actually encounter this issue in-the-wild. I think there are definitely ways to resolve it. I'll make a note for future reference!

PR #1634, opened 2024-04-30
Prune operators with low data in grammar search

ashayathalye:
This PR adds a flag `CFG.cluster_and_intersect_prune_low_data_pnads` to prune pnads whose datastores are small in the operator learner `cluster_and_intersect`. Specifically, this will keep pnads that have at least `CFG.cluster_and_intersect_min_datastore_fraction` fraction of the segments produced by the option in their NSRT. 

The motivation for this is that when using a VLM for predicate invention in the grammar search invention approach, the noise in the VLM's labeling makes it so that cluster_and_intersect learns a lot of pnads, many with tiny datastores, which significantly slows down hill climbing because there are just way too many operators.


NishanthJKumar (approved):
LGTM!

PR #1635, opened 2024-04-30
fix segmentation error with predicate invention from txt files or img demos

NishanthJKumar:
Previously, our approach would error out and die whenever we ran predicate invention via the newly-implemented txt file or img demos features. This is because segmentation called from `nsrt_learning_approach.py` tries to call `utils.abstract` on the proposed predicates, and can't (because they're VLM predicates and their classifiers haven't been implemented yet!). But of course, we know what the abstract states are for all the demonstrations, so this PR just passes that along correctly, and now `nsrt_learning` runs without erroring out!

### DIFF
--- a/predicators/approaches/nsrt_learning_approach.py
+++ b/predicators/approaches/nsrt_learning_approach.py
@@ -74,6 +74,27 @@ class NSRTLearningApproach(BilevelPlanningApproach):
             ground_atom_dataset = utils.create_ground_atom_dataset(
                 trajectories, self._get_current_predicates())
             utils.save_ground_atom_dataset(ground_atom_dataset, dataset_fname)
+        elif CFG.offline_data_method in ["demo+labelled_atoms", "img_demos"]:
+            # In this case, the annotations aren't basically ground atoms!
+            # We can use these to make GroundAtomTrajectories.
+            assert annotations is not None
+            assert len(annotations) == len(trajectories)
+            ground_atom_dataset = []
+            annotations_with_only_known_preds = []
+            known_preds = self._get_current_predicates()
+            for atoms_traj in annotations:
+                curr_known_preds_atoms_traj = []
+                for atoms_set in atoms_traj:
+                    curr_known_preds_atoms_set = set(
+                        atom for atom in atoms_set
+                        if atom.predicate in known_preds)
+                    curr_known_preds_atoms_traj.append(
+                        curr_known_preds_atoms_set)
+                annotations_with_only_known_preds.append(
+                    curr_known_preds_atoms_traj)
+            for ll_traj, atoms in zip(trajectories,
+                                      annotations_with_only_known_preds):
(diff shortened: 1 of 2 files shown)


predicators/approaches/nsrt_learning_approach.py
```diff
@@ -74,6 +74,27 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
             ground_atom_dataset = utils.create_ground_atom_dataset(
                 trajectories, self._get_current_predicates())
             utils.save_ground_atom_dataset(ground_atom_dataset, dataset_fname)
+        elif CFG.offline_data_method in ["demo+labelled_atoms", "img_demos"]:
+            # In this case, the annotations aren't basically ground atoms!
```
ashayathalye, line 78: ~~aren't~~ --> are?

predicators/approaches/nsrt_learning_approach.py
```diff
@@ -74,6 +74,27 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
...
+            # We can use these to make GroundAtomTrajectories.
+            assert annotations is not None
+            assert len(annotations) == len(trajectories)
+            ground_atom_dataset = []
+            annotations_with_only_known_preds = []
```
ashayathalye, line 83: I'm trying to understand the name/meaning of `annotations_with_only_known_preds`. I'm a bit generally confused. In the VLM setting, are there any annotations that aren't known predicates? 

I understand that: we have some given predicates (goal predicates), and then the VLM comes up with some other predicates, and we subselect out of those after grammar search invention approach is done, to get our final set of predicates. 
^ in this context, what is a known predicate?

tests/approaches/test_grammar_search_invention_approach.py
```diff
@@ -165,6 +167,34 @@ def test_labelled_atoms_invention():
...
+    env = IceTeaMakingEnv()
+    train_tasks = env.get_train_tasks()
+    loaded_dataset = create_dataset(env, train_tasks,
+                                    get_gt_options(env.get_name()))
+    approach = GrammarSearchInventionApproach(env.goal_predicates,
+                                              get_gt_options(env.get_name()),
+                                              env.types, env.action_space,
+                                              train_tasks)
+    approach.learn_from_offline_dataset(loaded_dataset)
+    assert len(approach._get_current_predicates()) == 1  # pylint:disable=protected-access
```
ashayathalye, line 195: Maybe add a comment explaining why there is only 1 predicate after learning? Not intuitive? (I'm also confused why.)

predicators/approaches/nsrt_learning_approach.py
NishanthJKumar: will rename!

tomsilver (approved):

PR #1638, opened 2024-05-01
fix some minor prompting + parsing issues

NishanthJKumar:
Previously, we weren't labelling one of the atom proposal prompts with "Skills executed in trajectory" (rather, we directly just vomited the skills there out of context), and also we didn't parse out any predicates that had objects with `_` in them (because this didn't happen in our example!). This quick PR fixes these before I forget


ashayathalye (commented):
I wonder if adding explicit numbers will help convey the order the skills are executing in, rather than just listing them in order -- something we can tweak later.

LGTM

NishanthJKumar:
Good suggestion; we should add that to our (long) queue of various things to look into!

tomsilver (approved):

PR #1640, opened 2024-05-01
Fix parsing of trajectory demonstration subfolders for macs.

ashayathalye:
Mac creates these annoying `.DS_STORE` files that customize how the application "finder" displays folders. The current code for parsing the subfolders that hold images for each demonstration doesn't account for any other files being in the folder other than the `traj_<number>` files. Slight update to the code to make it not crash in these scenarios.


NishanthJKumar (approved):
LGTM!

PR #1643, opened 2024-05-07
Test-time VLM Predicate evaluation

NishanthJKumar:
Implements a new `VLMPredicate` class that can be called at test time, makes sure that these are called during abstraction in a batched fashion (implements a new `utils` function for this), modifies execution monitor accordingly.

### DIFF
--- a/predicators/datasets/generate_atom_trajs_with_vlm.py
+++ b/predicators/datasets/generate_atom_trajs_with_vlm.py
@@ -20,7 +20,7 @@ from predicators.pretrained_model_interface import GoogleGeminiVLM, \
 from predicators.settings import CFG
 from predicators.structs import Action, Dataset, GroundAtom, \
     ImageOptionTrajectory, LowLevelTrajectory, Object, ParameterizedOption, \
-    State, Task, VLMPredicate, _Option
+    State, Task, _Option
 
 
 def _generate_prompt_for_atom_proposals(
@@ -274,12 +274,7 @@ def _parse_structured_state_into_ground_atoms(
         assert obj_type.name == "object"
     assert obj_type is not None
 
-    def _stripped_classifier(
-            state: State,
-            objects: Sequence[Object]) -> bool:  # pragma: no cover.
-        raise Exception("Stripped classifier should never be called!")
-
-    def _vlm_query_str(pred_name: str, objects: Sequence[Object]) -> str:
+    def _get_vlm_query_str(pred_name: str, objects: Sequence[Object]) -> str:
         return pred_name + "(" + ", ".join(
             str(obj.name) for obj in objects) + ")"  # pragma: no cover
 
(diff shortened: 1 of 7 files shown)


predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -278,6 +279,10 @@ def _stripped_classifier(
             objects: Sequence[Object]) -> bool:  # pragma: no cover.
         raise Exception("Stripped classifier should never be called!")
 
+    def _vlm_query_str(pred_name: str, objects: Sequence[Object]) -> str:
```
tomsilver, line 282: let's rename to `_get_vlm_query_str` to make clear this is a function

predicators/execution_monitoring/expected_atoms_monitor.py
```diff
@@ -30,7 +31,20 @@ def step(self, state: State) -> bool:
...
+            if isinstance(atom.predicate, VLMPredicate))
+        non_vlm_unsat_atoms = {
+            a
+            for a in (next_expected_atoms - next_expected_vlm_atoms)
+            if not a.holds(state)
+        }
+        vlm_unsat_atoms = set()
+        if len(next_expected_vlm_atoms) > 0:
+            vlm_unsat_atoms = utils.query_vlm_for_atom_vals(
+                sorted(list(next_expected_vlm_atoms)),
```
tomsilver, line 45: can we change `query_vlm_for_atom_vals` to take in a Collection of ground atoms, and do the sorting inside that function? from the user's perspective, it's not clear why the input to this function needs to be sorted.

also -- `sorted(next_expected_vlm_atoms)` should work (no `list` needed)

predicators/execution_monitoring/expected_atoms_monitor.py
```diff
@@ -30,7 +31,20 @@ def step(self, state: State) -> bool:
...
+                state)  # pragma: no cover.
```
tomsilver, line 46: nit: no period to end same-line comments

predicators/structs.py
```diff
@@ -301,6 +301,24 @@ def _negated_classifier(self, state: State,
...
+    def __lt__(self, other: Predicate) -> bool:
+        return str(self) < str(other)
+
+
+@dataclass(frozen=True, order=False, repr=False)
+class VLMPredicate(Predicate):
+    """Struct defining a predicate that calls a VLM as part of returning its
+    truth value.
+
+    NOTE: when instantiating a VLMPredicate, we typically pass in a 'Dummy'
```
tomsilver, line 313: can we make a small utils function that creates VLMPredicates but doesn't require the user to manually specify the dummy classifier? `create_vlm_predicate(name, types, get_vlm_query_str) -> VLMPredicate`

tomsilver (approved):
nice work!

predicators/structs.py
```diff
@@ -301,6 +301,24 @@ def _negated_classifier(self, state: State,
...
+    classifier (i.e., one that returns simply raises some kind of error instead
+    of actually outputting a value of any kind).
+    """
+    get_vlm_query_str: Callable[[Sequence[Object]], str]
+
+    def __hash__(self) -> int:
```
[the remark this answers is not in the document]
NishanthJKumar: turns out parent class doesn't inherit this method unless we set `eq=False` in the decorator for the dataclass. I just did this

PR #1644, opened 2024-05-08
[wip] adding pybullet version of coffee environment

tomsilver:
long overdue, not still not done. example command:

```
python predicators/main.py --approach oracle --env pybullet_coffee --seed 0   --num_train_tasks 0 --num_test_tasks 1 --coffee_num_cups_test '[3]' --debug --bilevel_plan_without_sim True --pybullet_control_mode reset --use_gui --make_failure_videos --pybullet_camera_width 1674 --pybullet_camera_height 900 --make_test_videos --video_fps 20
```


tomsilver:
I suspect there is some gimbal lock issue related to the conversion between euler and quaternion angles. We might see about getting rid of the euler angles entirely. The observed behavior sounds a lot like [this](https://en.wikipedia.org/wiki/Gimbal_lock#Robotics):

>An example of a wrist flip, also called a wrist singularity, is when the path through which the robot is traveling causes the first and third axes of the robot's wrist to line up. The second wrist axis then attempts to spin 180° in zero time to maintain the orientation of the end effector. The result of a singularity can be quite dramatic and can have adverse effects on the robot arm, the end effector, and the process.

PR #1645, opened 2024-05-09
Enable using CogMan to create demonstrations

NishanthJKumar:
Previously, we were generating training demonstrations without using CogMan. This is fine as long as a perceiver isn't required (which hasn't been necessary in the past). This PR implements a generalization of the old pipeline that handles the new use case.

At some point, we probably want to completely deprecate the `run_policy` and `run_policy_with_simulator` functions in `utils.py`. I tried, but that's unfortunately quite a big change, and probably not worth it right now. I updated the docstring for posterity though.


predicators/datasets/demo_only.py
```diff
@@ -154,63 +162,61 @@ def _generate_demonstrations(env: BaseEnv, train_tasks: List[Task],
     if annotate_with_gt_ops:
         annotations = []
     num_tasks = min(len(train_tasks), CFG.max_initial_demos)
-    rng = np.random.default_rng(CFG.seed)
-    if CFG.offline_data_bilevel_plan_without_sim is None:
-        bilevel_plan_without_sim = CFG.bilevel_plan_without_sim
-    else:
-        bilevel_plan_without_sim = CFG.offline_data_bilevel_plan_without_sim
+    # rng = np.random.default_rng(CFG.seed)
```
tomsilver, line 165: undo?

tests/test_cogman.py
```diff
@@ -72,3 +77,158 @@ def test_cogman_with_expected_atoms_monitor():
     next_obs = env.step(act)
     next_act = cogman.step(next_obs)
     assert not np.allclose(act.arr, next_act.arr)
+
+
+def testrun_episode_and_get_observations():
```
tomsilver, line 82: test_

tomsilver (approved):
great work!

PR #1647, opened 2024-05-14
Fix `kettle_only` oracle for kitchen

NishanthJKumar:
Previously, the robot would fail at the `kettle_only` task in kitchen ~50% of the time even with perfect samplers. This is because two plans were equally probable: pushing the kettle or pulling the kettle. This PR introduces a predicate to fix this and make oracle achieve a 100% success rate at the simple `kettle_only` goals.

Command (gets 10/10 now, and much less than 10/10 before):
```
python predicators/main.py --approach oracle --bilevel_plan_without_sim True --num_test_tasks 10 --kitchen_goals kettle_only --env kitchen --use_gui --seed 5 --kitchen_use_perfect_samplers True
```


tomsilver (approved):
oops wrong account

PR #1650, opened 2024-05-17
Create BurgerEnv to showcase geometric + VLM predicate invention

ashayathalye:
This environment is designed to showcase a predicate invention approach that learns geometric predicates that operate on the object-centric state and vision-language model predicates that operate on the visual rendering of the state. 

One quirk of this environment is that we want certain parts of the state to only be accessible by the oracle approach. This is because we want to invent predicates like `IsCooked` as a VLM predicate, but not a geometric predicate, so no information about how cooked an object is should be in the state. The work-around solution to give oracle approach access to this hidden state (so that we don't need to implement image-based predicates for oracle) to this is to hide certain state inside `State.simulator_state`. After the demonstrations are created by the oracle approach, we can erase this simulator state before we pass the demonstrations to the predicate invention approach to ensure it can't cheat. 

Example command to see demos created: 
```python predicators/main.py --env gridworld --approach grammar_search_invention --seed 0 --num_train_tasks 10 --option_model_terminate_on_repeat False --sesame_max_skeletons_optimized 1000 --timeout 80 --sesame_max_samples_per_step 1 --make_demo_videos --sesame_task_planner fdopt```

Notes:
- The reason we set `--option_model_terminate_on_repeat False` is that the `State.all_close()` check does not work when `State.simulator_state` is not `None`. 
- The default astar task planner is too slow -- fast downward is required to get it to solve in a reasonable amount of time. You will see that I made very specific oracle nsrts to try to make the branching factor smaller. 
- The oracle nsrts have no ignore effects. This is not ideal for this environment, which would benefit from oracle nsrts with ignore effects due to the various possibilities for being `Adjacent` to many things. However, the operator learner we are using after predicate invention is `cluster_and_intersect`, so it's helpful to have oracle nsrts that will most closely resemble the kind of things we will learn, for debugging purposes. 
- At the moment there is no variation in the training/test tasks -- I will add this in a subsequent PR. 

Remaining TODOs for this PR:
- Change the name from GridWorld to BurgerPrep
- Delete `State.simulator_state` after the demos are made 
- Add a flag to make the `State.all_close` function also compare the simulator state

### CODE REVIEW

File: predicators/ground_truth_models/gridworld/nsrts.py [hunk 1]
```diff
@@ -0,0 +1,605 @@
+"""Ground-truth NSRTs for the gridworld environment."""
```
NishanthJKumar, line 1: This file generally looks good. Some minor feedback:
* Remove all unnecessary NSRTs and variables that are currently commented out
* It could be useful to write a comment or two explaining why we have all the extra move and pick operators. This will mostly be useful for us in the future when we try to debug/modify these NSRTs

predicators/ground_truth_models/gridworld/options.py
```diff
@@ -0,0 +1,233 @@
...
+
+    @classmethod
+    def get_options(cls, env_name: str, types: Dict[str, Type],
+                    predicates: Dict[str, Predicate],
+                    action_space: Box) -> Set[ParameterizedOption]:
+
+        # Types
+        # top_bun_type = types["top_bun"]
+        # bottom_bun_type = types["bottom_bun"]
+        # cheese_type = types["cheese"]
```
NishanthJKumar, line 30: remove unnecessary comments here and below

predicators/envs/gridworld.py
```diff
@@ -0,0 +1,647 @@
...
+                next_state.set(held_item, "z", new_z)
+
+        # print("state: ", state)
+        # print("action: ", action)
+        # print("next state: ", next_state)
+        # print()
+        # top_bun = [obj for obj in state if obj.name == "top_bun"][0]
+        # if pickplace > 0 and state.simulator_state[top_bun]["is_held"] > 0.5:
+        #     print()
+        #     import pdb; pdb.set_trace()
```
NishanthJKumar, line 488: remove unnecessary comments

predicators/envs/gridworld.py
```diff
@@ -0,0 +1,647 @@
...
+    Predicate, State, Type
+
+
+class GridWorldEnv(BaseEnv):
+    """A simple gridworld environment where a robot prepares a burger, inspired
+    by https://github.com/portal-cornell/robotouille.
+
+    This environment is designed to showcase a predicate invention approach that
+    learns geometric predicates that operate on the object-centric state and
+    video-language model predicates that operate on the visual rendering of the
```
NishanthJKumar, line 31: nit: vision-language not video-language

predicators/envs/gridworld.py
```diff
@@ -0,0 +1,647 @@
...
+            GroundAtom(self._On, [cheese, patty]),
+            GroundAtom(self._On, [tomato, cheese]),
+            GroundAtom(self._On, [top_bun, tomato]),
+            GroundAtom(self._IsCooked, [patty]),
+            GroundAtom(self._IsSliced, [tomato]),
+        }
+
+        for _ in range(num):
+            state = utils.create_state_from_dict(state_dict)
+            state.simulator_state = hidden_state
```
NishanthJKumar, line 184: Note that the VLM predicate invention pipeline currently uses the simulator state to store renderings of the observation to pass to the VLM for predicate invention purposes. We can certainly modify this, and I don't think it's important for this specific PR, but we should probably chat about this and implement something that allows VLM stuff to be run on this env in the next PR

predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -341,7 +341,8 @@ def _quantified_types(self) -> List[Type]:
         ]
 
     def _classify_object(self, s: State, obj: Object) -> bool:
-        assert obj.type == self.body.types[self.free_variable_idx]
+        # assert obj.type == self.body.types[self.free_variable_idx]
+        assert obj.is_instance(self.body.types[self.free_variable_idx])
```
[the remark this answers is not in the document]
ashayathalye: The previous code does not support object type inheritance because it checks via `==`. Checking via `is_instance()` also checks if `obj`'s parent type matches. 

Do you want me to ^ write this explanation as a comment in the code? I think it's unnecessary.

predicators/ground_truth_models/gridworld/nsrts.py
ashayathalye: I had some extra NSRTs because we may need those ones once we make the goals in the environment more varied, e.g. via more randomization of initial locations of the various items and stations. So I started to write them because I was gonna use them, but then I wanted to make the first PR more simple and not have the variation in initial state. 

I could leave them in but add a comment explaining why they are left in there?

predicators/envs/gridworld.py
ashayathalye: I think it's fine to leave in there, since we only need to read things from the simulator state dictionary for a specific key. At the moment, I'm not comparing `compare_fn(state1.simualtor_state, state2.simulator_state)` anywhere. 

^ I would like to add this comparison possibility -- in which case maybe we can have a flag set the specific type of simulator state comparison function that is used (or better -- maybe we make a new `State.allclose2()` function that also takes in a comparison function as an argument for how to compare the simulator state. 

++ on dealing with that in the next PR

predicators/approaches/grammar_search_invention_approach.py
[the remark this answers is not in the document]
NishanthJKumar: nope - that's fine.

predicators/ground_truth_models/gridworld/nsrts.py
NishanthJKumar: leave it as a comment; that'd be great!

NishanthJKumar (approved):
Great work 🎉 !

PR #1654, opened 2024-05-22
Adds set-of-marks rendering for burners in the kitchen environment

NishanthJKumar:
Useful for VLM predicate invention. Note that I didn't test the utils functions I wrote because the only way to verify them is visually, and any test I could think to write was kinda just for the sake of running the code...

https://github.com/Learning-and-Intelligent-Systems/predicators/assets/12446934/b3484303-a3f4-4498-a712-60dea2fffbb9


tomsilver (approved):
exciting! lgtm when checks pass

PR #1656, opened 2024-05-23
Image-option diff-based VLM prompting

NishanthJKumar:
Introduces new diff-based prompting for atom labelling. Note that this only works at training time, and is not implemented at test time; which needs to be implemented in the future.


predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -98,14 +98,40 @@ def _generate_prompt_for_scene_labelling(
...
+            init_prompt += f"\n{atom_str}"
+        ret_list.append((init_prompt, traj.imgs[0]))
+        # Now, we use actual difference-based prompting for the second timestep
+        # and beyond.
+        # The prompt ends with a section for 'Predicates', so list these.
+        for atom_str in atoms_list:
+            prompt += f"\n{atom_str}"
+        for i in range(1, len(traj.imgs)):
+            curr_prompt = prompt[:]
+            # NOTE: same problem with ripping out images as in the above note.
```
ashayathalye, line 118: This comment refers to an earlier comment but the earlier comment is being deleted in this PR as of now, I think.

NishanthJKumar: yes good catch!

PR #1660, opened 2024-05-28
Support VLM predicate invention in BurgerEnv

ashayathalye:
Before we can get geometric + VLM predicate invention working in BurgerEnv, we need VLM predicate invention without geometric predicates working in BurgerEnv. This PR gets VLM predicate invention to run in BurgerEnv -- it does not yet work well (the predicate proposals are decent but the labeling isn't good enough yet) -- but at least it runs: we see `Cooked` and `Holding` generated as VLM predicates, but `Cooked` is not labeled accurately. 

For example, you can run `python predicators/main.py --env burger --approach grammar_search_invention --seed 0 --num_train_tasks 1 --num_test_tasks 1 --bilevel_plan_without_sim True --offline_data_method demo_with_vlm_imgs --vlm_model_name gemini-1.5-pro-latest --segmenter every_step --grammar_search_vlm_atom_proposal_prompt_type options_labels_whole_traj --grammar_search_vlm_atom_label_prompt_type img_option_diffs --grammar_search_task_planning_timeout 10.0 --sesame_max_skeletons_optimized 200 --disable_harmlessness_check True --sesame_task_planner fdopt --excluded_predicates all --option_model_terminate_on_repeat False --grammar_search_vlm_atom_proposal_use_debug False`.

Main changes:
- Now hashing the concatenated image ids that contribute to the cache foldername for the cached VLM output (was getting filename too long issues on my mac).
- Recall that in BurgerEnv we put state information that we want to provide only to the oracle and not to the agent, in `State.simulator_state` (well, now `State.simulator_state["state"]` as of this PR). The current VLM predicate invention pipeline is using `State.simulator_state` to store the images associated with each state. So we just needed to slightly modify `State.simulator_state` to hold both of these -- now the images go in `State.simulator_state["images"]`.
- Some funky code to convert matplotlib fig to PIL.Image.

### DIFF
--- a/predicators/envs/burger.py
+++ b/predicators/envs/burger.py
@@ -7,6 +7,8 @@ Nicole Thean (https://github.com/nicolethean).
 
 import copy
 import logging
+import io
+from PIL import Image
 from typing import Callable, List, Optional, Sequence, Set, Tuple
 
 import matplotlib
@@ -19,7 +21,7 @@ from predicators import utils
 from predicators.envs import BaseEnv
 from predicators.settings import CFG
 from predicators.structs import Action, EnvironmentTask, GroundAtom, Object, \
-    Predicate, State, Type, DefaultTask, Observation
+    Predicate, State, Type, DefaultTask, Observation, Video
 
 
 class BurgerEnv(BaseEnv):
@@ -185,7 +187,7 @@ class BurgerEnv(BaseEnv):
             state.simulator_state = hidden_state
             # A DefaultTask is basically a dummy task. Our render function
             # does not use the task argument, so this is ok.
-            state.simulator_state["image"] = self.render_state(state, DefaultTask)[0]
+            state.simulator_state["image"] = self.render_state(state, DefaultTask)
             # Recall that a EnvironmentTask consists of an Observation and a
             # GoalDescription, both of whose types are Any.
             tasks.append(EnvironmentTask(state, goal))
@@ -484,7 +486,7 @@ class BurgerEnv(BaseEnv):
                 next_state.set(held_item, "z", new_z)
 
         # Update the image
-        next_state.simulator_state["image"] = self.render_state(state, DefaultTask)[0]
+        next_state.simulator_state["image"] = self.render_state(state, DefaultTask)
 
         return next_state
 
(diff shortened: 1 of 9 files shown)


tests/envs/test_burger.py
```diff
@@ -30,6 +30,7 @@ def test_burger():
     for task in env.get_test_tasks():
         for obj in task.init:
             assert len(obj.type.feature_names) == len(task.init[obj])
+
```
NishanthJKumar, line 33: remove this line

tests/envs/test_burger.py
```diff
@@ -39,6 +40,7 @@ def test_burger():
     nsrts = get_gt_nsrts(env.get_name(), env.predicates, options)
     assert len(nsrts) == 15
     task = env.get_train_tasks()[0]
+
```
NishanthJKumar, line 43: remove

predicators/pretrained_model_interface.py
```diff
@@ -125,6 +127,11 @@ def sample_completions(self,
         cached_prompt, completion_strs = cache_str.split(_CACHE_SEP, 1)
         assert cached_prompt == prompt
         completions = completion_strs.split(_CACHE_SEP)
+
+        logging.debug(f"Querying model {model_id} with new prompt.")
+        # Query the model.
+        completions = self._sample_completions(prompt, imgs, temperature, seed,
+                                               stop_token, num_completions)
```
NishanthJKumar, line 134: woah woah - what is this? why are you querying the model again?

predicators/envs/burger.py
```diff
@@ -173,17 +178,24 @@ def _get_tasks(self, num: int,
...
             GroundAtom(self._On, [cheese, patty]),
-            GroundAtom(self._On, [tomato, cheese]),
-            GroundAtom(self._On, [top_bun, tomato]),
+            # GroundAtom(self._On, [tomato, cheese]),
+            # GroundAtom(self._On, [top_bun, tomato]),
             GroundAtom(self._IsCooked, [patty]),
-            GroundAtom(self._IsSliced, [tomato]),
+            # GroundAtom(self._IsSliced, [tomato]),
+            # GroundAtom(self._GoalHack, [bottom_bun, patty, cheese, tomato,
+            #     top_bun])
```
NishanthJKumar, line 186: why are all these commented out for now?

NishanthJKumar (commented):
looks generally good! Left some comments - the most serious of which is about querying `sample_completions` seemingly multiple times. Happy to approve after those are addressed

predicators/pretrained_model_interface.py
ashayathalye: I'd commented out the earlier code that calls this function and caches the result when the caching was throwing an error (because of the foldername being too long), and that's why I had this one here -- and I just forgot to remove this line after I uncommented the earlier code and fixed the caching issue.

predicators/envs/burger.py
ashayathalye: The pipeline works fine with the full goal, I just wanted to see the learned predicates/operators with something simpler so I commented those out to make the goal simpler. I'll uncomment them -- doesn't really matter but probably best to keep it as the full goal for what's on master.

NishanthJKumar (approved):
LGTM! Thanks for doing this; very exciting stuff!

PR #1662, opened 2024-06-03
Ask vlm for description before labeling

ashayathalye:
In many instances:

- if you ask the VLM to describe the changes between two images taken before and after a skill was executed, its description is accurate. 
- But if you ask the VLM to label predicates in the after image given labels of predicates in the before image, its labels are often incorrect. 

In an effort to deal with this, we ask the VLM to *first* describe the changes in the scene before it outputs any predicate values. This will then allow it to attend to its description, which will then help it generate correct labels -- I think. 

This enables the VLM to propose and consistently, accurately label `Cooked(patty)` in BurgerEnv when the goal is `On(patty, bottom_bun), On(cheese, patty), IsCooked(patty)` via 
```python predicators/main.py --env burger --approach grammar_search_invention --seed 0 --num_train_tasks 1 --num_test_tasks 1 --bilevel_plan_without_sim True --offline_data_method demo_with_vlm_imgs --vlm_model_name gemini-1.5-pro-latest --segmenter every_step --grammar_search_vlm_atom_proposal_prompt_type options_labels_whole_traj --grammar_search_vlm_atom_label_prompt_type img_option_diffs_label_history --grammar_search_task_planning_timeout 10.0 --sesame_max_skeletons_optimized 200 --disable_harmlessness_check True --sesame_task_planner fdopt --excluded_predicates all --option_model_terminate_on_repeat False --grammar_search_vlm_atom_proposal_use_debug False```

We could also ask the VLM to generate explanations for each predicate *before* labeling that predicate -- for the same reason that it can attend to that explanation while it labels the predicate.


NishanthJKumar (approved):
LGTM! exciting that this produces good results for us!

PR #1663, opened 2024-06-04
Allow different task goals for demonstrator vs agent

ashayathalye:
### Background:
When doing VLM predicate invention, we may want to invent goal predicates. In this case, we need some alternative way to describe the goal for the agent than the normal goal predicates. One (slightly hacky?) thing we can do is have a new goal predicate that evaluates to True when all of the actual goal predicates evaluate to True. We can't also have the demonstrator use this new goal predicate as their goal, because the oracle operators are defined in terms of the original goal predicates. Rather than rewrite the oracle operators in terms of this hacky new goal predicate, it makes sense to instead provide a way for the tasks to be in terms of different goal predicates for the agent versus the demonstrator. 

Conceptually, this seems ok. The demonstrator solved some tasks with some goals in mind. There's no reason we can't take those same trajectories and give it to an agent to learn from, but frame it in terms of goals that are described a bit differently than what the demonstrator had in mind. 

### Code-changes:
Goals for tasks are specified in an environment's `_get_tasks()` function. The alternative goals need to be set inside that function as well, because otherwise we wouldn't know how to ground the new goal predicates. We add a new field to the `EnvironmentTask` and `Task` classes for an alternative goal that can be set inside `_get_tasks()`. Then, in `main.py`, we replace the goal with the alternative goal for all the tasks that are given to the approach. We don't do this replacing for the tasks that are passed to the demonstrator. 

### Impact:
After this change, in BurgerEnv–– if we exclude `IsCooked`, we propose and consistently accurately label `Cooked(patty)` as a VLM predicate, and then solve an actual test task (which at the moment is the same as the train task, but at lest we go through the motions of querying the VLM at each timestep at test time). This is the first successful demonstration of the entire pipeline in BurgerEnv. 

^ to get this result, you can run:
`python predicators/main.py --env burger --approach grammar_search_invention --seed 0 --num_train_tasks 1 --num_test_tasks 1 --bilevel_plan_without_sim True --offline_data_method demo_with_vlm_imgs --vlm_model_name gemini-1.5-pro-latest --segmenter every_step --grammar_search_vlm_atom_proposal_prompt_type options_labels_whole_traj --grammar_search_vlm_atom_label_prompt_type img_option_diffs_label_history --grammar_search_task_planning_timeout 10.0 --sesame_max_skeletons_optimized 200 --disable_harmlessness_check True --sesame_task_planner fdopt --excluded_predicates IsCooked --option_model_terminate_on_repeat False --grammar_search_vlm_atom_proposal_use_debug False --allow_exclude_goal_predicates True`

### DIFF
--- a/predicators/envs/burger.py
+++ b/predicators/envs/burger.py
@@ -176,14 +176,25 @@ class BurgerEnv(BaseEnv):
         hidden_state[bottom_bun] = {"is_held": 0.0}
 
         goal = {
-            GroundAtom(self._On, [patty, bottom_bun]),
-            GroundAtom(self._On, [cheese, patty]),
-            GroundAtom(self._On, [tomato, cheese]),
-            GroundAtom(self._On, [top_bun, tomato]),
-            GroundAtom(self._IsCooked, [patty]),
-            GroundAtom(self._IsSliced, [tomato]),
+            # GroundAtom(self._On, [patty, bottom_bun]),
+            # GroundAtom(self._On, [cheese, patty]),
+            # GroundAtom(self._On, [tomato, cheese]),
+            # GroundAtom(self._On, [top_bun, tomato]),
+            # GroundAtom(self._IsCooked, [patty]),
+            # GroundAtom(self._IsSliced, [tomato]),
             # GroundAtom(self._GoalHack, [bottom_bun, patty, cheese, tomato,
             #     top_bun])
+
+            GroundAtom(self._On, [patty, bottom_bun]),
+            GroundAtom(self._On, [cheese, patty]),
+            GroundAtom(self._IsCooked, [patty]),
+        }
+
+        alt_goal = {
+            GroundAtom(self._On, [patty, bottom_bun]),
+            GroundAtom(self._On, [cheese, patty]),
+            GroundAtom(self._GoalHack, [bottom_bun, patty, cheese, tomato,
+                top_bun])
         }
 
         for _ in range(num):
@@ -196,7 +207,7 @@ class BurgerEnv(BaseEnv):
                 state, DefaultEnvironmentTask)
             # Recall that a EnvironmentTask consists of an Observation and a
             # GoalDescription, both of whose types are Any.
-            tasks.append(EnvironmentTask(state, goal))
+            tasks.append(EnvironmentTask(state, goal, _alt_goal_desc=alt_goal))
 
         return tasks
 
@@ -301,10 +312,10 @@ class BurgerEnv(BaseEnv):
         atoms = [
             self._On_holds(state, [patty, bottom]),
             self._On_holds(state, [cheese, patty]),
-            self._On_holds(state, [tomato, cheese]),
-            self._On_holds(state, [top, tomato]),
+            # self._On_holds(state, [tomato, cheese]),
+            # self._On_holds(state, [top, tomato]),
             self._IsCooked_holds(state, [patty]),
-            self._IsSliced_holds(state, [tomato])
+            # self._IsSliced_holds(state, [tomato])
         ]
         return all(atoms)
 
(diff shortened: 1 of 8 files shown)


predicators/envs/burger.py
```diff
@@ -176,14 +176,24 @@ def _get_tasks(self, num: int,
         hidden_state[bottom_bun] = {"is_held": 0.0}
 
         goal = {
```
ashayathalye, line 178: I've commented the real goal out because I'm never using the full goal during testing -- it's slower to plan to it and more prone to errors. Rather than just completely replace it for the next `N` PRs, I say we keep it commented out here. Since no one else is working on BurgerEnv except us, I think it's fine to leave it in this informal state for now.

predicators/main.py
```diff
@@ -309,7 +317,9 @@ def _generate_interaction_results(
 
 
 def _run_testing(env: BaseEnv, cogman: CogMan) -> Metrics:
-    test_tasks = env.get_test_tasks()
+    test_tasks = [
```
ashayathalye, line 320: Note: 
During training, we used the perceiver to turn all the environment tasks into tasks before passing the into the approach. 

But at test time, the "turning into a task" is done inside CogMan, where it's a bit more annoying to replace the goal with the alternative goal -- so I do it here, before we pass things into CogMan.

predicators/structs.py
```diff
@@ -430,6 +430,12 @@ class Task:
     """Struct defining a task, which is an initial state and goal."""
     init: State
     goal: Set[GroundAtom]
+    # Sometimes we want the task presented to the agent to have goals described
+    # in terms of predicates that are different than those describing the goal
+    # of the task presented to the demonstrator. In these cases, we will store
+    # an "alternative goal" in this field and replace the goal with the
+    # alternative goal before giving the task to the agent.
+    alt_goal: Optional[Set[GroundAtom]] = field(default_factory=set)
```
ashayathalye, line 438: I think "alternative goal" is a decent name -- not great, but not terrible. Would be great if you have any suggestions for a better name.

predicators/utils.py
```diff
@@ -2199,19 +2199,17 @@ def _stripped_classifier(state: State, objects: Sequence[Object]) -> bool:
 
 
 def strip_task(task: Task, included_predicates: Set[Predicate]) -> Task:
-    """Create a new task where any excluded predicates have their classifiers
-    removed."""
+    """Create a new task where any excluded goal predicates have their
+    classifiers removed."""
     stripped_goal: Set[GroundAtom] = set()
     for atom in task.goal:
-        # The atom's goal is known.
```
ashayathalye, line 2206: Removed this comment because it is confusing. The goal is indeed known -- a more precise commen would say if the predicate in this atom was excluded or not. But the code itself says that pretty clearly, so any comment will be pretty redundant.

predicators/main.py
NishanthJKumar: interesting - we should probably fix that in CogMan later

predicators/structs.py
NishanthJKumar: what about 'oracle goal', or 'demonstrator goal'?

NishanthJKumar (approved):
LGTM! Good work! Just one relatively minor comment :)

predicators/structs.py
ashayathalye: Well, the alternative goal is not for the oracle, it's for the agent. Calling it `agent_goal` is a possibility, but I think that is still confusing because then someone might think the normal goal is not for the agent. 

I think let's keep it `alt_goal` for now, and we can update it later if we want to.

predicators/main.py
ashayathalye: I don't know if there's anything to fix -- not sure. 

Currently, we only have CogMan for the test tasks, not for the train tasks. That seems to be by design, because CogMan is meant to manage episodes of environment interaction? 

I don't know if it makes sense to have the perceiver be external to the agent's cognitive manager for the train tasks but not for the test tasks. I feel like the agent is a black box and the input to it should be the environment tasks in both the train and test scenarios, esp if we're practically using the same perceiver to turn the environment tasks into tasks in both cases. 

Maybe it would make more sense if we used a different perceiver to create the train tasks, than the perceiver the agent was gonna use at test time.

PR #1665, opened 2024-06-06
Created rl_bridge_approach and rendering for grid_row_door

matrixbalto:
Set door position and light levels in grid_row_door to be fixed for consistency across train and test. Note that rl_bridge_approach does not have timeout feature

https://github.com/Learning-and-Intelligent-Systems/predicators/assets/102700848/f762019b-4f37-40d4-9013-c9e87cd02057

### DIFF
--- a/predicators/approaches/maple_q_approach.py
+++ b/predicators/approaches/maple_q_approach.py
@@ -61,10 +61,11 @@ class MapleQApproach(OnlineNSRTLearningApproach):
     def get_name(cls) -> str:
         return "maple_q"
 
+    # pylint: disable=arguments-differ
     def _solve(self,
                task: Task,
                timeout: int,
-               train_or_test: str = "train") -> Callable[[State], Action]:
+               train_or_test: str = "") -> Callable[[State], Action]:
 
         def _option_policy(state: State) -> _Option:
             return self._q_function.get_option(
@@ -112,6 +113,15 @@ class MapleQApproach(OnlineNSRTLearningApproach):
                      annotations: Optional[List[Any]]) -> None:
         # Start by learning NSRTs in the usual way.
         super()._learn_nsrts(trajectories, online_learning_cycle, annotations)
+        if CFG.approach == "active_sampler_learning":
+            # Check the assumption that operators and options are 1:1.
+            # This is just an implementation convenience.
+            assert len({
+                nsrt.option  # pragma: no cover.
+                for nsrt in self._nsrts
+            }) == len(self._nsrts)  # pragma: no cover.
+            for nsrt in self._nsrts:  # pragma: no cover.
+                assert nsrt.option_vars == nsrt.parameters  # pragma: no cover.
         # On the first cycle, we need to register the ground NSRTs, goals, and
         # objects in the Q function so that it can define its inputs.
         if not online_learning_cycle:
--- a/predicators/approaches/maple_q_approach.py
+++ b/predicators/approaches/maple_q_approach.py
@@ -61,14 +61,18 @@ class MapleQApproach(OnlineNSRTLearningApproach):
     def get_name(cls) -> str:
         return "maple_q"
 
-    def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+    def _solve(self,
+               task: Task,
+               timeout: int,
+               train_or_test: str = "train") -> Callable[[State], Action]:
 
         def _option_policy(state: State) -> _Option:
             return self._q_function.get_option(
                 state,
                 task.goal,
                 num_samples_per_ground_nsrt=CFG.
-                active_sampler_learning_num_samples)
+                active_sampler_learning_num_samples,
+                train_or_test=train_or_test)
 
         return utils.option_policy_to_policy(
             _option_policy, max_option_steps=CFG.max_num_steps_option_rollout)
@@ -108,11 +112,6 @@ class MapleQApproach(OnlineNSRTLearningApproach):
                      annotations: Optional[List[Any]]) -> None:
         # Start by learning NSRTs in the usual way.
         super()._learn_nsrts(trajectories, online_learning_cycle, annotations)
-        # Check the assumption that operators and options are 1:1.
-        # This is just an implementation convenience.
-        assert len({nsrt.option for nsrt in self._nsrts}) == len(self._nsrts)
-        for nsrt in self._nsrts:
-            assert nsrt.option_vars == nsrt.parameters
         # On the first cycle, we need to register the ground NSRTs, goals, and
         # objects in the Q function so that it can define its inputs.
         if not online_learning_cycle:
--- a/predicators/envs/grid_row.py
+++ b/predicators/envs/grid_row.py
@@ -180,7 +180,7 @@ class GridRowEnv(BaseEnv):
 class GridRowDoorEnv(GridRowEnv):
     """Simple variant on GridRow where there is also a door."""
 
-    #Properties for rendering
+    # Properties for rendering
     cell_width = 1
     robot_width = 0.5
     robot_height = 0.75
@@ -286,13 +286,14 @@ class GridRowDoorEnv(GridRowEnv):
                    rng: np.random.Generator) -> List[EnvironmentTask]:
         # There is only one goal in this environment: to turn the light on.
         goal = {GroundAtom(self._LightOn, [self._light])}
-        # The only variation in the initial state is the light target level.
         tasks: List[EnvironmentTask] = []
         while len(tasks) < num:
             state_dict = {
                 self._robot: {
                     "x": 0.5,
                 },
+                # Note: light level and door locations are fixed for now
+                # in order to maintain consistency across train and test
                 self._light: {
                     "x": len(self._cells) - 0.5,
                     "level": 0.0,
--- a/predicators/envs/grid_row.py
+++ b/predicators/envs/grid_row.py
@@ -3,6 +3,7 @@
 from typing import List, Optional, Sequence, Set
 
 import matplotlib
+import matplotlib.pyplot as plt
 import numpy as np
 from gym.spaces import Box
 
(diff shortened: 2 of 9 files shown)


predicators/approaches/maple_q_approach.py
```diff
@@ -108,11 +112,6 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
                      annotations: Optional[List[Any]]) -> None:
         # Start by learning NSRTs in the usual way.
         super()._learn_nsrts(trajectories, online_learning_cycle, annotations)
-        # Check the assumption that operators and options are 1:1.
-        # This is just an implementation convenience.
-        assert len({nsrt.option for nsrt in self._nsrts}) == len(self._nsrts)
-        for nsrt in self._nsrts:
-            assert nsrt.option_vars == nsrt.parameters
```
NishanthJKumar, line 115: we definitely no longer want this assumption, but we also don't want silent failures when the old code for our Planning to PRactice paper is run. Could you put back in this code, but put it under an `if` statement that says `if CFG.approach == "active_sampler_learning"`?

predicators/envs/grid_row.py
```diff
@@ -179,6 +180,12 @@ def _Adjacent_holds(self, state: State, objects: Sequence[Object]) -> bool:
 class GridRowDoorEnv(GridRowEnv):
     """Simple variant on GridRow where there is also a door."""
 
+    #Properties for rendering
```
NishanthJKumar, line 183: nit: add a space after '#' whenever you write a comment

predicators/envs/grid_row.py
```diff
@@ -193,6 +200,69 @@ def __init__(self, use_gui: bool = True) -> None:
     def get_name(cls) -> str:
         return "grid_row_door"
 
+    def render_state_plt(
```
NishanthJKumar, line 203: nicely done! Can you post an example video from this in the description of the PR?

predicators/envs/grid_row.py
```diff
@@ -226,10 +296,10 @@ def _get_tasks(self, num: int,
                 self._light: {
                     "x": len(self._cells) - 0.5,
                     "level": 0.0,
-                    "target": rng.uniform(0.5, 1.0),
+                    "target": 0.75,
                 },
                 self._door: {
-                    "x": rng.choice(range(2, len(self._cells))) - 0.5,
+                    "x": len(self._cells) // 2 + 0.5,
```
NishanthJKumar, line 302: maybe add a comment saying that the light target and door location are fixed for now, but in the future, we'd like to relax this assumption

tests/envs/test_grid_row.py
```diff
@@ -243,5 +244,6 @@ def test_grid_row_door():
         assert option.terminal(state)
         if all(a.holds(state) for a in ground_nsrt.add_effects):
             break
+    env.render_state(state, task)
```
NishanthJKumar, line 247: nit: instead of just calling `render_state`, you should assert that `render_state` isn't none and actually returns what it says it's returning (i.e. an array of some kind).

predicators/ml_models.py
```diff
@@ -1354,6 +1354,11 @@ def __init__(self,
         self._num_ground_nsrts = 0
         self._replay_buffer: Deque[MapleQData] = deque(
             maxlen=self._replay_buffer_max_size)
+        self._epsilon = 0.5
+        self._min_epsilon = 0.05
```
NishanthJKumar, line 1358: it's useful to have this, but instead of hard-coding this here, you should make these command-line flags by using `CFG`.

predicators/ml_models.py
```diff
@@ -1375,9 +1380,12 @@ def get_option(self,
                    state: State,
                    goal: Set[GroundAtom],
                    num_samples_per_ground_nsrt: int,
-                   epsilon: float = 0.0) -> _Option:
+                   train_or_test: str = "test") -> _Option:
```
NishanthJKumar, line 1383: so ideally, we wouldn't delete this functionality of being able to directly specify an `epsilon` value. Could we support both by instead introducing some class variable called `use_epsilon_annealing` which will be a boolean set by command line args. Then here, we can check whether `use_epsilon_annealing` is True, and if so, we can use your logic, otherwise we can take the epsilon value that's being passed in?

predicators/approaches/maple_q_approach.py
```diff
@@ -61,14 +61,18 @@ def __init__(self, initial_predicates: Set[Predicate],
     def get_name(cls) -> str:
         return "maple_q"
 
-    def _solve(self, task: Task, timeout: int) -> Callable[[State], Action]:
+    def _solve(self,
+               task: Task,
+               timeout: int,
+               train_or_test: str = "train") -> Callable[[State], Action]:
```
NishanthJKumar, line 67: can we not have a default value here? I'm nervous about that: whoever is calling this method should explicitly specify this.

predicators/approaches/maple_q_approach.py
```diff
@@ -61,14 +61,18 @@ def __init__(self, initial_predicates: Set[Predicate],
...
 
         def _option_policy(state: State) -> _Option:
             return self._q_function.get_option(
                 state,
                 task.goal,
                 num_samples_per_ground_nsrt=CFG.
-                active_sampler_learning_num_samples)
+                active_sampler_learning_num_samples,
+                train_or_test=train_or_test)
```
NishanthJKumar, line 75: nice - very clean way to pass around this `train_or_test` info!

tests/approaches/test_bridge_policy_approach.py
```diff
@@ -233,3 +234,61 @@ def _mock_human_demonstratory_policy(*args, **kwargs):
...
+    }
+    utils.reset_config(args)
+    env = get_or_create_env(CFG.env)
+    train_tasks = [t.task for t in env.get_train_tasks()]
+    approach = RLBridgePolicyApproach(env.predicates,
+                                      get_gt_options(env.get_name()),
+                                      env.types, env.action_space, train_tasks)
+    assert approach.get_name() == "rl_bridge_policy"
+    assert approach.is_learning_based
+    # interaction_requests = approach.get_interaction_requests()
```
NishanthJKumar, line 263: remove comment if you're no longer using this line

tests/test_ml_models.py
```diff
@@ -475,10 +475,12 @@ def test_maple_q_function():
...
     model.set_grounding(objects, [task.goal], ground_nsrts)
     # Test getting a random option.
-    sampled_option = model.get_option(task.init, task.goal, 1, epsilon=1.0)
+    sampled_option = model.get_option(task.init, task.goal, 1, \
+                                      train_or_test="test")
     assert sampled_option.initiable(task.init)
     # Test getting a non-random option.
-    sampled_option = model.get_option(task.init, task.goal, 1, epsilon=0.0)
+    sampled_option = model.get_option(task.init, task.goal, 1, \
+                                      train_or_test="test")
```
NishanthJKumar, line 483: make sure to modify this once you've made the changes to the `get_option` function

tests/approaches/test_bridge_policy_approach.py
```diff
@@ -233,3 +234,61 @@ def _mock_human_demonstratory_policy(*args, **kwargs):
...
+    gets_reward = False
+    for (_, _, _, _, reward, _) in approach.mapleq._q_function._replay_buffer:  # pylint: disable=protected-access
+        if reward > 0:
+            gets_reward = True
+    assert gets_reward
+
+    assert approach.learn_from_interaction_results([]) is None
+
+    # Evaluate approach after 5 online learning cycles.
+    # We should have learned correct policy by now
```
NishanthJKumar, line 292: comment seems outdated?

tests/approaches/test_bridge_policy_approach.py
```diff
@@ -233,3 +234,61 @@ def _mock_human_demonstratory_policy(*args, **kwargs):
...
+    results = _run_testing(env, cogman)
+    results["num_online_transitions"] = num_online_transitions
```
NishanthJKumar, line 294: what's the point of this line? is it doing something important?

tests/approaches/test_bridge_policy_approach.py
```diff
@@ -233,3 +234,61 @@ def _mock_human_demonstratory_policy(*args, **kwargs):
...
+    if not CFG.load_approach or CFG.restart_learning:
+        cogman.learn_from_interaction_results(interaction_results)
+    # We should be adding more data to replay buffer
+    assert len(approach.mapleq._q_function._replay_buffer) > 0  # pylint: disable=protected-access
+    # Test that reward is positive for some trial
+    gets_reward = False
+    for (_, _, _, _, reward, _) in approach.mapleq._q_function._replay_buffer:  # pylint: disable=protected-access
+        if reward > 0:
+            gets_reward = True
+    assert gets_reward
```
NishanthJKumar, line 287: nice way of having the test be fast but also test what you want :D

tests/approaches/test_bridge_policy_approach.py
```diff
@@ -233,3 +234,61 @@ def _mock_human_demonstratory_policy(*args, **kwargs):
...
+
+    assert approach.learn_from_interaction_results([]) is None
+
```
NishanthJKumar, line 290: nit: delete the new lines before and after this assertion

NishanthJKumar (commented):
Good work! Left a few comments that hopefully shouldn't be too hard to address!

predicators/approaches/maple_q_approach.py
matrixbalto: oh it looks like we need a default in order for the type checking to work (for it to be compatible with supertype BaseApproach)

predicators/ml_models.py
matrixbalto: Turns out when the epsilon parameter in get_option here is kept, we would have to have default train_or_test = "test" in rl_bridge's solve function. The tricky part is we want eps annealing to be True for rl_bridge, but then we'll override epsilon to be whatever the annealing epsilon is (which is not 0), since setting train_or_test = "" in Maple Q will overwrite the train_or_test: str = "test" here. We should able to specify epsilon in the command line argument though so hopefully that's ok?

NishanthJKumar: Yup, that should be okay! I think if an epsilon is set but a train or test value is also set, we should throw an error or something here just to make sure there isn't an unexpected silent failure

NishanthJKumar (approved):
Phenomenal work! Just left a minor comment

PR #1666, opened 2024-06-13
Enable predicate invention with both geometric and VLM predicates

ashayathalye:
### Summary
This basically just calls both the method to generate predicates from a grammar and the method to generate predicates from a VLM and then combines the atom datasets produced by both. 

This does not yet work "well" in BurgerEnv yet because we don't get the predicates we want in the grammar (we'll have to tweak the grammar a bit, and also be careful about what predicates we include/exclude). But the pipeline works in that the atom datasets are combined properly into something with both geometric and VLM ground atoms.

An important change in this is that we now make it possible to replace goal predicates (e.g. goal predicates that we might expect the demonstrator to use) with agent-specific goal predicates -- which is relevant when we want the agent to invent the corresponding goal predicate that the demonstrator used. Before, this was partially done by making it so that tasks' goals could be reframed in terms of alternative goal predicates, but this PR also makes this happen for the predicates that get passed into an approach's initial/known predicates. 

### Tests
At the time of this writing, (2024-06-14), there is no coverage for `offline_data_method="demo_with_vlm_imgs"` and `offline_data_method="saved_vlm_img_demos_folder"` because a VLM would need to be instantiated and called and that would be slow (there's already a comment in the code about this) -- the current code tests this separately and ignores the coverage issue?

### DIFF
--- a/predicators/approaches/bridge_policy_approach.py
+++ b/predicators/approaches/bridge_policy_approach.py
@@ -288,7 +288,7 @@ class BridgePolicyApproach(OracleApproach):
                 states = [traj.states[0]]
                 atoms = atom_traj
             else:
-                states = utils.segment_trajectory_to_state_sequence(
+                states = utils.segment_trajectory_to_start_end_state_sequence(
                     segmented_traj)
                 atoms = utils.segment_trajectory_to_atoms_sequence(
                     segmented_traj)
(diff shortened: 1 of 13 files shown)


NishanthJKumar (approved):
LGTM! Great work!

PR #1667, opened 2024-06-16
CallPlanner

matrixbalto:
Added CallPlanner action to RL_Bridge_Approach.

### DIFF
--- a/predicators/approaches/bridge_policy_approach.py
+++ b/predicators/approaches/bridge_policy_approach.py
@@ -42,6 +42,11 @@ Learned bridge policy in stick button:
 
 import logging
 import time
+import os
+import sys
+
+import numpy as np
+
 from typing import Callable, List, Optional, Sequence, Set
 
 from gym.spaces import Box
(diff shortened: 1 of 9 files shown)


predicators/approaches/bridge_policy_approach.py
```diff
@@ -486,13 +569,183 @@ def learn_from_interaction_results(
...
+            nsrt_plan,
+            goal=task.goal,
+            rng=self._rng,
+            necessary_atoms_seq=atoms_seq)
+
+    def _run_task_plan( #pylint: disable=dangerous-default-value
+        self, task: Task, nsrts: Set[NSRT],
+        preds: Set[Predicate], timeout: float, seed: int, \
+            init: Set[GroundAtom] = set(), **kwargs: Any
+    ) -> Tuple[List[_GroundNSRT], List[Set[GroundAtom]], Metrics]:
```
NishanthJKumar, line 729: okay - i'm not fully sure why you had to override this function here? Why not use the function from `planning.py`?

matrixbalto: I needed to override it to include the new init variable so we can start the plan from any state.

NishanthJKumar:
Closing in favor of #1670

PR #1669, opened 2024-06-18
Add task variation to BurgerEnv

ashayathalye:
- Changes BurgerEnv from a 4x4 grid to a 5x5 grid. 
- Adds variation in train and test tasks. All objects are placed on edges such that the robot cannot be adjacent to more than one object to avoid ignore affect issues from the `Adjacent` predicate.


NishanthJKumar (approved):
LGTM - phenomenal work!

PR #1670, opened 2024-06-18
Added CallPlanner action to RL_Bridge_Policy

matrixbalto:
A CallPlanner NSRT was added to RLBridgePolicyApproach, which does not change the low-level environment state but transfers control back to the planner. Bridge policy can use this NSRT to call the planner, allowing the method to resume online bilevel planning after the bridge policy is executed.

### DIFF
--- a/predicators/approaches/bridge_policy_approach.py
+++ b/predicators/approaches/bridge_policy_approach.py
@@ -56,7 +56,7 @@ from predicators.nsrt_learning.segmentation import segment_trajectory
 from predicators.option_model import _OptionModelBase
 from predicators.settings import CFG
 from predicators.structs import NSRT, Action, Array, BridgeDataset, \
-    DefaultState, DemonstrationQuery, DemonstrationResponse, GroundAtom, \
+    DefaultState, DemonstrationQuery, DemonstrationResponse, \
     InteractionRequest, InteractionResult, LiftedAtom, LowLevelTrajectory, \
     Object, ParameterizedOption, Predicate, Query, State, Task, Type, \
     Variable, _GroundNSRT, _Option
--- a/predicators/approaches/bridge_policy_approach.py
+++ b/predicators/approaches/bridge_policy_approach.py
@@ -549,8 +549,6 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
         # for efficiency.
         if not results:
             return None
-        all_states = []
-        all_actions = []
         policy_logs = self._policy_logs
         for i in range(len(results)):
             result = results[i]
@@ -568,8 +566,6 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
             mapleq_states.append(result.states[-1])
             new_traj = LowLevelTrajectory(mapleq_states, mapleq_actions)
             self._trajs.append(new_traj)
-            all_states.extend(mapleq_states)
-            all_actions.extend(mapleq_actions)
             policy_logs = policy_logs[len(result.states) - 1:]
 
         self.mapleq.get_interaction_requests()
--- a/predicators/approaches/bridge_policy_approach.py
+++ b/predicators/approaches/bridge_policy_approach.py
@@ -382,7 +382,7 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
         self.mapleq=MapleQApproach(self._get_current_predicates(), \
                                    self._initial_options, self._types, \
                                     self._action_space, self._train_tasks)
-        self._current_control = ""
+        self._current_control: Optional[str] = None
         option_policy = self._get_option_policy_by_planning(
             self._train_tasks[0], CFG.timeout)
         self._current_policy = utils.option_policy_to_policy(
@@ -391,7 +391,7 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
             raise_error_on_repeated_state=True,
         )
         self._bridge_called_state = State(data={})
-        self._policy_logs: List[str] = []
+        self._policy_logs: List[Optional[str]] = []
 
     def _Can_plan(self, state: State, _: Sequence[Object]) -> bool:
         if (self.mapleq._q_function._vectorize_state(state) !=  # pylint: disable=protected-access
@@ -428,17 +428,8 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
         ignore_effects: Set[Predicate] = set()
         call_planner_nsrt = NSRT("CallPlanner", parameters, preconditions,
                                  add_effects, delete_effects, ignore_effects,
-                                 option, option_vars, self.planner_sampler)
+                                 option, option_vars, utils.null_sampler)
         return call_planner_nsrt
-
-    def planner_sampler(self, state: State, goal: Set[GroundAtom],
-                        rng: np.random.Generator,
-                        objs: Sequence[Object]) -> Array:
-        """sampler for CallPlanner Option."""
-        del state, goal, objs  # unused
-        # Note: just return a random value from -1 to 1
-        return np.array([rng.uniform(0, len(self._train_tasks))],
-                        dtype=np.float32)
 
     @classmethod
     def get_name(cls) -> str:
@@ -498,13 +489,7 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
             self._maple_initialized = True
             self._init_nsrts()
 
-        # Prevent infinite loops by detecting if the bridge policy is called
-        # twice with the same state.
-        last_bridge_policy_state = DefaultState
-
         def _policy(s: State) -> Action:
-            nonlocal last_bridge_policy_state
-
             # Normal execution. Either keep executing the current option, or
             # switch to the next option if it has terminated.
             try:
--- a/predicators/approaches/bridge_policy_approach.py
+++ b/predicators/approaches/bridge_policy_approach.py
@@ -375,7 +375,7 @@ class RLBridgePolicyApproach(BridgePolicyApproach):
             "CallPlanner",
             types=None,
             policy=self.call_planner_policy,
-            params_space=Box(0, len(train_tasks), (1, )),
+            params_space=Box(low=np.array([]), high=np.array([]), shape=(0,)),
         )
         initial_options.add(self.CallPlanner)
         self._initial_options = initial_options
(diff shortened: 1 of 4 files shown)


predicators/ml_models.py
```diff
@@ -1540,6 +1542,7 @@ def _vectorize_option(self, option: _Option) -> Array:
             and tuple(n.objects) == tuple(option.objects)
         ]
         assert len(matches) == 1
+
```
NishanthJKumar, line 1545: revert

predicators/approaches/maple_q_approach.py
```diff
@@ -124,7 +124,7 @@ def _learn_nsrts(self, trajectories: List[LowLevelTrajectory],
                 assert nsrt.option_vars == nsrt.parameters  # pragma: no cover.
         # On the first cycle, we need to register the ground NSRTs, goals, and
         # objects in the Q function so that it can define its inputs.
-        if not online_learning_cycle:
+        if not online_learning_cycle and CFG.approach != "rl_bridge_policy":
```
NishanthJKumar, line 127: can you adda  comment here explaining why we need ` CFG.approach != "rl_bridge_policy"`?

predicators/approaches/bridge_policy_approach.py
```diff
@@ -54,11 +55,11 @@
 from predicators.nsrt_learning.segmentation import segment_trajectory
 from predicators.option_model import _OptionModelBase
 from predicators.settings import CFG
-from predicators.structs import NSRT, Action, BridgeDataset, DefaultState, \
-    DemonstrationQuery, DemonstrationResponse, InteractionRequest, \
-    InteractionResult, LowLevelTrajectory, ParameterizedOption, Predicate, \
-    Query, State, Task, Type, _GroundNSRT, _Option
-from predicators.utils import OptionExecutionFailure
```
NishanthJKumar, line 61: why did you get rid of this `OptionExecutionFailure` import?

predicators/approaches/bridge_policy_approach.py
```diff
@@ -365,10 +366,79 @@ def __init__(self,
...
+                                 add_effects, delete_effects, ignore_effects,
+                                 option, option_vars, self.planner_sampler)
+        return call_planner_nsrt
+
+    def planner_sampler(self, state: State, goal: Set[GroundAtom],
+                        rng: np.random.Generator,
+                        objs: Sequence[Object]) -> Array:
+        """sampler for CallPlanner Option."""
+        del state, goal, objs  # unused
+        # Note: just return a random value from -1 to 1
```
NishanthJKumar, line 439: could you just make the `call_planner_nsrt` not take in any continuous params, and then use a `NullSampler` (which I think we've implemented somewhere in the codebase?)

predicators/approaches/bridge_policy_approach.py
```diff
@@ -365,10 +366,79 @@ def __init__(self,
...
+            types=None,
+            policy=self.call_planner_policy,
+            params_space=Box(0, len(train_tasks), (1, )),
+        )
+        initial_options.add(self.CallPlanner)
+        self._initial_options = initial_options
         self.mapleq=MapleQApproach(self._get_current_predicates(), \
                                    self._initial_options, self._types, \
                                     self._action_space, self._train_tasks)
+        self._current_control = ""
```
NishanthJKumar, line 385: instead of an empty string, set this to None and then set the type of this variable to `Optional[str]`. In general, it's a good design pattern to use `None` when you don't have a value for a variable; makes errors easier to spot

predicators/approaches/bridge_policy_approach.py
```diff
@@ -410,48 +482,49 @@ def _solve(self,
...
+            self._maple_initialized = True
+            self._init_nsrts()
 
         # Prevent infinite loops by detecting if the bridge policy is called
         # twice with the same state.
         last_bridge_policy_state = DefaultState
 
         def _policy(s: State) -> Action:
-            nonlocal current_control, current_policy, last_bridge_policy_state
+            nonlocal last_bridge_policy_state
```
NishanthJKumar, line 506: where is this variable getting used?

predicators/approaches/bridge_policy_approach.py
```diff
@@ -365,10 +366,79 @@ def __init__(self,
...
+        self._current_control = ""
+        option_policy = self._get_option_policy_by_planning(
+            self._train_tasks[0], CFG.timeout)
+        self._current_policy = utils.option_policy_to_policy(
+            option_policy,
+            max_option_steps=CFG.max_num_steps_option_rollout,
+            raise_error_on_repeated_state=True,
+        )
+        self._bridge_called_state = State(data={})
+        self._policy_logs: List[str] = []
```
NishanthJKumar, line 394: as far as i can tell, you're only ever setting this value, never reading from it. What's the point of this? Are you logging/using it somewhere?

NishanthJKumar (commented):
looking good! Left some comments to be addressed; we can do another round of review after this.

Also, could you update the PR to be slightly more descriptive? In particular, could you say something about how `call_planner` has been added as an NSRT that the bridge policy can call?

predicators/approaches/bridge_policy_approach.py
matrixbalto: self._bridge_called_state is used for the precondition of CallPlanner, _Can_plan, which is False if self._bridge_called_state is the same as our current state, and True if they are different.

predicators/approaches/bridge_policy_approach.py
matrixbalto: when I set it to None I get an error from the type checker, should I just set it to "planner"?

predicators/approaches/bridge_policy_approach.py
NishanthJKumar: sorry - i meant `self._policy_logs`

predicators/approaches/bridge_policy_approach.py
```diff
@@ -410,48 +473,43 @@ def _solve(self,
...
+                action = self._current_policy(s)
+                if train_or_test == "train":
+                    self._policy_logs.append(self._current_control)
                 return action
-            except OptionExecutionFailure as e:
-                failed_option = e.info["last_failed_option"]
-                if failed_option is not None:
-                    all_failed_options.append(failed_option)
+            except utils.OptionExecutionFailure:
+                print("failed", self._current_control)
```
NishanthJKumar, line 501: in general, we've moved away from using `print` statements in the codebase. Instead, we use `logging.debug` or `logging.info` (you can grep for these and see how they're used elsewhere)

NishanthJKumar (commented):
almost there!

predicators/approaches/bridge_policy_approach.py
matrixbalto: `self._policy_logs` is used in `learn_from_interaction_results` to determine which actions were executed by bridge and which actions were executed by planner, so that bridge policy only learns from the actions that itself takes

predicators/approaches/bridge_policy_approach.py
```diff
@@ -486,13 +551,28 @@ def learn_from_interaction_results(
...
-            all_states.extend(result.states)
-            all_actions.extend(result.actions)
+            all_states.extend(mapleq_states)
+            all_actions.extend(mapleq_actions)
+            policy_logs = policy_logs[len(result.states) - 1:]
 
         self.mapleq.get_interaction_requests()
         self.mapleq._learn_nsrts(self._trajs, 0, [] * len(self._trajs))  # pylint: disable=protected-access
+        self._policy_logs = []
         return None
```
NishanthJKumar, line 578: could you remove the `return None` here?

predicators/approaches/bridge_policy_approach.py
```diff
@@ -486,13 +551,28 @@ def learn_from_interaction_results(
...
+                if policy_log[j] == "bridge"
+                or policy_log[max(j - 1, 0)] == "bridge"
+            ]
+            mapleq_states.append(result.states[-1])
+            new_traj = LowLevelTrajectory(mapleq_states, mapleq_actions)
             self._trajs.append(new_traj)
-            all_states.extend(result.states)
-            all_actions.extend(result.actions)
+            all_states.extend(mapleq_states)
+            all_actions.extend(mapleq_actions)
```
NishanthJKumar, line 572: are we using `all_states` or `all_actions` anywhere???

predicators/approaches/bridge_policy_approach.py
```diff
@@ -486,13 +551,28 @@ def learn_from_interaction_results(
...
-        for result in results:
-            new_traj = LowLevelTrajectory(result.states, result.actions)
+        policy_logs = self._policy_logs
+        for i in range(len(results)):
+            result = results[i]
+            policy_log = policy_logs[:len(result.states[:-1])]
+            mapleq_states = [
+                state for j, state in enumerate(result.states[:-1])
+                if policy_log[j] == "bridge"
+                or policy_log[max(j - 1, 0)] == "bridge"
```
NishanthJKumar, line 561: hold on - the `or` part of this statement is a bit concerning. If `j` starts at `0` and `j-1` is `-1`, wouldn't we index the list with `-1` as the first part of the `if` here, which is not what we want? Could you explain why you need this logic more?

NishanthJKumar (commented):
looking good; just a few more things to make sure it's as clean as possible before merge!

predicators/approaches/bridge_policy_approach.py
matrixbalto: the type checker gets mad without the return None :,)

predicators/approaches/bridge_policy_approach.py
matrixbalto: This is because of the way I set `self._current_control`. It ends up setting the action CallPlanner as if it was executed by the planner not the bridge, so I did this to account for the CallPlanner actions (if the action before this was done by bridge, then we set this one to bridge since I think this only happens CallPlanner is used)

matrixbalto: This is probably a bit bad though, lmk if you see a better way to do this

predicators/approaches/bridge_policy_approach.py
NishanthJKumar: oh gotcha, nw

predicators/approaches/bridge_policy_approach.py
NishanthJKumar: wait - but `max(-1, 0)` is `0`, so you won't look at the state right before it?

matrixbalto: if j=0, then it doesn't matter since on the first time step it's always the planner's control

NishanthJKumar: ah okay, i think i get it. Please add a comment to the code explaining this in case we need to revisit in the future

NishanthJKumar (approved):
LGTM! Great work! Just remember to add a comment about the indexing for checking the policy logs

PR #1671, opened 2024-06-24
Fix oracle approach on test tasks when an environment defines alternative goals and agent-specific goal predicates

ashayathalye:
Previously, in `main.py`, we 
- Replaced goal predicates with agent-specific goal predicates whenever the environment defined these, and these were passed into the approach's initial predicates (which messes up oracle's ground truth operators)
- Replaced task goals with alternative goals whenever the environment defined these (but oracle's ground truth operators won't usually be in terms of the alternative goals)
both of which caused oracle approach to not work at *test* time in BurgerEnv (it worked fine at train time). 

Now we make these two replacements only happen when the approach is not oracle. 

For example, now `python predicators/main.py --env burger --approach oracle --seed 0 --num_train_tasks 1 --num_test_tasks 1 --option_model_terminate_on_repeat False --sesame_max_skeletons_optimized 200 --timeout 80 --sesame_max_samples_per_step 1  --sesame_task_planner fdopt --bilevel_plan_without_sim True`
solves the test task, which was not happening previously.

### DIFF
--- a/predicators/main.py
+++ b/predicators/main.py
@@ -99,12 +99,16 @@ def main() -> None:
         env)
     # The known predicates are passed into the approach and into dataset
     # creation. In some cases, like when inventing geometric and VLM predicates,
-    # we want to hide certain goal predicates from the agent by replacing them
-    # with agent-specific goal predicates that the environment defines.
-    # Note that inside dataset creation, the known predicates are only used to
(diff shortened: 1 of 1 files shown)


NishanthJKumar (approved):

PR #1672, opened 2024-06-24
Fix geo and VLM predicate invention's use of annotations during nsrt learning. 

ashayathalye:
Want to make sure the annotations (the values of the VLM-predicate atoms) are used when learning nsrts (specifically, operators) for the geo+VLM predicate invention offline data method.


NishanthJKumar (approved):
lgtm! sorry this is not tested/hard to find!

PR #1673, opened 2024-06-25
Adds nicer error message to fd dr-reachability case

NishanthJKumar:
This way we won't need to go on wild goose hunts again hopefully


ashayathalye (approved):
LGTM

PR #1680, opened 2024-06-27
Continuous Door Opening Actions and Fixed Light in GridRowDoor

matrixbalto:
Added continuous door opening actions, MoveKey and TurnKey, and set the TurnOnLight action to always turn on the light successfully.

### DIFF
--- a/predicators/envs/grid_row.py
+++ b/predicators/envs/grid_row.py
@@ -189,7 +189,8 @@ class GridRowDoorEnv(GridRowEnv):
     def __init__(self, use_gui: bool = True) -> None:
         super().__init__(use_gui)
         # type door with features ['x', and 'open']
-        self._door_type = Type("door", ["x", 'open'])
+        self._door_type = Type("door",
+                               ["x", "open", "target", "open1", "target1"])
         self._door = Object("door", self._door_type)
 
         self._DoorInCell = Predicate("DoorInCell",
@@ -241,7 +242,12 @@ class GridRowDoorEnv(GridRowEnv):
                 facecolor="white")
         ax.add_patch(light)
         door_pos = state.get(self._door, "x")
-        if state.get(self._door, "open"):
+        door_open = state.get(self._door, "open")
+        door_target = state.get(self._door, "target")
+        door_open1 = state.get(self._door, "open1")
+        door_target1 = state.get(self._door, "target1")
+        if (door_target - 0.1 <= door_open <= door_target + 0.1 \
+    and door_target1 - 0.1 <= door_open1 <= door_target1 + 0.1):
             door = plt.Rectangle((door_pos - self.cell_width / 2.0, 0),
                                  self.cell_width,
                                  self.cell_width,
@@ -279,8 +285,8 @@ class GridRowDoorEnv(GridRowEnv):
 
     @property
     def action_space(self) -> Box:
-        # dx, dlight, ddoor
-        return Box(-np.inf, np.inf, (3, ))
+        # dx, dlight, ddoor, ddoor1
+        return Box(-np.inf, np.inf, (4, ))
 
     def _get_tasks(self, num: int,
                    rng: np.random.Generator) -> List[EnvironmentTask]:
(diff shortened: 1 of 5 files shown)


predicators/envs/grid_row.py
```diff
@@ -325,27 +337,34 @@ def simulate(self, state: State, action: Action) -> State:
...
-            next_state.set(self._door, "open", 1.0)
+        if robot_cell == door_cell and not (door_target - 0.1 \
+                            <= door_open <= door_target + 0.1 \
+    and door_target1 - 0.1 <= door_open1 <= door_target1 + 0.1):
+            new_door_level = np.clip(
+                state.get(self._door, "open") + ddoor, 0.0, 1.0)
+            next_state.set(self._door, "open", new_door_level)
+            new_door1_level = np.clip(
+                state.get(self._door, "open1") + ddoor1, 0.0, 1.0)
+            next_state.set(self._door, "open1", new_door1_level)
```
NishanthJKumar, line 351: Wait, so is it true that the robot could technically open the door with just one action (by setting the door level and door1 level correctly simultaneously?

predicators/envs/grid_row.py
```diff
@@ -325,27 +337,34 @@ def simulate(self, state: State, action: Action) -> State:
...
         ]
         assert len(light_cells) == 1
         light_cell = light_cells[0]
-        if robot_cell == light_cell:
-            new_light_level = np.clip(
-                state.get(self._light, "level") + dlight, 0.0, 1.0)
-            next_state.set(self._light, "level", new_light_level)
+        if robot_cell == light_cell and dlight == 0.75 and self._LightOff_holds(
+                state, [self._light]):
+            next_state.set(self._light, "level", 0.75)
```
NishanthJKumar, line 361: wait - so is it now true that the `dlight` part of the action vector is effectively useless? If that's the case, I'd prefer you just modify the sampler to be correct instead of handling the light in this way in the transition function: modifying the sampler to always succeed is more straightforward and clear.

NishanthJKumar: i see you've also modified the sampler, so why do you also need this line in the transition function?

predicators/envs/grid_row.py
```diff
@@ -301,7 +307,10 @@ def _get_tasks(self, num: int,
                 },
                 self._door: {
                     "x": len(self._cells) // 2 + 0.5,
-                    "open": 0.0
+                    "open": 0.0,
+                    "target": 0.5,
+                    "open1": 0.0,
+                    "target1": 0.75
```
NishanthJKumar, line 313: could you name these with more informative names so it's easier to grok what they are + what's generally going on here? e.g. could you rename `open1` to `key_turn_angle` or something (that's what it corresponds to right?)

NishanthJKumar (commented):
Great work! Left a few comments and questions just to make sure I understand everything/it's as simple as it can be (to help with debugging!)

predicators/envs/grid_row.py
matrixbalto: No, there isn't an action that handles both of the levels at once so it's forced to open the door through two actions (one TurnKey action and one MoveKey action)

predicators/envs/grid_row.py
matrixbalto: Wait sorry I don't quite understand, don't we still need to change the light level and to do so we first need to check that the robot is currently in the light cell, if a TurnOnLight was called (dlight == 0.75)? I guess I could technically get rid of the self._LightOff_holds(state, [self._light]), but I think we'll need to check that the light is off when we generalize to multiple lights

tests/envs/test_grid_row.py
```diff
@@ -219,31 +232,18 @@ def test_grid_row_door():
...
-    for _ in range(100):
-        assert all(a.holds(state) for a in ground_nsrt.preconditions)
-        option = ground_nsrt.sample_option(state, set(), rng)
-        assert option.initiable(state)
-        action = option.policy(state)
-        state = env.simulate(state, action)
-        assert option.terminal(state)
-        if all(a.holds(state) for a in ground_nsrt.add_effects):
-            break
+    print(state)
```
NishanthJKumar, line 240: remove print!

NishanthJKumar (approved):
Great work! Left one final comment - please address before merging!

tests/envs/test_grid_row.py
matrixbalto: oops sorry about that!!

PR #1682, opened 2024-06-28
Update pour option init to allow oracle to run when exclude all predicates

yichao-liang:


yichao-liang:
I don't understand why if the yapf test failed...

NishanthJKumar (approved):
LGTM, thanks for doing this!

PR #1685, opened 2024-07-02
Kitchen goal complexity

NishanthJKumar:
modify test-time goals for certain tasks in kitchen to make sure they're different from training time


NishanthJKumar (commented):
LGTM!

PR #1686, opened 2024-07-02
Parallelize vlm labeling

ashayathalye:


predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -173,33 +174,64 @@ def _sample_vlm_atom_proposals_from_trajectories(
...
+
+    indexed_trajectories = list(enumerate(trajectories))
+
+    if CFG.grammar_search_parallelize_vlm_labeling:
+        logging.info("Labeling trajectories in parallel.")
+        with ThreadPoolExecutor() as executor:
+            for traj_txt_outputs in executor.map(label_function,
+                                                 indexed_trajectories):
+                output_labelled_atoms_txt_list.append(traj_txt_outputs)
+        return output_labelled_atoms_txt_list
```
NishanthJKumar, line 229: very cool! Is there any way to control the number of parallel processes we fork? Or will the system just fork as many as it can? Would be nice to introduce a command line arg that dictates this if possible, otherwise a comment explaining what's happening with the number of forks would be good!

NishanthJKumar (approved):
LGTM! Just left one relatively minor comment

predicators/datasets/generate_atom_trajs_with_vlm.py
ashayathalye: This is only multithreading, not multiprocessing. Since the thing we need to parallelize here is a network I/O thing basically (waiting for response back from google's servers), multiprocessing isn't particularly helpful -- multithreading is sufficient. I think we make 1 thread for each trajectory, and we can't do better than that.

PR #1688, opened 2024-07-04
No grammar search vlm invention baseline

NishanthJKumar:
This is the implementation for a baseline that just uses all the proposed predicates instead of trying to subselect!


ashayathalye (approved):
LGTM

PR #1689, opened 2024-07-14
Bowen

Jaraxxus-Me:
Hi Tom and Nishanth,
In this PR. I modified:
(1) GNN-shooting approach, we break each try when the time ">" timed out. Previously it is "<", so that only the last shoot works.
(2) Added "satellites-markov" env and its GT option model/NSRT, where the low-level will also have Markov Property.

To check this (in a quick way), use the command:
```
predicators/main.py --env satellites-markov --approach gnn_option_policy --seed 0 --make_demo_videos --gnn_num_epochs 500 --num_train_tasks 10
```

Please review and comment.

Best,
Bowen

### DIFF
--- a/predicators/approaches/gnn_option_policy_approach.py
+++ b/predicators/approaches/gnn_option_policy_approach.py
@@ -199,7 +199,12 @@ class GNNOptionPolicyApproach(GNNApproach):
         start_time = time.perf_counter()
         memory: Dict = {}  # optionally updated by predict()
         # Keep trying until the timeout.
+        tries = 0
+        all_num_act = 0
         while time.perf_counter() - start_time < timeout:
+            tries += 1
+            if tries > 2:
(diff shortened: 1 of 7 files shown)


NishanthJKumar:
Hi Bowen,

Thanks for doing this! A few requests to get this merged.

1. Could you split this into 2 separate PR's, according to (1) and (2) you've mentioned above
2. Could you follow our [instructions for contributing](https://github.com/Learning-and-Intelligent-Systems/predicators?tab=readme-ov-file#instructions-for-contributing) and make sure for each of these PR's our various checks are passed? Might be harder for the satellites PR, but I can help/provide pointers on that if necessary

Jaraxxus-Me:
Oh, I see. Sure, let me try. I'll reach out if I meet with problems!

PR #1693, opened 2024-07-23
Added DoorKnobs Environment!

matrixbalto:
Added a DoorKnobs environment to test RL_Bridge. Subclass of Doors environment, with a new doorknob object that the agent must turn a certain amount to open the door. The DoorKnobs environment also does not have certain predicates from the Doors environment.

### DIFF
--- a/predicators/envs/doors.py
+++ b/predicators/envs/doors.py
@@ -14,7 +14,7 @@ from predicators import utils
 from predicators.envs import BaseEnv
 from predicators.settings import CFG
 from predicators.structs import Action, EnvironmentTask, GroundAtom, Object, \
-    Predicate, State, Type
+    Predicate, State, Type, Array
 from predicators.utils import Rectangle, StateWithCache, _Geom2D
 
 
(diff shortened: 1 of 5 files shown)


predicators/envs/doors.py
```diff
@@ -729,3 +729,447 @@ def _get_open_door_target_value(mass: float, friction: float,
...
+
+class DoorKnobsEnv(DoorsEnv):
+    """A 2D navigation environment with obstacles, rooms, and doors."""
+    # Types
+    _robot_type = Type("robot", ["x", "y"])
+    _door_type = Type("door", ["x", "y", "theta", "mass", "friction", "open"])
+    _knob_type = Type("knob", ["x", "y", "theta", "rot", "target_rot", "open"])
+    _room_type = Type("room", ["x", "y"])
+    _obstacle_type = Type("obstacle", ["x", "y", "width", "height", "theta"])
+    open_door_thresh: ClassVar[float] = 0.1
```
NishanthJKumar, line 742: looks like a significant amount of this is copied from the `DoorsEnv` superclass above. We want to minimize code duplication - especially when subclassing - so please only override/change the types that are different from the superclass

predicators/envs/doors.py
```diff
@@ -729,3 +729,447 @@ def _get_open_door_target_value(mass: float, friction: float,
...
+        # Hyperparameters from CFG.
+        self._room_map_size = CFG.doors_room_map_size
+        self._min_obstacles_per_room = CFG.doors_min_obstacles_per_room
+        self._max_obstacles_per_room = CFG.doors_max_obstacles_per_room
+        self._min_room_exists_frac = CFG.doors_min_room_exists_frac
+        self._max_room_exists_frac = CFG.doors_max_room_exists_frac
+        # See note in _sample_initial_state_from_map().
+        self._task_id_count = itertools.count()
+        self._door_to_knob: Dict[Object, Object] = {}
+        self._open_door_target_value = CFG.doorknobs_target_value
```
NishanthJKumar, line 775: ditto for the comment of not repeating things from the superclass

predicators/envs/doors.py
```diff
@@ -729,3 +729,447 @@ def _get_open_door_target_value(mass: float, friction: float,
...
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "doorknobs"
+
+    def simulate(self, state: State, action: Action) -> State:
```
NishanthJKumar, line 781: there's a bunch of code that's being shared between this method and the same method in the superclass. Could you refactor slightly to introduce a helper function that handles the mechanics of computing the next state so that the main difference is in the mechanics of the door?

predicators/envs/doors.py
```diff
@@ -729,3 +729,447 @@ def _get_open_door_target_value(mass: float, friction: float,
...
+
+    def _generate_train_tasks(self) -> List[EnvironmentTask]:
+        self._room_map_size = CFG.doors_room_map_size
+        return self._get_tasks(num=CFG.num_train_tasks, rng=self._train_rng)
+
+    def _generate_test_tasks(self) -> List[EnvironmentTask]:
+        self._room_map_size = CFG.test_doors_room_map_size
+        return self._get_tasks(num=CFG.num_test_tasks, rng=self._test_rng)
+
+    @property
```
NishanthJKumar, line 822: in general with this and below, it looks like there's a lot of duplicated code from the superclass. Could you cut down on this and make sure you only override things that need to be changed? I think that'll make review + future code management a lot easier.

predicators/ground_truth_models/doors/nsrts.py
```diff
@@ -171,3 +171,131 @@ def open_door_sampler(state: State, goal: Set[GroundAtom],
         nsrts.add(move_through_door_nsrt)
 
         return nsrts
+
+
+class DoorknobsGroundTruthNSRTFactory(DoorsGroundTruthNSRTFactory):
```
NishanthJKumar, line 176: looks like there's substantial code duplication here as well; I think there should be a way to merge/simplify things: see the nsrts.py file for the cover env for an example of making one NSRT factory that handles a number of different variants of the same env

predicators/ground_truth_models/doors/options.py
```diff
@@ -293,3 +294,233 @@ def _get_position_in_doorway(cls, room: Object, door: Object,
             ty = (y0 + y1) / 2
             position_cache[(room, door)] = (tx, ty)
         return position_cache[(room, door)]
+
+
+class DoorknobsGroundTruthOptionFactory(DoorsGroundTruthOptionFactory):
```
NishanthJKumar, line 299: ditto w.r.t comment about ground truth nsrts factory; please look at the ground truth options factory for the `cover` env and try to use that pattern to minimize code duplication

NishanthJKumar (commented):
good work overall! left a number of comments: my main feedback is around code duplication: I think you should be able to minimize this, and that should make review + future maintenance much easier.

Also, could you update the PR description to mention why you had to add the `DoorKnob` as a separate object and not just use the state space from the `Doors` env?

predicators/envs/doors.py
```diff
@@ -729,3 +738,354 @@ def _get_open_door_target_value(mass: float, friction: float,
         # A made up complicated function.
         return np.tanh(target_rot) * (np.sin(mass) +
                                       np.cos(friction) * np.sqrt(mass))
+
+
+class DoorKnobsEnv(DoorsEnv):
+    """A 2D navigation environment with obstacles, rooms, and doors."""
+    # Types
+    _robot_type = Type("robot", ["x", "y"])
```
NishanthJKumar, line 746: i think you can just inherit the `robot`, `room` and `obstacle` types from the superclass?

predicators/envs/doors.py
```diff
@@ -729,3 +738,354 @@ def _get_open_door_target_value(mass: float, friction: float,
...
+            task_cache: Dict[str, Dict] = {
+                "static_geom": {},
+                "door_to_rooms": {},
+                "room_to_doors": {},
+                "door_to_doorway_geom": {},
+                "position_in_doorway": {},
+            }
+            state_with_cache = utils.StateWithCache(state.data, task_cache)
+            if not self.state_has_collision(state_with_cache):
+                return state_with_cache
```
NishanthJKumar, line 1002: these two functions look identical to the superclass? Can you not just inherit?

predicators/envs/doors.py
```diff
@@ -729,3 +738,354 @@ def _get_open_door_target_value(mass: float, friction: float,
...
+
+        mass, friction, _ = rng.uniform(0.0, 1.0, size=(3, ))
+        return {
+            "x": x,
+            "y": y,
+            "theta": theta,
+            "mass": mass,
+            "friction": friction,
+            "open": 0.0,  # always start out closed
+        }
```
NishanthJKumar, line 1065: this also looks identical to the superclass?

NishanthJKumar (commented):
better! left more comments on the main doors env file. Will look at `options.py` and provide some pointers (if I can) about how to avoid code duplication there

predicators/envs/doors.py
matrixbalto: yes! sorry I changed this but I guess my changes didn't save :,)

predicators/envs/doors.py
matrixbalto: I got rid of the original  "rot" and "target_rot" features. tbh I could also j add them back since it doesn't make much of a difference in the learning process

NishanthJKumar (approved):
LGTM! Great work!

PR #1696, opened 2024-07-28
Bowen gnn

Jaraxxus-Me:
Hi Tom and Nishanth,
Following your suggestions, This PR only changes:
- The GNN-shooting methods, `predicators/approaches/gnn_option_policy_approach.py`, in every try of shooting, we break when the time `>` timeout. Previously it was `<`, so that only the last try is effective, all previous tries will only shoot one action.

Another small note is that:
```
                if time.perf_counter() - start_time > timeout:
                    break
```
Needs to happen before:
```
                if num_act == 0:
                    break
```
Otherwise, the `test_gnn_option_policy_approach_special_cases()` won't cover the timeout break.

Please review, thanks.
I'll proceed with the satellite-markov env once this is passed and merged.

### DIFF
--- a/predicators/approaches/gnn_option_policy_approach.py
+++ b/predicators/approaches/gnn_option_policy_approach.py
@@ -201,10 +201,10 @@ class GNNOptionPolicyApproach(GNNApproach):
         # Keep trying until the timeout.
         tries: int = 0
         all_num_act: int = 0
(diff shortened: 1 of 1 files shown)


NishanthJKumar (approved):
lgtm - thanks for doing this!

PR #1702, opened 2024-08-04
VLM Open-Loop Approach

NishanthJKumar:
Includes a new VLM-based planning approach [based on recent work](https://arxiv.org/pdf/2311.17842).

TODO's remaining:
1. Add some kind of explanation of what the different continuous parameters are; because otherwise it's kind of hopeless/ludicrous maybe to ask the VLM to output them?
2. Potentially test entire approach.

### DIFF
--- a/predicators/approaches/llm_open_loop_approach.py
+++ b/predicators/approaches/llm_open_loop_approach.py
@@ -138,8 +138,8 @@ class LLMOpenLoopApproach(NSRTMetacontrollerApproach):
         option_plan_with_cont_params = utils.\
             parse_model_output_into_option_plan(
             llm_prediction, objects, self._types, self._initial_options, False)
-        option_plan = [(elem[0], elem[1])
-                       for elem in option_plan_with_cont_params]
+        option_plan = [(option, objs)
+                       for option, objs, _ in option_plan_with_cont_params]
         return option_plan
 
     def learn_from_offline_dataset(self, dataset: Dataset) -> None:
--- a/predicators/approaches/llm_open_loop_approach.py
+++ b/predicators/approaches/llm_open_loop_approach.py
@@ -31,10 +31,10 @@ Easier setting:
 """
 from __future__ import annotations
 
-import logging
 from typing import Collection, Dict, Iterator, List, Optional, Sequence, Set, \
     Tuple
 
+from predicators import utils
 from predicators.approaches import ApproachFailure
 from predicators.approaches.nsrt_metacontroller_approach import \
     NSRTMetacontrollerApproach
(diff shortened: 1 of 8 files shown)


ashayathalye:
@ explanation of continuous parameters to VLM -- even with that it'll probably be a terrible sampler? The Look Before You Leap paper doesn't have any parameterized skills -- maybe we just enforce this to only run on environments with given samplers or null-sampler parameterized skills.

predicators/approaches/vlm_planning_prompts/few_shot.txt
```diff
@@ -0,0 +1,38 @@
+You are highly skilled in robotic task planning, breaking down intricate and long-term tasks into distinct primitive actions.
+Consider the following skills a robotic agent can perform. Note that each of these skills takes the form of a `ParameterizedOption` and may have both discrete arguments (indicated by the `types` field, referring to objects of particular types),
+as well as continuous arguments (indicated by `params_space` field, which is formatted as `Box([<param1_lower_bound>, <param2_lower_bound>, ...], [<param1_upper_bound>, <param2_upper_bound>, ...], (<number_of_params>,), <datatype_of_all_params>)`).
+
+{options}
+
+
+Here are a number of example demonstrations of solving a range of goals from an expert demonstrator. For each of these examples, you are provided 
+with the goal the demonstrator was trying to achieve, the plan (i.e., sequence of skills with discrete and continuous arguments specified), as well
+as images corresponding to the state of the environment at every timestep (starting at 0, when skill 0 is executed). Note that the state images
```
ashayathalye, line 10: "skill 0" could be interpreted as the first skill listed earlier in the prompt, rather than the first skill in the plan.

predicators/approaches/vlm_planning_prompts/few_shot.txt
```diff
@@ -0,0 +1,38 @@
+You are highly skilled in robotic task planning, breaking down intricate and long-term tasks into distinct primitive actions.
+Consider the following skills a robotic agent can perform. Note that each of these skills takes the form of a `ParameterizedOption` and may have both discrete arguments (indicated by the `types` field, referring to objects of particular types),
+as well as continuous arguments (indicated by `params_space` field, which is formatted as `Box([<param1_lower_bound>, <param2_lower_bound>, ...], [<param1_upper_bound>, <param2_upper_bound>, ...], (<number_of_params>,), <datatype_of_all_params>)`).
+
+{options}
+
+
+Here are a number of example demonstrations of solving a range of goals from an expert demonstrator. For each of these examples, you are provided 
+with the goal the demonstrator was trying to achieve, the plan (i.e., sequence of skills with discrete and continuous arguments specified), as well
```
ashayathalye, line 9: In what format are you providing "the goal the demonstrator was trying to achieve"? 

If we want to stay as close to the original paper as possible, which does things like "Stack all the blocks", "Put all the blocks in the bottom corner" -- then we won't want to give ground atoms, but instead a description of the task.

predicators/approaches/vlm_planning_prompts/few_shot.txt
```diff
@@ -0,0 +1,38 @@
+You are highly skilled in robotic task planning, breaking down intricate and long-term tasks into distinct primitive actions.
```
ashayathalye, line 1: They (see appendix A3) also seem to provide some more (environment-specific?) information about what to do / not do. E.g.
```
If the object is in sight, you need to directly manipulate it. If the object is not in sight, you need to use primitive skills to find the object
first. If the target object is blocked by other objects, you need to remove all the blocking objects before picking up the target object. At
the same time, you need to ignore distracters that are not related to the task.```

We might want to do this too, if we want to more closely replicate their approach. This additional info seems to guide the VLM to propose the right plan without explicitly telling it which skills to use to do something. 

An analagous thing in BurgerEnv might be something like "You have to use the grill to cook the patty."? It seems hard to figure out what additional information to give -- sometimes it gives away too much? Other times maybe it doesn't.

File: predicators/approaches/llm_open_loop_approach.py [hunk 4]
```diff
@@ -135,63 +135,11 @@ def _llm_prediction_to_option_plan(
...
-            # The types of the objects match, but we haven't yet checked if
-            # all arguments of the option have an associated object.
-            if len(objs_list) != len(option.types):
-                malformed = True
-            if not malformed:
-                option_plan.append((option, objs_list))
+        option_plan_with_cont_params = utils.\
+            parse_model_output_into_option_plan(
+            llm_prediction, objects, self._types, self._initial_options, False)
+        option_plan = [(elem[0], elem[1])
```
ashayathalye, line 141: `[(option, objs) for option, objs, cont_params in option_plan_with_cont_params`]

predicators/approaches/vlm_open_loop_approach.py
```diff
@@ -0,0 +1,192 @@
...
+                f"Goal: {str(sorted(traj_goal))}\n"
+            for state_num, state in enumerate(traj.states):
+                assert state.simulator_state is not None
+                assert len(
+                    state.simulator_state["images"]) == num_imgs_per_state
+                for img_num, img in enumerate(state.simulator_state["images"]):
+                    pil_img = PIL.Image.fromarray(img)  # type: ignore
+                    draw = ImageDraw.Draw(pil_img)
+                    img_font = utils.get_scaled_default_font(draw, 20)
+                    img_with_txt = utils.add_text_to_draw_img(
```
ashayathalye, line 95: We probably don't want to draw onto the image itself, in case the text overlaps with objects and/or labels for objects (e.g. our set of marks labels)?

It would be better to add a white rectangle to the top or bottom of the image and put this into into that white rectangle.

predicators/approaches/vlm_planning_prompts/few_shot.txt
NishanthJKumar: yeah - i feel this is unfair: this is basically trying to move all the motion-level reasoning up to the task-planning level. I think it can kinda work in certain cases, but doesn't feel reliable/trustable...

predicators/approaches/vlm_open_loop_approach.py
NishanthJKumar: yes - good idea!

predicators/approaches/vlm_planning_prompts/few_shot.txt
NishanthJKumar: i think it's fine to just not bother here; it's quite annoying to change this and feels not worth it imo (we do have a notion of `EnvironmentTask`, which has a `GoalDescription` that's usually in natural language, but it's really annoying to bring this down into the approach

NishanthJKumar:
re: explaning the continuous params: I feel like we can just do this naive thing for now (and know that it's going to fail). If we stick to envs that don't have continuous params, we'll be fine. If not, this baseline will fail, and we can talk about why. I think there's some possibility that the thing just pattern matches from the few-shot examples, which is basically what we'd be doing if we did our sampler learning: especially with small amounts of data

NishanthJKumar:
Runnable command: 
```
python predicators/main.py --env burger --approach vlm_open_loop --seed 0 --num_train_tasks 1 --num_test_tasks 1 --bilevel_plan_without_sim True --make_failure_videos --sesame_task_planner fdopt --debug --vlm_model_name gemini-1.5-pro-latest --vlm_open_loop_use_training_demos True
```

PR #1706, opened 2024-08-15
Create BurgerEnv variant without a dedicated move option/operator

### LINKED ISSUE #1704: Fix VLM atom proposal generation for multiple objects of same type
Should generate the same atoms for all objects when there are multiple objects of each type, e.g. `"patty1"` and `"patty2"` of type `"_patty_type"` in `BurgerEnv`.

ashayathalye:
This PR does the following:
- Adds cropped images into the ImageOptionTrajectory to supplement the uncropped images in BurgerEnv -- these seem to improve VLM labeling accuracy -- via `CFG.vlm_include_cropped_images`
- Adds a new BurgerEnv variant called BurgerNoMoveEnv that gets rid of the dedicated move option/operator, integrating the movement into each of the other options in order to minimize overfitting and ignore effects that arise from the predicates associated with the original move operator's effects (Adjacent, Facing, etc.).
- Adds a larger variety of train/test tasks into BurgerEnv and BurgerNoMoveEnv.
- Changes the tomato to a lettuce -- whole/cut lettuce labeled more accurately/consistently than whole/sliced tomato.
- Adds realistic images for the uncooked/cooked patty and the unchopped/chopped lettuce.
- Adds a VLM labeling prompt specific to burger.
- Adds an option to hardcode VLM atom proposals for specific environments via `CFG.use_hardcoded_vlm_atom_proposals`
- Adds an option to skip hill-climbing and select debug grammar predicates directly via `CFG.grammar_search_select_all_debug`
- Adds a debug grammar for VLM predicates
- Fixes #1704.

Note that BurgerNoMoveEnv is still a work in progress. 

Example command:
`python predicators/main.py --env burger_no_move --approach grammar_search_invention --seed 0 --num_train_tasks 3 --num_test_tasks 1 --bilevel_plan_without_sim True --offline_data_method geo_and_demo_with_vlm_imgs --vlm_model_name gemini-1.5-pro-latest --segmenter option_changes --grammar_search_vlm_atom_proposal_prompt_type options_labels_whole_traj --grammar_search_vlm_atom_label_prompt_type img_option_diffs_burger --grammar_search_task_planning_timeout 10.0 --sesame_max_skeletons_optimized 200 --disable_harmlessness_check True --sesame_task_planner fdopt --excluded_predicates IsCooked,IsSliced --option_model_terminate_on_repeat False --grammar_search_vlm_atom_proposal_use_debug False --allow_exclude_goal_predicates True --grammar_search_prune_redundant_preds False --grammar_search_predicate_cost_upper_bound 7 --allow_state_allclose_comparison_despite_simulator_state True --grammar_search_max_predicates 200 --grammar_search_parallelize_vlm_labeling True --grammar_search_use_handcoded_debug_grammar True --grammar_search_select_all_debug True --bilevel_plan_without_sim True --cluster_and_intersect_soft_intersection_for_preconditions True --vlm_include_cropped_images True --timeout 80 --grammar_search_vlm_atom_proposal_use_debug True --burger_no_move_task_type "more_stacks"`


predicators/approaches/grammar_search_invention_approach.py
```diff
@@ -1025,6 +1028,11 @@ def learn_from_offline_dataset(self, dataset: Dataset) -> None:
             score_function = create_score_function(
                 CFG.grammar_search_score_function, self._initial_predicates,
                 atom_dataset, candidates, self._train_tasks)
+            self._learned_predicates = set(
+                p for p in candidates.keys() if p.name in [
+                    "((0:robot).fingers<=[idx 0]0.5)", "Cooked0", "Whole0",
+                    "Diced0", "Cut0", "Sliced0", "Shredded0", "Chopped0"
+                ])
```
NishanthJKumar, line 1035: I'm confused: the point of adding this here is to avoid hill-climbing right? but from the code, it looks like hill-climbing is still being run? 
Also, is there a reason we couldn't just make a debug grammar here instead of doing this?

predicators/datasets/vlm_input_data_prompts/atom_labelling/img_option_diffs_label_history.txt
```diff
@@ -1,3 +1,3 @@
...
...o cropped images highlighting objects relevant to the skill's arguments (the third image is a crop of the before image, and the fourth image is a crop of the after image). You are also given a list of predicates and their values in the before image. Your task is to output the values of these predicates in the after image. First, describe the expected visual changes based on the executed skill, explicitly noting the skill performed. Second, describe the visual changes you observe between the before and after images, focusing on the objects involved in the skill's arguments. Finally, list each predicate value in the after image in the format `<predicate>: <truth_value>. <explanation>`, and separately list the predicates that changed. If an object is not visible in the cropped images, check the uncropped images. If you still cannot locate the object involved in a specific predicate, retain the predicate’s value from the previous timestep. Note that when a cook action is taken on any kind of food that is on the grill, it does not take time to finish cooking -- it cooks instantly. Note that just placing any kind of food on the grill without explicitly cooking it does not make it cooked.
```
NishanthJKumar, line 1: Could we separate out the burger-specific part of the prompt (i.e., the last part) and the rest of it? Just want to make sure we don't accidentally apply this to a new env where there is no notion of `grilling' and `cooking', etc

predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -690,8 +719,29 @@ def _generate_ground_atoms_with_vlm_pure_visual_preds(
             image_option_trajs, vlm, 1)
         logging.info("Done querying VLM for candidate atoms!")
         # We now parse and sanitize this set of atoms.
+        atom_proposals_set = set([
+            "Cooked(patty)", "Raw(patty)", "IsBrown(patty)", "IsPink(patty)",
+            "IsGrilled(patty)", "Cut(lettuce)", "Diced(lettuce)",
+            "Sliced(lettuce)", "Whole(lettuce)", "Shredded(lettuce)"
+        ])
```
NishanthJKumar, line 726: this feels dangerous; could you introduce a command-line flag - something like `--use_hardcoded_vlm_atom_proposals` - and do something like how we do the debug grammar? I don't want this/things like this accidentally left-in such that they might trigger unexpected/silent failures in other envs.

predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -282,7 +298,20 @@ def _parse_unique_atom_proposals_from_list(
                         atom_is_valid = False
                         break
             if atom_is_valid:
-                atoms_strs_set.add(atom)
+                # Create VLM atoms for all other possible groundings of this
```
NishanthJKumar, line 301: nice! i'd been meaning to add this feature, but it somehow slipped my mind

predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -958,16 +1008,72 @@ def create_ground_atom_data_from_generated_demos(
...
+                            min_row * approx_cell_size:(max_row + 1) *
+                            approx_cell_size,
+                            min_col * approx_cell_size:(max_col + 1) *
+                            approx_cell_size, :]
+                        cropped_imgs = [
+                            PIL.Image.fromarray(img_arr)  # type: ignore
+                            for img_arr in
+                            [cropped_curr_img, cropped_prev_img]
+                        ]
+                        cropped_state_imgs.append(cropped_imgs)
```
NishanthJKumar, line 1069: important: raise a `NotImplementedError` if the `CFG.env` isn't one of these for which we've implemented cropping!

predicators/envs/burger.py
```diff
@@ -104,6 +109,12 @@ def __init__(self, use_gui: bool = True) -> None:
                              self._On_holds)
         self._OnNothing = Predicate("OnNothing", [self._object_type],
                                     self._OnNothing_holds)
+        # When an argument should be an item or a station but not a robot,
+        # having the argument more narrowly scoped than the object type would be
+        # helpful. But this is causing errors so we'll avoid it for now.
+        # # self._Clear = Predicate("Clear", [self._item_or_station_type],
+        #                         self._Clear_holds)
+        # TODO(ashay): fix the errors associated with this.
```
NishanthJKumar, line 117: if this isn't a TODO anymore; remove the comment

predicators/envs/burger.py
```diff
@@ -279,25 +293,29 @@ def _get_tasks(self, num: int,
...
             }
-
             state = utils.create_state_from_dict(state_dict)
             state.simulator_state = {}
             state.simulator_state["state"] = hidden_state
             # A DefaultEnvironmentTask is a dummy environment task. Our render
             # function does not use the task argument, so this is ok.
             state.simulator_state["images"] = self.render_state(
                 state, DefaultEnvironmentTask)
+
```
NishanthJKumar, line 303: nit: delete this newline

predicators/envs/burger.py
```diff
@@ -610,8 +642,9 @@ def render_state_plt(
         # The DPI has to be sufficiently high otherwise when the matplotlib
         # figure gets converted to a PIL image, text in the image can become
         # blurry.
-        fig, ax = plt.subplots(1, 1, figsize=figsize, dpi=200)
-        plt.suptitle(caption, wrap=True)
+        fig, ax = plt.subplots(1, 1, figsize=figsize, dpi=216)
+        # plt.suptitle(caption, wrap=True)
```
NishanthJKumar, line 646: why did you take out captioning?

NishanthJKumar (commented):
This PR is a tour de force - great work! 🎉  Left a bunch of comments that should be fairly easy to address: but I think this is all mostly ready! Should be mergeable with 1-2 more rounds of feedback 🥇

predicators/datasets/generate_atom_trajs_with_vlm.py
ashayathalye: I was gonna remove this before the final PR -- but I left it in while I keep testing the PR.

But your proposal is better, will do that.

predicators/datasets/vlm_input_data_prompts/atom_labelling/img_option_diffs_label_history.txt
ashayathalye: Good idea, will do.

predicators/approaches/grammar_search_invention_approach.py
ashayathalye: My first reply: I was gonna remove this before the final PR -- I just have this here while I test the PR. 

But also -- remember that the implementation of debug grammar still runs hill climbing -- just on that specific hardcoded pool of predicates -- so it's still slower than just choosing the predicates you want right away. So, I didn't use the debug grammar for speed reasons. 

The right solution is to have two different ways the debug grammar is used -- (1) one where it actually runs hill climbing on the debug grammar pool, another (2) where it just takes the debug grammar predicates directly. 

We'd use (1) when we want to actually test the hill climbing search, and (2) when hill climbing itself is not important.

NishanthJKumar: oh good point. maybe worth quickly implementing this "right" solution? I'm sure it'll be useful in future envs as well!

predicators/envs/burger.py
ashayathalye: short answer: this is still a TODO, but it is not a high priority since we don't seem to have any problems right now the way it is. 

longer answer:
Ok, so I'll expand on why I was trying to make a new type that was an "item or a station". Like, I wanted the type hierarchy to be `_object_type` -> `_item_or_station_type`-> <`_item_type`, `_station_type`>. 

Basically, for some of the `GoalHack` predicates, they take in two arguments. I initially made the first argument `_object_type`, so e.g. one of the GoalHacks was true whenever <the patty was above ANY object + the patty was cooked> but then I was seeing this become true at times when I didn't want it to, because like, I saw `GoalHack(robot, patty)`. I temporarily dealt with this by just having 2 or more GoalHacks for the patty being on the cutting board + cooked, and another for the patty being on the grill + cooked., and another for being on top of the bottom bun + cooked, but it would be cleaner to have fewer goalhack predicates. 

But, I came across some weird error when trying to use `_item_or_station_type`. It's low priority so I think we should come back to it later, but I want to leave it as a comment since when we get to the final experiments for the paper, for cleanliness, we probably as few GoalHack predicates as possible.

predicators/envs/burger.py
ashayathalye: I don't remember... I think maybe when I was doing the cropped images. I want the matplotlib output to be the same dimension all the time and not have much else other than the grid itself (no white padding!) so that I can basically hardcode some cell-width values (so it makes it more reliable to automatically get a crop of a particular set of cells). 

I think maybe the caption was unnecessary and adds variable whitespace which can mess up the cropping. We aren't going to use the caption at all at least when creating demonstrations to train on.

predicators/datasets/vlm_input_data_prompts/atom_labelling/img_option_diffs_label_history.txt
ashayathalye: Rather than environment-specific additions to the prompt, I think because we might want multiple additions at different places in the orignal prompt (e.g. before or after the predicates are listed, before or after the beginning paragraph, etc.) it's better to just have an environment-specific prompt itself.

predicators/approaches/grammar_search_invention_approach.py
ashayathalye: Will do.

predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -103,13 +106,12 @@ def _generate_prompt_for_scene_labelling(
     except FileNotFoundError:
         raise ValueError("Unknown VLM prompting option " +
                          f"{CFG.grammar_search_vlm_atom_label_prompt_type}")
+
```
NishanthJKumar, line 109: nit: revert

predicators/datasets/vlm_input_data_prompts/atom_labelling/img_option_diffs_label_history.txt
```diff
@@ -1,3 +1,3 @@
...
```
NishanthJKumar, line 1: why does this still have a burger-specific line at the end? Don't we want this prompt to be fully general across all envs?

predicators/envs/burger.py
NishanthJKumar: sounds good! We can revisit this after we start running experiments + getting results in Burger (which will hopefully be v soon!)

NishanthJKumar (commented):
looks generally good! Left some comments/questions about the prompting though

predicators/datasets/vlm_input_data_prompts/atom_labelling/img_option_diffs_label_history.txt
ashayathalye: Mistake -- reverting to what that file is on `master` currently.

NishanthJKumar (approved):
Great work! Very excited we've hit this pretty-cool milestone! :tada: 
Minor thing: could you post some example commands in the PR description? will be useful for posterity

PR #1709, opened 2024-09-24
Test-time VLM atom labelling with history

NishanthJKumar:
This PR enables us to label atom values using history-based prompting at test-time. It also:
* using VLM predicates as part of goal predicates (by modifying `Task.goal_holds()`
* using the execution monitor with `grammar_search_invention` approach

IMPORTANT: while this PR sets up all the mechanics necessary to accomplish test-time labelling, this won't fully work until we are able to include the `state_history`, `skill_history` and `vlm_label_history` fields in every `state.simulator_state`. We will do this in a forthcoming PR!


predicators/settings.py
```diff
@@ -704,6 +704,9 @@ class GlobalSettings:
     # saved_vlm_img_demos_folder
     vlm_trajs_folder_name = ""
     vlm_predicate_vision_api_generate_ground_atoms = False
+    # At test-time, we will use the below number of states
+    # as part of labelling the current state's VLM atoms.
+    vlm_test_time_atom_label_prompt_type = "per_scene_naive"
```
ashayathalye, line 709: Maybe move these two 
```
grammar_search_vlm_atom_proposal_prompt_type = "options_labels_whole_traj"
grammar_search_vlm_atom_label_prompt_type = "per_scene_naive"
```
down to this area just to have all the vlm settings in the same place and rename them to just `vlm_[something]` without the `grammar_search` prefix for consistency.

PR #1721, opened 2024-10-24
Fix skill history usage in VLM prompt and add double checking VLM output with another call to VLM

ashayathalye:
This PR does two things. 

First, it fixes a bug that affected which skill was passed into the prompt given to the VLM at training time. After some refactoring we did, it turned out that the skill passed into the VLM in each labeling prompt was always the last skill in the trajectory, rather than the skill that was run just before the timestep being labeled. This fix was made in a clean way by only passing the skill history up to the current timestep in `_generate_prompt_for_scene_labelling()`.

Second, it adds an option to have the VLM double check its answer -- which often improves it. In this case, we ask the VLM to double check its reasoning w.r.t. how it uses the previous timestep's labels. 

WIth this PR (after adding double checking), the labeling in `--burger_no_move_task_type combo_burger` is good enough for the debug grammar to work!

For example, the command
```
python predicators/main.py --env burger_no_move --approach grammar_search_invention --num_test_tasks 10 --grammar_search_vlm_atom_proposal_prompt_type options_labels_whole_traj_diverse --offline_data_method geo_and_demo_with_vlm_imgs --grammar_search_invent_geo_predicates_only False --burger_no_move_task_type combo_burger --bilevel_plan_without_sim True --segmenter option_changes --grammar_search_vlm_atom_label_prompt_type img_option_diffs_label_history_burger --grammar_search_task_planning_timeout 10.0 --sesame_max_skeletons_optimized 200 --disable_harmlessness_check True --sesame_task_planner fdopt --excluded_predicates all --option_model_terminate_on_repeat False --grammar_search_vlm_atom_proposal_use_debug True --allow_exclude_goal_predicates True --grammar_search_prune_redundant_preds True --grammar_search_predicate_cost_upper_bound 5 --allow_state_allclose_comparison_despite_simulator_state True --grammar_search_max_predicates 100 --grammar_search_parallelize_vlm_labeling True --grammar_search_use_handcoded_debug_grammar True --grammar_search_select_all_debug True --cluster_and_intersect_soft_intersection_for_preconditions True --precondition_soft_intersection_threshold_percent 0.8 --vlm_include_cropped_images True --timeout 80 --grammar_search_grammar_includes_givens False --cluster_and_intersect_prune_low_data_pnads False --cluster_and_intersect_min_datastore_fraction 0.05 --num_train_tasks 8 --seed 0 --vlm_model_name gpt-4o --vlm_double_check_output True
```
gets 10/10 tasks solved.


predicators/datasets/generate_atom_trajs_with_vlm.py
```diff
@@ -101,9 +101,10 @@ def _generate_prompt_for_scene_labelling(
         yield utils.get_prompt_for_vlm_state_labelling(
             CFG.grammar_search_vlm_atom_label_prompt_type, atoms_list,
             label_history, traj.imgs[:i + 1], traj.cropped_imgs[:i + 1],
-            traj.actions)
+            traj.actions[:i])
```
NishanthJKumar, line 104: omg - can't believe we missed this earlier; phenomenal catch!

NishanthJKumar (approved):
LGTM! Great work - and thanks for a blazing fast turnaround on this (I definitely wouldn't have thought to do this double-check thing!) 🔥 👏

PR #1731, opened 2024-11-11
Break vila-with-fewshot in burger

ashayathalye:
There are many ways to go about breaking vila-with-fewshot in burger. One we settled on was having some test tasks start with the robot holding a patty -- something not seen in training. Then the VLM will fail to deal with this when pattern-matching to the demos.

### DIFF
--- a/predicators/approaches/grammar_search_invention_approach.py
+++ b/predicators/approaches/grammar_search_invention_approach.py
@@ -1105,14 +1105,10 @@ class GrammarSearchInventionApproach(NSRTLearningApproach):
                     CFG.grammar_search_score_function,
                     self._initial_predicates, atom_dataset, candidates,
                     self._train_tasks)
-                self._learned_predicates = set(
-                    p for p in candidates
-                    if p.name in ["cooked0", "free0", "chopped0"])
-                import pdb; pdb.set_trace()
-                # self._learned_predicates = \
-                #     self._select_predicates_by_score_hillclimbing(
-                #     candidates, score_function, self._initial_predicates,
-                #     atom_dataset, self._train_tasks)
+                self._learned_predicates = \
+                    self._select_predicates_by_score_hillclimbing(
+                    candidates, score_function, self._initial_predicates,
+                    atom_dataset, self._train_tasks)
         elif CFG.grammar_search_pred_selection_approach == "clustering":
             self._learned_predicates = self._select_predicates_by_clustering(
                 candidates, self._initial_predicates, dataset, atom_dataset)


predicators/approaches/vlm_open_loop_approach.py
```diff
@@ -211,6 +211,7 @@ def _query_vlm_for_option_plan(self, task: Task) -> Sequence[_Option]:
                 num_completions=1)
         plan_prediction_txt = vlm_output[0]
         option_plan: List[_Option] = []
+        # import pdb; pdb.set_trace()
```
NishanthJKumar, line 214: remove unnecessary comment

predicators/approaches/vlm_planning_prompts/few_shot.txt
```diff
@@ -9,7 +9,7 @@ Here are a number of example demonstrations of solving a range of goals from an
 with the goal the demonstrator was trying to achieve, the plan (i.e., sequence of skills with discrete and continuous arguments specified), as well
 as images corresponding to the state of the environment at every timestep (starting at 0, when the 0th skill is executed). Note that the state images
 are annotated with text indicating which demonstration and timestep they belong to. Additionally, note that a state might have more than one image
-associated with it.
+associated with it.
```
NishanthJKumar, line 12: revert space

predicators/envs/burger.py
```diff
@@ -1327,6 +1359,7 @@ def create_tasks_for_type_two(
                 patty2 = Object("patty2", self._patty_type)
                 state_dict[patty2] = {"row": r, "col": c, "z": 0}
                 hidden_state[patty2] = {"is_cooked": 0.0, "is_held": 0.0}
+
```
NishanthJKumar, line 1362: revert

predicators/pretrained_model_interface.py
```diff
@@ -171,8 +171,8 @@ def set_openai_key(self, key: Optional[str] = None) -> None:
             assert "OPENAI_API_KEY" in os.environ
             key = os.environ["OPENAI_API_KEY"]
 
-    @retry(wait=wait_random_exponential(min=1, max=60),
-           stop=stop_after_attempt(10))
+    # @retry(wait=wait_random_exponential(min=1, max=60),
+    #        stop=stop_after_attempt(10))
```
NishanthJKumar, line 175: revert - we need this!

scripts/configs/pred_invention_vlm.yaml
```diff
@@ -1,50 +1,50 @@
 # Experiments to test predicate invention with VLMs
 ---
 APPROACHES:
-  ours:
```
NishanthJKumar, line 4: revert all changes to this file, except getting rid of the comments on the fact that a different `num_train_tasks` is needed for VILA

predicators/envs/burger.py
```diff
@@ -1376,17 +1409,44 @@ def create_tasks_for_type_two(
...
+                    test_goal.add(GroundAtom(self._IsCooked, [patty3]))
+                    test_goal.add(GroundAtom(self._On, [patty3, bottom_bun2]))
                 alt_test_goal = {
                     GroundAtom(self._GoalHack2, [bottom_bun, patty]),
                     GroundAtom(self._GoalHack6, [patty, patty2]),
                     GroundAtom(self._On, [top_bun, patty2])
                 }
+                if i < CFG.burger_num_test_start_holding:
+                    alt_test_goal.add(
+                        GroundAtom(self._GoalHack2, [bottom_bun2, patty3]))
```
NishanthJKumar, line 1449: just to confirm my understanding here - is the goal when `i < CFG.burger_num_test_start_holding` that the agent should cook the patty it's holding and also put it on a bottom bun (in addition to the usual goal)? Why do we need to explicitly tell it to cook the thing it's holding - can we just not change the goal at all? Just curious

NishanthJKumar (commented):
great work! i think some changes are necessary before we can properly merge this though

predicators/envs/burger.py
ashayathalye: The goal is to have a flag that lets you control what % of test tasks have you start out holding a patty. The easy way to implement this was to have the first `N=CFG.burger_num_test_start_holding` test tasks have this property. 

My initial thought was:
- We need the patty to be cooked before the agent places it on the bottom bun because our "place" operator that is applicable here has a goalhack add effect that is "patty is cooked + patty is above bottom bun". 

But actually -- we don't have any goal related to that extra patty we start out holding, I don't think it should be a problem taht we don't actually satisfy the add effects of the operator upon execution, since we aren't doing execution monitoring at the moment. 

Can continue discussion on slack.

ashayathalye: Ok -- if we remove explicit goals related to the patty we start out holding, we still find a successful plan that starts out with placing that patty (and we cook it before placing it, as expected).

NishanthJKumar:
good to know our approach works without modifying the goals! However, given we ran experiments on the modified goals, could you put those back? I think that's probably the most fair thing for now

NishanthJKumar (approved):
LGTM!

